PR #18, opened 2018-09-13
Prepending kw namespace to exported functions.

RenatoGeh:
Co-authored-by: Matheus Tavares <matheusmalavasi@uol.com.br>

### DIFF
--- a/kw.sh
+++ b/kw.sh
@@ -16,7 +16,7 @@ external_script_path=${external_script_path:-"$HOME/.config/$EASY_KERNEL_WORKFLO
 # Export external variables required by kworkflow
 export EASY_KERNEL_WORKFLOW
 
-function kworkflow-help()
+function kw::help()
 {
   echo -e "Usage: kw [target] cmd"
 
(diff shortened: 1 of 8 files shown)


rodrigosiqueira:
First of all, thanks for your patch and your work! I really appreciate your effort. :)

I'm a little bit confused about the purpose of this PR. Do you intend to fix the problem that kw functions are visible for the whole system? If so, how add "kw::" to the beginning of each function solves the problem? From my point of view, the exportation problem is not fixed by adding "kw::" at the beginning of each function. Did I miss something?

If possible, test the version in the branch "rework_export_names".

Ps.: I'm using bash

RenatoGeh:
We intended on fixing what was taken care of in fdc4e45234bc0d1b39b59509d113526b1b80cea2. The idea was to handle function visibility by "obscuring" the exposed functions through a namespace. The assumption was that nobody would use the prepended namespace. It was a silly fix, and your solution is better. Should we close this PR?

Issue #17, opened 2018-09-12
Create a Host deploy option


giulianobelinassi:
Description Later

lms11:
Description: add feature to use kworkflow as host. This should allow deploying on host.

rodrigosiqueira:
There is a variable named TARGET which can be useful for this issue. Maybe this variable can receive the string "host" and change the normal behavior of kw. Remember that you have to keep kw flexible enough for working with a VM and local machine. If you have any question, do not hesitate to ask me ;)

Ps.: Please, write a detailed description of this feature in the commit message when you finish it ;)

rodrigosiqueira:
Ongoing on KernelInstall branch

PR #22, opened 2018-09-24
Implementation of host deploy option

capellaresumo:
Fixes Issue #17

### DIFF
--- a/kw.sh
+++ b/kw.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 # Set required variables
+TARGET="qemu" 
 EASY_KERNEL_WORKFLOW=${EASY_KERNEL_WORKFLOW:-"kw"}
 src_script_path=${src_script_path:-"$HOME/.config/$EASY_KERNEL_WORKFLOW/src"}
 external_script_path=${external_script_path:-"$HOME/.config/$EASY_KERNEL_WORKFLOW/external"}
@@ -12,6 +13,14 @@ function kw()
 {
   action=$1
   shift
+
+  while getopts 'lv' c
+  do
+    case $c in
+      l) TARGET="host" ;;
+      v) TARGET="qemu" ;;
+    esac
+  done
 
   case "$action" in
     mount)
@@ -123,3 +132,4 @@ function kw()
       ;;
   esac
 }
+
--- a/kw.sh
+++ b/kw.sh
@@ -123,7 +123,7 @@ function kw()
       ;;
     set-deploy-target)
       (
-
+        . $src_script_path/commons.sh --source-only
       
         set_deploy_target $1
       )
--- a/kw.sh
+++ b/kw.sh
@@ -133,3 +133,6 @@ function kw()
   esac
 }
 
+kw $@
+
+
--- a/kw.sh
+++ b/kw.sh
@@ -121,6 +121,13 @@ function kw()
         kworkflow-help
       )
       ;;
+    set-deploy-target)
+      (
+
+      
+        set_deploy_target $1
+      )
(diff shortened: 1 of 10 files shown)


src/cfg.sh
```diff
@@ -0,0 +1,15 @@
+# Author: therealfarfetchd
```
rodrigosiqueira: Please, move this code to `util.sh` and follows the current code style

kw.sh
```diff
@@ -66,8 +66,7 @@ function kw()
     new|n)
       (
         . $src_script_path/mk.sh --source-only
```
rodrigosiqueira: Add space, remember to following the current code style.

src/mk.sh
```diff
@@ -31,6 +31,38 @@ function vm_new_release_deploy
...
+    host_new_release_deploy
+  else
+    vm_new_release_deploy
+  fi
+}
+
+function host_kernel_install
+{
+  if [ -e "/etc/arch-release" ]
+  then
```
rodrigosiqueira: Remember to following the current code style.

src/deploy/arch.sh
```diff
@@ -0,0 +1,12 @@
+IMAGE_NAME=$1
```
rodrigosiqueira: Change the name of "deploy" to "specific_target_distro"

src/mk.sh
```diff
@@ -31,6 +31,38 @@ function vm_new_release_deploy
...
+    sudo bash $src_script_path/deploy/arch.sh linuxkw
```
rodrigosiqueira: make linuxkw global

rodrigosiqueira:
First of all, thanks for your PR! Follows my review

There are many problems on how these series of patches were separated. I will not comment on all of the issues related to this series; I will mention just some of them in order to provide examples. Before it, I want
to give some guidelines on how you have to prepare a series of patches,
keep in mind:

1. Do not think in terms of your changes. For example, the fixed typo commits that you sent does not make sense in the original code because the fix is related to your code.
2. Break commit in features, not in small pieces of code. For example, all
the code related to set-deploy-target should be part of a single commit
3. Take some time to write the commit messages.

Please, take a careful read at:
https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#describe-your-changes
https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#separate-your-changes

So, let's take a look in some of the commits:

```
Typo fix: add_to_bashrc in setup.sh (709f0f9027bf1)
Bug fix: wrong CONFIG_PATH in check_local_configuration (ec5fdbe50394c80b)
```

None of these commits are related to the kworkflow (upstream) code, they are related to your code. You have to make rebases in your git tree to get rid of this kind of commit.

About the commit:

```
Add set-deploy-target in help description
```

This commit changes many things, and this is not good. This commit only have to handle the help function, the other stuff does not belong to this patch.

About the commit:

"Adding set-deploy-target command"

This features should be in a single commit, for example, the function
set_deploy_target is defined in another commit.

Please, remove the "merge" commits.

What is the purpose of the commit:

" Set default value in file, remove from code."
? Why use "drm-misc"? 

This commit makes thousands of thing; this is not good.

Do not add extra white space in your code, always take a look on the diff
before committing the code.

Keep the focus on the issue. Some of the patches do not belong to the issue and could be extracted in other PR.

Where is the code documentation of the new feature?  How can I know in advance how to use your new feature? Remember to add the required information in the README.

rodrigosiqueira:
Hi,

any update about this issue?

rodrigosiqueira:

Issue #9, opened 2018-08-27
Replace Qemu options for libvirt


rodrigosiqueira:
I had a small talk with a Qemu maintainer, and he recommended me to remove the direct use of the Qemu option for support and instability reasons. He suggested me to use libvirt.

HanHaijiao:
Agree. Libvirt is more convenient, with some tools like virt-manager, virt-viewer, virt-install and so on, 
could ignore many details, focus on your businese。

rodrigosiqueira:
We already made a first attempt in the PR #23.

However, I think we have to improve other stuff before get back to this issue.
Ps.: Handling libvirt is a hard task.

rodrigosiqueira:
Just for keep tracking of the information, follows some of the commands that we used during our first attempt of use libvirt:

```
=============
sudo virsh --connect=qemu:///system net-define --file /home/siqueira/.config/kw/config_files/kworkflow-network-default.xml  
=============
virsh --connect=qemu:///system net-start kworkflow-network
=============
virt-install --connect=qemu:///system -n kworkflow-vm --ram 1024  --boot hd --vcpus=2 --cpu=host --os-type=linux --os-variant=virtio26 --disk=/home/siqueira/p/virty.qcow2,format=qcow2 --network network=kworkflow-network,mac=52:54:00:21:23:52 --filesystem /home/siqueira/linux,kw_share

virt-install --connect=qemu:///system -n kworkflow-vm --ram 1024 --boot hd --vcpus=2 --cpu=host --os-type=linux --os-variant=virtio26 --disk=/home/siqueira/p/virty.qcow2,format=qcow2 --network network=default --filesystem /home/siqueira/linux,kw_share

virsh --connect=qemu:///system net-list --all
virsh list --all

virsh --connect=qemu:///system undefine <VM_NAME>    
virsh start kworkflow-vm
virt-viewer --connect=qemu:///system

virsh net-update default add ip-dhcp-host \
          "<host mac='52:54:00:00:00:01' \
           name='bob' ip='192.168.122.45' />" \
           --live --config
====
# virsh net-define /usr/share/libvirt/networks/default.xml
Network default defined from /usr/share/libvirt/networks/default.xml
# virsh net-autostart default
Network default marked as autostarted
# virsh net-start default
Network default started
======
virsh net-list --all
virsh net-destroy default
virsh  net-undefine default
virsh --connect=qemu:///system destroy kworkflow-vm
[...]

PR #23, opened 2018-09-24
Libvirt work

marceloschmitt1:
Replacing qemu with libvirt, which solves issue #9.

### DIFF
--- a/config_files/kworkflow-network-default.xml
+++ b/config_files/kworkflow-network-default.xml
@@ -14,6 +14,7 @@ or other application using the libvirt API.
   <ip address='10.0.1.1' netmask='255.255.255.0'>
     <dhcp>
       <range start='10.0.1.2' end='10.0.1.2'/>
+      <host mac='52:54:00:21:23:52' name='kworflow-vm' ip='10.0.1.2'/>
     </dhcp>
   </ip>
 </network>
--- a/config_files/kworkflow-network-default.xml
+++ b/config_files/kworkflow-network-default.xml
@@ -7,7 +7,7 @@ or other application using the libvirt API.
 
 <network>
   <name>kworkflow-network</name>
-  <uuid>3f94494f-356f-4d9d-893e-4acd18d40a98</uuid>
+  <uuid>7d167334-7e36-475b-a0ae-9889b87848d7</uuid>
   <forward mode='nat'/>
   <bridge name='virbr-kworkflow' stp='on' delay='0'/>
   <mac address='52:54:00:8a:da:00'/>
--- a/config_files/kworkflow-network-default.xml
+++ b/config_files/kworkflow-network-default.xml
@@ -11,9 +11,9 @@ or other application using the libvirt API.
   <forward mode='nat'/>
   <bridge name='virbr-kworkflow' stp='on' delay='0'/>
   <mac address='52:54:00:8a:da:00'/>
-  <ip address='192.168.122.1' netmask='255.255.255.0'>
+  <ip address='10.0.1.1' netmask='255.255.255.0'>
     <dhcp>
-      <range start='192.168.122.2' end='192.168.122.2'/>
+      <range start='10.0.1.2' end='10.0.1.2'/>
     </dhcp>
   </ip>
 </network>
--- a/src/vm.sh
+++ b/src/vm.sh
@@ -25,6 +25,7 @@ function vm_umount
 
 function vm_boot
 {
+  #seems broken
   check_local_configuration
 
   $QEMU -hda $VDISK \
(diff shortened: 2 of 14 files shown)


rodrigosiqueira:
Just for let you know, I already started to review your patch. I'm just doing some research to understand why "kw register" break the VM boot. This problem really annoying me...

src/vm.sh
```diff
@@ -59,12 +94,71 @@ function vm_ssh
...
+  
+  local current_path=$PWD
+  local virt_net_config_path=$PWD/kworkflow-network.xml
+
+  #set network config to default if inexistent
+  if [ ! -f $virt_net_config_path ]; then
+    virt_net_config_path=$DEFAULT_CONFIG_PATH/kworkflow-network-default.xml
+  fi
+
+  if [ $(vm_status $VIRT_VM_NAME) != "inexistent" ] || \
```
rodrigosiqueira: 1. Remove the VM name validation from here (see the next review comment)
2. Check if we already have the network and, based on that, take some decision. We suggest something like this:

```sh
if [ $(net_status $VIRT_NET_NAME) != "inexistent" ]  ; then
  say "VM installation process."
  say "Creating libvirt network. Sudo is required here."
  sudo $VIRT_NET_DEFINE --file $virt_net_config_path  
  $VIRT_NET_START $VIRT_NET_NAME
fi
```

src/vm.sh
```diff
@@ -59,12 +94,71 @@ function vm_ssh
...
+    complain "Please run 'kw clean' to unregister both VM and Network."
+    return
+  fi
+
+  say "VM installation process."
+  say "Creating libvirt network. Sudo is required here."
+  sudo $VIRT_NET_DEFINE --file $virt_net_config_path  
+  $VIRT_NET_START $VIRT_NET_NAME
+  say "Creating libvirt vm..."
+  $VIRT_INSTALL -n $VIRT_VM_NAME \
```
rodrigosiqueira: We have to make some changes in this part. Follows:

1. We have capture the return status. If fails, inform the user... otherwise go ahead
2. Remove the mac address, let the libvirt handle it for you

src/vm.sh
```diff
@@ -59,12 +94,71 @@ function vm_ssh
...
+                --ram 1024 \
+                --boot hd \
+                --vcpus=2 \
+                --cpu=host \
+                --os-type=linux \
+                --os-variant=virtio26 \
+                --disk=${configurations[virt_path_image]},format=qcow2 \
+                --network network=$VIRT_NET_NAME,mac=52:54:00:21:23:52 \
+                --filesystem $HOME/linux,kw_share
```
rodrigosiqueira: We have to think a little bit more about this...

config_files/kworkflow-network-default.xml
```diff
@@ -0,0 +1,20 @@
...
+or other application using the libvirt API.
+-->
+
+<network>
+  <name>kworkflow-network</name>
+  <uuid>7d167334-7e36-475b-a0ae-9889b87848d7</uuid>
+  <forward mode='nat'/>
+  <bridge name='virbr-kworkflow' stp='on' delay='0'/>
+  <mac address='52:54:00:8a:da:00'/>
+  <ip address='10.0.1.1' netmask='255.255.255.0'>
```
rodrigosiqueira: <ip address='10.0.1.1' netmask='255.255.255.0'/>

config_files/kworkflow-network-default.xml
```diff
@@ -0,0 +1,20 @@
...
+    <dhcp>
```
rodrigosiqueira: Remove all the <dhcp> block

config_files/kworkflow-network-default.xml
```diff
@@ -0,0 +1,20 @@
+<!--
```
rodrigosiqueira: How about set static IP in each VM that Kw manages?
1. Mount vm disk (vm_mount)
2. Detect the distribution
3. Change the dhcp.conf by one of the IPs that we manage
4. Umount the vm disk (vm_umount)

Remember, we have to manage the IPs. Probably, we want to do it in a different file.

rodrigosiqueira:
Another comment:

If we handle multiple VMs, the ```kw down``` have to be update for this situation.

rodrigosiqueira:
Hey guys, sorry for the long time to reply you. However, use libvirt is extremely hard! I put as much effort as I could to figure out all the problems, sadly, it is hard to make libvirt stable on multiple distro. I'm convinced that it will take a long time to make this work well, in this sense, I decided to create a new work branch for this and close this PR. I will work on this; if you want to keep work on this, let me know ;)

Issue #14, opened 2018-09-05
Add a management mechanism for .config files


rodrigosiqueira:
During the kernel development activities, it is common to have multiple ```.config``` files for different targets. For example, one ```.config``` file for your host machine and other for your virtual machine. How about make kworkflow provide tools for manage this file? Something like:

```
kw register_config /PATH/TO/.config -d "This .config file works well on my x86 machine"
kw register_config /PATH/TO/.config -d "This .config file works for my arm device"
kw show_configs
 1) This .config file works well on my x86 machine
 2) This .config file works for my arm device
kw get_configs 1
```

PR #26, opened 2018-10-01
Add management mechanism for .config files

capellaresumo:
This commit solve issue #14. Added the following functions
  * show_configs    -- list all registered .config files
  * get_config      -- copy a registered .config to current directory
  * register_config -- register a specific .config file into the db

Note that our code allows to register is not limited to .config file,
that means, any file can be added into the database.

Sign-off-by Giuliano Belinassi <giulano.belinassi@usp.br>
Sign-off-by Gabriel Capella <gabriel@capella.pro>


src/config_manager.sh
```diff
@@ -0,0 +1,69 @@
...
+    if echo $line | grep -F = &>/dev/null
+    then
+      i=$(($i+1))
+      paths[$i]=$(echo $line | cut -d '=' -f 1 | tr -d '[:space:]')
+      comments[$i]=$(echo "$line" | cut -d '=' -f 2-)
+    fi
+  done < $DB_PATH
+}
+
+function register_config()
```
rodrigosiqueira: This feature is close to become nice... however, take a look in the following scenario:
```
cd my_kernel_branch/
kw rc .config "This is my working version of lala"

cd ../another_kernel_branch/
kw gc 1
```

This will not work well because you only save the path. The `kw` saving only the name of the config file. This feature will be correct if it saves the config file registered in the directory `config_files` with a name `config_<INDEX>`; then, the `kw gc` just take a copy of it in the current directory. With this change, we will have a real backup system of `.config` files.

Finally, it could be nice if we can have an alias option for each config file that we manager instead of using the index number. For example:
```
kw rc .config "This config file handle my Rasp" rasp
kw rc .config "This config file handle my x86 pc" x86host
kw sc
1) rasp "This config file handle my Rasp"
2) x86host "This config file handle my x86 pc"
kw gc rasp # or kw gc 1
kw gc x86host # or kw gc 2
```

src/config_manager.sh
```diff
@@ -0,0 +1,69 @@
...
+{
+  load_configs
+
+  if [ $# -eq 0 ] ; then
```
rodrigosiqueira: You have to be pickier in this validation, notice that verification pass if I just specify one or n elements (I suppose that you do not want it).

rodrigosiqueira:
First of all, thanks for your patch! Really nice one :)

Keep your branch synchronized with the unstable version; you are currently with the code from master. Note that the unstable branch changed the way to invoke a command, just to provide an example, take a look in the snippet of code below:

```bash
...
codestyle)
      (
      . $src_script_path/checkpatch_wrapper.sh --source-only

        execute_checkpatch $@
      )
      ;;
    maintainers)
      (
        . $src_script_path/get_maintainer_wrapper.sh --source-only

        execute_get_maintainer $@
      )
      ;;
    register_config|rc)
       (
        . $src_script_path/config_manager.sh

        register_config $@
       )
      ;;
...
```
Additionally, the help function was moved from the file `kw.sh` for the `utils.sh`.

I recommend you to remove `. $src_script_path/commons.sh --source-only` from the kw.sh file, and add it in the `config_manager.sh` (see the example above, on how to call the function after this change).

Remember to prettify your outputs.

rodrigosiqueira:

rodrigosiqueira:
Due to the lack of activity, I closed this PR. Thanks for your efforts, I will work on it; if you want to work on this issue let me know ;)

Issue #13, opened 2018-09-05
Encapsulate grep and find search


rodrigosiqueira:
A common task in the kernel development it is the task of finding a snippet of code via grep or a file via find. Usually, I'm using a command such as ```grep -nrwI PATH/ -e EXPRESSION```, we can encapsulate this sort os search. Maybe something like:

```
kw search EXPRESSION -> search the entire directory
kw search PATH EXPRESSION
```

It could be nice if this feature design becomes expansive for another sort of search, for example, git log.

PR #27, opened 2018-10-02
Adds 'kw explore'

RenatoGeh:
Adds a new feature 'kw explore' that has two purposes:

- Searches files and/or directories for a regular expression.
- Searches the git log for commits that contain a certain expression in
its diff or commit message.

This commit also adds a unit test for the 'explore' function.

Signed-off-by: Renato Lui Geh <renatogeh@gmail.com>

----
Refers to Issue #13

### DIFF
--- a/src/utils.sh
+++ b/src/utils.sh
@@ -1,3 +1,5 @@
+. $src_script_path/miscellaneous.sh --source-only
+
 function kworkflow-help()
 {
   echo -e "Usage: kw [target] cmd"
@@ -23,4 +25,32 @@ function kworkflow-help()
     "\t                             the mailing list. \"-a\" also\n" \
     "\t                             prints files authors\n" \
     "\thelp,h - displays this help mesage"
+    "\tcodestyle - Apply checkpatch on directory or file\n" \
+    "\tmaintainers - Return the maintainers and the mailing list\n" \
+    "\texplore,e - Search for expression on git log or directory\n" \
+    "\thelp"
+}
+
+function explore()
+{
+  if [[ "$#" -eq 0 ]]; then
+    complain "Expected path or 'log'"
+    exit 1
+  fi
+  case "$1" in
+    log)
+      (
+        git log -S"$2" "${@:3}"
+      );;
+    *)
+      (
+        local path=${@:2}
+        local regex=$1
+        if [[ $# -eq 1 ]]; then
+          path="."
+          regex=$1
+        fi
+        git grep -e $regex -nI $path
+      );;
+  esac
 }
--- a/tests/utils_test.sh
+++ b/tests/utils_test.sh
@@ -6,12 +6,25 @@
 function suite
 {
   suite_addTest "testHelp"
+  suite_addTest "testExplore"
 }
 
 function testHelp
 {
   HELP_OUTPUT=$(kworkflow-help | head -n 1)
   [[ $HELP_OUTPUT =~ Usage:\ kw.* ]]; assertTrue "Help text not displaying correctly." $?
+}
+
+MSG_OUT="GNU"
+LOG_OUT="Initial commit"
+
+function testExplore
+{
+  assertEquals "Expected an error message." "Expected path or 'log'" "$(explore)"
+  assertEquals "$MSG_OUT nopath" "$(explore LICENSE | grep "$MSG_OUT" | awk '{print $2}') nopath"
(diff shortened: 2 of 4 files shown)


src/utils.sh
```diff
@@ -20,5 +22,25 @@ function kworkflow-help()
     "\tup,u - Wake up vm\n" \
     "\tcodestyle - Apply checkpatch on directory or file\n" \
     "\tmaintainers - Return the maintainers and the mailing list\n" \
+    "\tsearch - Search for expression on git log or directory\n" \
     "\thelp"
 }
+
+function search()
+{
+  if [[ "$#" -eq 0 ]]; then
```
rodrigosiqueira: You want to be more strict about this, note that I can pass thousands of parameter with this validation. Additionally, add a unit test to check the parameters values.

src/utils.sh
```diff
@@ -20,5 +22,25 @@ function kworkflow-help()
...
+    complain "Expected path or 'log'"
+    exit 1
+  fi
+  if [[ "$1" == "log" ]]; then
+    git log | grep -nwI -e "$2"
```
rodrigosiqueira: There is a better way to do it. Take a look at:

https://git-scm.com/docs/git-grep
https://stackoverflow.com/questions/2928584/how-to-grep-search-committed-code-in-the-git-history/2929502
https://stackoverflow.com/questions/1337320/how-to-grep-git-commit-diffs-or-contents-for-a-certain-word
https://stackoverflow.com/questions/7124914/how-to-search-a-git-repository-by-commit-message/7124949#7124949

src/utils.sh
```diff
@@ -20,5 +22,25 @@ function kworkflow-help()
...
+  else
```
rodrigosiqueira: Maybe the "case" operator is better than the "if/else" in this case, because we want to expand this options soon.

rodrigosiqueira:
First of all, follows some general comments about the patch:
1) You forgot to update the README with an explanation on how to use this command.
2) Try to find a small name for the search operation. Suggestion (think about it): explore,e
3) Instead of adding a huge file for the test, I recommend you to use the kworkflow as a target.
The other comments can be found directly in the code.

rodrigosiqueira:
Hi @RenatoGeh 
I want to highlight that search with `git log -S<word> -p` or `git log -S<word>` are also extremely useful [1]. It could be awesome if we can support some variety of search that uses git grep and git log.

1. https://stackoverflow.com/a/1340245

rodrigosiqueira:

src/utils.sh
RenatoGeh: > You want to be more strict about this, note that I can pass thousands of parameter with this validation. 

As of this new patch, this is intended behavior. `kw explore` now accepts multiple files/directories when searching for expressions.

> Additionally, add a unit test to check the parameters values.

I don't see how this can be done, as parameters are either directories/files, or just a regular expression. The way I see it, any value can be passed no problem.

RenatoGeh:
Thank you for the review. I have addressed the requested changes and force pushed a new commit in place of the old to keep it nice and clean.

src/utils.sh
```diff
@@ -23,4 +25,32 @@ function kworkflow-help()
     "\t                             the mailing list. \"-a\" also\n" \
     "\t                             prints files authors\n" \
     "\thelp,h - displays this help mesage"
+    "\tcodestyle - Apply checkpatch on directory or file\n" \
```
rodrigosiqueira: you introduced a tiny bug here, if you run the tests you will see:

./src/utils.sh: line 29: \tcodestyle - Apply checkpatch on directory or file\n: command not found

Additionally, why you moved the other help descriptions? You just had to add you new entry here. Finally, I think you forgot to rebase with the unstable.

tests/utils_test.sh
```diff
@@ -14,4 +15,16 @@ function testHelp
   [[ $HELP_OUTPUT =~ Usage:\ kw.* ]]; assertTrue "Help text not displaying correctly." $?
 }
 
+MSG_OUT="GNU"
+LOG_OUT="Initial commit"
```
rodrigosiqueira: Why these variables are here, and not inside testExplore?

rodrigosiqueira:
Hi, thanks for your PR.

I made some changes in the commit based on the comments that I made.

Merged, look at:

https://github.com/rodrigosiqueira/kworkflow/commit/3e50c91a00829e73c5ba9eae994698db81e7e5e2

PR #34, opened 2018-10-08
Fix run_tests.sh's test failure check

MatheusBernardino:
Previously run_tests.sh categorized the test files that returned 1 as
failures and all others as success. This made, for example, test files
without execution permission to be considered as successfully 
executed. Now, only the ones that returns 0 are considered success
and all others are considered failures.

Also, the function report_results was printing the number of test
files that succeeded, failed and wasn't found, not the number of
tests. So the strings to be printed were changed from "tests" to "test
files".

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -101,15 +101,15 @@ function report_results
     elif [[ $success -eq $total ]]; then
       success $SEPARATOR
       success "Total: $total test file(s)"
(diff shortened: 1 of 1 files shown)


rodrigosiqueira:
Merged, See:

https://github.com/rodrigosiqueira/kworkflow/commit/3f7364e310c06a811c1a2a63ab86e37e2d4ca67c

PR #37, opened 2018-10-09
Command completion alert

### LINKED ISSUE #30: Add sonorous and visual alert
Sometimes, compilation and installation take time and we want to do something different while we wait for the end of the processes; as a result, we forgot to go back and check the compilation and wast a little bit of time. It could be nice if kw provides an option for making a sonorous and visual alert at the end of some processes.

MatheusBernardino:
This PR adds an option to the user to get notified when commands that may take longer to run complete their execution. The notification can be visual, sonorous or both, accordingly to the option given. If closes #30 .


MatheusBernardino:
In this PR, I used the terminal beep, for better portability, as I thought it would be more portable then relying on a player command such as aplay, play or paplay. But rethinking about it, the terminal beep is, to me, a very annoying sound (I have mine disabled). Don't you think it would be better if we packed a royalty free notification sound with kw and played that sound instead? The only "problem" would be to choose a more portable player... (maybe pulseaudios's paplay or ALSA's aplay)

kw.sh
```diff
@@ -5,9 +5,53 @@ EASY_KERNEL_WORKFLOW=${EASY_KERNEL_WORKFLOW:-"kw"}
 src_script_path=${src_script_path:-"$HOME/.config/$EASY_KERNEL_WORKFLOW/src"}
 external_script_path=${external_script_path:-"$HOME/.config/$EASY_KERNEL_WORKFLOW/external"}
 
+# Notification definitions
+declare -r ALERT_NONE=0
+declare -r ALERT_VISUAL=1
+declare -r ALERT_SOUND=2
+declare -r ALERT_BOTH=3
```
rodrigosiqueira: It is not a good idea to have these variables here since we export all of them. I recommend you to move these variables to common.sh.

kw.sh
```diff
@@ -5,9 +5,53 @@ EASY_KERNEL_WORKFLOW=${EASY_KERNEL_WORKFLOW:-"kw"}
...
+
 # Export external variables required by kworkflow
 export EASY_KERNEL_WORKFLOW
 
+# Alerts command completion to the user. First argument is the alert code,
+# and second is the command string to be printed in visual notification.
+function alert_completion()
```
rodrigosiqueira: Same for this function, move it for common.sh and do not export it for the whole environment

kw.sh
```diff
@@ -5,9 +5,53 @@ EASY_KERNEL_WORKFLOW=${EASY_KERNEL_WORKFLOW:-"kw"}
...
+{
+  local ALERT=$1
+  local COMMAND=$2
+  if [[ $(($ALERT & $ALERT_SOUND)) != 0 ]]; then
+    echo -e "\a\a"
```
rodrigosiqueira: Replace it for something more flexible. Take a look at my main comment in the PR

README.md
```diff
@@ -158,6 +158,22 @@ If you have everything set, just execute the command:
...
+```
+kw prepare --alert=vs
+```
+
+> There are four options to --alert=$: vs|sv, v, s, n.
+> v enables visual notification
+> s enables sound notification
+> vs or sv enables both
+> n disable notifications (default when --alert=$ is not given)
+
```
rodrigosiqueira: Remember to update the README for add paplay as a dependency.

kw.sh
```diff
@@ -5,9 +5,53 @@ EASY_KERNEL_WORKFLOW=${EASY_KERNEL_WORKFLOW:-"kw"}
...
+  fi
+  if [[ $(($ALERT & $ALERT_VISUAL)) != 0 ]]; then
+    notify-send -i checkbox -t 10000 "kw" "Command: \"$2\" completed!"
```
rodrigosiqueira: Do we need to update the README in order to inform this dependency? This command does not work on my environment, but I suppose this is ok since I'm using openbox.
Ps.: Can we think about a terminal visual alert if "notify-send" does not work? Just any idea.

kw.sh
```diff
@@ -5,9 +5,53 @@ EASY_KERNEL_WORKFLOW=${EASY_KERNEL_WORKFLOW:-"kw"}
...
+        ;;
+      vs|sv)
+        return $ALERT_BOTH
+        ;;
+      n)
+        return $ALERT_NONE
+        ;;
+     esac
+  fi
+  return $ALERT_NONE
```
rodrigosiqueira: Take care with the use of return in shell script, the use of it is restricted. Basically, return just inform values from 0-255 which represents the errno code. In shell just use return for return status.
If you want to get some value, I recommend you to do something like this:
```sh
function lala()
{
 ...
tmp="xpto"
echo $tmp
 ...
}
...
value=$(lala)
echo $value
...
```

This approach provide an extra advantage for you: you can save one line and one variable in the kw() function. You can do something like:

```sh
local alert=$(lala $@)
```

However, I'm asking to myself: can we simplify this code?

kw.sh
```diff
@@ -5,9 +5,53 @@ EASY_KERNEL_WORKFLOW=${EASY_KERNEL_WORKFLOW:-"kw"}
...
+    echo -e "\a\a"
+  fi
+  if [[ $(($ALERT & $ALERT_VISUAL)) != 0 ]]; then
+    notify-send -i checkbox -t 10000 "kw" "Command: \"$2\" completed!"
+  fi
+}
+
+# Check the options for "--alert=[v|s|vs|sv|n]" and returns the respectively
+# alert code.
+function check_alert_option()
```
rodrigosiqueira: Again, we don't want to make this function visible for the whole system.

kw.sh
```diff
@@ -46,35 +90,60 @@ function kw()
       (
         . $src_script_path/vm.sh --source-only
 
+        check_alert_option $@
+        local alert=$?
         vm_prepare
+        local ret=$?
+        alert_completion $alert "kw prepare"
```
rodrigosiqueira: Here, after add paplay or any other player, I recommend you to use:

```sh
alert_completition $alert "kw prepare" &
```
with this we detach the execution of the sound, and make things a little bit more "natural".

rodrigosiqueira:
Hi! First of all, thanks for your PR! It is a really nice feature :)

For simplicity sake, I will point out some of the problems in subsections and in the code. 

## Make the player used by kw flexible

As you said, using terminal sound is not good. I suggest you use "paplay" as a default option for two reasons: (1) it is really common to find this tiny player in most of the distros, and (2) we can add it as a dependency. Additionally, I suggest you make the "player" option as a value stored in a variable and by default, this value should be "paplay"; however, give the freedom for users to change it via config file.

Something like this:

0. Create a directory named "sounds"

You can add the following audio as a default:
https://github.com/rodrigosiqueira/fancy-beep/tree/master/sounds

1. in the kw.sh

```sh
...
sounds_path=${sounds_path:-"$HOME/.config/$EASY_KERNEL_WORKFLOW/sounds"}
sounds_app="paplay"
...
eval $sound_app $sound_path
...
```
2. in the setup.sh

```sh
...
declare -r SOUNDS="sounds"
...
rsync -vr $SOUNDS $INSTALLTO
...
```

3. In the kworkflow.conf

```
default_player=paplay
default_sound=PATH
```

## Exposes values

Your patch exports the following variables and functions: check_alert_option(), alert_completion(), ALERT_NONE, ...
This is not good, please do not expose these values. Take a look at the function kw() to get some ideas on how to not expose.

rodrigosiqueira:
I just test the idea of moving all of the alert functions to "common.sh", it worked well because:

1. `. $src_script_path/mk.sh --source-only` -> includes vm.sh -> includes `commons.sh`
2. `. $src_script_path/mk.sh --source-only` -> includes `commons.sh`

It work like a charm for me. Of course, you have to make some changes. I really  recommend to find a way to simplify or get rid of `check_alert_option`

MatheusBernardino:
> I just test the idea of moving all of the alert functions to "common.sh", it worked well because:
> 
>     1. `. $src_script_path/mk.sh --source-only` -> includes vm.sh -> includes `commons.sh`
> 
>     2. `. $src_script_path/mk.sh --source-only` -> includes `commons.sh`

Nice! I checked it here and seems to be a good option. I just thought of another one, let me know what you think: Since the alert functions are only used by kw.sh, putting then at common.sh will, in my view, increase a file that is imported by many others, with some code unused by the majority of them. What do you think about moving those functions to an "alert.sh" file?

> I really recommend to find a way to simplify or get rid of `check_alert_option`

This function is supposed to "parse" the alert command line options, and return a numerical value representing the user's wish (sound alert, visual alert, both or none). But I can try to improve this if it doesn't seem good. What would you like me to improve in this section?

rodrigosiqueira:
@MatheusBernardino will you finish this PR? You're really close to finish it.

MatheusBernardino:
> @MatheusBernardino will you finish this PR? You're really close to finish it.

Yes, I will. I'll need some time, thought, to get along with what we were doing here haha But I'll finish it

kw.sh
MatheusBernardino: Nice! I think it would be better to add the "&" in paplay execution (inside alert_completion function), in order to detach sound execution, directly, don't you think?

src/alert.sh
```diff
@@ -0,0 +1,22 @@
+# NOTE: src/commons.sh must be included before this file
+
+# Alerts command completion to the user. First argument is the alert code,
+# and second is the command string to be printed in visual notification.
```
rodrigosiqueira: Let's start using some rules for this sort of comment. In some way started to use it in the "commons.sh" file, I would like to discuss it with you later. Anyway, how about comment this function like this:

```
# Alerts command completion to the user.
# 
# @parameter First argument is the alert code
# @parameter Second argument is the command string to be printed in visual notification
#
# NOTE: src/commons.sh must be included before this file
```

Also, write two or three phrases that add some detail about this function.

src/alert.sh
```diff
@@ -0,0 +1,22 @@
...
+
+  local COMMAND=$1
+  local ALERT_OPT=$2
+
+  if [[ $# -gt 1 && "$2" =~ ^--alert= ]]; then
+    local opts="$(echo $2 | sed s/--alert=// )"
+
+    grep -o . <<< "$opts" | while read option;  do
+      if [ "$option" == "v" ]; then
+        eval "${configurations[visual_alert_command]}"
```
rodrigosiqueira: We should avoid eval command in the Kw, especially here. Why? The user can set the value loaded by kw, and here live the dangerous of this operation. A malicious user could set a command that "rm" the whole home for example. For extra information take a look at:

https://unix.stackexchange.com/questions/278427/why-and-when-should-eval-use-be-avoided-in-shell-scripts

Anyway, we can replace the eval command by simply call the player. Just remember to add the check if the program is available or not.

src/alert.sh
```diff
@@ -0,0 +1,22 @@
...
+      elif [ "$option" == "s" ]; then
+        eval "${configurations[alert_player]}" "${configurations[alert_sound_path]}" &
```
rodrigosiqueira: Add the "&" to the "v" option as well

rodrigosiqueira:
I made many comments along the patch, take a look on it. I just want to add some extra points here:

1. It could be super cool if we can enable the sound/visual alert by default in the configuration file. Something like this:

```
cat /PATH/kworkflow.config
# Default ssh ip
ssh_ip=localhost
...
# Make alert enable by default if you add v, s, or vs.
# If it's empty, the alert is not enable by default
alert=
```
2. How about move the function in the “alert.sh” to “kwio.sh”? In some way, the alert is some output and make sense to be in “kwio.sh” file.
3. I know this is boring, but how about adding a test for this feature?
4. I’m a little bit worried about this sound license; maybe we should double check it and add a reference to the sound license in the README file.

README.md
```diff
@@ -173,6 +179,22 @@ If you have everything set, just execute the command:
 kw prepare|p
 ```
 
+> Some commands take a considerable amount of time to run. So you can
+> do something else while it runs and get a notification alert when
```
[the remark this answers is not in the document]
MatheusBernardino: Hm, I think it's "get", because it agrees with "you", not "it". "You can [...] and get a notification"

etc/kworkflow.config
```diff
@@ -29,3 +31,14 @@ qemu_net_options=-net nic -net user,hostfwd=tcp::2222-:22,smb=/home/USERKW
...
+# Player to use for command completion alert
+alert_player=paplay
+
+# Sound to play on command completion alert
+alert_sound_path=INSTALLPATH/sounds/complete.oga
+
+# Command to be run for visual completion alert
+# Note: You may use $COMMAND, which will be replaced by the kw command
+#       whose conclusion the user wished to be alerted.
+visual_alert_command=notify-send -i checkbox -t 10000 "kw" "Command: \\"$COMMAND\\" completed!"
```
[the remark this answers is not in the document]
MatheusBernardino: Hmm, I don't get it. The problem is with the $COMMAND variable? I will describe the idea here and if this is a problem, you tell me, ok?

The `visual_alert_command` string will be the command executed by kw to display a visual alert.

The visual alerts informs the user some kw command they ran have completed. One might want to display the kw command (e.g. "kw bi") in the alert text, to know which command has finished. That's what the $COMMAND variable holds.

The $COMMAND usage is a easy way to let the user specify a custom visual alert command and still give him the option to have the string holding the kw command to be printed in the alert. He may not want to use $COMMAND, though, which is totally fine.

src/alert.sh
MatheusBernardino: I didn't understand the problem, can you explain it to me a little further? I mean, if we change `eval $COMMAND_SET_BY_USER` to `$COMMAND_SET_BY_USER`, won't the command, which can be an `rm` as you said, be executed anyway?

README.md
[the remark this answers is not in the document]
rodrigosiqueira: In this case, it could be better if you explicitly add the pronoun 'you'.

src/alert.sh
rodrigosiqueira: Good point here!

I'm worried about this kind of operation because I have plans to package kw for Debian and Arch. If we add kw as a real distro package (e.g., apt install kw), we have to take extra attention to not create a problem for users. In this sense, I want to avoid eval and other sorts of operations that can accidentally/maliciously compromise the system.

Now, I think that could be better if we only support paplay and send-notify. In the future, if someone asks we can add other tools.

README.md
```diff
@@ -173,6 +179,22 @@ If you have everything set, just execute the command:
...
+> they finish. The commands with this feature available are: prepare,
+> build, install, new and bi. To enable the notification, use the
```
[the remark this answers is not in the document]
MatheusBernardino: Can they take that much to run?

src/alert.sh
MatheusBernardino: Ok, I understand your security concern here. But the "malicious user" would be setting a command to run on his own machine. So he can't do harm to anyone but himself...

Is your concern more related to a situation where a ks user is downloading something from the internet that maliciously comes with a "kworkflow.config" file, set with a "rm" command? In this case, the user might run a "kw something" and execute arbitrary code not desired by him. But to try preventing this situation seems a little bit overkill to me. What do you thing? Should I go and remove the customization anyway?

src/alert.sh
rodrigosiqueira: Nice, I did know about that because I'm not using notify-send. Anyway, did you test it?

src/alert.sh
rodrigosiqueira: Ok, you're right. It's just paranoid from my side. Just ignore it, and go ahead.

src/alert.sh
MatheusBernardino: Yes, it's tested :) (but we don't have automatic tests for this feature yet)

MatheusBernardino:
Hi, @rodrigosiqueira . Suggestions and reviews were applied  :) Including alert tests, default alert behavior customizable with kworkflow.config, changed sound to one that I created and moved kwio.sh to alert.sh.

rodrigosiqueira:
Great Job!!!!! I reallyyyyyy liked this patchset.

PR #38, opened 2018-10-11
Make 'kw maintainers' work from any directory

### LINKED ISSUE #28: kw maintainers only work from the first directory of the Linux tree
It could be nice if `kw maintainers` works from any directory. For example:

```
cd drivers/staging/
kw maintainers iio/
```

MatheusBernardino:
This PR makes  'kw maintainers' work  from any directory. It fixes #28 

It does that looking for he kernel tree root, cd-ing to it before running get_maintainers and cd-ing back after.

### DIFF
--- a/src/get_maintainer_wrapper.sh
+++ b/src/get_maintainer_wrapper.sh
@@ -1,90 +1,231 @@
 . $src_script_path/kwio.sh --source-only
 
-
+# Prints the authors of a given file or files inside a given dir.
+#
+# @FILE_OR_DIR The argument is a file or directory path
 function print_files_authors()
 {
-    # TODO: currently only authors found in a single line MODULE_AUTHOR
-    # statement are captured and printed. In the future, it would be
-    # nice to look for multiline MODULE_AUTHOR statatements such as:
-    #
-    # MODULE_AUTHOR ("a_long_email@xx.com" \
-    #                "another_long@yy.com" )
-    # and:
-    #
-    # MODULE_AUTHOR ("a_long_email@xx.com"
-    #                "another_long@yy.com" )
-    #
-    # which are, both, valid C statements. In the currently
-    # implementation neither of these four emails will be printed.
-
-    local FILE_OR_DIR=$1
-    local files=( )
-    if [[ -d $FILE_OR_DIR ]]; then
-        for file in $FILE_OR_DIR/*; do
-            if [[ -f $file ]]; then
-                files+=($file)
-            fi
-        done
-    elif [[ -f $FILE_OR_DIR ]]; then
-        files+=($FILE_OR_DIR)
-    fi
-
-    local printed_authors_separator=false
-
-    for file in ${files[@]}; do
-        authors=$(grep -oE "MODULE_AUTHOR *\(.*\)" $file |
-                  sed -E "s/(MODULE_AUTHOR *\( *\"|\" *\))//g" |
-                  sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/, /g' )
-        if [[ ! -z $authors ]]; then
-            if [ $printed_authors_separator = false ]; then
-                say $SEPARATOR
-                say "MODULE AUTHORS:"
-                printed_authors_separator=true
-            fi
-            say -n "$(basename $file): "
-            echo "$authors"
-        fi
+  local FILE_OR_DIR=$1
+  local files=( )
+  if [[ -d $FILE_OR_DIR ]]; then
+    for file in $FILE_OR_DIR/*; do
+      if [[ -f $file ]]; then
+        files+=($file)
+      fi
     done
+  elif [[ -f $FILE_OR_DIR ]]; then
+    files+=($FILE_OR_DIR)
+  fi
+
+  local printed_authors_separator=false
+
+  for file in ${files[@]}; do
+    authors=$(grep -oE "MODULE_AUTHOR *\(.*\)" $file |
+              sed -E "s/(MODULE_AUTHOR *\( *\"|\" *\))//g" |
+              sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/, /g' )
+    if [[ ! -z $authors ]]; then
+      if [[ $printed_authors_separator = false ]]; then
+        say $SEPARATOR
+        say "MODULE AUTHORS:"
+        printed_authors_separator=true
+      fi
+      say -n "$(basename $file): "
+      echo "$authors"
+    fi
+  done
 }
 
-function execute_get_maintainer()
+# Checks if a directory is a kernel tree root
+#
+# @DIR A directory path
+#
+# Returns:
+# True if given dir is a kernel tree root and false otherwise.
+function is_kernel_root()
 {
-  local FILE_OR_DIR_CHECK
-  local print_authors
+  local -r DIR="$@"
+  # The following files are some of the files expected to be at a linux
+  # tree root and not expected to change. Their presence (or abscense)
+  # is used to tell if a directory is a linux tree root or not. (They
+  # are the same ones used by get_maintainer.pl)
+  if [[ -f "${DIR}/COPYING" &&
+        -f "${DIR}/CREDITS" &&
+        -f "${DIR}/Kbuild" &&
+        -e "${DIR}/MAINTAINERS" &&
+        -f "${DIR}/Makefile" &&
+        -f "${DIR}/README" &&
+        -d "${DIR}/Documentation" &&
+        -d "${DIR}/arch" &&
+        -d "${DIR}/include" &&
+        -d "${DIR}/drivers" &&
+        -d "${DIR}/fs" &&
+        -d "${DIR}/init" &&
+        -d "${DIR}/ipc" &&
+        -d "${DIR}/kernel" &&
+        -d "${DIR}/lib" &&
+        -d "${DIR}/scripts" ]]; then
+    return 0
+  fi
+  return 1
+}
 
-  local -r script="scripts/get_maintainer.pl"
-  local -r options="--separator , --nokeywords --nogit --nogit-fallback --norolestats "
-  local -r getmaintainers="perl $script $options"
+# Finds the root of the linux kernel repo containing the given file
+#
+# @FILE_OR_DIR The argument is a directory of file path
+#
+# Returns:
+# The path of the kernel tree root (string) which the file or dir belongs to, or
+# an empty string if no root was found.
+function find_kernel_root
+{
+  local -r FILE_OR_DIR="$@"
+  local current_dir
+  local kernel_root=""
 
-  # Check if the command was invoked from the kernel root tree
-  if [ ! -f $script ]; then
-    complain "You have to execute this command from the linux tree"
+  if [[ -f "$FILE_OR_DIR" ]]; then
+    current_dir="$(dirname $FILE_OR_DIR)"
+  else
+    current_dir="$FILE_OR_DIR"
+  fi
+
+  # Find the kernel tree root
+  if is_kernel_root "$current_dir"; then
+    kernel_root="$current_dir"
+  else
+    while [[ "$current_dir" != "/" ]]; do
+      current_dir="$(dirname $current_dir)"
+      if is_kernel_root "$current_dir"; then
+        kernel_root="$current_dir"
+        break
+      fi
+    done
+  fi
+
+  echo "$kernel_root"
+}
+
+# Checks if the given path is a patch file
+#
+# @FILE_PATH The argument is the path
+#
+# Returns:
+# True if given path is a patch file and false otherwise.
+function is_a_patch
+{
+  local -r FILE_PATH="$@"
+
+  if [[ ! -f "$FILE_PATH" ]]; then
     return 1
   fi
 
+  local file_content=`cat "$FILE_PATH"`
+
+  # The following array stores strings that are expected to be present
+  # in a patch file. The absence of any of these strings makes the
+  # given file be considered NOT a patch
+  local -ar PATCH_EXPECTED_STRINGS=(
+    "diff --git"
+    "Subject:"
+    "Date:"
+    "From:"
+    "---"
+    "@@"
+  )
+
+  for expected_str in "${PATCH_EXPECTED_STRINGS[@]}"; do
+    if [[ ! "$file_content" =~ "$expected_str" ]]; then
+      return 1
+    fi
+  done
+
+  return 0
+}
+
+# Executes get_maintainer with the given file or dir
+#
+# @1 First argument can be "--authors" or "a" to enable authors
+#    printing. It is optional.
+# @2 Second argument is the given file/dir. It is optional and, if not
+#    given, the current working directory will be used.
+#
+# Returns:
+# False uppon error and true otherwise.
+#
+# This function also handle the cases where:
+# - A patch file is given
+# - The working dir and/or given path is not the root of a linux kernel repo
+# - The working dir and/or given path is not inside a linux kernel repo
+function execute_get_maintainer()
+{
+  local FILE_OR_DIR
+  local print_authors
+  local is_file_a_patch=true
+  local is_file_inside_kernel_tree=true
+
+  local -r script="scripts/get_maintainer.pl"
+  local options="--separator , --nokeywords --nogit --nogit-fallback --norolestats "
+
+  local -r original_working_dir=$PWD
+  local kernel_root=""
+  local path_from_kernel_root=""
+
+  # Check function options
   if [[ $# -ge 1 && ( $1  = "--authors" || $1 = "-a" ) ]]; then
-      FILE_OR_DIR_CHECK=$2
-      print_authors=true
+    FILE_OR_DIR=$2
+    print_authors=true
   else
-      FILE_OR_DIR_CHECK=$1
-      print_authors=false
+    FILE_OR_DIR=$1
+    print_authors=false
+  fi
+
+  # If no file is given, assume "."
+  if [[ -z $FILE_OR_DIR ]]; then
+    FILE_OR_DIR="."
   fi
 
   # Check if is a valid path
-  if [ ! -d $FILE_OR_DIR_CHECK -a ! -f $FILE_OR_DIR_CHECK ]; then
+  if [[ ! -d $FILE_OR_DIR && ! -f $FILE_OR_DIR ]]; then
     complain "Invalid path"
+    return 1
+  fi
+
+  FILE_OR_DIR="$(realpath $FILE_OR_DIR)"
+
+  # if given path is not a patchfile, add -f to get_maintainer.pl options
+  if ! is_a_patch "$FILE_OR_DIR"; then
+    is_file_a_patch=false
+    options="$options -f "
+  fi
+
+  # try to find kernel root at given path
+  kernel_root="$(find_kernel_root $FILE_OR_DIR)"
+  if [[ -z "$kernel_root" ]]; then
+    is_file_inside_kernel_tree=false
+    # fallback: try to find kernel root at working path
+    kernel_root="$(find_kernel_root $original_working_dir)"
+  fi
+
+  # Check if kernel root was found.
+  if [[ -z "$kernel_root" ]]; then
+    complain "Neither the given path nor the working path is in a kernel tree."
+    return 1
+  fi
+
+  # If file is not a patch and outside a kernel tree, it must be an user's
+  # mistake. Although get_maintainer.pl can handle this, it's better to abort
+  # because it is most likely a user's mistake. So better let the user know.
+  if ! $is_file_a_patch && ! $is_file_inside_kernel_tree; then
+    complain "The given file is not a patch and is outside a kernel tree."
     return 1
   fi
 
   say $SEPARATOR
   say "HERE:"
-  $getmaintainers $FILE_OR_DIR_CHECK
+  cd $kernel_root
+  eval perl $script $options "$FILE_OR_DIR"
+  cd $original_working_dir
 
-  if [ $? -ne 0 ]; then
-    return $?
-  fi
-
-  if [ $print_authors = true ]; then
-      print_files_authors $FILE_OR_DIR_CHECK
+  if $print_authors; then
+    print_files_authors $FILE_OR_DIR
   fi
 }
--- a/tests/get_maintainer_wrapper_test.sh
+++ b/tests/get_maintainer_wrapper_test.sh
@@ -11,52 +11,37 @@ function suite
   suite_addTest "testPrintFileAuthorForDir"
 }
 
-# Hold the the lines print_files_authors should print when given the file
-# samples/print_file_author_test_dir directory and
-# samples/print_file_author_test_dir/code1.c file, respectively
-CORRECT_DIR_MSG=(
-    "========================================================="
-    "MODULE AUTHORS:"
-    "code1.c: John Doe <johndoe@community.com>, Caesar Doe <caesar@community.com>, Michael Doe <michael@community.com>"
-    "code2.c: Bob Hilson <bob@opensource.com>"
-)
-CORRECT_FILE_MSG=(
-    "========================================================="
-    "MODULE AUTHORS:"
-    "code1.c: John Doe <johndoe@community.com>, Caesar Doe <caesar@community.com>, Michael Doe <michael@community.com>"
-)
+# The following variables hold the the lines print_files_authors should
+# print when given the file samples/print_file_author_test_dir directory
+# and samples/print_file_author_test_dir/code1.c file, respectively.
+CORRECT_DIR_MSG="=========================================================
+MODULE AUTHORS:
+code1.c: John Doe <johndoe@community.com>, Caesar Doe <caesar@community.com>, Michael Doe <michael@community.com>
+code2.c: Bob Hilson <bob@opensource.com>"
+CORRECT_FILE_MSG="=========================================================
+MODULE AUTHORS:
+code1.c: John Doe <johndoe@community.com>, Caesar Doe <caesar@community.com>, Michael Doe <michael@community.com>"
 
 function testPrintFileAuthorForFile
 {
-  local counter=0
-  while read -r line
-  do
-    local expected=${CORRECT_FILE_MSG[$counter]}
-    if [[ "$line" != "$expected" ]]; then
-        fail "Expecting line $counter to be:\n\"$expected\"\nBut got:\n\"$line\""
-        true # Reset return value
-        return
-    fi
-    ((counter+=1))
-  done < <(print_files_authors "tests/samples/print_file_author_test_dir/code1.c")
+  local -r ret=$(print_files_authors "tests/samples/print_file_author_test_dir/code1.c")
+  if [[ "$ret" != "$CORRECT_FILE_MSG" ]]; then
+    local -r expected_prefixed=$(prefix_multiline "$CORRECT_FILE_MSG")
+    local -r got_prefixed=$(prefix_multiline "$ret")
+    fail "Expecting return:\n$expected_prefixed\nBut got:\n$got_prefixed"
+  fi
   true # Reset return value
 }
 
 function testPrintFileAuthorForDir
 {
-    local counter=0
-    while read -r line
-    do
-      local expected=${CORRECT_DIR_MSG[$counter]}
-      if [[ "$line" != "$expected" ]]; then
-          fail "Expecting line $counter to be:\n\"$expected\"\nBut got:\n\"$line\""
-          true # Reset return value
-          return
-      fi
-      ((counter+=1))
-    done < <(print_files_authors "tests/samples/print_file_author_test_dir")
-    true # Reset return value
+  local -r ret=$(print_files_authors "tests/samples/print_file_author_test_dir")
+  if [[ "$ret" != "$CORRECT_DIR_MSG" ]]; then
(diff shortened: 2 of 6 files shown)


src/get_maintainer_wrapper.sh
```diff
@@ -47,6 +47,31 @@ function print_files_authors()
     done
 }
 
+function is_linux_tree_root()
```
rodrigosiqueira: how about put this function on "commons.sh"? This can be useful for a new version of the function handler for checkpatch (see #40)

MatheusBernardino: Seems good! But don't you think it would be nicer to leave the function here, for now, and move it in a PR for #40 ? (Just thinking about that idea you told us of "Do what is sufficient for now, not for the future", you know?)

MatheusBernardino:
Rebased and added tests, as requested. Also, improved get_maintainers.pl behavior to handle even more corner cases, as described in the commit message.

MatheusBernardino:
> First of all, thanks for your PR! I really enjoyed reviewing this one :-)
> 
> I just have two comments:
> 
>     1. Rebase you code with the branch "unstable"
> 
>     2. Could you add some tests?

Rebased and added tests, as requested. Also, improved get_maintainers.pl behavior to handle even more corner cases, as described in the commit message.

src/get_maintainer_wrapper.sh
```diff
@@ -47,40 +48,168 @@ function print_files_authors()
     done
 }
 
+# Returns true if given dir is a kernel tree root. Argument is the dir.
```
rodrigosiqueira: Use spaces instead of tabs.

src/get_maintainer_wrapper.sh
```diff
@@ -47,40 +48,168 @@ function print_files_authors()
...
+{
+  local -r DIR="$@"
+  # The following files are some of the files expected to be at a linux
+  # tree root and not expected to change. Their presence (or abscense)
+  # is used to tell if a directory is a linux tree root or not. (They
+  # are the same ones used by get_maintainer.pl)
+  if [[ -f "${DIR}/COPYING" &&
+  		-f "${DIR}/CREDITS" &&
+		-f "${DIR}/Kbuild" &&
+		-e "${DIR}/MAINTAINERS" &&
```
rodrigosiqueira: Why you use '-e' instead of '-f' here?

src/get_maintainer_wrapper.sh
```diff
@@ -47,40 +48,168 @@ function print_files_authors()
...
+  fi
+  # Check if kernel root was found.
+  if [ -z "$kernel_root" ]; then
+		complain "Neither the given path nor the working path is in a kernel tree."
+		return 1
+  fi
+
+	# If file is not a patch and outside a kernel tree, it must be an user's
+	# mistake. Although get_maintainer.pl can handle this, and would print
+	# the kernel root maintainer, linux torvalds, it's better to abort
```
rodrigosiqueira: I did not understood what you meant by "linux torvalds", but in any case, I prefer no name references in the code.

tests/get_maintainer_wrapper_test.sh
```diff
@@ -3,60 +3,145 @@
...
+# respectively.
+CORRECT_TMP_MSG="=========================================================
+HERE:
+Jane Doe <jane@email.com>,kernel@list.org"
+CORRECT_TMP_FS_MSG="=========================================================
+HERE:
+John Doe <john@email.com>,Jane Doe <jane@email.com>,fs@list.org,kernel@list.org"
+
+
+
```
rodrigosiqueira: get rid of the extras spaces. Just one line is enough.

rodrigosiqueira:
Hi, first of all: thanks a lot for your PR :)

We are close to finish this. I have some comments:

1. Try command below produce an infinity loop:

```
kw m ../la/lu.txt
```

2. Take care with tabs, and white spaces. Take a look in your editor (I always recommend use vim ;). We use space as default in kworkflow.

3. Keep the 'if' condition patterns. Sometimes you use '[]', '[[ ]]' and sometimes you use nothing. I recommend the use of '[[ ]]' every time. (Actually... I'm surprised that any error message was not pop up)

4. I really liked these functions: is_kernel_root(), find_kernel_root(), and is_a_patch(). It could be better to move all of them to `commons.sh` as the first commit in this patchset. Then, you replace the include of `miscellaneous.sh` by `commons.sh`. These functions could be used to improve the `checkpatch` in the future, because of this I recommend you to replace this in the `commons`.

5. There is some weird behavior associated with the new tests. I executed the test two or three times, and I got the following error:

```
Running test [get_maintainer_wrapper_test]
testPrintFileAuthorForFile
testPrintFileAuthorForDir
testLinuxRootCheck
testGetMaintainers
./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/' doesn't appear to be a patch.  Add -f to options?
ASSERT:Expecting return:\n> =========================================================
> HERE:
> Jane Doe <jane@email.com>,kernel@list.org\nBut got:\n> =========================================================
> HERE:\n(Calling lkr from outside lkr)
./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/' doesn't appear to be a patch.  Add -f to options?
ASSERT:Expecting return:\n> =========================================================
> HERE:
> Jane Doe <jane@email.com>,kernel@list.org\nBut got:\n> =========================================================
> HERE:\n(Calling lkr from lkr)
./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/fs/' doesn't appear to be a patch.  Add -f to options?
ASSERT:Expecting return:\n> =========================================================
> HERE:
> John Doe <john@email.com>,Jane Doe <jane@email.com>,fs@list.org,kernel@list.org\nBut got:\n> =========================================================
> HERE:\n(Calling lkr/fs from lkr)
./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/' doesn't appear to be a patch.  Add -f to options?
ASSERT:Expecting return:\n> =========================================================
> HERE:
> Jane Doe <jane@email.com>,kernel@list.org\nBut got:\n> =========================================================
> HERE:\n(Calling lkr from lkr/fs)
./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/fs/' doesn't appear to be a patch.  Add -f to options?
ASSERT:Expecting return:\n> =========================================================
> HERE:
> John Doe <john@email.com>,Jane Doe <jane@email.com>,fs@list.org,kernel@list.org\nBut got:\n> =========================================================
> HERE:\n(Calling lkr/fs from lkr/fs)

Ran 4 tests.

FAILED (failures=5)
```
However... for some reason, it started to pass.

src/get_maintainer_wrapper.sh
MatheusBernardino: Oh, nice question. The get_maintainer.pl script uses this exactly same tests to check for a kernel root. So I used '-e' instead of '-f' to keep this function consistent with get_maintainer.pl

MatheusBernardino: Oh, nice question. The get_maintainer.pl script uses this exactly same tests to check for a kernel root. So I used '-e' instead of '-f' to keep this function consistent with get_maintainer.pl.

MatheusBernardino: Oh, nice question. The get_maintainer.pl script uses this exactly same tests to check for a kernel root. So I used '-e' instead of '-f' to keep this function consistent with get_maintainer.pl.

MatheusBernardino:
Hi, @rodrigosiqueira . I finally got back to this PR (really sorry for such a long delay). I corrected most of the things but there are some I want a new feedback from you, since it has been such a long time since the review.

>     1. Try command below produce an infinity loop:
> ```
> kw m ../la/lu.txt
> ```

I couldn't reproduce it here. Can you make me a more concrete example?

>     2. Keep the 'if' condition patterns. Sometimes you use '[]', '[[ ]]' and sometimes you use nothing. I recommend the use of '[[ ]]' every time. (Actually... I'm surprised that any error message was not pop up)

Ok, I can replace the '[]' by '[[]]' (since there seems to be a compatibility), but as for the `if ! $bla;`, I would have to use something like `if [[ $bla -ne true ]];`, which seems to complicate the statement. Do you think it's a good idea? Or do you know if there is any other way to do a "if not" using '[[]]'?

Also, I'm reading a lot of people recommending to use '[]' instead of '[[]]', since the last is bash-only and the first is POSIX (although '[[]]' seems to be a great enhancement over '[]'). Would we like to make kw more portable in the near future, so that it can be installed on zsh and others?

>     3. I really liked these functions: is_kernel_root(), find_kernel_root(), and is_a_patch(). It could be better to move all of them to `commons.sh` as the first commit in this patchset. Then, you replace the include of `miscellaneous.sh` by `commons.sh`. These functions could be used to improve the `checkpatch` in the future, because of this I recommend you to replace this in the `commons`.
> 

Since the review, we changed a lot of things related to `commons` and `miscellaneous` (the latter doesn't even exist anymore), but the overall idea is still doable. Even so, `commons` became a file containing exclusively local/global configurations handlers. And, IMO, it would be a very good design to leave it that way, i.e., not to mix other things in this file (I would even suggest to rename `commons` to `config_handler`, or something like this). I know `find_kernel_root` may be used in another PR to fix issue #40, but IMO, we should leave it at `get_maintainger_wrapper` for now, so we can get this patch merged, and when someone (possibly me :) goes to fix #40, we discuss were to move this function to. What do you think?

>     4. There is some weird behavior associated with the new tests. I executed the test two or three times, and I got the following error:
> 
> 
> ```
> Running test [get_maintainer_wrapper_test]
> =========================================================
> testPrintFileAuthorForFile
> testPrintFileAuthorForDir
> testLinuxRootCheck
> testGetMaintainers
> ./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
> scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/' doesn't appear to be a patch.  Add -f to options?
> ASSERT:Expecting return:\n> =========================================================
> > HERE:
> > Jane Doe <jane@email.com>,kernel@list.org\nBut got:\n> =========================================================
> > HERE:\n(Calling lkr from outside lkr)
> ./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
> scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/' doesn't appear to be a patch.  Add -f to options?
> ASSERT:Expecting return:\n> =========================================================
> > HERE:
> > Jane Doe <jane@email.com>,kernel@list.org\nBut got:\n> =========================================================
> > HERE:\n(Calling lkr from lkr)
> ./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
> scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/fs/' doesn't appear to be a patch.  Add -f to options?
> ASSERT:Expecting return:\n> =========================================================
> > HERE:
> > John Doe <john@email.com>,Jane Doe <jane@email.com>,fs@list.org,kernel@list.org\nBut got:\n> =========================================================
> > HERE:\n(Calling lkr/fs from lkr)
> ./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
> scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/' doesn't appear to be a patch.  Add -f to options?
> ASSERT:Expecting return:\n> =========================================================
> > HERE:
> > Jane Doe <jane@email.com>,kernel@list.org\nBut got:\n> =========================================================
> > HERE:\n(Calling lkr from lkr/fs)
> ./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
> scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/fs/' doesn't appear to be a patch.  Add -f to options?
> ASSERT:Expecting return:\n> =========================================================
> > HERE:
> > John Doe <john@email.com>,Jane Doe <jane@email.com>,fs@list.org,kernel@list.org\nBut got:\n> =========================================================
> > HERE:\n(Calling lkr/fs from lkr/fs)
> 
> Ran 4 tests.
> 
> FAILED (failures=5)
> ```
> 
> However... for some reason, it started to pass.

Again, I couln't reproduce it here :( Was it just the first time you executed the test? Do you have any clue of what happened or how I may be able to replicate it? (By the output, it may be something related to the '[]' / '[[]]' / 'nothing' thing, we talked earlier.)

MatheusBernardino:
> I just did:
> 
> ```
> cd /tmp
> kw m ../la/lu.txt
> ```
> 
> I tried it again, and I still get this look.
> 

Oh, now I was able to replicate! I was creating the `la` and `lu` dirs and `lu.txt` file, but the infinity loop happens exactly when the dirs/files does not exits. Thanks, I'll fix that!

> For me using [[ -eq ]] is totally fine. It is just a syntax pattern. Anyway, you can try other alternatives. 

Ok. So can I use `if [[ $BLA -eq true]]` to test wether a variable is true? It stills seems a little too verbose to me, since we could just use `if $BLA` (But I really don't know about the disadvantages of using it. Maybe the way you suggested, with '[[]]' is more secure?).

> How about using one of these patterns:
> 
> ```
> command && another_command
> command || other_command
> ```

Yes, I'm already using this patterns, as in `if ! $is_file_a_patch && ! $is_file_inside_kernel_tree; then`, for example. But I understood you prefer not to have if's without '[]' or '[[]]', is that right? Or should I leave if's like the above as they are?  


> Yes, I know. However, take a look at the top of our files, and you’ll see “#!/bin/bash” (i.e., we already rely on bash). Additionally, it is a real pain to make a big set os shell scripts portable, and IMHO we should avoid it. Finally, the shebang forces the current terminal to use bash.
> 

I know we use `#!/bin/bash`, bash style functions and other things that would have to me changed for a POSIX kw, but just wanted to know wether that's a plan for the future or not. I agree with you that it would be too much hard work and it's not worthy for now.

> > > ```
> > > 4. There is some weird behavior associated with the new tests. I executed the test two or three times, and I got the following error:
> > > ```
> > > ```
> > > Running test [get_maintainer_wrapper_test]
> > > =========================================================
> > > testPrintFileAuthorForFile
> > > testPrintFileAuthorForDir
> > > testLinuxRootCheck
> > > testGetMaintainers
> > > ./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
> > > scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/' doesn't appear to be a patch.  Add -f to options?
> > > ASSERT:Expecting return:\n> =========================================================
> > > > HERE:
> > > > Jane Doe <jane@email.com>,kernel@list.org\nBut got:\n> =========================================================
> > > > HERE:\n(Calling lkr from outside lkr)
> > > ./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
> > > scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/' doesn't appear to be a patch.  Add -f to options?
> > > ASSERT:Expecting return:\n> =========================================================
> > > > HERE:
> > > > Jane Doe <jane@email.com>,kernel@list.org\nBut got:\n> =========================================================
> > > > HERE:\n(Calling lkr from lkr)
> > > ./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
> > > scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/fs/' doesn't appear to be a patch.  Add -f to options?
> > > ASSERT:Expecting return:\n> =========================================================
> > > > HERE:
> > > > John Doe <john@email.com>,Jane Doe <jane@email.com>,fs@list.org,kernel@list.org\nBut got:\n> =========================================================
> > > > HERE:\n(Calling lkr/fs from lkr)
> > > ./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
> > > scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/' doesn't appear to be a patch.  Add -f to options?
> > > ASSERT:Expecting return:\n> =========================================================
> > > > HERE:
> > > > Jane Doe <jane@email.com>,kernel@list.org\nBut got:\n> =========================================================
> > > > HERE:\n(Calling lkr from lkr/fs)
> > > ./src/get_maintainer_wrapper.sh: line 178: [: is_a_patch: unary operator expected
> > > scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kworkflow/tests/.tmp/fs/' doesn't appear to be a patch.  Add -f to options?
> > > ASSERT:Expecting return:\n> =========================================================
> > > > HERE:
> > > > John Doe <john@email.com>,Jane Doe <jane@email.com>,fs@list.org,kernel@list.org\nBut got:\n> =========================================================
> > > > HERE:\n(Calling lkr/fs from lkr/fs)
> > > 
> > > Ran 4 tests.
> > > 
> > > FAILED (failures=5)
> > > ```
> > > However... for some reason, it started to pass.
> > 
> > 
> > Again, I couln't reproduce it here :( Was it just the first time you executed the test? Do you have any clue of what happened or how I may be able to replicate it? (By the output, it may be something related to the '[]' / '[[]]' / 'nothing' thing, we talked earlier.)
> 
> Sorry, I don't remember about this as well. I recommend you to rebase your branch with the unstable branch, and send a new version. I will test it again ;)

The branch in this PR is already rebased with unstable :) If you can test and give me some feedback on this test error, I would really appreciate :)

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
+  local FILE_OR_DIR=$1
+  local files=( )
+  if [[ -d $FILE_OR_DIR ]]; then
+    for file in $FILE_OR_DIR/*; do
+      if [[ -f $file ]]; then
+        files+=($file)
+      fi
+    done
+  elif [[ -f $FILE_OR_DIR ]]; then
+    files+=($FILE_OR_DIR)
```
rodrigosiqueira: How about adding an else operation for the case that something weird is passed in $FILE_OR_DIR? Another question, how about check if the user has permission to the directory and file?
IMHO this can be good to provide precise information for users.

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
 . $src_script_path/kwio.sh --source-only
 
-
+# Prints the authors of a given file or files inside a given dir.
+#
+# @parameter The argument is a file or directory path
```
rodrigosiqueira: How about use "@FILE_OR_DIR The ..." instead of @parameter? The Kernel documentation use this style.

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
-    local files=( )
-    if [[ -d $FILE_OR_DIR ]]; then
-        for file in $FILE_OR_DIR/*; do
-            if [[ -f $file ]]; then
-                files+=($file)
-            fi
-        done
-    elif [[ -f $FILE_OR_DIR ]]; then
-        files+=($FILE_OR_DIR)
+  # TODO: currently only authors found in a single line MODULE_AUTHOR
```
rodrigosiqueira: How about create an issue for this?

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
+      fi
+      say -n "$(basename $file): "
+      echo "$authors"
     fi
+  done
+}
+
+# Checks if a directory is a kernel tree root
+#
+# @parameter A directory path
```
rodrigosiqueira: How about "@DIR"?

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
+        -d "${DIR}/Documentation" &&
+        -d "${DIR}/arch" &&
+        -d "${DIR}/include" &&
+        -d "${DIR}/drivers" &&
+        -d "${DIR}/fs" &&
+        -d "${DIR}/init" &&
+        -d "${DIR}/ipc" &&
+        -d "${DIR}/kernel" &&
+        -d "${DIR}/lib" &&
+        -d "${DIR}/scripts" ]]; then
```
rodrigosiqueira: I’m a little bit worried about this part since it makes the “kw m” fragile. If the Linux Kernel changes a single file, “kw m” will fail. Also, this hardcoded option could be a little bit more flexible.

My suggestions:
1. Add three or four files for checking if the directory is a kernel root or not. I suggest you inspect for get_maintainers.pl, MAINTAINERS, codestyle.pl, and kernel directory.
2. Add a second if, that loads an extra variable from etc/kworkflow.config. This variable could have other files specified by the user; with this, if ‘kw m’ fails to match the root kernel, the user can add a fine-grained level for an extra check.

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
-                say $SEPARATOR
-                say "MODULE AUTHORS:"
-                printed_authors_separator=true
-            fi
-            say -n "$(basename $file): "
-            echo "$authors"
-        fi
+# Finds the root of the linux kernel repo containing the given file
+#
+# @parameter The argument is a directory of file path
```
rodrigosiqueira: @FILE_OR_DIR

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
+  local kernel_root=""
+
+  if [[ -f "$FILE_OR_DIR" ]]; then
+    current_dir="$(dirname $FILE_OR_DIR)"
+  else
+    current_dir="$FILE_OR_DIR"
+  fi
+
+  # Find the kernel tree root
+  if is_kernel_root "$current_dir"; then
```
rodrigosiqueira: ```if [[ $(is_kernel_root "$current_dir") ]]; then``` ?

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
+    kernel_root="$current_dir"
+  else
+    while [[ "$current_dir" != "/" ]]; do
+      current_dir="$(dirname $current_dir)"
+      if is_kernel_root "$current_dir"; then
```
rodrigosiqueira: Use [[ ]]?

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
+        kernel_root="$current_dir"
+        break
+      fi
     done
+  fi
```
rodrigosiqueira: newline

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
 
+# Checks if the given path is a patch file
+#
+# @parameter The argument is the path
+#
+# Returns:
+# True if given path is a patch file and false otherwise.
+function is_a_patch
+{
+  local -r FILE_PATH="$@"
```
rodrigosiqueira: newline

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
+        break
+      fi
     done
+  fi
+  echo "$kernel_root"
 }
 
+# Checks if the given path is a patch file
+#
+# @parameter The argument is the path
```
rodrigosiqueira: @FILE_PATH

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
+    "diff --git"
+    "Subject:"
+    "Date:"
+    "From:"
+    "deletions(-)"
+    "insertions(+)"
+  )
+
+  for expected_str in "${PATCH_EXPECTED_STRINGS[@]}"; do
+    if ! grep -F "$expected_str" "$FILE_PATH" &> /dev/null; then
```
rodrigosiqueira: [[ ]]

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
+  done
+
+  return 0
+}
+
+# Executes get_maintainer with the given file or dir
+#
+# @parameter First argument can be "--authors" or "a" to enable authors
+#            printing. It is optional.
+# @parameter Second argument is the given file/dir. It is optional and, if not
```
rodrigosiqueira: Just @FILE_OR_DIR and merge the comments?

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
 
-  say $SEPARATOR
-  say "HERE:"
-  $getmaintainers $FILE_OR_DIR_CHECK
+  FILE_OR_DIR="$(realpath $FILE_OR_DIR)"
 
-  if [ $? -ne 0 ]; then
-    return $?
+  # if given path is not a patchfile, add -f to get_maintainer.pl options
+  if ! is_a_patch "$FILE_OR_DIR"; then
```
rodrigosiqueira: [[ ]]

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
 
-  if [ $print_authors = true ]; then
-      print_files_authors $FILE_OR_DIR_CHECK
+  # try to find kernel root at given path
+  kernel_root="$(find_kernel_root $FILE_OR_DIR)"
+  if [[ -z "$kernel_root" ]]; then
+    is_file_inside_kernel_tree=false
+    # fallback: try to find kernel root at working path
+    kernel_root="$(find_kernel_root $original_working_dir)"
+  fi
```
rodrigosiqueira: newline

src/get_maintainer_wrapper.sh
```diff
@@ -1,90 +1,240 @@
...
+  # Check if kernel root was found.
+  if [[ -z "$kernel_root" ]]; then
+    complain "Neither the given path nor the working path is in a kernel tree."
+    return 1
+  fi
+
+  # If file is not a patch and outside a kernel tree, it must be an user's
+  # mistake. Although get_maintainer.pl can handle this, it's better to abort
+  # because it is most likely a user's mistake. So better let the user know.
+  if ! $is_file_a_patch && ! $is_file_inside_kernel_tree; then
```
rodrigosiqueira: [[ ]]

tests/get_maintainer_wrapper_test.sh
```diff
@@ -3,60 +3,141 @@
...
+  local -r expected_tmp_prefixed=$(prefix_multiline "$CORRECT_TMP_MSG")
+  local -r expected_tmp_fs_prefixed=$(prefix_multiline "$CORRECT_TMP_FS_MSG")
+
+  setupGetMaintainers
+
+  ret="$(execute_get_maintainer tests/.tmp)"
+  if [[ "$ret" != "$CORRECT_TMP_MSG" ]]; then
+    got_prefixed=$(prefix_multiline "$ret")
+    fail "Expecting return:\n$expected_tmp_prefixed\nBut got:\n$got_prefixed\n(Calling lkr from outside lkr)"
+  fi
```
rodrigosiqueira: newline

tests/get_maintainer_wrapper_test.sh
```diff
@@ -3,60 +3,141 @@
...
+  cd tests/.tmp
+  ret="$(execute_get_maintainer .)"
+  if [[ "$ret" != "$CORRECT_TMP_MSG" ]]; then
+    got_prefixed=$(prefix_multiline "$ret")
+    fail "Expecting return:\n$expected_tmp_prefixed\nBut got:\n$got_prefixed\n(Calling lkr from lkr)"
+  fi
```
rodrigosiqueira: newline

tests/get_maintainer_wrapper_test.sh
```diff
@@ -3,60 +3,141 @@
...
+  ret="$(execute_get_maintainer fs)"
+  if [[ "$ret" != "$CORRECT_TMP_FS_MSG" ]]; then
+    got_prefixed=$(prefix_multiline "$ret")
+    fail "Expecting return:\n$expected_tmp_fs_prefixed\nBut got:\n$got_prefixed\n(Calling lkr/fs from lkr)"
+  fi
```
rodrigosiqueira: newline

tests/get_maintainer_wrapper_test.sh
```diff
@@ -3,60 +3,141 @@
...
+  cd fs
+  ret="$(execute_get_maintainer ..)"
+  if [[ "$ret" != "$CORRECT_TMP_MSG" ]]; then
+    got_prefixed=$(prefix_multiline "$ret")
+    fail "Expecting return:\n$expected_tmp_prefixed\nBut got:\n$got_prefixed\n(Calling lkr from lkr/fs)"
+  fi
```
rodrigosiqueira: newline

src/get_maintainer_wrapper.sh
MatheusBernardino: Ok, created and comment removed from code.

MatheusBernardino:
@rodrigosiqueira All changes made! :)

PR #44, opened 2018-10-29
Add unit test for check_local_configuration

MatheusBernardino:
This PR adds a unit test for the function check_local_configuration of src/commons.sh.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ tests/shunit2
 .checkpatch-camelcase.git.
 .checkpatch-camelcase.date.*
 tests/external
+tests/.tmp_commons_test
--- a/tests/commons_test.sh
+++ b/tests/commons_test.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+. ./tests/utils --source-only
+
+function suite
+{
+  suite_addTest "testConfigFileNotFound"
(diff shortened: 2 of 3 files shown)


rodrigosiqueira:
Merged, see:

https://github.com/rodrigosiqueira/kworkflow/commit/1e9d4393428afc2866eb8b812c662fb2a0f3c696

Ps.: I made some changes in the original patch

MatheusBernardino:
> Merged, see:
> 
> [1e9d439](https://github.com/rodrigosiqueira/kworkflow/commit/1e9d4393428afc2866eb8b812c662fb2a0f3c696)
> 
> Ps.: I made some changes in the original patch

@rodrigosiqueira , using `pushd/popd` instead of `cd` causes the directories to be printed when running the test. Don't you think that pollutes the output?

MatheusBernardino:
> Merged, see:
> 
> [1e9d439](https://github.com/rodrigosiqueira/kworkflow/commit/1e9d4393428afc2866eb8b812c662fb2a0f3c696)
> 
> Ps.: I made some changes in the original patch

@rodrigosiqueira , using `pushd/popd` instead of `cd` causes the directories to be printed when running the test. Don't you think that pollutes a little the output?

Issue #5, opened 2018-08-27
Add unit test


rodrigosiqueira:
There is no unit test in the kworkflow. Add the basic infrastructure using shunit2, follow some suggestion:

1) If you never used shunit2, take a look at: https://github.com/rodrigosiqueira/flextrace/tree/master/tests and https://github.com/rodrigosiqueira/flextrace/blob/master/run_tests.sh
2) Test functions that handle kworkflow.config 
3) Test the build option
4) Test codestyle
5) Test maintainers

PR #45, opened 2018-10-29
Adds test for kw up

RenatoGeh:
This patch adds a unit test for function src/vm.sh:vm_up.

Signed-off-by: Renato Lui Geh <renatogeh@gmail.com>
Signed-off-by: Gabriel Capella <gabriel@capella.pro>

---

Somewhat related to #5.

### DIFF
--- a/tests/vm_test.sh
+++ b/tests/vm_test.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+. ./src/vm.sh --source-only
+. ./tests/utils --source-only
+
+# TODO: unit test for execute_get_maintainer
+
+function suite
+{
+  suite_addTest "testVmUp"
+}
+
+function testVmUp
+{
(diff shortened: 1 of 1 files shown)


rodrigosiqueira:
This test is really cool as an integration testing but not suitable as a unit test because it depends on an external file (the vm image). Additionally, we have to install kw to make the test work. We could use it in the future when we add the structure for it (how about creating an issue about it?). Anyway, the correct way to make this unit test is mocking the result of:

```
$QEMU ${configurations[qemu_hw_options]} \
        ${configurations[qemu_net_options]} \
        ${configurations[qemu_path_image]}
```

Find a way to intercept the final output of the above code, and make some tests on it. Try some variations...

rodrigosiqueira:
Du to the lack of update, I closed this PR. Thanks for the contribution

Issue #41, opened 2018-10-18
Refactor miscellaneous.sh, commons.sh and utils.sh


MatheusBernardino:
In the src folder, kw has a commons.sh, a miscellaneous.sh and a utils.sh, which can make a little difficult to understand the project structure for a newcomer. Can't we refactor the code to join all these files, or rename them with more meaningful names?

(If you chose to make this, keep in mind that tests uses miscellaneous.sh)

rodrigosiqueira:
You are totally right about that. What do you think about this idea to fix the problem:

1. Move the code from utils.sh to commons (the helper)
2. Get rid of utils.sh
3. Rename miscellaneous.sh to something like kwio.sh
4. Update the required places

Do you want to work on this?

PR #48, opened 2018-10-31
Refactor utils, miscellaneous and commons

MatheusBernardino:
This PR closes issue #41. We made the modifications suggested by @rodrigosiqueira , with an exception: Instead of moving utils.sh's functions to commons.sh, we splited utils.sh in explore.sh and help.sh. We thought this structure would be nicer because commons.sh is used by many files, but the function in utils.sh are only used by kw.sh.


rodrigosiqueira:
Merged:

https://github.com/rodrigosiqueira/kworkflow/commit/d72712dce03ca3329374904568024b16c5986762

I made some tiny modifications

PR #52, opened 2018-11-14
Add 'kw-completion'

shayenne:
Add autocomplete for kw commands whole words.
Update setup.sh to add source for the new file.

Signed-off-by: Shayenne da Luz Moura <shayenneluzmoura@gmail.com>

### DIFF
--- a/kw-completion.sh
+++ b/kw-completion.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+complete -W "export build install prepare new ssh mail mount
+	     umount boot vars up codestyle maintainers help" kw
--- a/kw-completion.sh
+++ b/kw-completion.sh
@@ -1,3 +1,0 @@
-#!/bin/bash
-complete -W "export build install prepare new ssh mail mount
-	     umount boot vars up codestyle maintainers help" kw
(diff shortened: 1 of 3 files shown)


kw-completion.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+complete -W "export build install prepare new ssh mail mount
+	     umount boot vars up codestyle maintainers help" kw
```
rodrigosiqueira: I recommend you to put this line in the file "kw.sh" on top of "function kw()", with this change you can get rid of the file "kw-completion.sh". Additionally, I recommend you add the short commands in the list as well.

rodrigosiqueira:
First of all, thanks for your patch! This is a really nice pacth :)

Follows my comments in topics:

* You don't need the kw-completion.sh, take a look on my inline comment.
* Could you check if the "comple" command is a default? If not, could you update the readme by add this dependency?
* Update your commit message because it is not clear. The message "Add 'kw-completion", probably could be replaced by something like "Add autocomplete for kw". Finally, describe a little the changes in the commit message.

kw-completion.sh
shayenne: Okay! Regarding the complete command, I found that it is a builtin function. Does it need to add in the dependence?

shayenne: https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html

rodrigosiqueira:
Merged:

https://github.com/rodrigosiqueira/kworkflow/commit/bc43e82dfa46f4b7830d017a4bda7bffc3f5610d

PR #54, opened 2019-01-11
Rework the config file manipulation

rodrigosiqueira:
Kworkflow has a messed mechanism for handling global and local variable.
Additionally, there are too many global variables that make things
complicated. This patch uses a single place to store all the global
variables in the same fashion of Unix systems; under
“$HOME/.config/kw/” it is added the directory “etc/” with a global
“kworkflow.config”. Finally, this patch also takes care of the local
config file.

### DIFF
--- a/etc/kworkflow.config
+++ b/etc/kworkflow.config
@@ -1,0 +1,29 @@
+# Default configuration file
+# Attention: the string USERKW will be replaced by the current user name if you
+# use "./setup.sh". You can manually change this behavior.
+
+# Default ssh ip
+ssh_ip=localhost
+
+# Default ssh port
+ssh_port=2222
+
+# Specify the default architecture used by KW
+arch=x86_64
+
+# Defines the virtualization tool that should be used by kw. Current, we only
+# support QEMU
+virtualizer=qemu-system-x86_64
+
+# Defines the kw mount point, this directory is used by libguestfs during the
+# mount/umount operation of a VM
+mount_point=/home/USERKW/p/mount
+
+# Sets basic QEMU options
+qemu_hw_options=-enable-kvm -daemonize -smp 2 -m 1024
+
+# Defines the network configuration
+qemu_net_options=-net nic -net user,hostfwd=tcp::2222-:22,smb=/home/USERKW
+
+# Specify the VM image path
+qemu_path_image=/home/USERKW/p/virty.qcow2
--- a/src/commons.sh
+++ b/src/commons.sh
@@ -1,63 +1,45 @@
 . $src_script_path/kwio.sh --source-only
 
 BASE=$HOME/p/linux-trees
-MOUNT_POINT=$HOME/p/mount
 BUILD_DIR=$BASE/build-linux
-
-QEMU_ARCH="x86_64"
-QEMU="qemu-system-${QEMU_ARCH}"
-QEMU_OPTS="-enable-kvm -smp 2 -m 1024"
-VDISK="$HOME/p/virty.qcow2"
-QEMU_MNT="/mnt/qemu"
-DEFAULT_PORT="2222"
-DEFAULT_IP="127.0.0.1"
 
 TARGET="qemu"
 
-BASHPATH=/bin/bash
-
 # Default configuration
-declare -A configurations=( ["ip"]="127.0.0.1" ["port"]="2222" )
+declare -A configurations
 
+# This function is used to show the current set up used by kworkflow.
 function show_variables()
 {
-  say "Global values:"
+  local has_local_config_path="No"
 
-  echo -e "\tBASE: $BASE"
-  echo -e "\tBUILD_DIR: $BUILD_DIR"
-  echo -e "\tQEMU ARCH: $QEMU_ARCH"
-  echo -e "\tQEMU COMMAND: $QEMU"
-  echo -e "\tQEMU MOUNT POINT: $QEMU_MNT"
-  echo -e "\tTARGET: $TARGET"
-
-  check_local_configuration
-
-  if [ $? -eq 1 ] ; then
-    say "There is no kworkflow.conf, adopt default values for:"
-    echo -e "\tQEMU OPTIONS: $QEMU_OPTS"
-    echo -e "\tVDISK: $VDISK"
+  if [ -f "$PWD/kworkflow.config" ] ; then
+    has_local_config_path="Yes"
   else
-    say "kw found a kworkflow.conf file. Read options:"
-    echo -en "\tQEMU OPTIONS: ${configurations[qemu_hw_options]}"
-    echo     "${configurations[qemu_net_options]}"
-    echo -e "\tVDISK: ${configurations[qemu_path_image]}"
+    has_local_config_path="No"
   fi
+
+  say "Variables:"
+  echo -e "\tLocal config file: $has_local_config_path"
+  echo -e "\tTarget arch: ${configurations[arch]}"
+  echo -e "\tMount point: ${configurations[mount_point]}"
+  echo -e "\tVirtualization tool: ${configurations[virtualizer]}"
+  echo -e "\tQEMU options: ${configurations[qemu_hw_options]}"
+  echo -e "\tQEMU Net options: ${configurations[qemu_net_options]}"
+  echo -e "\tVdisk: ${configurations[qemu_path_image]}"
 }
 
-function check_local_configuration()
+# This function read the configuration file and make the parser of the data on
+# it. For more information about the configuration file, take a look at
+# "etc/kworkflow.config" in the kworkflow directory.
+# @parameter: This function expects a path to the configuration file.
+function parse_configuration()
 {
-  local config_path=$PWD/kworkflow.config
+  local config_path=$1
+  local filename=$(basename $config_path)
 
-  # File does not exist, use default configuration
-  if [ ! -f $config_path ] ; then
-    configurations=(
-      [qemu_path_image]=$VDISK
-      [qemu_hw_options]=$QEMU_OPT
-      [qemu_net_options]=""
-      [port]=$DEFAULT_PORT
-      [ip]=$DEFAULT_IP
-    )
-    return 1
+  if [ ! -f $1 ] || [ "$filename" != "kworkflow.config" ] ; then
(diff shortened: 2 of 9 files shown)


src/commons.sh
```diff
@@ -17,47 +17,42 @@ TARGET="qemu"
...
 {
-  say "Global values:"
-
-  echo -e "\tBASE: $BASE"
-  echo -e "\tBUILD_DIR: $BUILD_DIR"
-  echo -e "\tQEMU ARCH: $QEMU_ARCH"
-  echo -e "\tQEMU COMMAND: $QEMU"
-  echo -e "\tQEMU MOUNT POINT: $QEMU_MNT"
-  echo -e "\tTARGET: $TARGET"
+  local local_config_path=$PWD/kworkflow.config
```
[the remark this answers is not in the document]
rodrigosiqueira: Totally agree!

etc/kworkflow.config
```diff
@@ -0,0 +1,25 @@
+# Default ip for try ssh connection
+ssh_ip=localhost
+
+# Default port for try to connect via ssh
```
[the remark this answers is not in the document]
rodrigosiqueira: Agree

src/commons.sh
```diff
@@ -17,47 +17,42 @@ TARGET="qemu"
 BASHPATH=/bin/bash
```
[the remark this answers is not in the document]
rodrigosiqueira: Actually this is an unused variable, I want to make a new PR for remove this and other variables; because of this, I did not move this variable to /etc/kworkflow.config.

rodrigosiqueira:
Thanks a lot for the great review! I applied most of your suggestions. If you fill ok with this new version, feel free to push the merge button ;)

MatheusBernardino:
Hi, @rodrigosiqueira . Everything seems nice, to me! I just still don't quite like two things:
1)  the call to `check_local_configuration` at `show_variables` (because the function is doing more than it's duty)
2) The pushd/popd polluting the test's output.

Can you tell me your opinion on this? :)

Also, since you changed `src/commons.sh` at line 83 (the call to `check_local_configuration`), please, don't forget to update the comment above it (because it is still saying "default config") :)

One last thing: I will open a issue on the global variables needed to be removed, just so as we don't forget it latter, ok? :) Please, feel free to edit the issue's text, as you know more about what is needed to be done than me.

rodrigosiqueira:
> Hi, @rodrigosiqueira . Everything seems nice, to me! I just still don't quite like two things:
> 
>     1. the call to `check_local_configuration` at `show_variables` (because the function is doing more than it's duty)

I kept  'check_local_configuration' here because it is important for users to know what is the real value of the variables in use. Keep in mind, that local config file changes the behavior of kw and in a debug situation it is useless to show global value if we have local values that may change the overall operation.

>     2. The pushd/popd polluting the test's output. (but I understand if you don't feel like changing it, since it's not, directly, part of the PR)
> 

I did not change it, because this is not part of this PR. We should create an issue and make it in a separate PR.

> Can you tell me your opinion on this? :)
> 
> Also, since you changed `src/commons.sh` at line 83 (the call to `check_local_configuration`), please, don't forget to update the comment above it (because it is still saying "default config") :)

Fixed

> One last thing: I will open a issue on the global variables needed to be removed, just so as we don't forget it latter, ok? :) Please, feel free to edit the issue's text, as you know more about what is needed to be done than me.

Nice! Thanks for open the issue.

MatheusBernardino:
> I kept 'check_local_configuration' here because it is important for users to know what is the real value of the variables in use. Keep in mind, that local config file changes the behavior of kw and in a debug situation it is useless to show global value if we have local values that may change the overall operation.

Yes, I agree with you about showing the real value of variables. But since `check_local_configuration` is always called when `commons.sh` is imported, won't the variables be already set up when we call `show_variables`?

rodrigosiqueira:
> > I kept 'check_local_configuration' here because it is important for users to know what is the real value of the variables in use. Keep in mind, that local config file changes the behavior of kw and in a debug situation it is useless to show global value if we have local values that may change the overall operation.
> 
> Yes, I agree with you about showing the real value of variables. But since `check_local_configuration` is always called when `commons.sh` is imported, won't the variables be already set up when we call `show_variables`?

I got it now, and you're totally right.
However, I realized that use "check_local_configuration" is not a good idea and change it back to "load_configuration $config_files_path/kworkflow.config". Why?
Basically, the local "kworkflow.config" may not define all the variables (e.g., arch) and if we load the gloabal config first, we can guarantee that all the required variables gonna there (even if the user forgot to add it).

> Also, I just noted another very important thing, now: The following variables define settings already moved to the global configuration file system, but are still being used in KW src files: `MOUNT_POINT`, `QEMU`, `QEMU_OPTS` and `VDISK`. (They should be substituted for `configurations[foo]`.)
> Shouldn't this be also fixed in this PR, since this makes KW state inconsistent? I mean, `show_variables` will print the configurations from config file, but the configurations being used will be others...
> 
> And the variables `QEMU_MNT` and `BUILD_DIR` weren't moved to the global config file. Will they stay as hard coded variables, for now?

Reallyyyyy nice catch! Actually, this is one of the main points of this PR. Tahnks a lot for this! I just updated the PR with your comments.

MatheusBernardino:
Hi, @rodrigosiqueira . I think we are almost there. Just a few things more:

1) The commons_test.sh is now failing, because of the changes on etc/kworkflow.config and commons.sh. But it's a very simple thing to correct :)
2) As we are already removing, from commons.sh, some global variables that are now configurations at kworkflow.config, I don't understand why we don't just go ahead and remove `DEFAULT_IP` and `DEFAULT_PORT`. As they were moved, I think their removal should be part of this PR.
3) I understand why you had to change the `check_local_configuration` call  back to `load_configuration $config_files_path/kworkflow.config`. (And really nice catch, by the way :) But I still like the idea of having the configurations (local and/or global) loaded automatically when commons.sh is imported. Don't you thing we should load both global and local config file at the commons.sh's last line? This way, we can remove the call to `check_local_configuration` at `show_variables`. Or even better, `check_local_configuration` could automatically import global file before importing local file, so that all configurations get correctly attributed.

rodrigosiqueira:
> Hi, @rodrigosiqueira . I think we are almost there. Just a few things more:
> 
>     1. The commons_test.sh is now failing, because of the changes on etc/kworkflow.config and commons.sh. But it's a very simple thing to correct :)

Done :)

>     2. As we are already removing, from commons.sh, some global variables that are now configurations at kworkflow.config, I don't understand why we don't just go ahead and remove `DEFAULT_IP` and `DEFAULT_PORT`. As they were moved, I think their removal should be part of this PR.

Done :)

>     3. I understand why you had to change the `check_local_configuration` call  back to `load_configuration $config_files_path/kworkflow.config`. (And really nice catch, by the way :) But I still like the idea of having the configurations (local and/or global) loaded automatically when commons.sh is imported. Don't you thing we should load both global and local config file at the commons.sh's last line? This way, we can remove the call to `check_local_configuration` at `show_variables`. Or even better, `check_local_configuration` could automatically import global file before importing local file, so that all configurations get correctly attributed.
> 
>     4. (still in the same topic as 3) The user can change the global configuration file and remove some variable. This would break kw commands. (So loading global config file before local wouldn't help much on this issue...) We could make kw check wether some configuration is present or not and exit nicely (with a message and error code) when not. But I think a simpler way would be to just add a commentary at `etc/kworkflow.config` telling the user that no variable should be removed from this file or kw will have unwanted behavior.

I made the changes based on your comments, and improved it. Now I think you gonna like :)

tests/commons_test.sh
```diff
@@ -51,22 +72,20 @@ function testConfigFileNotFound
...
       [key3]="value3"
-      [ip]="127.0.0.1"
-      [port]="2222"
     )
 
     cp tests/samples/kworkflow.config tests/.tmp_commons_test/
 
     pushd tests/.tmp_commons_test
-    check_local_configuration
+    load_configuration
```
MatheusBernardino: I think we should remove pushd/popd and change the call to `load_configuration` here to `parse_configuration tests/.tmp_commons_test/kworkflow.config`. Why? `load_configuration` will try to load the global configuration`file, which is `$config_files_path/kworkflow.config`,  prior to local. Since `kw.sh` was not imported, `$config_files_path` equals to none. So it will try to load "/kworkflow.config". If the user testing kw have a local `kworkflow.config` at his/her root "/", them this test will fail. The test shouldn't depend on the user's environment to pass.

src/commons.sh
```diff
@@ -1,63 +1,47 @@
...
-
-  echo -e "\tBASE: $BASE"
-  echo -e "\tBUILD_DIR: $BUILD_DIR"
-  echo -e "\tQEMU ARCH: $QEMU_ARCH"
-  echo -e "\tQEMU COMMAND: $QEMU"
-  echo -e "\tQEMU MOUNT POINT: $QEMU_MNT"
-  echo -e "\tTARGET: $TARGET"
+  local has_local_config_path="No"
 
   check_local_configuration
```
MatheusBernardino: This line should be removed since this function doesn't exist anymore and common.sh automatic loads both global and local configuration files.

etc/kworkflow.config
```diff
@@ -0,0 +1,29 @@
+# Default configuration file
```
MatheusBernardino: I would like to add a comment here saying that no variables should be removed from this file, in order to kw work properly, just to warn users that want to customize their global configuration files.

tests/commons_test.sh
rodrigosiqueira: Actually, I changed the test because it does not make sense anymore. Now, I'm testing the local config file with different parameters.

src/commons.sh
rodrigosiqueira: Done

etc/kworkflow.config
rodrigosiqueira: I did not add this, because this is not true. For example, if the user uses libvirt as a virtualized, the QEMU command has no effect. However, now I'm think about adding some hardcoded variable just in case that a variable was not loaded from the config file. After this PR we could work to make kw stable and generic from the point of view of the configuration variables.

MatheusBernardino:
Already merged, see bc43e82

Issue #66, opened 2019-01-29
Support multiple shells


rodrigosiqueira:
Ideas:
 - We can have a single alias in all *rc
 - Make a driver
 - Add to .profile

P.s.: We should take a look at the auto completion 

This issue depends on #50

rodrigosiqueira:
See: https://github.com/rodrigosiqueira/kworkflow/commit/649a6141224a06c0db343ca1027bec9ca6bcede3

PR #76, opened 2019-02-23
Turns `kw.sh` into a driver

giulianobelinassi:
This commits makes `kw.sh` a script that do not requires sourcing to be
used. Although this commit does not explicitly add support to other
shells than bash, it makes `kw` callable from other shells. As a proof
of concept, a new test was added calling `kw` from dash.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,17 @@ following dependencies:
 * libguestfs
 * Qemu
 * Ansible
+* Bash
 
 > If you want to use the default alert system (for commands that may take longer
 to run), you will also need to install:
 
 * paplay
 * notify-send
+
+> For development, you will need the additional packages:
+
+* dash
 
 ## Recommendations
 
@@ -48,8 +53,12 @@ your `.bashrc`:
 
 ```
 # kw
-source /home/<user>/.config/kw/kw.sh
+PATH=$PATH:/home/<user>/.config/kw
+source /home/<user>/.config/kw/src/bash_autocomplete.sh
 ```
+
+> If you use another shell (`ksh`, for example), you will need to manually add
+the path to `kw` to `PATH` environment variable.
 
 > To check if the installations was ok, type:
 
--- a/kw
+++ b/kw
@@ -9,36 +9,6 @@ config_files_path=${config_files_path:-"$HOME/.config/$EASY_KERNEL_WORKFLOW/etc"
 # Export external variables required by kworkflow
 export EASY_KERNEL_WORKFLOW
 
-_kw_autocomplete()
-{
-    local current_command previous_command kw_options
-    COMPREPLY=()
-    current_command="${COMP_WORDS[COMP_CWORD]}"
-    previous_command="${COMP_WORDS[COMP_CWORD-1]}"
-    kw_options="export e build b bi install i prepare p new n ssh s
-                mail mount mo umount um boot bo vars v up u codestyle c
-                maintainers m help h"
-
-    # By default, autocomplete with kw_options
-    if [[ ${previous_command} == kw ]] ; then
-        COMPREPLY=( $(compgen -W "${kw_options}" -- ${current_command}) )
-        return 0
-    fi
-
-    # TODO:
-    # Autocomplete in the bash terminal is a powerful tool which allows us to
-    # make many interesting things. In the future, we could add an
-    # autocompletion for subcommands, the code below illustrates an example
-    # that tries to add this feature for the ‘maintainers’ options.
-    #
-    # For maintainers and m options, autocomplete with folder
-    # if [ ${previous_command} == maintainers ] || [ ${previous_command} == m ] ; then
-    #   COMPREPLY=( $(compgen -d -- ${current_command}) )
-    #   return 0
-    # fi
-}
-complete -o default -F _kw_autocomplete kw
-
 function kw()
 {
   action=$1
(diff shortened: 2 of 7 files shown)


giulianobelinassi:
I forgot to mention that this is a somewhat closer step for #66

rodrigosiqueira:
Hi, @giulianobelinassi good to see you around :)

First of all, thanks for your patch! This is a really good one!!!

Follow my comments:

1. About the commit message

I think you can improve the body of the commit message a little bit, how about something like this:

> This commit removes the dependency on sourcing `kw.sh`. Although this commit does not explicitly add support to other shells, such as ksh and fish, it makes `kw` callable from other shells. As a proof of concept, a new test was added with the goal of call `kw` from dash.

2. About the new tests

For make your new test to pass, it is required to install “dash” which is ok. However, I think you should document it on the readme. How about that:

a. Split the “Package Dependencies” in two parts: the dependencies for running kw and the dependence for development.
b. In the development dependencies, you should add the ‘dash’ application.

3. About “*autocomplete.sh”

I prefer that any source code lives inside ‘src’ directory. Additionally, I think you should add a suitable handler for ‘autocomplete’ sources in this PR. Some ideas on how you can do it:

a. Create a file named ‘autocomplete.sh’ in the ‘src’ directory.
b. In the autocomplete, you should add a mechanism for detecting the current bash in order to use a peculiar autocomplete. P.s: you just need to add support for bash, since we currently use it.

4. Something changed in the ‘kw_test’ output

Before your patch, ‘kw_test’ does not show any message from kw. However, after you patch the following output is displayed:

```sh
Running test [kw_test]
Invalid option
Usage: kw [target] cmd

The current supported targets are:
         host - this machine
         qemu - qemu machine
         arm - arm machine

Commands:
        build,b - Build Kernel and modules
        install,i - Install modules
        bi - Build and install modules
        prepare,p - Deploy basic environment in the VM
        new,n - Install new Kernel image
        ssh,s - Enter in the vm
        mount,mo - Mount partition with qemu-nbd
        umount,um - Umount partition created with qemu-nbd
        vars,v - Show variables
        up,u - Wake up vm
        codestyle,c - Apply checkpatch on directory or file
        maintainers,m [-a|--authors] - Return the maintainers and
                                     the mailing list. "-a" also
                                     prints files authors
        explore,e - Search for expression on git log or directory
        help,h - displays this help mesage
testVariables
testExported
```

giulianobelinassi:
Hi. Thank you for your quick review!

> For make your new test to pass, it is required to install “dash” which is ok. However, I think you should document it on the readme. How about that:
> 
> a. Split the “Package Dependencies” in two parts: the dependencies for running kw and the dependence for development.
> b. In the development dependencies, you should add the ‘dash’ application.

I think a better solution is instead calling `dash` directly, call the POSIX `sh` shell. All POSIX-compliant systems have it installed and therefore nothing needs to be installed.

> 
>     About “*autocomplete.sh”
> 
> I prefer that any source code lives inside ‘src’ directory. Additionally, I think you should add a suitable handler for ‘autocomplete’ sources in this PR. Some ideas on how you can do it:
> 
> a. Create a file named ‘autocomplete.sh’ in the ‘src’ directory.
> b. In the autocomplete, you should add a mechanism for detecting the current bash in order to use a peculiar autocomplete. P.s: you just need to add support for bash, since we currently use it.

I think it is better to keep the autocompletion scripts separately, i. e, one file to bash, another for ksh. Autocompletion feature is not POSIX standard and each shell might implement it their way. Then i think that setup might detect the shell and `source` it.

rodrigosiqueira:
> Hi. Thank you for your quick review!
> 
> > For make your new test to pass, it is required to install “dash” which is ok. However, I think you should document it on the readme. How about that:
> > a. Split the “Package Dependencies” in two parts: the dependencies for running kw and the dependence for development.
> > b. In the development dependencies, you should add the ‘dash’ application.
> 
> I think a better solution is instead calling `dash` directly, call the POSIX `sh` shell. All POSIX-compliant systems have it installed and therefore nothing needs to be installed.

I totally agree with you. However, my point here is different :)

If we add your PR we can start to dream about make kw portable for different shells. Additionally, we can add a development dependency which is different from production dependency. With your PR and the change described in my first comment, we can start to think about have unit tests for different shells. For example, we can have one test per shell:

```
 tests/kw_dash.sh
 tests/kw_ksh.sh
 tests/kw_fish.sh
```

> 
> > ```
> > About “*autocomplete.sh”
> > ```
> > I prefer that any source code lives inside ‘src’ directory. Additionally, I think you should add a suitable handler for ‘autocomplete’ sources in this PR. Some ideas on how you can do it:
> > a. Create a file named ‘autocomplete.sh’ in the ‘src’ directory.
> > b. In the autocomplete, you should add a mechanism for detecting the current bash in order to use a peculiar autocomplete. P.s: you just need to add support for bash, since we currently use it.
> 
> I think it is better to keep the autocompletion scripts separately, i. e, one file to bash, another for ksh. Autocompletion feature is not POSIX standard and each shell might implement it their way. Then i think that setup might detect the shell and `source` it.

Agree

giulianobelinassi:
I've updated the PR with your suggestions. :-)

rodrigosiqueira:
Merged! See: https://github.com/rodrigosiqueira/kworkflow/commit/649a6141224a06c0db343ca1027bec9ca6bcede3

PR #84, opened 2019-04-07
README: fix typos and improve wording

chriscool:


README.md
```diff
@@ -28,10 +28,10 @@ to run), you will also need to install:
 
 If you want to use Qemu, we recommend the following steps:
 
-1) Create an Qemu image as a qcow2;
+1) Create an Qemu image in the qcow2 format;
```
MatheusBernardino: As we are already here, it would be nice to change "Create an Qemu [...]" for "Create a Qemu [...]".

README.md
```diff
@@ -93,30 +93,30 @@ qemu_path_image=$HOME/p/virty.qcow2
...
 
-> Build a Kernel and install in the Qemu image:
+> Build a Kernel and install it in the Qemu image:
 
 ```
 kw bi
 ```
 
-> Mount Qemu image to transfer data:
+> Mount a Qemu image to transfer data:
```
MatheusBernardino: IMHO, we should say "the" instead of "a" since we are talking about the specific Qemu image defined in the kworkflow.config file (global or local). Maybe we should indeed add this information here, to make it cleaner...

README.md
```diff
@@ -93,30 +93,30 @@ qemu_path_image=$HOME/p/virty.qcow2
...
 
 ```
 kw mount|mo
 ```
 
-> Umount Qemu image:
+> Umount a Qemu image:
```
MatheusBernardino: Same as before (a -> the)

README.md
```diff
@@ -128,7 +128,7 @@ kw umount|um
 kw vars|v
 ```
 
-> Turn on vm:
+> Turn on a VM:
```
MatheusBernardino: Same as before (a -> the)

README.md
chriscool: Yeah, it seems that the rule is the following: "Use “a” before words that start with a consonant sound and “an” before words that start with a vowel sound"

And Qemu indeed starts with a consonant sound.

README.md
chriscool: Yeah, I don't know the tool, so it's difficult to tell for me.

chriscool:
@MatheusBernardino I just forced pushed the branch after adding my S-o-B to the commit message and making the changes you suggest.

It seems that the PR has already been changed to target the unstable branch.

chriscool:
I just saw that I could still change the PR to target the unstable branch, so I did that.

README.md
chriscool: Sure.

README.md
chriscool: Sure.

MatheusBernardino:
Thanks, @chriscool , merged (93ab696)

PR #87, opened 2019-04-08
Remove depecrated commands from bash autocomplete

lucasoshiro:
Remove the following deprecated commands completions for bash: mail, boot, bo.

There's a test case (testExplore) that fails. However, that case fails even in the unstable branch on my machine. I'm using shunit 2.7.1 and bash 5.0.2.


rodrigosiqueira:
Hi, thanks for your patch :)

Really nice catch!

PR #88, opened 2019-04-10
[WIP] Add fish shell completion

lucasoshiro:
Here is a script that declares autocomplete suggestions for fish shell. In this commit, I only created completion rules for the kw subcommands. However, other rules for options of these subcommands can be easily done, but I'll do it in the next commits after a first review :).

Before going ahead, I have these questions: 

- I don't know exactly where to put the completion file. Currently, it is at `etc/fish_completion`, is it a good place for it? 

- There's an auxiliary file written in fish script that helps the tests for this feature. Its name is `_fish_completion.fish`, is it a good name for it? Is it a good idea creating that script in another file or it would be cleaner if it was hardcoded in `fish_completion.sh`?

- Is it ok to add fish as a dependency for development?

- The autocomplete script needs to be placed in one of the directories listed in the `fish_complete_path` environment variable on fish. Which of these options is better?

  - The user copies the script to one of these directories manually

  - The `setup.sh` script copies it to one of these directories

  - The `setup.sh` script creates a new directory, adds it to `fish_complete_path`, and puts the autocomplete script inside it

Here's a screenshot of how it works when you double-tab after typing `kw` on fish:

image


etc/fish_completion/kw.fish
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env fish
+
+# Test if the kw command has subcommands
+function __fish_kw_no_commands
+    set cmd (commandline -opc)
+    if [ "$cmd" = "kw" ]
+    	return 0
```
rodrigosiqueira: There's a tab here, replace it by spaces

etc/fish_completion/kw.fish
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env fish
+
+# Test if the kw command has subcommands
+function __fish_kw_no_commands
```
rodrigosiqueira: Why did you add "__" instead of "_"? Also, why did you not used `_kw_autocomplete` to maintain the pattern?

etc/fish_completion/kw.fish
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env fish
```
rodrigosiqueira: Does not make sense to add this file at `/etc/fish_completition/kw.fish`. Why did you add it in the `src` directory? I recommend you to follow the same patter used by bash.

etc/fish_completion/kw.fish
```diff
@@ -0,0 +1,30 @@
...
+complete -c kw -n "__fish_kw_no_commands" -a "new n" -d "Install new Kernel image"
+complete -c kw -n "__fish_kw_no_commands" -a "ssh s" -d "Enter in the vm"
+complete -c kw -n "__fish_kw_no_commands" -a "mount mo" -d "Mount partition with qemu-nbd"
+complete -c kw -n "__fish_kw_no_commands" -a "umount um" -d "Umount partition created with qemu-nbd"
+complete -c kw -n "__fish_kw_no_commands" -a "vars v" -d "Show variables"
+complete -c kw -n "__fish_kw_no_commands" -a "up u" -d "Wake up vm"
+complete -c kw -n "__fish_kw_no_commands" -a "codestyle c" -d "Apply checkpatch on directory or file"
+complete -c kw -n "__fish_kw_no_commands" -a "maintainers m" -d "Return the maintainers and the mailing list"
+complete -c kw -n "__fish_kw_no_commands" -a "explore e" -d "Search for expression on git log or directory"
+complete -c kw -n "__fish_kw_no_commands" -a "help h" -d "Display this help mesage"
```
rodrigosiqueira: These commands look extremely similar which is not good for code maintenance. How about that:
1. Create a bash hash, with the command and the description
2. Create a `for` loop and iterate over the hash and build the command

With this approach, it is going to be easy to expand the fish competition in the future.

tests/fish_completion_test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+. ./tests/utils --source-only
+
+function suite() {
```
rodrigosiqueira:
```
function suite()
{
...
}
```

tests/_fish_completion.fish
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env fish
```
rodrigosiqueira: Hmmmm... I'm thinking about this file. For me, it should be inside `external` or `samples`... but I don't have strong opinion about that. What do you think @MatheusBernardino ?

rodrigosiqueira:
Hi @lucasoshiro ,

First of all, thank you so much for your PR; I'm glad to see `kw` with fish support :)

I just have some tiny comments inline, take a look on it.

tests/fish_completion_test.sh
lucasoshiro: oops, sorry

etc/fish_completion/kw.fish
lucasoshiro: This code was based on this (this was made for docker):
https://github.com/docker/cli/blob/master/contrib/completion/fish/docker.fish. I tried to follow the same conventions used in it.

About the name of the file, it must be kw.fish and be placed on one of the directories listed on `fish_complete_path`, otherwise fish won't recognize as the script containing the completions for `kw`. In the source file it can be different, of course, but the user (or the setup script) will need to rename it.

etc/fish_completion/kw.fish
lucasoshiro: This answers one of my questions :-). I didn't put it in `src` because it's not part of kw itself and I thought it could mess that repository. About using the same pattern used by bash (maybe `fish_autocomplete.fish` ?), I discussed that in the previous conversation.

etc/fish_completion/kw.fish
lucasoshiro: I totally agree with you, but, sadly, fish script doesn't have hashes natively, and I didn't find a way to create matrices in that language... I can create two arrays, one with the commands, and the other with the descriptions, but I don't think it would be clear...

lucasoshiro:
@rodrigosiqueira Thanks for the review! I'll change the commit title later in order to not lose your comments by now.

etc/fish_completion/kw.fish
rodrigosiqueira: If use "__" is a fish shell constraints, that's ok. However, if it is a pattern used by docker I prefer not to use it.

About the path, I prefer to move it to `src` dir because it is a code feature. However, I'm open to new ideas :)

etc/fish_completion/kw.fish
rodrigosiqueira: arrrgh... I've to admit that I really dislike this in the fish shell... anyway, just go ahead. We don't want to change the world, just add support for fish shell :P

rodrigosiqueira:
> @rodrigosiqueira Thanks for the review! I'll change the commit title later in order to not lose your comments by now.

Thanks for your reply.
I'm looking forwarding for the next version :)

etc/fish_completion/kw.fish
lucasoshiro: About "__": yeah, I found that it is used by these scripts packed with fish: https://github.com/fish-shell/fish-shell/tree/master/share/completions, as you can see, for example, at the completion scripts for apt-get, npm, git and valgrind (all of them use subcommands, just like kw).

tests/_fish_completion.fish
lucasoshiro: I'll remove this file, and I'll summarize it into a string inside `fish_completion_test.sh`. It will also give more flexibility to this piece of code.

lucasoshiro:
Luckily, GitHub kept the comments, I thought they would disappear

rodrigosiqueira:
@lucasoshiro any update?

lucasoshiro:
I''m closing this, and I'll submit a new and updated PR

PR #89, opened 2019-04-13
src: vm: Add support to execute commands/scripts via ssh

rodrigosiqueira:
This commit adds the capability of executing commands or scripts via kw
s. We have two options:

 1. Commands: the command to be executed inside the VM
 2. Scripts: A script in the host machine to be executed in the VM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,14 @@ kw vars|v
 ```
 kw up|u
 ```
+> Based on the kworkflow.config file tries to perform ssh operation such as
+> login into the VM or execute a host script in the VM.
+> Note: The `--script` parameter, only evaluate bash scripts
+
+```
+kw ssh|s [--script|-s="SCRIPT PATH"]
+kw ssh|s [--command|-c="COMMAND"]
+```
 
 > Run checkpatch in a target (directory of file):
 
--- a/src/vm.sh
+++ b/src/vm.sh
@@ -51,11 +51,52 @@ function vm_up
         ${configurations[qemu_path_image]}
 }
 
+# This function manages ssh operations. Currently, we have three basic actions:
+# connect to the VM, execute a command in the VM, and perform a local script
+# inside the VM.
+#
+# @opts Expects a parameter to be passed for the ssh command (it could be
+#       --command, -c, --script, -s). If this parameter receives a null value
+#       this function will make a simple ssh operation; otherwise, a command or
+#       script will be attempted to execute.
 function vm_ssh
 {
-  say "SSH"
-  say "-> Port: " ${configurations[ssh_port]} " IP: " ${configurations[ssh_ip]}
-  ssh -p ${configurations[ssh_port]} ${configurations[ssh_ip]}
+  local opts=$@
+  local port=${configurations[ssh_port]}
+  local target=${configurations[ssh_ip]}
+
+  # Mandatory parameter
+  if [ -z "$target" ]; then
+    complain "Invalid argument: $@"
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -145,6 +145,13 @@ kw vars|v
 kw up|u
 ```
 
+> Based on the kworkflow.config tries to perform ssh operation such as login
+> into the VM or execute a command/script.
+
+```
+kw ssh|s [(--script|-s|--coomand|-c)="COMMAND OR SCRIPT PATH"]
```
MatheusBernardino: typo: coomand -> command

src/vm.sh
```diff
@@ -51,11 +51,50 @@ function vm_up
         ${configurations[qemu_path_image]}
 }
 
+# This function manages ssh operations. Currently, we can have three basic
```
MatheusBernardino: maybe "we have" or  instead of "we can have"

tests/ssh_test.sh
```diff
@@ -0,0 +1,120 @@
...
+function setupSsh
+{
+  local -r current_path=$PWD
+
+  rm -rf $TEST_PATH
+
+  mkdir -p $TEST_PATH
+
+  cp -f tests/samples/kworkflow.config tests/.tmp/
+  cp -f tests/samples/dmesg tests/.tmp/
```
MatheusBernardino: Use `$TEST_PATH` instead of explicit "tests/.tmp/" in those two previous lines.

tests/ssh_test.sh
```diff
@@ -0,0 +1,120 @@
...
+  cd $TEST_PATH
+  load_configuration
+  cd $current_path
+}
+
+function tearDownSsh
+{
+  local -r test_path=$TEST_PATH
+
+  rm -rf $test_path
```
MatheusBernardino: Why assigning TEST_PATH to a local test_path variable, here?

src/vm.sh
```diff
@@ -51,11 +51,50 @@ function vm_up
...
+    exit 22 # EINVAL
+  fi
+
+  if [[ $# -gt 0 ]]; then
+    if [[ "$opts" =~ ^(--command|-c)= ]]; then
+      opts="$(echo $opts | cut -d = -f2 )"
+    elif [[ "$opts" =~ ^(--script|-s)= ]]; then
+      local script_path=$(echo $opts | cut -d = -f2 )
+
+      if [[ ! -f $script_path ]]; then
```
MatheusBernardino: Running something like `kw --script=~/kworkflow/tests/samples/dmesg` will failure this test, but I really don't understand why... I know that bash won't expand `~` for quoted strings such as `"~/"`, but we are not adding `"` here...

tests/ssh_test.sh
```diff
@@ -0,0 +1,120 @@
...
+  assertTrue "We expected a substring \"$INVALID_ARG: $args\", but we got \"$ret\"" '[[ $ret =~ "$INVALID_ARG: $args" ]]'
+
+  args="-s="
+  ret=$(vm_ssh $args)
+  assertTrue "We expected a substring \"$NO_SUCH_FILE\", but we got \"$ret\"" '[[ $ret =~ "$NO_SUCH_FILE" ]]'
+
+  args="--script="
+  ret=$(vm_ssh $args)
+  assertTrue "We expected a substring \"$NO_SUCH_FILE\", but we got \"$ret\"" '[[ $ret =~ "$NO_SUCH_FILE" ]]'
+
```
MatheusBernardino: IMHO, we could add "-c=" to these tests.

PR #93, opened 2019-04-16
src: doc: Add man documentation

rodrigosiqueira:
This commit adds a man documentation by using a tool named rst2man, from
python-docutils package. Additionally, it removes the duplicated
documentation from the README file.

### DIFF
--- a/documentation/man/kw.rst
+++ b/documentation/man/kw.rst
@@ -1,0 +1,305 @@
+=====
+ kw
+=====
+
+--------------------------------------------
+Inglorious kernel developer workflow scripts
+--------------------------------------------
+
+:Author: Rodrigo Siqueira <rodrigosiqueiramelo@gmail.com>
+:Author: Matheus Tavares <matheus.bernardino@usp.br>
+:Date: 2018-05-18
+:Copyright: GPLv2+
+:Manual section: 1
+:Manual group: Kernel Workflow
+
+SYNOPSIS
+========
+
+**kw** *COMMAND* [*OPTION* ...] 
+
+DESCRIPTION
+===========
+
+**kw** is a set of scripts that have a single mission: reduce the overhead
+related with infrastructure project setup in projects that have a similar
+workflow of Linux Kernel.
+
+COMMANDS
+========
+
+Here we have a list of commands tied to **kw**; target audiences and
+functionality below group the commands. Many of the commands have default
+aliases.
+
+COMMANDS FOR USING QEMU
+=======================
+
+Projects that use QEMU in the development cycle usually requires the developers
+to use some specific commands. In this section, you can find all commands
+available in **kw** to manage some daily tasks related to QEMU. Notice, that
+some rules are specific for Linux Kernel.
+
+mo, mount
+---------
+Mount a QEMU image in a specific directory, based on the data available in the
+**kworkflow.config**. Notice, that the configuration file have the information
+about the source and destiny.
+
+.. note::
+  **Just run this command after you turn off your VM**.
+
+um, umount
+----------
+Umount a QEMU image previously mounted with the command **kw mount**.
+
+u, up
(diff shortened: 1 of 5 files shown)


documentation/man/kw.rst
```diff
@@ -0,0 +1,217 @@
...
+**-net nic -net user,hostfwd=tcp::2222-:22,smb=/home/USER**
+
+qemu_path_image=PATH
+--------------------
+Specify the VM image path. By default, **kw** uses
+**/home/USERKW/p/virty.qcow2**
+
+alert=[vs|s|v|n]
+----------------
+Default alert options, you have:
```
MatheusBernardino: I think we should give the user a little more background of this variable. We should say that the value specified here will be used by kw with commands that use alert notification when the user doesn't specify a --alert= option. Also, we lost the description of the --alert= option in the transition from README to this man page. We should say, somewhere, what this option does and which commands accept it.

documentation/man/kw.rst
```diff
@@ -0,0 +1,217 @@
...
+bi
+--
+Build and install a new module version. It is a combination of **build** and
+**install**.
+
+h, help
+-------
+Show basic help.
+
+ABOUT kworflow.config
```
MatheusBernardino: Although it is implicit, I think we should explicitly say at the text that the user should add local config files for his/her projects as desired, but he can also change the global config file to have a custom kw configuration running for every directory he/she whishes to invoke the application. Also, we should mention where the global file is.

kw
```diff
@@ -149,6 +149,25 @@ function kw()
         explore "$@"
       )
       ;;
+    man)
+      (
```
MatheusBernardino: You have to import kwio.sh here, since you are using complain below. Also, maybe should we move this whole code to a man.sh file and just call a function at it, as we do with other commands? I don't really know.

documentation/man/kw.rst
```diff
@@ -0,0 +1,217 @@
...
+=====
+
+-----------------------------------------------
+Inglorious kernel developer workflow scripts
+-----------------------------------------------
+
+:Author: Rodrigo Siqueira <rodrigosiqueiramelo@gmail.com>
+:Author: Matheus Tavares <matheus.bernardino@usp.br>
+:Date: 2018-05-18
+:Copyright: GPLv2+
```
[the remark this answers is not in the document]
rodrigosiqueira: No GPLv2+ :)

documentation/man/kw.rst
```diff
@@ -0,0 +1,217 @@
...
+ kw
+=====
+
+-----------------------------------------------
+Inglorious kernel developer workflow scripts
+-----------------------------------------------
+
+:Author: Rodrigo Siqueira <rodrigosiqueiramelo@gmail.com>
+:Author: Matheus Tavares <matheus.bernardino@usp.br>
+:Date: 2018-05-18
```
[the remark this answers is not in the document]
rodrigosiqueira: Yes, do you have another suggestion?

setup.sh
```diff
@@ -78,6 +79,7 @@ function synchronize_files()
   rsync -vr $SRCDIR $INSTALLTO
   rsync -vr $DEPLOY_DIR $INSTALLTO
   rsync -vr $SOUNDS $INSTALLTO
+  rsync -vr $DOCUMENTATION $INSTALLTO
```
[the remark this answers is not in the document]
rodrigosiqueira: Unfortunately, no :(

Usually, man files are kept at `/usr/share/man`. However, don't worry, one day kw will be package for Debian and Arch; in that time, man kw will work \o/

kw
rodrigosiqueira: Nice catch! :)

Do you really think that we need to create `man.sh`? I think we can put it in another place, how about `help.sh`?

PR #106, opened 2019-10-06
src: mk: use of ${configurations[arch]} instead of $MAKE_OPTS

### LINKED ISSUE #63: Use the arch variable in the kw b
Now we have the information about the compilation architecture in the kworkflow.config. How about starting to use it?

rodrigorsdc:
Remove $MAKE_OPTS and replace by ${configurations[arch]}
in mk_build function

Fixes #63 

Signed-off-by: Rodrigo Ribeiro Carvalho <rodrigorsdc@gmail.com>

### DIFF
--- a/src/mk.sh
+++ b/src/mk.sh
@@ -43,8 +43,8 @@ function mk_build
     PARALLEL_CORES=$(grep -c ^processor /proc/cpuinfo)
   fi
 
-  say "make -j$PARALLEL_CORES $MAKE_OPTS"
-  make -j$PARALLEL_CORES $MAKE_OPTS
+  say "make ARCH="${configurations[arch]}" -j$PARALLEL_CORES"
+  make ARCH="${configurations[arch]}" -j$PARALLEL_CORES
 }
 
 function mk_install


src/mk.sh
```diff
@@ -39,8 +39,8 @@ function mk_build
 
   PARALLEL_CORES=$(( $PARALLEL_CORES * 2 ))
 
-  say "make -j$PARALLEL_CORES $MAKE_OPTS"
-  make -j$PARALLEL_CORES $MAKE_OPTS
+  say "make -j$PARALLEL_CORES ARCH=${configurations[arch]}"
+  make -j$PARALLEL_CORES ARCH=${configurations[arch]}
```
matheustavares: A reeealy nitpick: when using `${configurations}` for bash commands, it's better if we protect it with single quotes so that strings with spaces get correctly parsed by the command. In this case, this is veeeeery unlikely as we don't expect `${configurations[arch]}` to be a string with spaces. But just to make sure, I think we could write this as `... ARCH='${configurations[arch]}'`.

matheustavares:
Hi, @rodrigorsdc . Thanks for working on this!

There are 2 extra commits in your PR, could you please rebase it with rodrigosiqueira/unstable and send just the last one?

Also, thanks for the issue reference in the commit message. But could you also document this change a little more, please? For example, I think it would be nice to mention that `$MAKE_OPTS` was outdated as this variable is never defined. In addition, we now have a `${configurations[arch]}` which defines the arch to build but we don't currently pass it to `make` invokation. So, since this is the only compiling option we allow the user to define right now, we remove the outdated `$MAKE_OPTS` and give the correct setting to `make`.

rodrigorsdc:
Hi @matheustavares, thanks for reviewing!

 I improved commit message. What do you think?
Also, I rabase my commit with rodrigosiqueira/unstable

thanks

matheustavares:
Hi, @rodrigorsdc . Thanks for the new version. Another nitpick that I forgot earlier is to use the imperative form in the message's title. I also think you could be a little more verbose in the message, to make it clear you are correcting two bugs. My suggestion is:

```
src: mk: use ${configurations[arch]} instead of $MAKE_OPTS

We currently pass the $MAKE_OPTS variable to make when 'kw build' is invoked. However, this variable is outdated as it is never defined anywhere. In addition, we now have a ${configurations[arch]} user setting which defines the arch to build but we don't currently pass it to make. So, solving both problems at once, remove the outdated $MAKE_OPTS and give the architecture option instead.

Fixes #63

Signed-off-by: ...
```

What do you think?

src/mk.sh
```diff
@@ -45,8 +45,8 @@ function mk_build
 
   PARALLEL_CORES=$(( $PARALLEL_CORES * 2 ))
```
rodrigosiqueira: If you have time, can you fix this in another PR? See: https://github.com/rodrigosiqueira/kworkflow/issues/85

rodrigosiqueira:
@rodrigorsdc 
Do you have any update here?

rodrigorsdc:
> @rodrigorsdc
> Do you have any update here?

Hi @rodrigosiqueira,

I am going to send a new version solving those nitpick.

> @rodrigorsdc @matheustavares
> 
> Hi,
> 
> First of all, thanks for the patch and the review :)
> 
> Always remember to validate the new feature, if you try `kw b` in a kernel repository, you will see the following error:
> 
> ```
> make -j32 ARCH='x86_64'
> Makefile:670: arch//Makefile: No such file or directory
> make: *** No rule to make target 'arch//Makefile'.  Stop.
> make: *** Waiting for unfinished jobs....
> Makefile:577: arch//Makefile: No such file or directory
> make[1]: *** No rule to make target 'arch//Makefile'.  Stop.
> make: *** [Makefile:695: include/config/auto.conf.cmd] Error 2
> make: *** [include/config/auto.conf.cmd] Deleting file 'include/config/tristate.conf'
> ```
> 
> Probably, you have to add another commit on top of this one, replacing `x86_64` in the following files:
> 
> ```
> documentation/man/kw.rst
> etc/kworkflow.config
> tests/commons_test.sh
> ```

I solve this problem by just replacing `''` by `""` on `make -j$PARALLEL_CORES ARCH='${configurations[arch]}'`. Thanks for pointing this bug.

rodrigosiqueira:
Merged: https://github.com/kworkflow/kworkflow/commit/9d9f09e616eaf15999de222b065ae12cbb7e993c

Thank you for your patch!

PR #108, opened 2019-10-25
src: explore: add support to work with multiple strings

### LINKED ISSUE #91: kw e does not work with two or more strings
`kw e` does not work as expected if we use two or more string as a search pattern. For example:
```bash
kw e "struct lala"
```

rodrigorsdc:
Now "kw explore" supports multiple strings as patterns to be used
    on "git grep". The new rules for "kw explore" are as follow:
    1) If you type two or more arguments and the last argument is a
    valid file/directory, it will perform a git-grep using all arguments
    but last as patterns and the last argument as pathspec. If the last
    argument not a valid file/directory, all argument will be used as
    patterns and pathspec will be ".".
    2) If you type just one argument, it will perform a git-grep using
    that argument as pattern and "." as pathspec.
    
Fixes #91

### DIFF
--- a/src/explore.sh
+++ b/src/explore.sh
@@ -1,4 +1,5 @@
 . $src_script_path/kwio.sh --source-only
+. $src_script_path/kwlib.sh --source-only
 
 function explore()
 {
@@ -15,11 +16,12 @@ function explore()
       (
         local path=${@:2}
         local regex=$1
+	# If user only set regex value
         if [[ $# -eq 1 ]]; then
           path="."
           regex=$1
         fi
-        git grep -e $regex -nI $path
+        cmd_manager "git grep -e \"$regex\" -nI $path"
       );;
   esac
 }
--- a/tests/explore_test.sh
+++ b/tests/explore_test.sh
@@ -17,6 +17,8 @@ function testExplore
   MSG_OUT="camel_case(void)"
   assertEquals "$MSG_OUT" "$(explore "camel_case" tests/samples | cut -d ' ' -f2 | sed -n -e 2p)"
   assertEquals "$LOG_OUT" "$(explore log LICENSE | grep "Initial commit" | awk '{print $1, $2}')"
+  MSG_OUT="Free Software Foundation"  
+  assertEquals "$MSG_OUT" "$(explore "Free Software Foundation" ./LICENSE | grep "Free Software Foundation" -o | head -n 1)"
   true
 }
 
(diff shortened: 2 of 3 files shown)


src/explore.sh
```diff
@@ -13,13 +13,24 @@ function explore()
...
+	  fi
+	else
+	  path="."
+	  expr=$@
+	fi
+	local regex=""
+	for i in $regex; do
+	  regex+="-e ${i} "
+	done
+        git grep $regex -nI $path
```
rodrigosiqueira: Use `cmd_manager` instead of call it here

rodrigosiqueira:
@rodrigorsdc any update here?

matheustavares:
Hi, @rodrigorsdc and @rodrigosiqueira 

Not really following this PR, but let me give my two cents here: `git-grep` has a collection of "logical" flags to specify how to do the search when multiple patterns are given. So, for example, if you run `git grep -e "regexA" -e "regexB"` you will have all the occurrences of the former \*\*or\*\* the latter. If you want only lines where \*\*both\*\* of them appear, you would have to run it as `git grep -e "regexA" --and -e "regexB"`.

As `--and`, there's also `--or` and `--not`, to compose more complex searches. I'm wondering if it's worthy, thought, to recreate all this flags in KW or just use `git-grep` directly when the user wants to have this finer control. What do you think, @rodrigosiqueira ?

rodrigorsdc:
> @rodrigorsdc any update here?

Hi @rodrigosiqueira, I've sent a new version. I hope that this
works well now.

tests/explore_test.sh
```diff
@@ -17,6 +17,8 @@ function testExplore
   MSG_OUT="camel_case(void)"
   assertEquals "$MSG_OUT" "$(explore "camel_case" tests/samples | cut -d ' ' -f2 | sed -n -e 2p)"
   assertEquals "$LOG_OUT" "$(explore log LICENSE | grep "Initial commit" | awk '{print $1, $2}')"
+  MSG_OUT="Free Software Foundation"
```
rodrigosiqueira: Enable your editor to show extra white space or take a look at git diff.

tests/explore_test.sh
```diff
@@ -17,6 +17,8 @@ function testExplore
   MSG_OUT="camel_case(void)"
   assertEquals "$MSG_OUT" "$(explore "camel_case" tests/samples | cut -d ' ' -f2 | sed -n -e 2p)"
   assertEquals "$LOG_OUT" "$(explore log LICENSE | grep "Initial commit" | awk '{print $1, $2}')"
+  MSG_OUT="Free Software Foundation"  
+  assertEquals "$MSG_OUT" "$(explore "Free Software Foundation" ./LICENSE | grep "Free Software Foundation" -o | head -n 1)"
```
rodrigosiqueira: There is no reason to rewrite "Free Software Foundation", just reuse `$MSG_OUT`

PR #109, opened 2019-10-27
src: mk: remove unnecessary multiplication of PARALLEL_CORES

### LINKED ISSUE #85: Fix wrong arguments for make invocation at mk.sh
At `mk.sh`, the function `mk_build()` tries to check the number of processors the user has so that it can invoke `make` with the maximum number of threads possible. However it multiplies this number by 2, which was probably done thinking about hyper-threading. Since the number returned by `nproc` and `grep -c ^processor /proc/cpuinfo` already accounts for this, we should not do this multiplication.

rodrigorsdc:
Since 'nproc --all' and 'grep -c ^processor /proc/cpuinfo' return
the correct number of CPU's even if hype-threading is enable, it
is not necessary to multiply varible PARALLEL_CORES by 2.

Fixes #85

Signed-off-by: Rodrigo Ribeiro Carvalho <rodrigorsdc@gmail.com>

### DIFF
--- a/src/mk.sh
+++ b/src/mk.sh
@@ -43,8 +43,6 @@ function mk_build
     PARALLEL_CORES=$(grep -c ^processor /proc/cpuinfo)
   fi
 
-  PARALLEL_CORES=$(( $PARALLEL_CORES * 2 ))
-
   say "make -j$PARALLEL_CORES $MAKE_OPTS"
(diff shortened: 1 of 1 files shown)


rodrigosiqueira:
Merged, see: https://github.com/rodrigosiqueira/kworkflow/commit/ab9f2369dd7ca8685f771fc4910035a20dec6485

Thanks!

PR #121, opened 2019-11-12
Fix some bugs at setup.sh

matheustavares:
Fix a bug with a failing sed command and a hard-coded string in fish synchronization.


rodrigosiqueira:
Merged, see https://github.com/rodrigosiqueira/kworkflow/commits/unstable

PR #123, opened 2019-11-12
Fix some bugs at checkpatch test

### LINKED ISSUE #122: Fix checkpatch_wrapper test
The `tests/checkpatch_wrapper.sh` has currently some problems which prevents it from working correctly. The minor one is replacing `assertFail` for `assertFalse`. The major one is that the test file didn't learned that kw now uses the local checkpatch.pl script. In this sense, the tests still assume a `~/.config/kw/external/checkpatch.pl` script and should in fact, be failing. The fact that it's succeeding is a bug.

To better understand the problem, we can print the `$res` variable at function `checkpatch`. It always contains `"Neither the given path nor the working path is in a kernel tree."`, not the expected strings.

matheustavares:
This fixes #122

### DIFF
--- a/tests/checkpatch_wrapper_test.sh
+++ b/tests/checkpatch_wrapper_test.sh
@@ -24,9 +24,25 @@ declare -A MSG=( \
     ["check"]=CHECK_MSG \
 )
 
+FAKE_KERNEL="tests/.tmp"
+
+function oneTimeSetUp
+{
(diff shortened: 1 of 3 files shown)


rodrigosiqueira:
Merged https://github.com/rodrigosiqueira/kworkflow/commits/unstable

rodrigosiqueira:
Thank you very much for these fixes!

PR #125, opened 2019-11-18
src: kwio: Make notification run in background

rodrigosiqueira:
We noticed that the command `notify-send` behaves differently in Debian
and ArchLinux. After inspecting the code, we saw that kw does not put
the command in background; for this reason, this commit make kw put a
notification execution in background.


matheustavares:
Nice fix! Merged, see 58cdaff

PR #131, opened 2019-11-26
src: explore: Rework "explore" feature

rodrigosiqueira:
This commit reworks the explore function by:
1. Break the feature in multiple atomic functions
2. Add unit tests that better cover its use
3. Add and update documentation

### DIFF
--- a/documentation/man/kw.rst
+++ b/documentation/man/kw.rst
@@ -101,16 +101,15 @@ This command is a wrapper for **checkpatch**, with the goal of simplifying the
 use of this tool; notice that you can specify a single file or an entire
 directory.
 
-e, explore [log] [*EXPRESSION*] [-p] [*DIRECTORY|FILE*]
+e, explore [--log] [*EXPRESSION*] [-p] [*DIRECTORY|FILE*]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-The *explore* command is based on git grep. It can search for string matches in
-either the git repository contents or in the git log messages. For example, you
-can use **kw e functionName** to find *functionName* in the source directory;
-If you want to search for a composed string, you have to quote your search
-(e.g., **kw e "str1 str2"**). You can also search the git log history by using
-*log* after the *e*; for instance, **kw e log STRING_MATCH**.  Additionally,
-you can use *-p* to see the
-diff in the search.
+The *explore* command is a wrapper to git grep. It can search for string
+matches in either the git repository contents or in the git log messages. For
+example, you can use **kw e functionName** to find *functionName* in the source
+directory; If you want to search for a composed string, you have to quote your
+search (e.g., **kw e "str1 str2"**). You can also search the git log history by
+using *--log* after the *e*; for instance, **kw e --log STRING_MATCH**.
+Additionally, you can use *-p* to see the diff in the search.
 
 m, maintainers [*-a|--authors*] [*DIRECTORY|FILE*]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tests/explore_test.sh
+++ b/tests/explore_test.sh
@@ -5,21 +5,116 @@
 
 function suite
 {
-  suite_addTest "testExplore"
+  suite_addTest "explore_files_under_git_repo_Test"
+  suite_addTest "explore_git_log_Test"
+  suite_addTest "explore_parser_Test"
 }
 
-function testExplore
-{
-  local MSG_OUT="GNU"
-  local LOG_OUT="Initial commit"
+declare -r test_path="tests/.tmp"
 
-  assertEquals "Expected an error message." "Expected path or 'log'" "$(explore)"
-  MSG_OUT="camel_case(void)"
-  assertEquals "$MSG_OUT" "$(explore "camel_case" tests/samples | cut -d ' ' -f2 | sed -n -e 2p)"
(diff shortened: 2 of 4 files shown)


tests/explore_test.sh
```diff
@@ -5,21 +5,111 @@
...
+  suite_addTest "explore_git_log_Test"
+  suite_addTest "explore_parser_Test"
 }
 
-function testExplore
+function setupExplore()
 {
-  local MSG_OUT="GNU"
-  local LOG_OUT="Initial commit"
+  local -r test_path="tests/.tmp"
```
matheustavares: Could we move this variable to outside the function? This way it can also be used in`tearDownExplore()` and `explore_git_log_Test()` without needing to redefine.

tests/explore_test.sh
```diff
@@ -5,21 +5,111 @@
...
+  ID=1
+  MSG_OUT="Expected string or 'log'"
+  output=$(explore)
+  assertEquals "($ID) - Expected an error message." "$MSG_OUT" "$output"
+
+  ID=2
   MSG_OUT="camel_case(void)"
-  assertEquals "$MSG_OUT" "$(explore "camel_case" tests/samples | cut -d ' ' -f2 | sed -n -e 2p)"
-  assertEquals "$LOG_OUT" "$(explore log LICENSE | grep "Initial commit" | awk '{print $1, $2}')"
+  output=$(explore "camel_case" tests/samples | cut -d ' ' -f2 | sed -n -e 2p)
```
matheustavares: Since we now have a fake repo created by `setupExplore()` can we use it instead of kw's repo? I think it will make the test more robust.

tests/explore_test.sh
```diff
@@ -5,21 +5,111 @@
...
+  assertEquals "($ID)" "$MSG_OUT" "$output"
+
+  ID=3
   MSG_OUT="Free Software Foundation"
-  assertEquals "$MSG_OUT" "$(explore "$MSG_OUT" ./LICENSE | grep "$MSG_OUT" -o | head -n 1)"
-  true
+  output=$(explore "$MSG_OUT" ./LICENSE | grep "$MSG_OUT" -o | head -n 1)
```
matheustavares: Same here.

tests/explore_test.sh
```diff
@@ -5,21 +5,111 @@
...
+  ID=2
+  output=$(explore_parser "something")
+  ret="$?"
+  assertEquals "($ID)" "2" "$ret"
+
+  # Others
+  ID=3
+  output=$(explore_parser --logljkl)
+  ret="$?"
+  assertEquals "($ID)" "2" "$ret"
```
matheustavares: I think we should consider all `--[A-Za-z]` as invalid options. Then, if the user really wants to use it as a search regex, they can use `\--[A-Za-z]`. The advantage is that it will be then possible to search for the string "\--log", for example, without confusing it with the option.

documentation/man/kw.rst
```diff
@@ -101,16 +101,15 @@ This command is a wrapper for **checkpatch**, with the goal of simplifying the
 use of this tool; notice that you can specify a single file or an entire
 directory.
 
-e, explore [log] [*EXPRESSION*] [-p] [*DIRECTORY|FILE*]
+e, explore [--log] [*EXPRESSION*] [-p] [*DIRECTORY|FILE*]
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The *explore* command is based on git grep. It can search for string matches in
```
matheustavares: While we are here, what do you think of changing this line to "*explore* command is a wrapper to git grep", instead of "based on" (which gives the idea that we re-implemented the search)?

tests/explore_test.sh
rodrigosiqueira: I did not get your idea. The setup function creates a fake project with fake commits, and it does not create any real file for simulate search. Notice that we search for something in a file in this test, I can't see any robustness advantage by making this file isolate. I believe that we are safer by using our samples.

rodrigosiqueira:
Hi @matheustavares , thank you very much for your review. I made some improvements in the test based on your comments, and I replied to some of your questions inline.

tests/explore_test.sh
matheustavares: What I meant is that this test performs a search in kw's repository and not in the fake repository we created for the tests. Since we already have the fake repo, I suggest changing the above lines for something like:

```
MSG_OUT="file_1.txt:1:This is file 1"
output=$(explore "is file 1" "$test_path/1")
```

The advantage is that we don't have to rely on kw's repo for the test, so if we have to change the `tests/samples` or `LICENSE` (as in the other test bellow) we don't have to adjust the tests as well.

tests/explore_test.sh
matheustavares: Any thought on this idea, @rodrigosiqueira ?

tests/explore_test.sh
rodrigosiqueira: I've one idea about this, I'll implement it today and update this PR

tests/explore_test.sh
rodrigosiqueira: Try `kw e --log "--log"`in the linux kernel, it works :)
Actually, I'm just testing the correct use of the flag `--log` in this test.

tests/explore_test.sh
rodrigosiqueira: Done :)

rodrigosiqueira:
Applied: https://github.com/kworkflow/kworkflow/commit/6f48f1796510571e6e2f74d5d9bcdb784f2b75e9

PR #132, opened 2019-11-29
src: checkpatch: Rework checkpatch_wrapper

### LINKED ISSUE #43: Improve 'kw c' output
If there is no code style problem in the file, kw shows something like this per file:

```

```

We have to handle it with two approaches:

1. Add a verbose option in the code style, that produces something like this:
```
kw c -v lalala.c
lalala.c: No checkpatch problem
```

2. Show nothing if the file is clean:

```
> kw c xpto.c
>
```

rodrigosiqueira:
We had many improvements in kw, and this commit brings some of these
enhancements to checkpath_wrapper. By doing it, we also rework part of
the code logic, added documentation, and improve the test coverage.

Closes #43

### DIFF
--- a/src/checkpatch_wrapper.sh
+++ b/src/checkpatch_wrapper.sh
@@ -1,9 +1,16 @@
+# Checkpatch is a useful tool provided by Linux, and the main goal of the code
+# in this file is to handle this script in a way to make this tool easier for
+# users.
+
 . $src_script_path/commons.sh --source-only
 . $src_script_path/kwlib.sh --source-only
 
+# Runs checkpatch in the given path, which might be a file or directory.
+#
+# @FILE_OR_DIR_CHECK Target path for running checkpatch script
 function execute_checkpatch()
 {
-  local FILE_OR_DIR_CHECK=$1
+  local FILE_OR_DIR_CHECK="$1"
 
   #TODO: Note that codespell file is not specified yet because of the poluted
   # output. It could be nice if we can add another option just for this sort
@@ -12,54 +19,58 @@ function execute_checkpatch()
 
   local -r options="--terse --no-tree --color=always -strict --file "
   local -r script="scripts/checkpatch.pl $options"
-  local -r original_working_dir=$PWD
+  local -r original_working_dir="$PWD"
   local kernel_root=""
 
-  if [[ -z $FILE_OR_DIR_CHECK ]]; then
+  if [[ -z "$FILE_OR_DIR_CHECK" ]]; then
     FILE_OR_DIR_CHECK="."
   fi
 
-  FILE_OR_DIR_CHECK="$(realpath $FILE_OR_DIR_CHECK)"
   # Check if is a valid path
-  if [[ ! -d $FILE_OR_DIR_CHECK && ! -f $FILE_OR_DIR_CHECK ]]; then
+  if [[ ! -d "$FILE_OR_DIR_CHECK" && ! -f "$FILE_OR_DIR_CHECK" ]]; then
     complain "Invalid path"
-    return 1
+    return 22 # EINVAL
   fi
 
-  # try to find kernel root at given path
+  # Get realpath for using inside checkpatch
+  FILE_OR_DIR_CHECK="$(realpath $FILE_OR_DIR_CHECK)"
+
+  # Try to find kernel root at given path
   kernel_root="$(find_kernel_root $FILE_OR_DIR_CHECK)"
   if [[ -z "$kernel_root" ]]; then
-    # fallback: try to find kernel root at working path
+    # Fallback: try to find kernel root at working path
     kernel_root="$(find_kernel_root $original_working_dir)"
   fi
 
-  # Check if kernel root was found.
+  # Check if kernel root was found
   if [[ -z "$kernel_root" ]]; then
     complain "Neither the given path nor the working path is in a kernel tree."
-    return 1
+    return 22 # EINVAL
   fi
+
   # Build a list of file to apply check patch
-  FLIST=`find $FILE_OR_DIR_CHECK -type f ! -name '*\.mod\.c' | grep "\.[ch]$" `
+  FLIST=`find $FILE_OR_DIR_CHECK -type f ! -name '*\.mod\.c' | grep "\.[ch]$"`
+
   say "Running checkpatch.pl on: $FILE_OR_DIR_CHECK"
-  say $SEPARATOR
+  say "$SEPARATOR"
 
-  for current_file in $FLIST
-  do
-    file=$current_file
+  cmd="perl $script "
+  for current_file in $FLIST; do
+    file="$current_file"
 
-    if [ ! -e "$file" ]
-    then
-      echo "$file does not exist."; echo
+    if [[ ! -e "$file" ]]; then
+      echo "$file does not exist."
       continue
     fi
 
-    cd $kernel_root
-    eval perl $script $file
-    cd $original_working_dir
+    cd "$kernel_root"
 
-    if [ $? != 0 ]; then
-      say $SEPARATOR
+    cmd_manager "SILENT" "$cmd $file"
+
+    if [[ "$?" != 0 ]]; then
+      say "$SEPARATOR"
     fi
 
+    cd "$original_working_dir"
   done
 }
--- a/tests/checkpatch_wrapper_test.sh
+++ b/tests/checkpatch_wrapper_test.sh
@@ -5,16 +5,21 @@
 
 function suite
 {
-  suite_addTest "testWarning"
-  suite_addTest "testError"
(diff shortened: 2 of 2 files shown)


src/checkpatch_wrapper.sh
```diff
@@ -12,54 +20,56 @@ function execute_checkpatch()
...
   if [[ -z "$kernel_root" ]]; then
     complain "Neither the given path nor the working path is in a kernel tree."
-    return 1
+    return 22 # EINVAL
   fi
+
   # Build a list of file to apply check patch
-  FLIST=`find $FILE_OR_DIR_CHECK -type f ! -name '*\.mod\.c' | grep "\.[ch]$" `
+  cmd="find $FILE_OR_DIR_CHECK -type f ! -name '*\.mod\.c' | grep \"\.[ch]$\""
+  FLIST=$(cmd_manager "SILENT" "$cmd")
```
matheustavares: Since this is not a command we will present to the user nor use in tests, shouldn't we perhaps just invoke it directly? (without `cmd_manager`)

src/checkpatch_wrapper.sh
```diff
@@ -12,54 +20,56 @@ function execute_checkpatch()
...
-    cd $kernel_root
-    eval perl $script $file
-    cd $original_working_dir
+    cd "$kernel_root"
 
-    if [ $? != 0 ]; then
-      say $SEPARATOR
-    fi
+    cmd_manager "SILENT" "$cmd $file"
+    say "$SEPARATOR"
```
matheustavares: In #43 you mentioned no output would be preferable when there's no codestyle errors. I.e. something like:

```
$ kw c drivers/gpu/drm/vkms
$
```

However, with the patch, we may get outputs such as:

```
$ kw c drivers/gpu/drm/vkms
Running checkpatch.pl on: linux/drivers/gpu/drm/vkms
```

For me it's OK like that! I just wanted to comment this to see if it is OK for you too.

tests/checkpatch_wrapper_test.sh
```diff
@@ -26,7 +31,19 @@ declare -A MSG=( \
     ["check"]=CHECK_MSG \
 )
 
-FAKE_KERNEL="tests/.tmp"
+# Note:
+# This function explores some exciting features from bash that deserve a
+# highlight. Here we have the use of indirect references in ${!MSG[$1]}
+# (indicated by "!"), we basically access values in the array MSG that, in its
+# turn, evaluate the constant value (WARNING_MSG, ERROR_MSG, etc).
```
matheustavares: Nice note haha :)

tests/checkpatch_wrapper_test.sh
```diff
@@ -42,31 +59,73 @@ function oneTimeTearDown
...
+  build_fake_path="$RANDOM/$RANDOM/$RANDOM/xptolala"
+
+  output=$(execute_checkpatch "$build_fake_path")
+  ret="$?"
+
+  [[ "$ret" != 22 ]] && fail "We forced an invalid path and we expect an error"
+}
+
+function no_kernel_directory_Test
+{
```
matheustavares: I really like this test! What do you think of adding one more to test `kw c` execution when the working dir is a kernel dir, but the given path is not? I.e. suppose I run `kw c /tmp/0001-testing-commit.patch` inside a kernel directory. Kw should succeed, using the `checkpatch.pl` script from the said dir. It would be nice to test this behavior, as well.

(We already test the case where the working dir is not a kernel dir but the given path is)

tests/checkpatch_wrapper_test.sh
```diff
@@ -42,31 +59,73 @@ function oneTimeTearDown
...
+  local build_fake_path
+  local output
+  local ret
+
+  build_fake_path="$RANDOM/$RANDOM/$RANDOM/xptolala"
+
+  output=$(execute_checkpatch "$build_fake_path")
+  ret="$?"
+
+  [[ "$ret" != 22 ]] && fail "We forced an invalid path and we expect an error"
```
matheustavares: This test fails in my machine. Probably it is because of my shunit version, which interprets tests returning false as failed. As a solution, you could add a line with `true` in the end. Or, to avoid this hacky solution, refactor the above to:
```
[[ "$ret" == 22 ]]; assertTrue "We forced an invalid path and we expect an error" $?
```

tests/checkpatch_wrapper_test.sh
```diff
@@ -42,31 +59,73 @@ function oneTimeTearDown
...
+
+  # Reference: https://www.tutorialkart.com/bash-shell-scripting/bash-split-string/
+  s="$output$delimiter"
+  while [[ "$s" ]]; do
+    array+=( "${s%%"$delimiter"*}" );
+    s=${s#*"$delimiter"};
+  done;
+
+  size="${#array[@]}"
+  [[ "$size" -lt "2" ]] && fail "We could not find more then two SEPARATOR sequence"
```
matheustavares: Same as previously.

tests/checkpatch_wrapper_test.sh
rodrigosiqueira: I think that I did not get your idea. We already test a valid case indirectly in the first set of tests (error_Test, correct_Test, etc). Maybe I got the wrong idea? Are you talking about the path in front of `kw c`?

matheustavares: In the fist set of tests we run `kw c` on a file _inside_ the kernel tree but _invoking from outside_ . It would be nice to have another test for the other way arround: i.e. running `kw c` _from within_ a kernel tree and passing a file that is _outside_. Something like:

```shell
# pseudocode:
local fake_kernel_tree = tests/.tmp
local not_a_kernel_tree = tests/.tmp2
mk_fake_kernel_root "$fake_kernel_tree"
cp -r tests/samples "$not_a_kernel_tree"

cd "$fake_kernel_tree"
$(execute_checkpatch) "$not_a_kernel_tree"
```

tests/checkpatch_wrapper_test.sh
```diff
@@ -42,31 +59,80 @@ function oneTimeTearDown
...
+  local ret
+
+  build_fake_path="$RANDOM/$RANDOM/$RANDOM/xptolala"
+
+  output=$(execute_checkpatch "$build_fake_path")
+  ret="$?"
+  if [[ "$ret" != 22 ]]; then
+    assertFalse "We forced an invalid path and we expect an error" "$?"
+  fi
+}
```
matheustavares: Hmm, it seems a little odd to use `assertFalse` inside the if statement, since the condition was already checked. You could replace `assertFalse` simply by `fail <message>` since the code will only get inside it when the test fails.

However, I would prefer to refactor this to:
```shell
execute_checkpatch "$build_fake_path"
ret = $?
assertTrue "We forced an invalid path and we expect an error" '[[ $ret == 22 ]]'
```

(I.e. to use `assertTrue`, giving the condition string. Also removed `output` since it's not used and removed the quotes from `$ret` and `$?` since they will always be a single number)

tests/checkpatch_wrapper_test.sh
```diff
@@ -42,31 +59,80 @@ function oneTimeTearDown
...
+
+  # We want to force an unexpected condition, because of this we change the
+  # basic setup but we rebuild it at the end of the test
+  oneTimeTearDown
+
+  output=$(execute_checkpatch "$sample_one")
+  ret="$?"
+  if [[ "$ret" != 22 ]]; then
+    assertFalse "We forced an invalid path and we expect an error" "$?"
+  fi
```
matheustavares: Same thing here, i.e.:

```shell
execute_checkpatch "$build_fake_path"
ret = $?
assertTrue "We forced an invalid path and we expect an error" '[[ $ret == 22 ]]'
```
(and removing `output`)

tests/checkpatch_wrapper_test.sh
```diff
@@ -42,31 +59,80 @@ function oneTimeTearDown
...
+    array+=( "${s%%"$delimiter"*}" );
+    s=${s#*"$delimiter"};
+  done;
+
+  size="${#array[@]}"
+  # We use three here because we expect one $SEPARATOR from the beginning and
+  # other from s="$output$delimiter"
+  if [[ "$size" -lt "3" ]]; then
+    assertTrue "We could not find more then two SEPARATOR sequence"
+  fi
```
matheustavares: And here. I.e. replace to:

```shell
assertTrue "We could not find more then two SEPARATOR sequence" '[[ "$size" -lt "3" ]]'
```

rodrigosiqueira:
Hi @matheustavares I applied all your suggestions for this reason I applied the patch, see: 22c9d2f06d1c33f446795f80fdbb1d040a3c58f2

Thanks a lot for the great review!

PR #134, opened 2019-12-13
Implement system-wide installation

### LINKED ISSUE #50: Make kw instalation closer to unix standards
kw is currently being installed at ~/.config/kw, which dosen't seem so good (since it's a dir for configurations, not code). Shouldn't we install kw somewhere else, such as ~/.kw, for example?

Also, we could add kw to PATH (or create an alias) instead of loading it to bash. This change would also make the subshells in kw.sh unnecessarily and the variables "src_script_path" and "external_script_path" not visible externally, which is good.

(proposed by @capellaresumo and I)

rodrigorsdc:
This patch add a Makefile with a few targets in order to add
a system-wide installation.

Because of the new install destination, we modify several files
to ensure a correct files paths and correctness of tests.

Now every call of 'kw' will trigger a verification whether user's HOME
has a .config/kw directory. If not, that dir will be create with
kworkflow.config file.

Fixes #50

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,42 @@
+PREFIX = ${DESTDIR}/usr
+STDDIR = ${PREFIX}/share/kw
+MANDIR = ${PREFIX}/share/man
+ACDIR = ${PREFIX}/share/bash-completion/completions
+
+all:
+	@echo "kworkflow is a shell script program. Try \"make install\""
+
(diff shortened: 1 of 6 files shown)


rodrigosiqueira:
@rodrigorsdc I'll close this PR based on your interaction with @matheustavares

PR #137, opened 2019-12-17
docs: typo-fix and update outdated example on config file

matheustavares:


rodrigosiqueira:
Merged, see https://github.com/kworkflow/kworkflow/commits/unstable

Issue #127, opened 2019-11-20
Make `tests/explore_test.sh` don't rely on kw's own repo


matheustavares:
This test file check `kw e` using kw's repository. However, in a shallow clone (as done by travis-ci) the whole log is not available. Also, it would be better if we didn't rely in our own commit history for the test, but instead, create a fake repository.

rodrigosiqueira:
see https://github.com/kworkflow/kworkflow/pull/131

Issue #126, opened 2019-11-20
Fix `tests/kwlib_test.sh` usage of `pwd`


matheustavares:
We test `pwd --help` output at `tests/kwlib_test.sh`. However, this command returns "pwd: pwd [-LP] ..." on newer versions of and "bash: pwd: --: invalid option" on older ones, making the test fail. In this sense, we should refactor this test to be more robust.

rodrigosiqueira:
See: https://github.com/kworkflow/kworkflow/commit/d4b1d33bbcec62a8b0caf7717a5f9499bb2f9bf2

matheustavares:
The main goal with this issue was to allow travis (which used to run on an older Ubuntu version, with our old configurations) to correctly run the said test case. With d4b1d33, however, @rodrigosiqueira already fixed this problem by upgrading the Ubuntu version in our `.travis.yaml` config file.

However, it would still be nice to avoid using the mentioned `pwd` execution as it would still fail in older Ubuntu's. So let's re-open the issue, but removing the kworkflow-alpha milestone (since it's no longer so urgent) and updating the description.

PR #144, opened 2020-01-11
.travis.yml: re-enable email notifications

matheustavares:
We previously disabled them because of issues #126 and #127, which would
make travis-ci always fail with the state of the codebase at that time.
Since the problem was already solved (even though #126 is still to be
solved), let's revoke the email notification prohibition.

Note: we should also consider whether we want email notifications on failures. I personally quite like it, but what do you think, @rodrigosiqueira ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,8 @@ script:
   - ./setup.sh --html
   - ./setup.sh -i
 
-# Disable notifications by email for now are there're some issues to be solved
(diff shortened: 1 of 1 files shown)


rodrigosiqueira:
First of all, thank you very much for this patch!

Take a look at your editor because you left a bunch of white spaces at the end of each line. I made the fixes and merged the commit.
See 741b655e6c7f95a150fbb5a7f20c4017d83f8aed
Thanks!

PR #145, opened 2020-01-11
Add logo and build status to README

matheustavares:
See the result here: https://github.com/matheustavares/kworkflow/blob/logo-readme/README.md

Note: second commit perhaps introduces a not so pretty hack. Take a look at commit message.


README.md
```diff
@@ -1,4 +1,4 @@
-# Kworkflow
+[![kworkflow](images/kw_logo.svg)](https://github.com/kworkflow/kworkflow/)
```
rodrigosiqueira: Why not use png? Additionally, how about use kw instead of kworkflow? We'll change it soon.

rodrigosiqueira:
Thanks!

Merged, see: https://github.com/kworkflow/kworkflow/commit/1109eb06934c647cddf11feca9738eece5d703ee

PR #167, opened 2020-03-11
kwlib: Rework get_from_colon to make it generic

rodrigosiqueira:
We have the function get_from_colon() which is intended to handle
strings that have ':' character, this is useful for handling IP:PORT
format. However, the nature of this function could be easily promoted to
something more generic in order to make it more powerful; for this
reason, this commit rename get_from_colon() to get_based_on_delimiter()
and adds a new parameter that indicates the character delimiter that
should be used. Rework this function also solve the issue of handling
hostname in the parameter `ssh_ip` on `kworkflow.config`, after this
change users can finally use `ssh_ip=username@hotname`.

### DIFF
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -1,33 +1,44 @@
 # NOTE: src/kw_config_loader.sh must be included before this file
 
+# A common task used inside kw is a string separation based on a delimiter, for
+# this reason, this function tries to handle this scenario by getting a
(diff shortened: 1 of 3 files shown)


rodrigosiqueira:
Merged

https://github.com/kworkflow/kworkflow/commit/06b07190fcdd908f118abf11fdd83c203d999cd6

PR #168, opened 2020-03-28
src: mk: Rework parse operation

rodrigosiqueira:
Currently, the parse option is spread around multiple functions which
make the code highly coupled and hard to maintain. This commit reworks
the parser option and other details inside the mk file in order to make
it simple.


rodrigosiqueira:
Merged e99d0d92185f898324afc8e933c448fda4e7d6c9

PR #174, opened 2020-04-07
mk: Add cleanup function for kw deploy

### LINKED ISSUE #170: kw deploy takes a lot of disk space
`kw deploy` creates a directory for centralizing all files that we need to use for updating a kernel in a remote machine. After one user gave me feedback, I realize that `kw` does not clean these files which explode the disk usage after a long time of use. Make `kw deploy` clean all temporary files after a kernel deploy finish.

rodrigosiqueira:
kw deploy creates a directory for centralizing all files that we need to
use for updating a kernel in a remote machine. We don't need to keep
these files in the hard disk for a long time, at the end of the deploy
we can remove it. This commit adds a function that cleans all temporary
files after a kernel deploys finish.

Closes #170


src/remote.sh
```diff
@@ -121,15 +121,15 @@ function prepare_host_deploy_dir()
   # We should expect the setup.sh script create the directory $HOME/kw.
   # However, does not hurt check for it and create in any case
   if [[ ! -d "$kw_dir" ]]; then
-    mkdir "$kw_dir"
+    mkdir -p "$kw_dir"
```
matheustavares, line 124: `.cache` is indeed an ephemeral directory (users might manually delete its contents at any time). To avoid problems in the future, when other commands might want to use cache, what if we always create the dir at kw startup? Not only `.cache/kw` but also `.cache` itself (i.e. `mkdir -p `$HOME/.cache/kw`). Then, we could remove the dir creation from setup.sh (which IMHO, makes sense, since the cache is supposed to exist only when running kw)

It would also be nice to have something like `kw clear-cache`, to delete all `.cache/kw` when the user desires to (for storage saving or debugging, for example)

src/mk.sh
```diff
@@ -75,6 +75,26 @@ function get_kernel_release
...
 }
 
+# When kw deploy a new kernel it creates temporary files to be used for moving
+# to the target machine. There is no need to keep those files in the user
+# machine, for this reason, this function is in charge of cleanup the temporary
+# files at the end.
+#
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/kwlib.sh` function `cmd_manager`
+function cleanup_after_deploy
```
matheustavares, line 85: If the user aborts execution (with `ctrl-c`, for example), the cache doesn't get removed and the next execution might get dirty files. I think we could add two changes to help avoiding this:

1. Use bash's `trap` command, to trap termination signals and cleanup cache before exiting.
2. Always call `cleanup_after_deploy` **before** deploy as well (and maybe rename the func?)

src/remote.sh
rodrigosiqueira: Really nice idea! I add this in the next version

src/mk.sh
rodrigosiqueira: Uouuu I love the idea of handling trap! However, I believe that add traps should be a separate patch since we have to handle other cases. I'm going to create an issue for support to trap.

rodrigosiqueira:
Merged
See 36650382c244ea97561c001f81fa1ca83b776c5d

PR #176, opened 2020-05-04
kw: small issue on u/mount alert message

melissawen:
After u/mount command an alert is exibited, however the alert
showed a message of -build (sic) completion-. This patch solve this
small issue.

### DIFF
--- a/kw
+++ b/kw
@@ -34,7 +34,7 @@ function kw()
 
         vm_mount
         local ret=$?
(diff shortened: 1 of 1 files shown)


melissawen:
Hi! I found a small problem on alert message exhibited after the conclusion of two commands (mount and umount). @rodrigosiqueira

rodrigosiqueira:
Thanks for your patch.
Merged.

Issue #139, opened 2019-12-28
Deploy new Kernel image for QEMU VM with libguestfs


rodrigosiqueira:
Currently, we use libguestfs for mounting a QEMU VM as a file system for install modules; however, do the same for update the Kernel image is not a straightforward task. We need to execute the following steps:
1. Mount the target image;
2. Copy the binary file to `/boot/`;
3. Generate the initramfs file;
4. Update grub;

Step 3 and 4 have a bunch of complications. For generating an initramfs in Arch Linux without root we can use something like this:

`$ mkinitcpio -k 5.4.0-rc4-VKMS-EXPERIMENTS2+ -r /home/siqueira/p/mount -c ~/p/mount/etc/mkinitcpio.conf -g /tmp/XPTO/initramfs-5.4.0-rc4-VKMS-EXPERIMENTS2+.img`

However, I'm facing a weird problem during the boot because the `img` file uses my current user for building the binary; however, it should use root. If I do it as a root user, everything works fine. I don't know how to fix it yet, and I also don't know if `update-initramfs` has better support for this sort of operation.

Another problem is the Grub update, more precisely, this command:
`grub-mkconfig -o /boot/grub/grub.cfg`This command does not have many options for generating a `grub.cfg` based on other mount points or similar. My workaround for this consists of manually updating the `grub.cfg` from the VM, and when the user updates the VM system, grub will automatically detect the initramfs and kernel image.

Related to: #133

PR #177, opened 2020-05-09
Vm update grub

melissawen:
Hi, I'm working in overcome a problem on vm deploy. I suggest the use of guestfish setup and commands to run update grub command; however, this patch still lacks generalization and was applied in a debian-to-debian environment. 
I think it could help to solve the issue #139 
@rodrigosiqueira


rodrigosiqueira:
We already discussed this pull request on another communication channel. For this reason, I'm going to close this PR.

Issue #61, opened 2019-01-25
Improve kw e by adding alternative tools in case that git cannot handle the situation


rodrigosiqueira:
"kw e" relies on git for searching things which is great. However, sometimes we want to explore things in a bunch of files that not use git. In this sense, it could be nice if "kw e" support alternative tools for search things. For example, we can have`grep -nrwI /PATH -e EXPRESSION` if the current directory is not under git control.

Issue #92, opened 2019-04-15
Improve kw explore semantics.


rodrigosiqueira:
Current, `kw explore` work by following this pattern:

```
kw e, explore [log] [*EXPRESSION*] [-p] [*DIRECTORY|FILE*] 
```

Note that [-p] depends on [log], but the current implementation does not force this dependency. We want something like this:

```
kw e, explore [log [-p]] [*EXPRESSION*] [*DIRECTORY|FILE*] 
```

PR #178, opened 2020-05-14
src: add grep utility to explore feature

melissawen:
This patch expands the feature -explore- adding the grep utility
for searching plain-text data in files that are not under git.
Related to #61 

### DIFF
--- a/src/explore.sh
+++ b/src/explore.sh
@@ -19,7 +19,15 @@ function explore()
       shift 1 # Remove 'log' string
       explore_git_log "$@"
     ;;
-    2) # Search in a git repository
+    2) # Use GNU GREP
+      shift 1 # Remove 'grep' string
+      explore_files_gnu_grep "$@"
+    ;;
+    3) # Search in directories controlled or not by git
+      shift 1
+      explore_all_files_git "$@"
+    ;; 
+    4) # Search in files under git control
       explore_files_under_git "$@"
     ;;
     *)
@@ -35,20 +43,37 @@ function explore()
 # Returns:
 # This function returns the following values:
 #   1 if --log is passed
-#   2 for something different from --log
+#   2 if --grep is passed
+#   3 if --all is passed
+#   4 for something different from --log or --grep
 #   22 for invalid operation
 function explore_parser()
 {
+  local option="$1"
+
   if [[ "$#" -eq 0 ]]; then
-    complain "Expected string or 'log'"
+    complain "Expected string or parameter. See man for detail."
     exit 22 # EINVAL
   fi
 
-  if [[ "$1" =~ ^--log$ && ! -z "$2" ]]; then
-    return 1
+  if [[ -z "$2" ]]; then
+    return 4
   fi
 
-  return 2
+  case "$option" in
+    --log|-l)
+      return 1
+    ;;
+    --grep|-g)
+      return 2
+    ;;
+    --all|-a)
+      return 3
+    ;;
+    *)
+      return 4
+    ;;
+  esac
 }
 
 # This function is responsible for handling the search in the log history.
@@ -69,8 +94,7 @@ function explore_git_log()
   cmd_manager "$flag" "git log --grep=\"$search_string\" $path"
 }
 
-# This function manages the search inside files and under the git repository
-# control.
+# This function searches string in files under git control.
 #
 # @regex Specifies the regex that we want to search in the files
 # @path Narrow down the search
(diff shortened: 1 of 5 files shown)


melissawen:
@rodrigosiqueira

src/explore.sh
```diff
@@ -40,15 +44,19 @@ function explore()
...
-    complain "Expected string or 'log'"
+    complain "Expected string or 'log' or 'grep'"
     exit 22 # EINVAL
   fi
 
   if [[ "$1" =~ ^--log$ && ! -z "$2" ]]; then
     return 1
   fi
+  
+  if [[ "$1" =~ ^--grep$ && ! -z "$2" ]]; then
```
rodrigosiqueira, line 55: in this case, it could be nice if we also support `-g`, i.e., `kw e -g lala` or `kw e --grep lala`.

src/explore.sh
```diff
@@ -90,3 +98,18 @@ function explore_files_under_git()
...
+  local path="$2"
+  local flag="$3"
+
+  # Silent by default
+  flag=${flag:-"SILENT"}
+
+  # If user only set regex value
+  path=${path:-"."}
+
+  cmd_manager "$flag" "grep -nrwI $path -e \"$regex\""
```
rodrigosiqueira, line 114: Since you decided to use `cmd_manager` you can use the flag `TEST_MODE` for implementing a new test that validate the command correctness. Take a look at `mk_test`, test `mk_kernel_uninstall_Test` for reference.

tests/explore_test.sh
```diff
@@ -37,6 +38,8 @@ function setUp()
     git add "$file_name" &> /dev/null
     git commit -m "Commit number $commit" &> /dev/null
   done
+  
+  echo "##regex_sample##" > e_grep_test.txt
```
rodrigosiqueira, line 42: To make this test a little bit more robust, I recommend you to add a full text in `tests/samples` and use it for validating your grep function.

rodrigosiqueira:
Also, since you're working at this part of the code could you also take a look at this issue: https://github.com/kworkflow/kworkflow/issues/92

src/explore.sh
matheustavares, line 114: We currently don use the `-w` flag in the "git grepping mode". So, for consistency, I think we should remove it here or add it in the git-grep search as well. Personally, I prefer not to use `-w`, since we wouldn't allow users to search for "non-words". And they can always resort to something like `"\bpattern\b"`, to simulate the same effect, without `-w`.

matheustavares:
Hi, @melissawen and @rodrigosiqueira . First of all, thanks for the new feature :)

Two ideas came to my mind, regarding implementation. Please, let me know what you think about them:

- `git-grep` has an option to search outside Git repositories: the `--no-index` flag. So we could keep using `git-grep` for this new case as well. One advantage is that we can ensure the same behavior among all our "explore" cases, since we would be using the same program. Also, I think (but not sure) `git-grep` should be faster than `GNU grep` for some cases.

- What if, instead of providing `--grep`, we automatically detect when we are inside/outside a repository's working tree and use the respective grepping option, in either case? We could do this test with a function like the following:

```bash
function in_git_work_tree()
{
  local ret="$(git rev-parse --is-inside-work-tree 2>&1)"
    if [[ $? == 0 && "$ret" == "true" ]]; then
      return 0
    fi
    return 1
}
```

src/explore.sh
melissawen: I agree with the homogeneity for the feature as a whole. So, I will remove -w in my next version

melissawen:
Hi @matheustavares,

I appreciate your suggestion with --no-index. For me, I can refactor this patch (or do another one) to use this parameter, as I don't see another employment for grep utility other than searching in files that are not under git control. 
Siqueira, what do you think?

rodrigosiqueira:
Hi @melissawen and @matheustavares

Great idea Matheus! I believe that `in_git_work_tree
` is a very useful function and should be in `kwlib` for making it available for other parts of kw.

Additionally and IMHO it does not hurt to provide an extra option for letting users explicitly say that they want to search something not under git control. For example, some times, I want to search on some metadata in the Linux kernel (under git control) that I know that is not under git control.

melissawen:
I agree. I think the user should decide to search in files under or not under git control. But I think we can handle it only using git tool, with --no-index (without adding grep tool). Is it good?

rodrigosiqueira:
I recommend that you follow Matheus's suggestion, but also add the option for users to ignore git repository (optional). Finally, I also agree to use --no-index.

matheustavares:
Oh, there is one caveat I just noticed: `--no-index` greps all files in the current directory, but if there is a `.git`, it ignores it. So if we want to allow users to search inside `.git`, as well, we might have to use `GNU grep`.

matheustavares:
Oops, I accidentally hit "Close and Comment". Sorry for that! Reopening...

melissawen:
Hi,

Just feedback:
I'm working on it; however, I saw a problem in_git_work_tree, since rev-parse returns a fatal error (instead of false) when in a folder that is not under git control (as reported at https://groups.google.com/forum/#!topic/git-users/dWc23LFhWxE). Maybe I will take a while to find a solution. :)

matheustavares:
Hi @melissawen. Yes, `git rev-parse` will die with an exit code different than 0 when outside a Git repository. But the `is_git_work_tree` function  should already deal with this case, in the `if [[ $? == 0 ..` section, right? Or is this check falling in your tests?

melissawen:
Hi @matheustavares, the original function failed my tests. However, I adjusted it to deal with the problem :) I will send everything after applying all the suggestions. Thanks :)

rodrigosiqueira:
Hi @melissawen 

Thanks a lot for your patch! It is already merged, see at:

https://github.com/kworkflow/kworkflow/commit/d131585d7911d2ec58c23a6c364b92cf7502464c

PR #194, opened 2020-09-26
Setup improvement

rodrigosiqueira:


setup.sh
```diff
@@ -60,6 +82,49 @@ function check_dependencies()
...
+function update_path()
+{
+  local new_path=""
+
+  IFS=':' read -ra ALL_PATHS <<< "$PATH"
+  for path in "${ALL_PATHS[@]}"; do
+    [[ "$path" -ef "$binpath" ]] && return
+  done
+
+  echo "PATH=$HOME/.local/bin:\$PATH" >> "$HOME/.bashrc"
```
melissawen, line 120: Would be 'echo "PATH=\$PATH:$kwbinpath" ' ?

melissawen: Even after modifying it, as mentioned above, kw command is not found. It is still looking for this kind of pattern /home/$USER/.config/kw/ ... not sure if there is some trash from the current version or what, but I got stuck on it.

melissawen (commented):
Hi! I got stuck on setup (install/uninstall) and had little progress. Could you check what I have pointed out?

melissawen:
Hi,
I checked the commands: maintainers, code style, mount, umount, explore, up, ssh, build, and deploy; and all of them worked smoothly. For this, I call kw directly, I mean, $.local/bin/kw
Let me know if you checked the setup issues that I have pointed out before (or if something is related to my environment)

melissawen:
Hi,

I've tested with the last update, and it works smoothly.

rodrigosiqueira:
Merge, see https://github.com/kworkflow/kworkflow/commit/bdcb009b4d5bb37133bdd204a7144a5640647fd9

PR #196, opened 2020-12-05
src: mk: Add support for cross-compilation

### LINKED ISSUE #95: Cross-compile option
kw could have support for cross-compilation, we should start by adding arm compilation. We had some legacy code like the one described below that tried to handle this scenario.

```
  if [ $TARGET == "arm" ] ; then
    export ARCH=arm CROSS_COMPILE="ccache arm-linux-gnu-"
  fi
```

rodrigosiqueira:
This commit introduces cross-compilation support to kw by handling new
entries in the kworkflow.config and a new build option for dealing with
menuconfig.

Closes #95


rodrigosiqueira:
Merged, see 4a56cde97f8d564d3c7289a2c6d68507b8c6f648

PR #201, opened 2020-12-26
src: config_manager: Update list and remove API

rodrigosiqueira:
Currently, we use `--ls` and `--rm`, which does not make sense when
compared with other kw commands. This commit replaces `--ls` for
`--list` or `-l` and `--rm` for `--remove` or `-rm`.


rodrigosiqueira:
Merged

PR #203, opened 2021-01-16
plugins/kernel_install: small fix for vm deployment

melissawen:
Copy the kernel image to the correct path where the vm is mounted

### DIFF
--- a/src/plugins/kernel_install/arch.sh
+++ b/src/plugins/kernel_install/arch.sh
@@ -133,7 +133,7 @@ function install_kernel()
 
   if [[ ! -z "$target" ]]; then
     [[ -z "$architecture" ]] && architecture="x86_64"
-    cmd="$sudo_cmd cp -v arch/$architecture/boot/$kernel_image_name /boot/vmlinuz-$name"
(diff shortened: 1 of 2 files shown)


melissawen:
Hi @rodrigosiqueira,

I recently reinstalled the upstream version and noted the kernel image was not updated when running kw d --vm due to a permission error in the copy image step. Looks like the problem was caused by an error in the destination path.

I use Debian on VM and host, and this fix works properly; due to the similarity, I also applied this fix to arch env.

rodrigosiqueira:
Hi @melissawen , first of all, thanks for your patch!

Your fix makes sense to me. Could you just update the commit message with:
1. In the title, use: `plugins: kernel_install: `
2. Add the Fixed tag with the commit hash that introduced this regression (the same way that we usually do in the kernel)

melissawen:
Thanks for the review,
I just applied your suggestions.

rodrigosiqueira:
Thanks for your fix. I already applied your patch to the unstable branch. See:
https://github.com/kworkflow/kworkflow/commit/c3dfee06ef996b5253ac582e9ce95ad0511bae7d

PR #205, opened 2021-03-20
Tests refactor

rodrigosiqueira:


src/plugins/kernel_install/utils.sh
```diff
@@ -61,67 +109,174 @@ function reboot_machine()
...
+  local distro="$2"
+  local cmd_grub="$3"
+  local cmd_init="$4"
+  local setup_grub="$5"
+  local grub_install="$6"
+  local flag="$7"
+  local cmd=''
+  # We assume Debian as a default option
+  local mount_root=": mount /dev/sda1 /"
+  local mkdir_grub=": mkdir-p /boot/grub"
```
melissawen, line 188: I checked that lines 188 and 203 switched places. If we assume Debian as a default option, we will use the variable mkdir_init as defined in line 203. In the arch case, we use the mkdir_grub.

src/plugins/kernel_install/arch.sh
```diff
@@ -5,90 +5,20 @@
...
-}
 
 # Update boot loader API
-function update_boot_loader()
+function update_arch_boot_loader()
 {
   local name="$1"
   local target="$2"
   local flag="$3"
+  local cmd_grub='grub-mkconfig -o /boot/grub/grub.cfg'
```
melissawen, line 16: It is not needed here. cmd_grub should not be passed to update_boot_loader and we also define it inside the update_boot_loader.

src/plugins/kernel_install/arch.sh
```diff
@@ -5,90 +5,20 @@
...
-
-  cmd="$sudo_cmd grub-mkconfig -o /boot/grub/grub.cfg"
-
-  # Update grub
-  if [[ "$target" == 'vm' ]] ; then
-    vm_update_boot_loader "$name" "$cmd"
-  else
-    cmd_manager "$flag" "$cmd"
-  fi
+  update_boot_loader "$name" 'arch' "$target" "$cmd_grub" "$cmd_init" "$setup_grub" "$grub_install"  "$flag"
```
melissawen, line 21: Here, cmd_grub should be removed, ie: update_boot_loader "$name" 'arch' "$target" "$cmd_init" "$setup_grub" "$grub_install"  "$flag"

melissawen (changes requested):
Hello @rodrigosiqueira,
I found some issues with VM deployment. Please see inline comments.
I have also verified that VM deployment on Debian and Arch works correctly after the changes attached here.
[PR205_patch.txt](https://github.com/kworkflow/kworkflow/files/6218105/PR205_patch.txt)
Thanks :)

rodrigosiqueira:
This patch was a draft, and I already applied most of the changes available here.

PR #207, opened 2021-03-23
plugins: arch: fix the misuse of mkinitcpio on vm deployment

melissawen:
The mkinitcpio is not used for vm deployment, but used for other
targets.

Fixes: 5d75cfa3 ("Add support for VM deploy")


rodrigosiqueira:
Nice catch! Merged

PR #209, opened 2021-03-27
src: get_maintainers_wrapper: Add `--update-patch` option

### LINKED ISSUE #53: Add an option for `kw m` for `format-patch` output
The `kw m` command outputs the name and email of maintainers, however it must be parsed for `git format-patch`. Kw could do it for us.

joaoseckler:
Include an `--update-patch/-u` option to `kw m`. It allows the user to include the maintainers of a given patch as recipients of that same patch, via the `To:` field in it's header. This PR includes:

- Logic implementation in `src/get_maintainer_wrapper` (including handling cases where the `To:` field is already present)
- Refactoring of parsing of user options in `src/get_maintainer_wrapper`
- Documentation update (both in `src/get_maintainer_wrapper` and `documentation/man/kw.rst`)
- Tests update

Fixes #53


src/get_maintainer_wrapper.sh
```diff
@@ -88,8 +112,12 @@ function execute_get_maintainer()
 
   # if given path is not a patchfile, add -f to get_maintainer.pl options
   if ! is_a_patch "$FILE_OR_DIR"; then
+    if $update_patch; then
```
rodrigosiqueira, line 115: Add double-quotes.

src/get_maintainer_wrapper.sh
```diff
@@ -114,12 +142,28 @@ function execute_get_maintainer()
...
   fi
 
-  say $SEPARATOR
-  say "HERE:"
   cd $kernel_root
-  eval perl $script $options "$FILE_OR_DIR"
+  local -r script_output="$(eval perl $script $script_options "$FILE_OR_DIR")"
   cd $original_working_dir
 
+  say $SEPARATOR
```
rodrigosiqueira, line 149: Add double-quote.

For a reference, take a look at:

https://github.com/koalaman/shellcheck/wiki/Sc2086

src/get_maintainer_wrapper.sh
```diff
@@ -38,40 +38,62 @@ function print_files_authors()
...
 #
 # This function also handle the cases where:
-# - A patch file is given
 # - The working dir and/or given path is not the root of a linux kernel repo
 # - The working dir and/or given path is not inside a linux kernel repo
+# - Option --authors/-a is given. It will print the name of the module's
+# authors
+# - Option --update-patch/-u is given. It will include a "To:" field in
+# the header of the patch with the contact of the maintainers associated
+# with that patch
```
rodrigosiqueira, line 53: Nit-pick:
I recommend you to ident the second line with the first character of the bullet.

src/get_maintainer_wrapper.sh
```diff
@@ -38,40 +38,62 @@ function print_files_authors()
...
   local FILE_OR_DIR
-  local print_authors
+  local print_authors=false
+  local update_patch=false
   local is_file_a_patch=true
   local is_file_inside_kernel_tree=true
 
   local -r script="scripts/get_maintainer.pl"
-  local options="--separator , --nokeywords --nogit --nogit-fallback --norolestats "
+  local script_options="--separator , --nokeywords --nogit --nogit-fallback --norolestats "
```
rodrigosiqueira, line 66: It would be nice if we get this configuration from kworkflow.config...

Just a random thought... feel free to ignore it.

rodrigosiqueira (changes requested):
Hi Joao,

First of all, thanks a lot for your patch. That's a very good piece of work :)

You can see a feel inline comments. Additionally, could you squash all of these patches in a single one? You're adding a new option, which makes sense to be a single commit instead of a commit per change; in other words, all commits are directly related to the same feature.

Finally, It would be nice if we also fix the maintainer's output to show their names correctly; I mean, something similar to the way you did for the --update-patch. However, you don't need to do it for this series.

Thanks

joaoseckler:
Hi Siqueira! Thanks!

I squashed everything into one commit and made the changes you suggested above. I just didn't quite understand what you meant with
> It would be nice if we also fix the maintainer's output to show their names correctly; I mean, something similar to the way you did for the --update-patch. However, you don't need to do it for this series.
Do you mean changing the `HERE:` message?

Thanks!

tests/get_maintainer_wrapper_test.sh
```diff
@@ -3,13 +3,14 @@
 . ./src/get_maintainer_wrapper.sh --source-only
 . ./tests/utils --source-only
 
+configurations[get_maintainer_options]="--separator , --nokeywords --nogit --nogit-fallback --norolestats "
```
joaoseckler, line 6: Is this correct? It seems like the test suite doesn't load the default kw configurations. Should get_maintainer_wrapper.sh handle the case where this config is empty?

rodrigosiqueira:
> > It would be nice if we also fix the maintainer's output to show their names correctly; I mean, something similar to the way you did for the --update-patch. However, you don't need to do it for this series.
> 
> Do you mean changing the `HERE:` message?

My bad here, just ignore my comment.

tests/get_maintainer_wrapper_test.sh
rodrigosiqueira: Hi Joao,

You don't need to add this change for this series; if you want to do that, I recommend doing it in another PR.

Anyway, about your question, you can try:
1. In the setup function, you can add:
```
parse_configuration "$KW_CONFIG_SAMPLE"
```
The `parse_configuration` function will parse the kworkflow.config file and populate the configurations hash based on the file information.

2. Add your `get_maintainer_options` in the sample file, for example:
Open -> `tests/samples/kworkflow.config`
Add -> get_maintainer_options="SOMETHING"

3. Update some of the tests and other potential failures.

Notice that the `tests/get_maintainer_wrapper_test.sh` is a little bit outdated when compared with other tests. For example, instead of using `setUp` function, it uses `setupGetMaintainers`.

joaoseckler: Ok! I reverted the changes referring to the config file.

joaoseckler:
I had forgotten to update the `src/help.sh` file with documentation on the new feature.

rodrigosiqueira:
Thanks for your patch!

I merged your change, see:

https://github.com/kworkflow/kworkflow/commit/4dae830deeb3a98cec1bee68b5c70bc94c206d49

Issue #80, opened 2019-02-27
Make kw work well in the zsh


rodrigosiqueira:
@giulianobelinassi make the kw file work as a driver (See: https://github.com/rodrigosiqueira/kworkflow/commit/649a6141224a06c0db343ca1027bec9ca6bcede3).

It could be nice if we have explicit support zsh :)

@RenatoGeh how about it?

rodrigosiqueira:
@alan-barzilay @kwy95
I suppose we can close this issue, right? If yes, feel free to close it.

alan-barzilay:
Yeah, with tab completion done I can´t think of anything else that is lacking

PR #212, opened 2021-03-30
Add suport for zsh tab completion

alan-barzilay:
This PR should help with zsh support in #80


setup.sh
```diff
@@ -298,13 +298,23 @@ function synchronize_files()
   setup_config_file
   ASSERT_IF_NOT_EQ_ZERO "Config file failed" "$?"
 
-  if [ -f "$HOME/.bashrc" ]; then
-      # Add to bashrc
-      echo "# $app_name" >> "$HOME/.bashrc"
-      echo "source $libdir/$BASH_AUTOCOMPLETE.sh" >> "$HOME/.bashrc"
-      update_path
+  if [ -f "$HOME/.bashrc" ] || [ -h "$HOME/.bashrc" ]; then
```
rodrigosiqueira, line 301: Why not use `[[ -f "..." || -h "..." ]]`?

setup.sh
```diff
@@ -298,13 +298,23 @@ function synchronize_files()
...
+    update_path
   else
-      warning "Unable to find a shell."
+    warning "Unable to find a bash shell."
+  fi
+
+  if [ -f "$HOME/.zshrc" ] || [ -h "$HOME/.zshrc" ]; then 
+    # Add to zshrc
+    echo "# Enable bash completion for zsh" >> "$HOME/.zshrc"
+    echo "autoload bashcompinit && bashcompinit" >> "$HOME/.zshrc"
```
rodrigosiqueira, line 313: Is this bashcompinit a default option in the zshrc?

setup.sh
```diff
@@ -298,13 +298,23 @@ function synchronize_files()
...
+    echo "# $app_name" >> "$HOME/.zshrc"
+    echo "source $libdir/$BASH_AUTOCOMPLETE.sh" >> "$HOME/.zshrc"
+  else
+    warning "Unable to find a zsh shell."
```
rodrigosiqueira, line 317: This code is very similar to the one above it; maybe you can avoid code duplication by making ".bashrc" and ".zshrc" values in a variable.

rodrigosiqueira (changes requested):
Hi @alan-barzilay 
I just added a few comments. Just for curiosity, did you test it in the zsh terminal?

setup.sh
alan-barzilay: This file was originally using `if [ -f "$HOME/.bashrc" ]`  and the codestyles file didnt mention anything about using [] or [[]] so I tried to stay consistent by keeping []. To be honest I prefer [[]] haha should I change it?

setup.sh
alan-barzilay: yes

setup.sh
alan-barzilay: I changed it to [[ -f "..." || -h "..." ]]

setup.sh
alan-barzilay: I made another commit where i extracted these 2 lines into a function, let me know if this was what you had in mind. We could decrease the code duplication even further by extracting the whole tab completion process into a function, but I think this would unnecessarily increase complexity.

rodrigosiqueira:
@alan-barzilay 
Just for curiosity, did you test it in the zsh terminal? For example, did you try kw deploy or similar?

alan-barzilay:
@rodrigosiqueira I'm not sure if I understood what you want to know. This PR is just about tabcompletion so in theory this shouldn't affect anything else besides the completions. I tested the tab completions after reinstalling kw and they work. 
Other commands such as kw deploy have the same same behavior in bash and zsh as far as I could tell.

rodrigosiqueira:
@alan-barzilay 
I'm not a zsh user; this is why I asked if you tried it.

alan-barzilay:
@rodrigosiqueira  I tried it out and as far as I could tell there were no differences and everything worked as it should

setup.sh
```diff
@@ -298,13 +298,21 @@ function synchronize_files()
...
+    append_bashcompletion ".bashrc"
+    update_path
   else
-      warning "Unable to find a shell."
+    warning "Unable to find a .bashrc file."
+  fi
+
+  # Add tabcompletion to zshrc
+  if [[ -f "$HOME/.zshrc" ||  -h "$HOME/.zshrc" ]]; then 
+    echo "# Enable bash completion for zsh" >> "$HOME/.zshrc"
```
rodrigosiqueira, line 311: Please, correct me if I'm wrong, but looks like that `setup.sh` will trigger a warning in a pure zsh system, right? Or zsh also relies on `.bashrc`? Maybe you want an `elif` here, e.g,:

```
if [[ -f "$HOME/.bashrc" || -h "$HOME/.bashrc" ]]; then
...
elif [[ -f "$HOME/.zshrc" || -h "$HOME/.zshrc" ]]; then
...
else
 complain about something
fi
```

alan-barzilay: You are correct, it would also complain in a pure bash system if it couldn't find a .zshrc file. 
My idea was to complain in both cases just to let the user know that it wasn't able to add tabcompletion, maybe this could explicitly mention the tabcompletions.
I use mainly zsh but I also have bash and would like to have tabcompletions for both, I wouldn't like  if the setup.sh silently failed to install tab completion to one of my shells.
Your suggestion would also cause the tabcompletions to be installed for bash but never for zsh if a bashrc file is present for the user, so I believe we should keep it as is. What do you think?

rodrigosiqueira: Hi @alan-barzilay 
I see your point.
However, I'll feel annoyed if I see zsh warnings in my bash system. In my case, I'm not a zsh user; what is the point of showing me such a warning? Other users might feel confused and think that maybe their installation did not complete correctly. Perhaps you can try to alleviate this issue by checking if the users have zsh installed in their system before sending a warning message.

alan-barzilay: @rodrigosiqueira We could create a better message to let the user know that this is not necessarily an error, but I agree that directly checking if the user has zsh installed is a better approach. 
Do you know of a way to easily confirm if the user has installed a given program? I couldn't think of a good and general way of doing this

rodrigosiqueira: Maybe something like:

`[ -x "$(command -v zsh)" ]`

alan-barzilay: nice! never thought of using command this way, done

setup.sh
```diff
@@ -298,16 +298,28 @@ function synchronize_files()
...
+    if [[ -f "$HOME/.bashrc"  ||  -h "$HOME/.bashrc" ]]; then
+      append_bashcompletion ".bashrc"
       update_path
-  else
-      warning "Unable to find a shell."
+    else
+      warning "Unable to find a .bashrc file."
+    fi
+  fi
+
```
rodrigosiqueira, line 310: Remove this extra space...

setup.sh
```diff
@@ -298,16 +298,28 @@ function synchronize_files()
...
-  else
-      warning "Unable to find a shell."
+    else
+      warning "Unable to find a .bashrc file."
+    fi
+  fi
+  
+  if [[ -x "$(command -v zsh)" ]]; then
+    # Add tabcompletion to zshrc
+    if [[ -f "$HOME/.zshrc" ||  -h "$HOME/.zshrc" ]]; then
```
rodrigosiqueira, line 313: Please remove the extra space at the end of the line.

setup.sh
```diff
@@ -318,6 +330,14 @@ function synchronize_files()
   warning " -> For a better experience with kw, please, open a new terminal."
 }
 
+function append_bashcompletion()
+{
+  local shellrc="$1"
+  echo "# $app_name" >> "$HOME/$shellrc"
+  echo "source $libdir/$BASH_AUTOCOMPLETE.sh" >> "$HOME/$shellrc"
+
```
rodrigosiqueira, line 338: Remove this extra line

setup.sh
```diff
@@ -318,6 +330,14 @@ function synchronize_files()
   warning " -> For a better experience with kw, please, open a new terminal."
 }
 
+function append_bashcompletion()
+{
+  local shellrc="$1"
```
rodrigosiqueira, line 335: Add one extra line to separate the variable declaration from the rest of the code.

rodrigosiqueira:
Hi @alan-barzilay ,
Thanks for your updates. Please, remove all the extra spaces and rebase your code in the latest code from the unstable branch. I notice that you forget some extra spaces in your PRs; before submitting your changes, use `git log -p` to double-check if everything looks good.

Thanks.

rodrigosiqueira (changes requested):

rodrigosiqueira:
Hi @alan-barzilay , any update here?

setup.sh
```diff
@@ -298,16 +298,28 @@ function synchronize_files()
   setup_config_file
   ASSERT_IF_NOT_EQ_ZERO "Config file failed" "$?"
 
-  if [ -f "$HOME/.bashrc" ]; then
-      # Add to bashrc
-      echo "# $app_name" >> "$HOME/.bashrc"
-      echo "source $libdir/$BASH_AUTOCOMPLETE.sh" >> "$HOME/.bashrc"
+  if [[ -x "$(command -v bash)" ]]; then
```
rodrigosiqueira, line 301: btw, @magalilemes added a helper named `command_exists()` (latest code in the unstable branch); please use that helper in your new PR.

alan-barzilay:
> Hi @alan-barzilay , any update here?

Hi @rodrigorsdc, just pushed a new version

PR #217, opened 2021-04-01
Fix tests for the case when shunit2 is cloned into the tests folder

### LINKED ISSUE #216: Bug: can't run tests by cloning shunit2 into the tests folder
If shunit2 is not detected in the system, it checks for the existence of a file called shunit2 in the tests folder and tries to source it:
```
  elif [[ -f ./tests/shunit2 ]]; then
    . ./tests/shunit2
```

However, since the documentation suggests cloning shunit2 into the tests folder, I believe the correct behavior should be to check if the shunit2 folder exists and if it does, source the shunit2 file inside it:
```
  elif [[ -d ./tests/shunit2 ]]; then
    . ./tests/shunit2/shunit2
```

alan-barzilay:
Closes #216

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -159,7 +159,7 @@ function strip_path
 
 if [[ "$#" -eq 0 ]]; then
   check_required_files
-  files_list=$(find ./tests -name "*_test.sh")
+  files_list=$(find ./tests -name "*_test.sh" | grep --invert-match "/shunit2/" )
   # Note: Usually we want to use double-quotes on bash variables, however,
   # in this case we want a set of parameters instead of a single one.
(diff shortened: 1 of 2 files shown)


rodrigosiqueira:
Hi @alan-barzilay ,

I already applied your change into the unstable branch (see b43f65c60644238a644ec9bdb6a237a557c3cdc3). I manually change two things in your commit message:

1. Make sure that your commit lines do not pass 71 characters;
2. Avoid adding the issue number in the commit title. Also, you can close an issue by using "Closes" in your commit message. Take a look at: https://github.blog/2013-01-22-closing-issues-via-commit-messages/

PR #222, opened 2021-04-03
src: mk: add build information option

### LINKED ISSUE #183: Provide build info
It is common to check the build info before deploy or just for checking something. For example, developers use:

```bash
make kernelrelease
cat modules.order | wc -l
```

Add a `kw build --info` that provides a build summary, something like:

```bash
kw b -i
Release: 5.6.0-COOL-NAME+
Total modules to be compiled: 158
...
```

magalilemes:
This pull request adds the '--info' option to the 'kw build' command.

Closes #183


src/mk.sh
```diff
@@ -525,6 +525,14 @@ function mk_build()
   IFS=' ' read -r -a options <<< "$raw_options"
   for option in "${options[@]}"; do
     case "$option" in
+      --info|-i)
```
rodrigosiqueira, line 528: Let's capture the kernel version:
make kernelversion

src/mk.sh
```diff
@@ -525,6 +525,14 @@ function mk_build()
...
+        local kernel_version=$(get_kernel_release "$flag")
+        local compiled_modules=$(cat modules.order | wc -l)
```
rodrigosiqueira, line 530: You are assuming that `modules.order` always exists, which is not true. This file is generated after users compile their kernel; in other words, do not `cat` directly to `modules.order`. You need to check if this file exists first, also, add a test to capture this case.

src/mk.sh
```diff
@@ -525,6 +525,14 @@ function mk_build()
...
+        
+        echo -e "Release: $kernel_version"
+        echo -e "Total modules to be compiled: $compiled_modules"
```
rodrigosiqueira, line 533: Why not use the function `say`?

src/mk.sh
```diff
@@ -525,6 +525,14 @@ function mk_build()
...
+        exit
+        ;;
```
rodrigosiqueira, line 535: I prefer to isolate the information code in a single function instead of adding it in the parser option. This can be useful for the following reasons:
1. It is easier to expand the build information feature;
2. We can have a dedicated test for this function.

tests/mk_test.sh
```diff
@@ -144,6 +147,15 @@ function mk_build_Test()
...
   ID=5
+  declare -a expected_cmd=(
+    "$release"
+    "$modules"
+  )
+  touch "$PWD/modules.order"
+  output=$(mk_build 'TEST_MODE' '--info')
+  compare_command_sequence expected_cmd[@] "$output" "$ID"
+
+  ID=6
```
rodrigosiqueira, line 158: As I suggested in the previous comment, isolate the information in a single function and add proper test coverage.

rodrigosiqueira (changes requested):
Hi @magalilemes,

First of all, thanks a lot for your contribution.

In general, you're are on the right path. Please, take a look at my comments.

Thanks

src/mk.sh
magalilemes: Should the kernel version be a new information provided alongside the kernel release and the total number of compiled modules or should it replace the kernel release?

src/mk.sh
magalilemes: I didn't really think this through, I guess it does make sense to use `say`. Thanks!

src/mk.sh
rodrigosiqueira: Yes

src/mk.sh
rodrigosiqueira: Hmmmm... I was wondering, instead of relying on `cat modules.order | wc -l` (it require compilation), can we derive the total of drivers from being compiled from the `.config` file?

magalilemes: If that's the case, what about doing  `grep -c =m .config`? to get the number of modules that will be compiled?

rodrigosiqueira: Yes, I think this grep command probably works fine. Maybe compare `cat modules.order | wc -l` and `cat modules.order | wc -l` output.

src/mk.sh
```diff
@@ -489,6 +503,26 @@ function kernel_deploy
   fi
 }
 
+# This function retrieves and prints information related to the building
+# process of the kernel.
```
rodrigosiqueira, line 507: How about:

> This function retrieves and prints information related to the kernel that will be compiled.

In some way, "building process" sounds weird because the sort of information we show is not related to the building process.

src/mk.sh
```diff
@@ -489,6 +503,26 @@ function kernel_deploy
...
+function build_info()
+{
+  if [[ -f ".config" ]]; then
```
rodrigosiqueira, line 510: /".config"/'.config'/

src/mk.sh
```diff
@@ -489,6 +503,26 @@ function kernel_deploy
...
+    local kernel_name=$(get_kernel_release "$flag")
```
rodrigosiqueira, line 511: You forgot to declare "$flag".

src/mk.sh
```diff
@@ -489,6 +503,26 @@ function kernel_deploy
...
+    local kernel_version=$(get_kernel_version "$flag")
+    local compiled_modules=$(grep -c "=m" .config)
```
rodrigosiqueira, line 513: /"=m"/'=m'/

src/mk.sh
```diff
@@ -489,6 +503,26 @@ function kernel_deploy
...
+
+    say "Kernel source information"
+    echo -e "\tName: $kernel_name"
+    echo -e "\tVersion: $kernel_version"
+    echo -e "\tTotal modules to be compiled: $compiled_modules"
+
+  else
+    complain ".config file was not found. You have to be inside a kernel"\
+    "repository with a .config file to run this command"
+  fi
```
rodrigosiqueira, line 523: We don't need to complain about anything if the user does not have a local `.config` file; we just don't want to try the grep command.

src/mk.sh
```diff
@@ -489,6 +503,26 @@ function kernel_deploy
...
+function build_info()
+{
+  if [[ -f ".config" ]]; then
+    local kernel_name=$(get_kernel_release "$flag")
+    local kernel_version=$(get_kernel_version "$flag")
+    local compiled_modules=$(grep -c "=m" .config)
+
+    say "Kernel source information"
+    echo -e "\tName: $kernel_name"
+    echo -e "\tVersion: $kernel_version"
```
rodrigosiqueira, line 517: This should be outside of this condition. The same is true for `local kernel_version=$(get_kernel_version "$flag")` and `local kernel_name=$(get_kernel_release "$flag")`. The only thing that needs to be inside this function is the operations that deal with `.config`.

tests/mk_test.sh
```diff
@@ -653,4 +660,36 @@ function cleanup_after_deploy_Test
...
+  local ID
+  local original="$PWD"
+  local release="5.4.0-rc7-test"
+  local version="5.4.0-rc7"
+  local release_output="Name: $release"
+  local version_output="Version: $version"
+  local modules="Total modules to be compiled: 5"
+
+  cd "$FAKE_KERNEL"
+  ID=1
```
rodrigosiqueira, line 674: Let's remove ID and use $LINENO.

tests/mk_test.sh
```diff
@@ -653,4 +660,36 @@ function cleanup_after_deploy_Test
...
+  output=$(mk_build 'TEST_MODE' '--info')
+  expected_result=".config file was not found. You have to be inside a kernel "
+  expected_result+="repository with a .config file to run this command"
+  assertEquals "($ID)" "$expected_result" "${output}"
```
rodrigosiqueira, line 678: e.g.,
```
assertEquals "($LINENO)" "$expected_result" "${output}"
```

src/help.sh
```diff
@@ -25,6 +25,7 @@ function kworkflow-help()
 
   echo -e "kw build:\n" \
     "\tbuild - Build kernel \n" \
+    "\tbuild [--info|-i] - Display build information\n" \
```
rodrigosiqueira, line 28: Now we have a specific help function inside mk.sh, please add this information in that part of the code (latest unstable).

rodrigosiqueira (changes requested):
Hi,
In general, it looks good. Please look at my inline comments and rebase your code with the latest code from the unstable branch. Notice that the code that handles the help function changed.

rodrigosiqueira:
Hi @magalilemes ,
Thanks a lot for your patch! Merged to unstable branch, see: https://github.com/kworkflow/kworkflow/commit/aaa0f2319a821fb10a5d58f2982455ddaff54ba3

PR #225, opened 2021-04-05
Implement include guards

### LINKED ISSUE #197: Implement and use an include manager mechanism
Currently, we don't have a well-organized way to manage file inclusions. For example, we rely on things like `. utils.sh --source-only`. We can do better than that...

alan-barzilay:
This PR creates an include guard for all files being sourced in the project, thus preventing them from being sourced twice. This PR also uses the same strategy to prevent remote.sh from being sourced before kw_config_loader.sh.

I suggest including this strategy into the codestyle of the project, this PR also includes a suggestion for a section in codingstyle.rst about include guards.

Closes #197

### DIFF
--- a/documentation/content/codingstyle.rst
+++ b/documentation/content/codingstyle.rst
@@ -180,6 +180,14 @@ $PATH, you should use 'type ', instead of 'which '.
 The output of 'which' is not machine parsable and its exit code
 is not reliable across platforms.
 
+Sourcing files
+---------------
+Every file that may be sourced must have an import guard of the form:  
+`[ -n "$FILENAME_IMPORTED" ] && return || readonly FILENAME_IMPORTED=1`.
+This guarantees that no file may be imported twice.
+
+
+
 Test name
 ---------
 
--- a/src/help.sh
+++ b/src/help.sh
@@ -1,3 +1,5 @@
+[ -n "$HELP_IMPORTED" ] && return || readonly HELP_IMPORTED=1
+
 . "$KW_LIB_DIR/kwio.sh" --source-only
 
 function kworkflow-help()
--- a/src/kw_config_loader.sh
+++ b/src/kw_config_loader.sh
@@ -1,3 +1,5 @@
+[ -n "$KW_CONFIGLOADER_IMPORTED" ] && return || readonly KW_CONFIGLOADER_IMPORTED=1
+
 . "$KW_LIB_DIR/kwio.sh" --source-only
 
 CONFIG_FILENAME=kworkflow.config
--- a/src/kwio.sh
+++ b/src/kwio.sh
@@ -1,5 +1,8 @@
 # NOTE: src/kw_config_loader.sh must be included before this file
 
+[ -n "$KWIO_IMPORTED" ] && return || readonly KWIO_IMPORTED=1
+
+
 declare -r BLUECOLOR="\033[1;34;49m%s\033[m"
 declare -r REDCOLOR="\033[1;31;49m%s\033[m"
 declare -r YELLOWCOLOR="\033[1;33;49m%s\033[m"
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -1,5 +1,7 @@
 # NOTE: src/kw_config_loader.sh must be included before this file
 
+[ -n "$KWLIB_IMPORTED" ] && return || readonly KWLIB_IMPORTED=1
+
 # A common task used inside kw is a string separation based on a delimiter, for
 # this reason, this function tries to handle this scenario by getting a
 # delimiter character followed by the position that the users want to retrieve.
--- a/src/remote.sh
+++ b/src/remote.sh
@@ -2,6 +2,10 @@
 
 # NOTE: It is recommended that src/kw_config_loader.sh be included before this
 # file
+if [[ -z "$KW_CONFIGLOADER_IMPORTED" ]]; then
+  . "$KW_LIB_DIR/kw_config_loader.sh" --source-only
+fi
+
 
 # We now have a kw directory visible for users in the home directory, which is
 # used for saving temporary files to be deployed in the target machine.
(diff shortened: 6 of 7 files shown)


joaoseckler:
Nice! I was wondering if it wouldn't be easier for future development to have a wrapper function around these imports, similar to how we handle commands with the `cmd_manager` function. Maybe something in the lines of

```
function kw_import() {
  local filename="$1"
  local varname=${filename// /_}
  varname=${varname^^}_IMPORTED

  if [ -n "${!varname}" ]; then
    return 255 
  else    
    ${varname}=1
    . "$@" --source-only # can return 0, 1 or 2
  fi
}
```

The we would include using this function instead of `.`. If the the caller needs to treat the special case in which a file has already been included or does not exist, it can rely on the the return value of this function; otherwise, the usage is the same as the dot include. 

What do you think?

alan-barzilay:
@joaoseckler  That's an interesting idea! But I wonder if we wouldn't run into a chicken and egg situation. To use this function it would have to be sourced, but we should use this function to source files and we can't use it  to source itself. We would end up still having to use `. file.sh --sources-only` to source the file that carries this function.

I think it would be simpler to just check the FILENAME_IMPORTED variable directly instead of trying to source it and check the exit code. I did something similar in remote.sh to ensure that kw_config_loader.sh would be sourced before remote.sh.

The user may also just want to check if a given file was imported without necessarily importing it, in which case they would use the variable directly and avoid using the function.

What do you think? Does it make sense?

joaoseckler:
> I think it would be simpler to just check the FILENAME_IMPORTED variable directly instead of trying to source it and check the exit code. 

The function I described does that! But it abstracts the name of the variable away from us and leaves the reason why a include possibly failed transparent via return values  

> I did something similar in remote.sh to ensure that kw_config_loader.sh would be sourced before remote.sh.

This made me think of something else that I will comment on the appropriate commit.

But you are right! Chicken and egg. It would only make sense if the file including a function like the one I described were to be included once at the very beginning of code execution, maybe sourced by the file kw itself. But that is a bigger change to the execution flow. So I think we're on a good path here.

rodrigosiqueira:
Hi @alan-barzilay  and @joaoseckler 
After following your discussion, I really like the idea of having a function for these reasons:
1. We can easily add a test for this function;
2. It will be easier to maintain file inclusion because any new change will require a  change in a single function instead of multiple files;
3. A function named "include" can be easily understood;

For trying to solve the chicken and egg problem, maybe we can try something like this:

We can create a file named "include.sh" to handle file inclusion. In the `kw` file, we use something like:

```
. "$KW_LIB_DIR/include.sh" --source-only
```

In theory, `kw` file is the entry point for kw operations. Because of that, all other files will see the include function (I did not test it).

Anyway, that's an interesting discussion, and we can probably evolve some ideas.

alan-barzilay:
Ok, looks good to me @rodrigosiqueira. 
@joaoseckler in your suggestion if I'm not mistaken we are handling the case in which there is a space character in the file name and we change it to an underscore. Is this necessary? None of our files have spaces in their names and they probably shouldn't, maybe we should explicitly add to the codestyle that file names should be snakecased and just drop this line. What do you think?

joaoseckler:
@alan-barzilay Nice! I agree that that line is unnecessary.  It also occurred to me that we should remove .sh from that variable name, maybe with something in the lines of `varname=${varname%.sh}`

joaoseckler:
A quick test for the current files in `src` reveals that this method only fails at `kw.fish`, but that file is not currently being imported anywhere:

```
$ for file in *; do varname="${file%.sh}"; varname="${varname^^}_IMPORT"; echo $varname; done
BASH_AUTOCOMPLETE_IMPORT
CHECKPATCH_WRAPPER_IMPORT
CONFIG_MANAGER_IMPORT
DIFF_IMPORT
EXPLORE_IMPORT
GET_MAINTAINER_WRAPPER_IMPORT
HELP_IMPORT
INIT_IMPORT
KW_CONFIG_LOADER_IMPORT
KW.FISH_IMPORT
KWIO_IMPORT
KWLIB_IMPORT
MK_IMPORT
PLUGINS_IMPORT
REMOTE_IMPORT
STATISTICS_IMPORT
VERSION_IMPORT
VM_IMPORT
```

alan-barzilay:
@rodrigosiqueira I may have found a problem. Our tests don't go through kw, so they won't have access to the include.sh imported there and we are back at the chicken and egg situation.
Do you have a preference on how to deal with this? AlI solutions I could think of feel hacky.

rodrigosiqueira:
> @rodrigosiqueira I may have found a problem. Our tests don't go through kw, so they won't have access to the include.sh imported there and we are back at the chicken and egg situation.
> Do you have a preference on how to deal with this? AlI solutions I could think of feel hacky.

How about adding the include file to `tests/utils`? IIRC, all tests include this file.

rodrigosiqueira:
@alan-barzilay @joaoseckler 
Can we have a new version of this patch with the new ideas? It will be easier to discuss things by looking at actual code.

alan-barzilay:
@rodrigosiqueira 

> How about adding the include file to tests/utils? IIRC, all tests include this file.

I dont know, it feels weird to have in every test something like:
```
. utils --sources-only
include something.sh
include something2.sh
```
I feel like it kinda defeats the purpose of having the include function, its confusing to have 2 different ways to import something in the same file.

>  Can we have a new version of this patch with the new ideas? It will be easier to discuss things by looking at actual code.

I was writing a new version with the suggestions when I realized this problem haha but I couldn't think of a way to make it work elegantly

rodrigosiqueira:
@alan-barzilay 
We only run tests via `run_tests.sh` script. Why not add the `. file` in this script? This approach is similar to the one used by the `kw` file.

alan-barzilay:
@rodrigosiqueira I believe that nothing that is imported in run_tests.sh is present during the tests, that's why we reimport `utils` in every test.
I'm not really sure why that is, I guessing the tests are run as independent sub-processes, I'm still trying to understand this

rodrigosiqueira:
Since there is another PR, can we close this one?

PR #227, opened 2021-04-05
tests/get_maintainer_wrapper_test: improve

### LINKED ISSUE #224: Fix problems with get_maintainer_wrapper_test
There is some problem with `get_maintainer_wrapper_test` as can be seen below:

```
Running test [get_maintainer_wrapper_test]
print_files_authors_Test
print_files_authors_from_dir_Test
execute_get_maintainer_Test
scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kw/kworkflow/tests/.tmp/' not found in version control 
scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kw/kworkflow/tests/.tmp/' not found in version control 
scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kw/kworkflow/tests/.tmp/fs/' not found in version control 
scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kw/kworkflow/tests/.tmp/' not found in version control 
scripts/get_maintainer.pl: file '/home/siqueira/Documents/Repositories/kw/kworkflow/tests/.tmp/fs/' not found in version control 
```

We need to take a look at that.

joaoseckler:
This PR makes two improvements to the `get_maintainer_wrapper_test` file: i) some refactoring and reduction of code duplication and ii) introduction of a fake `.git` folder to avoid some warnings by the `get_maintainer.pl` kernel script.

Fixes #224

### DIFF
--- a/tests/get_maintainer_wrapper_test.sh
+++ b/tests/get_maintainer_wrapper_test.sh
@@ -57,12 +57,19 @@ function oneTimeSetUp
   # dir is also created inside tests/.tmp so that get_maintainer.pl thinks
   # it is a git repo. This is done in order to avoid some warnings that
   # get_maintainer.pl prints when no .git is found.
+  local original_dir="$PWD"
   mk_fake_kernel_root "$FAKE_KERNEL"
-  mkdir -p "$FAKE_KERNEL/.git"
   cp -f tests/samples/MAINTAINERS "$FAKE_KERNEL"/MAINTAINERS
   cp -f tests/external/get_maintainer.pl "$FAKE_KERNEL"/scripts/
   cp -f tests/samples/update_patch_test{_model,}{,2}.patch "$FAKE_KERNEL"/
+  cd "$FAKE_KERNEL"
+  touch fs/some_file
+  git init --quiet
+  git add fs/some_file
+  git commit --quiet -m "Test message"
+  cd "$original_dir"
 }
+
 
 function oneTimeTearDown
 {
--- a/tests/get_maintainer_wrapper_test.sh
+++ b/tests/get_maintainer_wrapper_test.sh
@@ -10,6 +10,7 @@ function suite
   suite_addTest "print_files_authors_Test"
   suite_addTest "print_files_authors_from_dir_Test"
   suite_addTest "execute_get_maintainer_Test"
+  suite_addTest "execute_get_maintainer_patch_Test"
 }
 
 # The following variables hold the the lines print_files_authors should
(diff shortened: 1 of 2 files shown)


tests/get_maintainer_wrapper_test.sh
```diff
@@ -50,147 +51,91 @@ Maintainers already in \"To:\" field of update_patch_test.patch"
...
-function setupGetMaintainers
+function oneTimeSetUp
 {
   # This creates tests/.tmp which should mock a kernel tree root. A .git
   # dir is also created inside tests/.tmp so that get_maintainer.pl thinks
   # it is a git repo. This is done in order to avoid some warnings that
   # get_maintainer.pl prints when no .git is found.
   mk_fake_kernel_root "$FAKE_KERNEL"
-  mkdir -p "$FAKE_KERNEL/.git"
+  cp -rf tests/samples/fake-git/ "$FAKE_KERNEL"/.git
```
rodrigosiqueira, line 61: How about introducing a new helper function for creating a fake git repository? Something similar to mk_fake_kernel_root. Maybe you can add this in a separate commit introduced before this one.

rodrigosiqueira (changes requested):
Hi @joaoseckler

That's a really nice improvement :)

I added a single inline comment. Also, in the commit that says "avoid warnings", I recommend you add the warning that your commit fixes.

tests/get_maintainer_wrapper_test.sh
joaoseckler: The `.git` fake folder created in this commit is very specific to the warning it's trying to stop. It has one single commit with one single file under control: `fs/nothing`. So I don't think it's general enough to become it's own function. On the other hand, most test modules that call `mk_fake_kernel_root` also create an empty `.git` folder (the only exception is the `checkpath_wrapper_test` module). My proposal: this particular case remains as is in my commit and we include the creation of a `.git` folder as a part of the `mk_fake_kernel_root` function, in another PR. Which makes sense, after all, the kernel wouldn't be the kernel weren't it for git. What do you think, @rodrigosiqueira?

rodrigosiqueira: > Which makes sense, after all, the kernel wouldn't be the kernel weren't it for git.

Actually, some people can download the kernel tarball and compile/install without any git requirement :)

Instead of statically create this fake git folder, can we at least generate it dynamically in the oneTimeSetup function?

joaoseckler: Okay! In this new version we dynamically generate a `.git` folder inside the `oneTimeSetup` function.

rodrigosiqueira:
Thanks for your patch, merged to the unstable branch. See: https://github.com/kworkflow/kworkflow/commit/c801af0934e3c1d72f8a52e2096c6325a4456493

PR #230, opened 2021-04-10
src: mk: Remove unnecessary sound notification

magalilemes:
I noticed that when we run 'kw deploy --unknown-option' (and 'unknown-option' can be any invalid option name), we still get a notification when this command finished running. I believe this is an unexpected behaviour, as it doesn't happen with other commands that make use of the 'alert_completion' function.
To reproduce this, you have to be inside a kernel repository and run the command as described above.

### DIFF
--- a/src/mk.sh
+++ b/src/mk.sh
@@ -404,7 +404,7 @@ function kernel_deploy
   deploy_parser_options "$@"
   if [[ "$?" -gt 0 ]]; then
     complain "Invalid option: ${options_values['ERROR']}"
-    return 22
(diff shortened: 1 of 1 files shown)


rodrigosiqueira:
Hi,

Very nice catch! I already applied your change, see:
https://github.com/kworkflow/kworkflow/commit/ee0ced2caa912298a3e16a51d24858ece71d001d

Also, you forgot to add Signed-off-by: in the commit. Use `git commit -s` instead of `git commit`.

PR #232, opened 2021-04-11
src: mk: warn user if 'kw d' is executed as root

### LINKED ISSUE #164: Warn user if `kw d --remote` is executed as a sudo or root
`kw deploy --local` should not be executed as sudo or root, otherwise, users will see something like this:

```
/home/USER/.config/kw/kw: line 75: /root/.config/kw/src/mk.sh: No such file or directory
/home/USER/.config/kw/kw: line 77: kernel_deploy: command not found
/home/USER/.config/kw/kw: line 79: alert_completion: command not found
```

We should avoid this sort of problem by let the user know.

harnaman-hk:
Executing 'kw deploy' as root leads to undefined commands. This PR adds a warning for it in kernel_deploy. This is to display the warning even if other commands invoking kernel_deploy are run as root/sudo, for example `kw bd`. 

Closes #164

### DIFF
--- a/src/mk.sh
+++ b/src/mk.sh
@@ -307,6 +307,11 @@ function kernel_install
       fi
 
       # Local Deploy
+      if [[ $(id -u) == 0 ]]; then
+        complain "kw deploy --local should not be run as root"
+        exit 1 # EPERM
+      fi
+
       . "$KW_PLUGINS_DIR/kernel_install/utils.sh" --source-only
       . "$KW_PLUGINS_DIR/kernel_install/$distro.sh" --source-only
       install_kernel "$name" "$distro" "$kernel_img_name" "$reboot" "$arch_target" 'local' "$flag"
--- a/documentation/man/kw.rst
+++ b/documentation/man/kw.rst
@@ -109,7 +109,7 @@ Follows the summary of the options:
 the network.
 
 2. --local: Deploy the Kernel image and modules in the host machine, you will
-need root access.
+need root access. 'kw deploy –local' should not be executed with sudo or root.
 
 3. --vm: Deploy the Kernel image and modules to QEMU vm.
 
--- a/tests/mk_test.sh
+++ b/tests/mk_test.sh
@@ -39,6 +39,11 @@ function get_kernel_version_mock()
   echo '5.4.0-rc7'
 }
 
+function root_id_mock()
+{
+  echo "0"
+}
+
 function setUp
 {
   local create_mkinitcpio="$1"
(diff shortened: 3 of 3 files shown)


src/mk.sh
```diff
@@ -401,6 +401,10 @@ function kernel_deploy
   local runtime=0
   local ret=0
 
+  if [[ $(id -u) == 0 ]]; then
+    warning "kw deploy should not be run as root"
```
rodrigosiqueira, line 405: Instead of use `warning`, use `complain`. Also, we probably want to exit with `EPERM` (1) or similar.

Additionally, I think it makes more sense to add it under the local deploy part (LOCAL_TARGET).

rodrigosiqueira (changes requested):
Hi harnaman-hk,

First of all, thanks for your patch.

I added one inline comment. Also, expand your patch with:
1. Add a test that covers the new condition that you added.
2. Expand kw documentation by adding that we should not run kw deploy --local with sudo or root.

Thanks

harnaman-hk:
@rodrigosiqueira  I had a few points to discuss regarding the test for this condition
1. Tests probably should not be with kernel_deploy in [mk_test.sh](../blob/unstable/tests/mk_test.sh) because the complain is in response to commands like `sudo kw d` and not `sudo kernel_deploy`. Would [kw_test.sh](../blob/unstable/tests/kw_test.sh) be a suitable file to add this test?

2. I faced this in ubuntu, would probably be the case for many distros. If we just want to set it up for travis, we can do that and  this point can be ignored.
Kw setup installs it in the local bin which is not in the sudo PATH env variable by default. Adding a test may lead to unnecessary failed tests due to `sudo: kw command not found` for developers.

src/mk.sh
harnaman-hk: Yes, it will be better under local deploy. I'll do the changes.

rodrigosiqueira:
Hi @harnaman-hk , do you have plans to update this PR?

harnaman-hk:
@rodrigosiqueira Yes.
I am facing some trouble in the test for this change. I added a test with kernel_install_local but then the test file needs to be run with sudo. Another way could be to export all variables and do sudo bash -c. I am not able to export all variables, I keep facing an unset configurations array. 
Can you guide me?

rodrigosiqueira:
> @rodrigosiqueira Yes.
> I am facing some trouble in the test for this change. I added a test with kernel_install_local but then the test file needs to be run with sudo. Another way could be to export all variables and do sudo bash -c. I am not able to export all variables, I keep facing an unset configurations array.
> Can you guide me?

Hmmmm... good point, I forgot about this. I have one idea, how mock the return from `$(id -u)`? Maybe you can have something like this in your test:

```
function id_mock()
{
 return 1
}

...
shopt -s expand_aliases
alias id='id_mock'
...

your test
```

Take a look at this page:
https://advancedweb.hu/how-to-mock-in-bash-tests/

and to `mk_test.sh`.

rodrigosiqueira:
Hi @harnaman-hk ,
I made some minors changes to your patch, and I applied it to the unstable branch. See: https://github.com/kworkflow/kworkflow/commit/8c950262231dc90d39dbd30b867a1b5af37c295f

Thanks a lot for your patch!

PR #240, opened 2021-04-21
Implement include function for file sourcing

alan-barzilay:
This PR is an alternative for PR #225 and also closes issue #197.
I've implemented the include function following some of @joaoseckler suggestions.

This is still a draft, the kw_include.sh file that is sourced in run_tests.sh is not present during any of the tests since apparently they are independent environments. Any suggestions on how to deal with this issue would be greatly appreciated

### DIFF
--- a/documentation/content/codingstyle.rst
+++ b/documentation/content/codingstyle.rst
@@ -180,6 +180,14 @@ $PATH, you should use 'type ', instead of 'which '.
 The output of 'which' is not machine parsable and its exit code
 is not reliable across platforms.
 
+How to include/import files
+---------------
+Do not source code using `.` or `source`. We have a helper function for that named
+ `kw_include` in `include.sh` and it should be used any and everytime a file needs to be sourced,
+`. file.sh --source-only` should only be used to source `include.sh` itself.
+The `include` function guarantees us that no file will be sourced twice,
+making the kw dev life easier with one thing less to worry about.
+
 Test name
 ---------
 
--- a/kw
+++ b/kw
@@ -29,6 +29,10 @@ KW_CACHE_DIR="$HOME/.cache/kw"
 
 # Export external variables required by kworkflow
 export KWORKFLOW
+# This is the one and only time a file will be sourced this way.
+# The include function (sourced from this file) should always be used for file sourcing.
+. "$KW_LIB_DIR/kw_include.sh" --source-only
+
 
 function kw()
 {
@@ -40,7 +44,7 @@ function kw()
   case "$action" in
     init)
       (
-        . "$KW_LIB_DIR/init.sh" --source-only
+        include "$KW_LIB_DIR/init.sh"
 
         init_kw
         return "$?"
@@ -48,7 +52,7 @@ function kw()
       ;;
     mount|mo)
       (
-        . "$KW_LIB_DIR/vm.sh" --source-only
+        include "$KW_LIB_DIR/vm.sh"
 
         vm_mount
         local ret="$?"
@@ -58,7 +62,7 @@ function kw()
       ;;
     umount|um)
       (
-        . "$KW_LIB_DIR/vm.sh" --source-only
+        include "$KW_LIB_DIR/vm.sh"
 
         vm_umount
         local ret="$?"
@@ -68,14 +72,14 @@ function kw()
       ;;
     up|u)
       (
-        . "$KW_LIB_DIR/vm.sh" --source-only
+        include "$KW_LIB_DIR/vm.sh"
 
         vm_up
       )
       ;;
     build|b)
       (
-        . "$KW_LIB_DIR/mk.sh" --source-only
+        include "$KW_LIB_DIR/mk.sh"
 
         mk_build '' "$@"
         local ret="$?"
@@ -85,7 +89,7 @@ function kw()
       ;;
     deploy|d)
       (
-        . "$KW_LIB_DIR/mk.sh" --source-only
+        include "$KW_LIB_DIR/mk.sh"
 
         kernel_deploy "$@"
         local ret="$?"
@@ -95,7 +99,7 @@ function kw()
       ;;
     bd)
       (
-        . "$KW_LIB_DIR/mk.sh" --source-only
+        include "$KW_LIB_DIR/mk.sh"
 
         mk_build && kernel_deploy "$@"
         local ret="$?"
@@ -104,76 +108,76 @@ function kw()
       )
       ;;
     diff|df)
-        . "$KW_LIB_DIR/diff.sh" --source-only
+        include "$KW_LIB_DIR/diff.sh"
 
         diff_manager "$@"
       ;;
     ssh|s)
       (
-        . "$KW_LIB_DIR/vm.sh" --source-only
+        include "$KW_LIB_DIR/vm.sh"
 
         vm_ssh $@
       )
       ;;
     vars)
       (
-        . "$KW_LIB_DIR/kw_config_loader.sh" --source-only
+        include "$KW_LIB_DIR/kw_config_loader.sh"
 
         show_variables
       )
       ;;
     codestyle|c)
       (
-      . "$KW_LIB_DIR/checkpatch_wrapper.sh" --source-only
+        include "$KW_LIB_DIR/checkpatch_wrapper.sh"
 
         execute_checkpatch "$@"
       )
       ;;
     maintainers|m)
       (
-        . "$KW_LIB_DIR/get_maintainer_wrapper.sh" --source-only
+        include "$KW_LIB_DIR/get_maintainer_wrapper.sh"
 
         execute_get_maintainer $@
       )
       ;;
     configm|g)
       (
-        . "$KW_LIB_DIR/config_manager.sh" --source-only
+        include "$KW_LIB_DIR/config_manager.sh"
 
         execute_config_manager "$@"
       )
       ;;
     help|h)
       (
-        . "$KW_LIB_DIR/help.sh" --source-only
+        include "$KW_LIB_DIR/help.sh"
 
         kworkflow_help
       )
       ;;
     explore|e)
       (
-        . "$KW_LIB_DIR/explore.sh" --source-only
+        include "$KW_LIB_DIR/explore.sh"
 
         explore "$@"
       )
       ;;
     man)
       (
-        . "$KW_LIB_DIR/help.sh" --source-only
+        include "$KW_LIB_DIR/help.sh"
 
         kworkflow_man
       )
       ;;
     version|--version|-v)
       (
-        . "$KW_LIB_DIR/help.sh" --source-only
+        include "$KW_LIB_DIR/help.sh"
 
         kworkflow_version
       )
       ;;
     statistics)
       (
-        . "$KW_LIB_DIR/statistics.sh" --source-only
+        include "$KW_LIB_DIR/statistics.sh"
 
         statistics "$@"
       )
@@ -187,21 +191,21 @@ function kw()
       ;;
 
     clear-cache)
-        . "$KW_LIB_DIR/mk.sh" --source-only
+        include "$KW_LIB_DIR/mk.sh"
 
         cleanup
       ;;
     # Subsystems support
     drm)
       (
-        . "$KW_PLUGINS_DIR/subsystems/drm/drm.sh" --source-only
+        include "$KW_PLUGINS_DIR/subsystems/drm/drm.sh"
 
         drm_manager "$@"
       )
       ;;
     *)
       (
-        . "$KW_LIB_DIR/help.sh" --source-only
+        include "$KW_LIB_DIR/help.sh"
 
         complain "Invalid option"
         kworkflow_help
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 
-. ./tests/utils --source-only
-. ./src/kwio.sh --source-only
+. ./src/kw_include.sh --source-only
+include './tests/utils'
+include './src/kwio.sh'
 
 declare -r PATH_TO_TESTS_EXTERNALS="tests/external"
 
--- a/src/kw_include.sh
+++ b/src/kw_include.sh
@@ -1,0 +1,31 @@
+#Used for file sourcing.
+#This function should used for file sourcing instead of `. file.sh --source-only`
+#
+# @filepath Path to the file to be sourced
+#
+# Returns:
+# 0 on a succesful import, 1 if the file can't be read and 2 if it can´t be found
+function include()
+{
+  local filepath="$1"
+  local varname=$(basename "$filepath" .sh)
+
+  if [[ ! -e "$filepath" ]]; then
+    echo "File $filepath could not be found, check your file path."
+    return 2 # ENOENT
+  fi
+
+  if [[ ! -r "$filepath" ]]; then
+    echo "File $filepath could not be read, check your file permissions."
+    return 1 # EPERM
+  fi
+
+  varname=${varname^^}_IMPORTED # capitalize and append "_IMPORTED"
+
+  if [[ -v "${varname}" ]]; then
+    return 0
+  fi
+
+  declare -g "${varname}"=1
+  . "$filepath" --source-only
+}
--- a/src/remote.sh
+++ b/src/remote.sh
@@ -1,5 +1,6 @@
 # NOTE: It is recommended that src/kw_config_loader.sh be included before this
 # file
+include "$KW_LIB_DIR/kw_config_loader.sh"
 
 # We now have a kw directory visible for users in the home directory, which is
 # used for saving temporary files to be deployed in the target machine.
--- a/tests/checkpatch_wrapper_test.sh
+++ b/tests/checkpatch_wrapper_test.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-. ./src/checkpatch_wrapper.sh --source-only
-. ./tests/utils --source-only
+include './src/checkpatch_wrapper.sh'
+include './tests/utils'
 
 function suite
 {
--- a/tests/configm_test.sh
+++ b/tests/configm_test.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-. ./src/config_manager.sh --source-only
-. ./tests/utils --source-only
+include './src/config_manager.sh'
+include './tests/utils'
 
 LS_TITLE="Name\t\tDescription"
 COMMAND_MSG_UNKNOWN="Unknown option"
--- a/tests/diff_test.sh
+++ b/tests/diff_test.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-. ./src/diff.sh --source-only
-. ./tests/utils --source-only
+include './src/diff.sh'
+include './tests/utils'
 
 function suite
 {
--- a/tests/kw_include_test.sh
+++ b/tests/kw_include_test.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+include './tests/utils'
+
+function suite()
+{
+  suite_addTest "include_Test"
+  suite_addTest "include_twice_Test"
+  suite_addTest "include_wrong_path_Test"
+}
+
+function include_Test()
+{
+  include ./src/kwio.sh
+  output="$KWIO_IMPORTED"
+  assertEquals "($LINENO)" 1 "$output"
+}
+
+function include_twice_Test() {
+  include ./src/kwlib.sh
+  include ./src/kwlib.sh
+  ret="$?"
+  assertEquals "($LINENO)" 0 "$ret"
+}
+
+function include_wrong_path_Test()
+{
+  output=$(include ./src/batata.sh)
+  ret="$?"
+  assertEquals "($LINENO)" 2 "$ret"
+}
+
+invoke_shunit
--- a/tests/kw_test.sh
+++ b/tests/kw_test.sh
@@ -1,8 +1,10 @@
 #!/bin/bash
 
-. ./tests/utils --source-only
+include './tests/utils'
 unset -v KW_LIB_DIR  # to be able to test the developer mode
-. ./kw --source-only > /dev/null
+include './kw' > /dev/null
+# when imported kw prints the help function and we don´t want
+# to polute our test results, so we redirect its output to /dev/null
 
 function suite
 {
--- a/tests/kwio_test.sh
+++ b/tests/kwio_test.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
-. ./tests/utils --source-only
-. ./src/kwio.sh --source-only
-. ./src/kwlib.sh --source-only
+include './tests/utils'
+include './src/kwio.sh'
+include './src/kwlib.sh'
 
 # NOTE: All executions off 'alert_completion' in this test file must be done
 # inside a subshell (i.e. "$(alert_completion ...)"), because this function
--- a/documentation/content/installanduninstall.rst
+++ b/documentation/content/installanduninstall.rst
@@ -16,7 +16,7 @@ following dependencies:
 
 * libguestfs
 * Qemu
-* Bash
+* Bash >= 4.4
 * git
 * tar
 * python-docutils
--- a/src/checkpatch_wrapper.sh
+++ b/src/checkpatch_wrapper.sh
@@ -2,8 +2,8 @@
 # in this file is to handle this script in a way to make this tool easier for
 # users.
 
-. "$KW_LIB_DIR/kw_config_loader.sh" --source-only
-. "$KW_LIB_DIR/kwlib.sh" --source-only
+include "$KW_LIB_DIR/kw_config_loader.sh"
+include "$KW_LIB_DIR/kwlib.sh"
 
 # Runs checkpatch in the given path, which might be a file or directory.
 #
--- a/src/config_manager.sh
+++ b/src/config_manager.sh
@@ -1,7 +1,7 @@
-. "$KW_LIB_DIR/kwio.sh" --source-only
+include "$KW_LIB_DIR/kwio.sh"
 
-declare -r metadata_dir="metadata"
-declare -r configs_dir="configs"
+declare -gr metadata_dir='metadata'
+declare -gr configs_dir='configs'
 
 function config_manager_help()
 {
--- a/src/diff.sh
+++ b/src/diff.sh
@@ -1,8 +1,8 @@
-. "$KW_LIB_DIR/kwio.sh" --source-only
-. "$KW_LIB_DIR/kwlib.sh" --source-only
+include "$KW_LIB_DIR/kwio.sh"
+include "$KW_LIB_DIR/kwlib.sh"
 
 # Hash containing user options
-declare -A diff_options
+declare -gA diff_options
 
 function diff_manager()
 {
--- a/src/explore.sh
+++ b/src/explore.sh
@@ -2,8 +2,8 @@
 # command it unifies the way that we search for things in the project, by
 # things, you can understand from files to messages in git log.
 
-. "$KW_LIB_DIR/kwio.sh" --source-only
-. "$KW_LIB_DIR/kwlib.sh" --source-only
+include "$KW_LIB_DIR/kwio.sh"
+include "$KW_LIB_DIR/kwlib.sh"
 
 # The main entry point for the explore feature, if you want to add another
 # search mechanism, you probably wish to start by this function
--- a/src/get_maintainer_wrapper.sh
+++ b/src/get_maintainer_wrapper.sh
@@ -1,5 +1,5 @@
-. "$KW_LIB_DIR/kw_config_loader.sh" --source-only
-. "$KW_LIB_DIR/kwlib.sh" --source-only
+include "$KW_LIB_DIR/kw_config_loader.sh"
+include "$KW_LIB_DIR/kwlib.sh"
 
 # Prints the authors of a given file or files inside a given dir.
 #
--- a/src/help.sh
+++ b/src/help.sh
@@ -1,4 +1,4 @@
-. "$KW_LIB_DIR/kwio.sh" --source-only
+include "$KW_LIB_DIR/kwio.sh"
 
 function kworkflow_help()
 {
--- a/src/init.sh
+++ b/src/init.sh
@@ -1,7 +1,7 @@
 # The init.sh keep all the operations related to the `kworkflow.config`
 # initialization. The initialization feature it is inspired on `git init`.
 
-. "$KW_LIB_DIR/kwio.sh" --source-only
+include "$KW_LIB_DIR/kwio.sh"
 
 # This function is responsible for creating a local kworkflow.config based in a
 # template available in the etc directory.
--- a/src/kw_config_loader.sh
+++ b/src/kw_config_loader.sh
@@ -1,4 +1,4 @@
-. "$KW_LIB_DIR/kwio.sh" --source-only
+include "$KW_LIB_DIR/kwio.sh"
 
 CONFIG_FILENAME=kworkflow.config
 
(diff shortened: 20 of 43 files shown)


src/remote.sh
```diff
@@ -1,5 +1,6 @@
 # NOTE: It is recommended that src/kw_config_loader.sh be included before this
 # file
+include $KW_LIB_DIR/kw_config_loader.sh
```
rodrigosiqueira, line 3: I did not understand why this change deserves a single commit. See if it makes sense to squash this commit with the previous one; if not, please elaborate a little bit more on the commit message.

documentation/content/codingstyle.rst
```diff
@@ -180,6 +180,12 @@ $PATH, you should use 'type ', instead of 'which '.
 The output of 'which' is not machine parsable and its exit code
 is not reliable across platforms.
 
+Sourcing files
```
rodrigosiqueira, line 183: Maybe something like:
Do not source code using `.` or `source`. We have a helper function for that named `include`...

Also, we don't need a single commit for this change. Please, squash it with the first commit.

tests/kw_include_test.sh
```diff
@@ -0,0 +1,19 @@
...
+}
+
+function include_twice_Test()
+{
+  ID=1
+  include ./src/kw_include.sh
+  output=$(include ./src/kw_include.sh)
+  expected_result="255"
+  assertEquals "($ID)" "$expected_result" "$output" 
+}
```
rodrigosiqueira, line 17: I think you can expand this test a little bit more... Some examples:
1. Test the wrong path;
2. Include case that works;
3. Double inclusion;
4. Wrong path parameter;

Finally, squash this change in the first commit.

src/kw_include.sh
```diff
@@ -0,0 +1,14 @@
+# Used for file sourcing,
+# accepts as argument the path for the file to be sourced
+function include() {
```
rodrigosiqueira, line 3: `{` should be in the next line.

tests/kw_include_test.sh
rodrigosiqueira: Also, take care of extra white space at the end of the line.

documentation/content/codingstyle.rst
rodrigosiqueira: Also, avoid adding extra white space at the end of the line.

rodrigosiqueira (changes requested):
Hi @alan-barzilay,

First of all, thanks for your PR. I added a couple of inline comments; please take a look at that.

About include in the run_tests.sh, we invoke code as described in the below code:
```
function run_tests()
{
...
      say "$SEPARATOR"
      (
        init_env
        "$target"
      )
...
}
```
If you want that the include function works per test, you probably need to load it after the `init_env`.

rodrigosiqueira:
Also, remember to rebase your branch with the latest code from the unstable branch.

src/kw_include.sh
```diff
@@ -0,0 +1,14 @@
...
+  local filename="$1"
+  local varname=${filename%.sh}
+  varname=${varname^^}_IMPORTED
+
+  if [ -n "${!varname}" ]; then
```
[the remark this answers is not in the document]
joaoseckler: @rodrigosiqueira notice the `!` indicating an indirect expansion. If the file is not sourced, the variable whose name is defined by `$varname` will not be defined. So `${!varname}` will expand to `""` and thus this test will not be true.

src/kw_include.sh
```diff
@@ -0,0 +1,14 @@
...
+    return 255
+  else
+    printf -v "${varname}" '%s' "1"
```
joaoseckler, line 11: Wouldn't `declare -g` be more readable here? See [documentation](https://www.gnu.org/software/bash/manual/bash.pdf) at pages 53, 54. Both are bash builtins, so it shouldn't matter too much. Curiosity: in my system `declare` was is about 20% faster than `printf` for this purpose.

src/kw_include.sh
```diff
@@ -0,0 +1,14 @@
+# Used for file sourcing,
+# accepts as argument the path for the file to be sourced
+function include() {
+  local filename="$1"
+  local varname=${filename%.sh}
+  varname=${varname^^}_IMPORTED
+
+  if [ -n "${!varname}" ]; then
+    return 255
```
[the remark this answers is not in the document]
joaoseckler: The ideal return error here would be something like "file already included". But there seems to be nothing like that in the errno list. Maybe this isn't even an error at all and should return 0. What do you think? Commenting here since this number was originally my suggestion

src/kw_include.sh
alan-barzilay: I agree it would be more readable, but older versions of bash don't have the -g option, @rodrigosiqueira should we care about supporting older versions or may I change this?

src/kw_include.sh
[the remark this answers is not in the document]
alan-barzilay: I like the idea of just returning 0. The user is only interested in a guarantee that the file has been sourced, I see no use in knowing if the file was sourced in this function call or if it was already sourced in the past, the result is the same and has no impact for the user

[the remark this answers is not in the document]
joaoseckler: @alan-barzilay I agree

src/kw_include.sh
```diff
@@ -0,0 +1,14 @@
...
+  else
+    printf -v "${varname}" '%s' "1"
+    . "$filename" --source-only # can return 0, 1 or 2
```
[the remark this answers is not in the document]
joaoseckler: Relevant [documentation](https://www.gnu.org/software/bash/manual/bash.html#Bourne-Shell-Builtins):

> The return status is the exit status of the last command executed,  or zero if no commands are executed.  If filename is not found, or cannot be read, the return status is non-zero.

src/kw_include.sh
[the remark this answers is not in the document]
rodrigosiqueira: Hi @alan-barzilay and @joaoseckler,

> notice the ! indicating an indirect expansion. If the file is not sourced, the variable whose name is defined by $varname will not be defined. So ${!varname} will expand to "" and thus this test will not be true.

Hmmm... interesting...
I checked in the documentation, and the only reference for `!` expansion that I found was:

> ${!prefix*}
> ${!prefix@}
> Expands to the names of variables whose names begin with prefix, separated by the first character of the IFS special variable. When ‘@’ is used and the expansion appears within double quotes, each variable name expands to a separate word.
> 
> ${!name[@]}
> ${!name[*]}
> If name is an array variable, expands to the list of array indices (keys) assigned in name. If name is not an array, expands to 0 if name is set and null otherwise. When ‘@’ is used and the expansion appears within double quotes, each key expands to a separate word.
[[1](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html)]

Notice that the above references are different from the ones that you used here. Did I miss something? Could you point me to a better reference?

Additionally, with this code, I'm getting this error message: `invalid variable name`.

Furthermore, we will always get this condition set to true due to the way that you declared `varname`:

```
varname=${varname^^}_IMPORTED
```

It does not matter the result from `${varname^^}`, `varname` will always have `_IMPORTED`. 

Finally, when I tested this PR, kw was totally broke; I could not see any feature working as expected. Install this PR and try a simple command like `kw c PATH/` OR `kw m PATH/`.

src/kw_include.sh
rodrigosiqueira: I think `declare -g` is a little bit different in this context because `varname` is a local variable (and should be). Fwiu, `-g` will affect global variables, am I wrong?

Also, you said 20% fast. How did you benchmark this? Also, in terms of ms, what is the difference?

src/kw_include.sh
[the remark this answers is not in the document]
rodrigosiqueira: Based on your comment, let's handle this error.

rodrigosiqueira:
@alan-barzilay , @joaoseckler any progress here?

tests/kw_include_test.sh
alan-barzilay: What is the difference between 1 and 4?

alan-barzilay:
> If you want that the include function works per test, you probably need to load it after the init_env.

@rodrigosiqueira I don't know if I'm doing something wrong but this doesn't seem to work, I still get include command not found messages

      (
        init_env
        . ./src/kw_include.sh --source-only
        "$target"
      )

src/kw_include.sh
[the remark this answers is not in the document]
alan-barzilay: I have found a better way of doing this, newer versions of bash have a -v flag that checks if a variable is set

rodrigosiqueira:
> > If you want that the include function works per test, you probably need to load it after the init_env.
> 
> @rodrigosiqueira I don't know if I'm doing something wrong but this doesn't seem to work, I still get include command not found messages
> 
> ```
>   (
>     init_env
>     . ./src/kw_include.sh --source-only
>     "$target"
>   )
> ```

Hi @alan-barzilay , did you try something different?

alan-barzilay:
> Hi @alan-barzilay , did you try something different?

No, I couldn't think of anything else. I don't really understand why that didn't work

src/kw_include.sh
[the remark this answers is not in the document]
rodrigosiqueira: Please, read all my previous comments carefully. Your new patch still has some of the problems that I pointed before.

rodrigosiqueira:
> > Hi @alan-barzilay , did you try something different?
> 
> No, I couldn't think of anything else. I don't really understand why that didn't work, I'm open to suggestions

Try this test:

`./run_tests.sh test tests/explore_test.sh`

After that, make this change:

```
  (
    init_env
    include "$target"
  )
```

And try to run the test again. Please, notice that this is not a fix for your issue, just a direction; your patch still has many issues. Try to debug your code behavior by adding a couple of `echo`; try to understand what's going on.

tests/kw_include_test.sh
alan-barzilay: @rodrigosiqueira could you clarify what is the difference between cases 1 and 4?

src/kw_include.sh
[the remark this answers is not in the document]
joaoseckler: About indirect expansions, the documentation can be found (here)[https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion]: 

> If the first character of parameter is an exclamation point (!), and parameter is not a nameref, it introduces a level of indirection. Bash uses the value formed by expanding the rest of parameter as the new parameter; this is then expanded and that value is used in the rest of the expansion, rather than the expansion of the original parameter. This is known as indirect expansion.

By the way, [this feature is in kw's codebase](https://github.com/kworkflow/kworkflow/blob/7b7b8e44d5f65e7f29188958423b044edfafebf7/run_tests.sh#L55). But @alan-barzilay found a more elegant solution  using `-v` inside the double brackets.

src/kw_include.sh
joaoseckler: @rodrigosiqueira You are rigth that the `-g` option affects global variables, but that is exactly the case here. We need some way to remember which files were already included, and these global variables serve this purpose. This function should assert that if a file has been included, there is a global variable associated with that variable whose value is set to 1.

joaoseckler:
Hey @rodrigosiqueira @alan-barzilay,

Below a list of problems I found with the current version of this PR.

1. The include function is not being exported to subprocess calls of test functions. To amend this, I would suggest simply adding the line `export -f include` in the function `init_env` inside `test/utils`. This does forget which files have already been included, but that is coherent with the fact that the test scripts have to include files again anyway (since they are subprocesses);
2. The standard behavior of the builtin `declare` is to declare a local variable. When the sourcing of a file is made inside a function (as it is with this new system), all the variables declared with `declare` inside the sourced files will be considered variables local to the `include` function. The only solution I can see to this is to explicitly include the `-g` option in the entire codebase whenever it is the case that the variable is global;
3. The test function in `tests/kw_include_test` doesn't work since *i)* it calls `include` in a subprocess to get it's return value, instead of calling it in the function and getting the return value with `$?` and *ii)* the return values have been changed since this PR's first version and the test has not been changed accordingly. Suggestion: test if the correct variable has been created when importing an existent file.
4. My suggestion, accepted by Alan, of remembering if a file has already been included by setting a global variable has problems: it requires that we transform the filename into a variable name. In the current way we do this, if a file has the same `basename` but is in a different folder, it will not be included, since our code will create the same variable name for both (*). Another issue is that if the file has another format than `*.sh` or has a dot in the middle of it's name, our code fails with the error `bash: printf: 'EXAMPLE.FILE_IMPORTED': not a valid identifier`. I have two solutions in mind: use a proper hash table to remember if a file has already been included or change the variable name creation a bit to avoid these problems.

I propose modifications to address these issues, except for (*), in a branch in my fork: https://github.com/joaoseckler/kworkflow/tree/include_manager. These modifications are also in this patch file: [patch.txt](https://github.com/kworkflow/kworkflow/files/6533587/patch.txt)

alan-barzilay:
> The include function is not being exported to subprocess calls of test functions. To amend this, I would suggest simply adding the line export -f include in the function init_env inside test/utils. This does forget which files have already been included, but that is coherent with the fact that the test scripts have to include files again anyway (since they are subprocesses);


I like this approach, I had tried to make the include function visible by doing `  (
    init_env
    include "$target"
  )`  and `  (
    init_env
    . "$target"
  )` but these approaches will soil the test environment with everything imported in run_tests.sh while your idea keeps everything clean.



> The standard behavior of the builtin declare is to declare a local variable. When the sourcing of a file is made inside a function (as it is with this new system), all the variables declared with declare inside the sourced files will be considered variables local to the include function. The only solution I can see to this is to explicitly include the -g option in the entire codebase whenever it is the case that the variable is global;

Good catch! This solved all the esoteric errors we were having 

    

> My suggestion, accepted by Alan, of remembering if a file has already been included by setting a global variable has problems: it requires that we transform the filename into a variable name. In the current way we do this, if a file has the same basename but is in a different folder, it will not be included, since our code will create the same variable name for both (*). Another issue is that if the file has another format than *.sh or has a dot in the middle of it's name, our code fails with the error bash: printf: 'EXAMPLE.FILE_IMPORTED': not a valid identifier. I have two solutions in mind: use a proper hash table to remember if a file has already been included or change the variable name creation a bit to avoid these problems.

I was thinking about something similar, I believe we can use the `basename` function to solve some of these issues and to make our code more readable. About the file having a . in its name besides the one for .sh, I think we should just disallow this sort of name in our codestyles. I cant think of a legitimate reason for doing this

alan-barzilay:
Using the suggestions made by @joaoseckler I finished implementing the include function, it now has 3 tests and error handling for the cases when the file to be sourced cant be found or cant be read. After re-basing onto the unstable branch, the year_statistics_Test started failing on my machine, but it still passes on travis so I´m not sure if I just messed up my environment somehow. I would appreciate if anyone could test this PR and let me know if they also fail this test

magalilemes:
@alan-barzilay, I applied your patch to the current version of the unstable branch and the tests are running fine, except for this message `./src/kw_include.sh: line 3: /kwio.sh: No such file or directory` at the beginning of the output after running the tests.

alan-barzilay:
@magalilemes thank you, I must have messed up my environment somehow.

>  except for this message ./src/kw_include.sh: line 3: /kwio.sh: No such file or directory at the beginning of the output after running the tests.

You are right, I hadn´t noticed that. I think I understand why this is happening, we need to import kwio before using the include function since we need it for the "error handling" part, but in `run_tests.sh`, before we run the tests, the environment variable  `KW_LIB_DIR`has not been set. This causes an error because the file can´t be found.

I see 2 solutions, stop using "complain" inside include, or change the KW_LIB_DIR logic. @rodrigosiqueira  do you have a preference?

alan-barzilay:
> I see 2 solutions, stop using "complain" inside include, or change the KW_LIB_DIR logic. @rodrigosiqueira do you have a preference?

Or we could merge the include and kwio files

rodrigosiqueira:
> @magalilemes thank you, I must have messed up my environment somehow.
> 
> > except for this message ./src/kw_include.sh: line 3: /kwio.sh: No such file or directory at the beginning of the output after running the tests.
> 
> You are right, I hadn´t noticed that. I think I understand why this is happening. We need to import kwio before using the include function since we need it for the "error handling" part, but in `run_tests.sh`, before we run the tests, the environment variable `KW_LIB_DIR`has not been set. This causes an error because the file can´t be found.
> 
> I see 2 solutions, stop using "complain" inside include, or change the KW_LIB_DIR logic. @rodrigosiqueira do you have a preference?

Let's remove "complain" from include.

alan-barzilay:
@rodrigosiqueira, done

joaoseckler (approved):

src/kw_include.sh
```diff
@@ -0,0 +1,25 @@
+# Used for file sourcing,
+# accepts as argument the path for the file to be sourced
```
rodrigosiqueira, line 2: Please, expand this documentation and follow the pattern used in other files (see `remote.sh`).

src/kw_include.sh
```diff
@@ -0,0 +1,25 @@
...
+function include()
+{
+  local filename="$1"
```
rodrigosiqueira, line 5: Add an empty line between local and `if`

rodrigosiqueira, line 5: Declare `varname` at the beginning of the function.

src/kw_include.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "File $filename could not be read, check your file permissions."
+    return 1
+  fi
+
+  local varname=$(basename "$filename" .sh)
+  varname=${varname^^}_IMPORTED # capitalize and append "_IMPORTED"
+
+  if [[ -v "${varname}" ]]; then
+    return 0
+  else
```
rodrigosiqueira, line 21: You don't need this else.

src/kw_include.sh
```diff
@@ -0,0 +1,25 @@
...
+    declare -g "${varname}"=1
+    . "$filename" --source-only # can return 0, 1 or 2
```
rodrigosiqueira, line 23: From bash documentation:
> If filename is not found, or cannot be read, the return status is non-zero.

In this sense, you already checked the status in the previous if condition, so... no need for the extra comment.

src/kw_include.sh
```diff
@@ -0,0 +1,25 @@
+# Used for file sourcing,
+# accepts as argument the path for the file to be sourced
+function include()
+{
+  local filename="$1"
+  if [[ ! -e "$filename" ]]; then
+    echo "File $filename could not be found, check your file path."
+    return 2
```
rodrigosiqueira, line 8: Add a comment:
`# ENOENT`

src/kw_include.sh
```diff
@@ -0,0 +1,25 @@
...
+  fi
+
+  if [[ ! -r "$filename" ]]; then
+    echo "File $filename could not be read, check your file permissions."
+    return 1
```
rodrigosiqueira, line 13: Add a comment `EPERM`

kw
```diff
@@ -25,6 +25,8 @@ statistics_path=${statistics_path:-"$KW_DATA_DIR/statistics"}
 
 # Export external variables required by kworkflow
 export KWORKFLOW
+. "$KW_LIB_DIR/kw_include.sh" --source-only
```
rodrigosiqueira, line 28: Add a comment here that highlights how "special" it is this `.`.

run_tests.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
 
-. ./tests/utils --source-only
-. ./src/kwio.sh --source-only
+. ./src/kw_include.sh --source-only
+include "./tests/utils"
+include "./src/kwio.sh"
 
-declare -r PATH_TO_TESTS_EXTERNALS="tests/external"
+declare -gr PATH_TO_TESTS_EXTERNALS="tests/external"
```
rodrigosiqueira, line 7: I don't think we need to make this global.

run_tests.sh
```diff
@@ -156,7 +157,7 @@ function run_tests
   report_results "$total" "$success" "$notfound" "$fail" "$test_failure_list"
 }
 
-declare -a TESTS
+declare -ga TESTS
```
rodrigosiqueira, line 160: Same here. We don't need to add `g` here.

tests/kw_include_test.sh
```diff
@@ -0,0 +1,40 @@
...
+  suite_addTest "include_Test"
+  suite_addTest "include_twice_Test"
+  suite_addTest "include_wrong_path_Test"
+}
+
+function include_Test()
+{
+  ID=1
+  include ./src/kwio.sh
+  expected_result="1"
```
rodrigosiqueira, line 16: You don't need this variable; put it directly to the `assertEquals` (same this entire file).

tests/kw_include_test.sh
```diff
@@ -0,0 +1,40 @@
...
+  output=$KWIO_IMPORTED
+  assertEquals "($ID)" "$expected_result" "$output"
+}
+
+function include_twice_Test()
+{
+  ID=2
+  include ./src/kwlib.sh
+  include ./src/kwlib.sh
```
rodrigosiqueira, line 25: Just for curiosity, what happens if I include the include?

tests/kwio_test.sh
```diff
@@ -1,15 +1,15 @@
...
+include "./src/kwio.sh"
+include "./src/kwlib.sh"
 
 # NOTE: All executions off 'alert_completion' in this test file must be done
 # inside a subshell (i.e. "$(alert_completion ...)"), because this function
 # invokes other commands in the background. So if not done inside a subshell,
 # the function will return before the background commands finish.
 
-declare -A configurations
+declare -gA configurations
```
rodrigosiqueira, line 12: I don't think you need `g` here.

tests/plugins/kernel_install/arch_test.sh
```diff
@@ -11,7 +11,7 @@ function suite
   suite_addTest 'generate_arch_temporary_root_file_system_Test'
 }
 
-declare -r TEST_ROOT_PATH="$PWD"
+declare -gr TEST_ROOT_PATH="$PWD"
```
rodrigosiqueira, line 14: I don't think you need `g` here.
(Same for all tests that you changed)

rodrigosiqueira (changes requested):
Hi @alan-barzilay and @joaoseckler 
Thanks a lot for this new version; we had great progress here. Please, see my inline comments.

src/remote.sh
```diff
@@ -1,5 +1,6 @@
 # NOTE: It is recommended that src/kw_config_loader.sh be included before this
 # file
+include "$KW_LIB_DIR/kw_config_loader.sh"
```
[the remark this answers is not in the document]
alan-barzilay: I think it would be good to keep it in order to give some context, its not obvious that  config loader needs to be imported before remote and this makes it explicit

tests/kw_include_test.sh
alan-barzilay: it will source itself again, but if you try it again than it wont be sourced twice(or trice, actually). We can declare a INCLUDE_IMPORTED global variable as well to keep it from sourcing it the first time but I think it is unnecessary. Nobody should even consider including include and this won't protect it from being sourced twice with ` . kw_include etc`. If we are worried about this then we should use an include guard like the original idea for this issue.

rodrigosiqueira: Don't worry about that... I just asked for curiosity, maybe explain this on the documentation page.

alan-barzilay:
@rodrigosiqueira  continuing the discussion from #286, since bash 4.4 https://github.com/bminor/bash/blob/a0c0a00fc419b7bc08202a79134fcd5bc0427071/NEWS  there is support for the declare -g option. According to wikipedia, 4.4 was released in 2016

tests/kw_include_test.sh
```diff
@@ -0,0 +1,40 @@
...
+  ret=$?
+  expected_result="0"
+  assertEquals "($ID)" "$expected_result" "$ret"
```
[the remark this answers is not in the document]
alan-barzilay: ok, but I just copied the scheme from another test, is this something that in the future should be changed in all tests in kw?

src/kw_include.sh
```diff
@@ -0,0 +1,31 @@
...
+  fi
+
+  varname=${varname^^}_IMPORTED # capitalize and append "_IMPORTED"
+
+  if [[ -v "${varname}" ]]; then
+    return 0
+  fi
+
+  declare -g "${varname}"=1
+  . "$filepath" --source-only > /dev/null
```
alan-barzilay, line 30: I added a `> /dev/null` because the kw_test was including kw and messing up the tests output:

```
Running test [kw_test]
Invalid option
Usage: kw [options]

The current supported targets are:
         Host - this machine
         Qemu - qemu machine
         Remote - machine reachable via the network

Commands:
        bd - Build and install modules
        build|b - Build kernel
        clear-cache - Clear files generated by kw
        codestyle,c - Apply checkpatch on directory or file
        configm,g - Manage config files
        deploy,d - Deploy a new kernel image to a target machine
        diff,df - Diff files
        drm - Set of commands to work with DRM drivers 
        explore,e - Explore string patterns
        help,h - displays this help mesage
        init - Initialize kworkflow config file
        maintainers,m - Get maintainers and mailing list
        man - Show manual
        mount,mo - Mount partition with qemu-nbd
        pomodoro,p - kw pomodoro support
        ssh,s - SSH support
        statistics - Provide basic statistics related to daily development
        umount,um - Umount partition created with qemu-nbd
        up,u - Wake up vm
        vars,v - Show variables
        version,--version,-v - show kw version

validate_global_variables_Test
check_kworkflow_global_variable_Test
set_KW_LIB_DIR_in_dev_mode_Test

Ran 3 tests.
 ```

File: documentation/content/codingstyle.rst [hunk 25]
```diff
@@ -180,6 +180,13 @@ $PATH, you should use 'type ', instead of 'which '.
...
+---------------
+Do not source code using `.` or `source`.
+We have a helper function for that named `include` in `include.sh` and it
+should be used any and everytime a file needs to be sourced,
+`. file.sh --source-only` should only be used to source `include.sh` itself.
```
rodrigosiqueira, line 188: the file name is `kw_include.sh` not `include.sh`.

documentation/content/codingstyle.rst
```diff
@@ -180,6 +180,13 @@ $PATH, you should use 'type ', instead of 'which '.
 The output of 'which' is not machine parsable and its exit code
 is not reliable across platforms.
 
+Sourcing files
+---------------
+Do not source code using `.` or `source`.
```
rodrigosiqueira, line 185: Why break the line here? It looks weird to me.

documentation/content/codingstyle.rst
```diff
@@ -180,6 +180,13 @@ $PATH, you should use 'type ', instead of 'which '.
 The output of 'which' is not machine parsable and its exit code
 is not reliable across platforms.
 
+Sourcing files
```
rodrigosiqueira, line 183: Maybe something like "How to include code" or something similar that gives the idea of inclusion operation.

documentation/content/codingstyle.rst
rodrigosiqueira: Also, explain why use `include` and its advantage here.

kw
```diff
@@ -193,21 +197,21 @@ function kw()
...
-        . "$KW_LIB_DIR/mk.sh" --source-only
+        include "$KW_LIB_DIR/mk.sh"
 
         cleanup
       ;;
     # Subsystems support
     drm)
       (
-        . "$KW_PLUGINS_DIR/subsystems/drm/drm.sh" --source-only
+        include "$KW_LIB_DIR/subsystems/drm/drm.sh"
```
rodrigosiqueira, line 207: Why did you replace `$KW_PLUGINS_DIR` with `$KW_LIB_DIR`?

run_tests.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
 
-. ./tests/utils --source-only
-. ./src/kwio.sh --source-only
+. ./src/kw_include.sh --source-only
+include "./tests/utils"
+include "./src/kwio.sh"
```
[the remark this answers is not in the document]
rodrigosiqueira: Please, change this in your next patch.

src/kw_include.sh
```diff
@@ -0,0 +1,31 @@
+#Used for file sourcing.
+#This function should used for file sourcing instead of `. file.sh --source-only`
+#
+# @filepath Path to the file to be sourced
+#
+# Returns:
+# 0 on a succesful import, 1 if the file can't be read and 2 if it can´t be found
+function include() {
```
rodrigosiqueira, line 8: https://siqueira.tech/doc/kw/content/codingstyle.html#functions

tests/checkpatch_wrapper_test.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-. ./src/checkpatch_wrapper.sh --source-only
-. ./tests/utils --source-only
+include "./src/checkpatch_wrapper.sh"
+include "./tests/utils"
```
rodrigosiqueira, line 4: For your next version, prefer a single quote here.

tests/configm_test.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-. ./src/config_manager.sh --source-only
-. ./tests/utils --source-only
+include "./src/config_manager.sh"
+include "./tests/utils"
```
rodrigosiqueira, line 4: For your next version, prefer a single quote here.

tests/diff_test.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-. ./src/diff.sh --source-only
-. ./tests/utils --source-only
+include "./src/diff.sh"
+include "./tests/utils"
```
rodrigosiqueira, line 4: For your next version, prefer a single quote here.
(please, review all of these quotes in your series).

tests/kw_include_test.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+include "./tests/utils"
+
+function suite() {
```
rodrigosiqueira, line 5: https://siqueira.tech/doc/kw/content/codingstyle.html#functions

tests/kw_include_test.sh
```diff
@@ -0,0 +1,30 @@
...
+  suite_addTest "include_Test"
+  suite_addTest "include_twice_Test"
+  suite_addTest "include_wrong_path_Test"
```
rodrigosiqueira, line 8: single-quote

tests/kw_include_test.sh
```diff
@@ -0,0 +1,30 @@
...
+}
+
+function include_Test() {
+  include ./src/kwio.sh
+  output=$KWIO_IMPORTED
```
rodrigosiqueira, line 13: double-quote: `"$KWIO_IMPORTED"`

tests/kw_include_test.sh
```diff
@@ -0,0 +1,30 @@
...
+  assertEquals "($LINENO)" "1" "$output"
```
rodrigosiqueira, line 14: Maybe drop quotes around numbers in this entire test.

tests/kw_test.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 
-. ./tests/utils --source-only
+include "./tests/utils"
 unset -v KW_LIB_DIR  # to be able to test the developer mode
-. ./kw --source-only > /dev/null
+include "./kw" > /dev/null
```
rodrigosiqueira, line 5: This looks like a corner case... at least document this behavior in the include function and the documentation.

rodrigosiqueira (changes requested):
@alan-barzilay 
About the bash version, please document the minimum bash version required for `kw` in the documentation. Additionally, in the commit message, add a link reference to the `NEW` file that you pointed out (please find the original file link or at least something more official, do not use the mirror link).

kw
alan-barzilay: uou, nice catch. I was probably blindly making the adaption and didn´t realise the path for drm was different. Sorry about that

run_tests.sh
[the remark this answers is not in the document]
rodrigosiqueira: @alan-barzilay why did you ignore all the single-quote comments?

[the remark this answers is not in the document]
alan-barzilay: I wanted to talk about this on our meeting tomorrow, I was planning on arguing that we should keep it as is. Also I'm lazy and was leaving it for last because its going to be a boring manual process haha

rodrigosiqueira:
Hi @alan-barzilay and @joaoseckler ,

Thanks a lot for all your effort to make this include function possible and to deliver a high-quality code. I really appreciate all the effort that both of you put in here. Thanks a lot!

Merged, see: https://github.com/kworkflow/kworkflow/commit/cf8bc05c69ff8250727c7a2cdc59cfc13cf6e35a

PR #247, opened 2021-04-27
Add error message for wrong commands set in the alert variables

### LINKED ISSUE #219: notify-send package error message
Whenever a user invokes a command that sends notifications when finished running, the 'alert' variable in the kworkflow.config file is set to either 'v' or 'vs' and the notify-send package isn't installed, an error message like this is displayed:

`/home/magali/.local//lib/kw/kwio.sh: line 31: notify-send: command not found`

It'd be better if this message wasn't displayed.

magalilemes:
This PR adds a message in case the commands assigned to the alert variables 'sound_alert_command' and 'visual_alert_command' in the kworkflow.config file don't work.

I changed the testing command from "echo" to "/bin/echo", since the latter isn't a built-in, but a file in fact. This was done, so that the command_exists function returns 0 to the testing command. I don't know if this is the best approach, though.

Closes #219

### DIFF
--- a/src/kwio.sh
+++ b/src/kwio.sh
@@ -28,9 +28,23 @@ function alert_completion()
 
   grep -o . <<< "$opts" | while read option;  do
     if [ "$option" == "v" ]; then
-      eval "${configurations[visual_alert_command]} &"
+      if command_exists "${configurations[visual_alert_command]} &"; then
+        eval "${configurations[visual_alert_command]} &"
+      else
+        warning "The following command set in the visual_alert_command variable" \
+        "couldn't be run:"
+        warning "${configurations[visual_alert_command]}"
+        warning "Check if the necessary packages are installed."
+      fi
     elif [ "$option" == "s" ]; then
-      eval "${configurations[sound_alert_command]} &"
+      if command_exists "${configurations[sound_alert_command]} &"; then
+        eval "${configurations[sound_alert_command]} &"
+      else
+        warning "The following command set in the sound_alert_command variable" \
+        "couldn't be run:"
+        warning  "${configurations[sound_alert_command]}"
+        warning "Check if the necessary packages are installed."
+      fi
     fi
   done
 }
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -313,3 +313,17 @@ function store_statistics_data
 
   echo "$label $value" >> "$day_path"
 }
+
+# This function checks if a certain command can be run
+#
+# @command The whole command that is meant to be executed
+function command_exists()
+{
+  local command="$1"
+  local package=( $command )
+
+  if [[ -x "$(command -v $package)" ]]; then
+    return 0
+  fi
+  return 22 # EINVAL
+}
--- a/tests/kwio_test.sh
+++ b/tests/kwio_test.sh
@@ -2,6 +2,7 @@
 
 . ./tests/utils --source-only
 . ./src/kwio.sh --source-only
+. ./src/kwlib.sh --source-only
 
 # NOTE: All executions off 'alert_completion' in this test file must be done
 # inside a subshell (i.e. "$(alert_completion ...)"), because this function
@@ -17,6 +18,7 @@ function suite
   suite_addTest "alert_completion_options_Test"
   suite_addTest "alert_completition_validate_config_file_options_Test"
   suite_addTest "alert_completion_visual_alert_Test"
+  suite_addTest "alert_completion_sound_alert_Test"
 }
 
 function setUp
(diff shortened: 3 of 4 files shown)


src/kwlib.sh
```diff
@@ -313,3 +313,17 @@ function store_statistics_data
...
+
+# This function checks if a certain command can be run
+#
+# @command The whole command that is meant to be executed
+function command_exists()
+{
+  local command="$1"
+  local package=( $command )
+
+  if [ -x "$(command -v $package)" ]; then
```
rodrigosiqueira, line 325: Use this code style here:
```
[[ -x "$(command -v $package)" ]]
```

src/kwlib.sh
```diff
@@ -313,3 +313,17 @@ function store_statistics_data
...
   echo "$label $value" >> "$day_path"
 }
+
+# This function checks if a certain command can be run
+#
+# @command The whole command that is meant to be executed
+function command_exists()
+{
+  local command="$1"
+  local package=( $command )
```
rodrigosiqueira, line 323: Just for curiosity, why do we need `( $command )` why not just `"$command"`?

src/kwio.sh
```diff
@@ -28,9 +28,19 @@ function alert_completion()
 
   grep -o . <<< "$opts" | while read option;  do
     if [ "$option" == "v" ]; then
-      eval "${configurations[visual_alert_command]} &"
+      if command_exists "${configurations[visual_alert_command]} &"; then
+        eval "${configurations[visual_alert_command]} &"
+      else
+        complain "The command set in the visual_alert_command variable" \
+        "couldn't be run. Check if the necessary packages are installed."
```
rodrigosiqueira, line 35: Use `warning` instead of `complain`.

I think it would be nice to show the value inside `${configurations[visual_alert_command]}` for the user. Same for the sound alert.

src/kwlib.sh
```diff
@@ -313,3 +313,17 @@ function store_statistics_data
...
+
+  if [ -x "$(command -v $package)" ]; then
+    return 0
+  fi
+  return 1
```
rodrigosiqueira, line 328: Maybe `22` (EINVAL).

tests/kwio_test.sh
```diff
@@ -104,10 +106,19 @@ function alert_completition_validate_config_file_options_Test
...
   assertEquals "Variable $v should exist." "$ret" "$expected"
   true
 }
 
+function alert_completion_sound_alert_Test
+{
+  local expected="TESTING COMMAND"
+  configurations["sound_alert_command"]="/bin/echo \$COMMAND"
+  ret="$(alert_completion "$expected" "--alert=s")"
+  assertEquals "Variable $s should exist." "$ret" "$expected"
```
rodrigosiqueira, line 120: `$s`? Is that correct?

tests/kwio_test.sh
```diff
@@ -104,10 +106,19 @@ function alert_completition_validate_config_file_options_Test
...
   ret="$(alert_completion "$expected" "--alert=v")"
   assertEquals "Variable $v should exist." "$ret" "$expected"
   true
 }
 
+function alert_completion_sound_alert_Test
+{
+  local expected="TESTING COMMAND"
+  configurations["sound_alert_command"]="/bin/echo \$COMMAND"
+  ret="$(alert_completion "$expected" "--alert=s")"
```
rodrigosiqueira, line 119: Maybe `output` instead of `ret`. Also, declare it as a `local` variable and all other variables that belong to this specific test.

tests/kwio_test.sh
```diff
@@ -104,10 +106,19 @@ function alert_completition_validate_config_file_options_Test
...
+  assertEquals "Variable $s should exist." "$ret" "$expected"
+  true
```
rodrigosiqueira, line 121: Do we need this?

tests/kwlib_test.sh
```diff
@@ -355,4 +356,21 @@ function statistics_manager_Test
   assertTrue "($ID) Database day" '[[ ! -f "$statistics_path/$this_year_and_month/$today" ]]'
 }
 
+function command_exists_Test
+{
+  local ID
+  local fake_command="a-non-existent-command -p"
+  local real_command="mkdir"
+
+  ID=1
```
rodrigosiqueira, line 365: We want to avoid the ID variable now. You can drop it from this test and replace all occurrences of `$ID` with `$LINENO`.

rodrigosiqueira (changes requested):
Hi @magalilemes,
That's a very nice PR! Thanks a lot.

Please, see my inline comments.

src/kwlib.sh
magalilemes: `( $ command )` retrieves the first word of the whole command. So, for instance, suppose we have a command like
`notify-send -i checkbox -t 10000 "kw" "Command: \"$COMMAND\" completed!" &`
If I do `package="$command"`, I'll have the command entirely, and I believe we only need the first word of it. And to get the first word, I did ( $command ). 
Does that make sense?

tests/kwio_test.sh
magalilemes: What's "this" exactly? hehe :-)

tests/kwio_test.sh
magalilemes: I tried to follow the test done by the `alert_completion_visual_alert_Test` function right above this one. But looking at it now, I can't seem to find any variable named `s`. My bad!
That also makes me wonder why the `alert_completion_visual_alert_Test` function mentions a `v` variable and where it comes from.

rodrigosiqueira:
Thanks a lot for your patch.
Merged, see https://github.com/kworkflow/kworkflow/commit/6b49e1268c0f3b4c5bf3784f4738fa081d42ccfb

PR #248, opened 2021-04-27
run_tests: remove warning when running ./run_tests.sh list

magalilemes:
When `./run_tests.sh list` is run, the following warning appears:
`./run_tests.sh: line 179: local: can only be used in a function`
This PR fixes the warning by declaring the 'index' variable globally.


rodrigosiqueira:
Nice catch! Merged.

PR #266, opened 2021-05-18
Run uninstalled

terceiro:
This PR adds support for running kw directly from the source code, without installing it first.

### DIFF
--- a/kw
+++ b/kw
@@ -7,6 +7,16 @@ KW_INSTALL_PREFIX=${KW_INSTALL_PREFIX:-"##KW_INSTALL_PREFIX_TOKEN##"}
 
 KWORKFLOW=${KWORKFLOW:-"kw"}
 
+##BEGIN-DEV-MODE##
+KW_BIN="$(readlink -f "${BASH_SOURCE[0]}")"
+KW_BASE_DIR="$(dirname "${KW_BIN}")"
+if [ -f "${KW_BASE_DIR}/src/kwlib.sh" ]; then
+  # running from source directory
+  KW_LIB_DIR="${KW_BASE_DIR}/src"
+  KW_INSTALL_PREFIX="${KW_BASE_DIR}"
+fi
+##END-DEV-MODE##
+
 # Global paths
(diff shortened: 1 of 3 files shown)


kw
```diff
@@ -7,6 +7,16 @@ KW_INSTALL_PREFIX=${KW_INSTALL_PREFIX:-"##KW_INSTALL_PREFIX_TOKEN##"}
 
 KWORKFLOW=${KWORKFLOW:-"kw"}
 
+##UNINSTALLED##
```
rodrigosiqueira, line 10: I I think that "UNINSTALLED" is not a good name for this functionality. Fwiu, you proposed this change to simplify the development mode cycle by enabling developers to use `./kw` and avoid `setup -i`. In this sense, I think that a name like "DEV_MODE" (or similar) better describes the nature of this code.

rodrigosiqueira (changes requested):
Hi Terceiro,

First of all, thanks a lot for your PR; that's a significant improvement in the development workflow.

In general, it looks good to me. You will see a comment in the code, and I also have two generic comments:

1. For new PRs, use the unstable branch and send PR to this branch. I recently merged the unstable branch into the master; for this reason, they look similar. Nevertheless, the development happens in the unstable branch, and you will see a gap between these two branches soon. For more information, see https://siqueira.tech/doc/kw/content/howtocontribute.html
2. We have a simple test for `kw` file (we need to improve it). Since you're adding this new code path, could you also add a test for this part? Just check if the variables are correctly set in this "uninstalled" mode. See: https://github.com/kworkflow/kworkflow/blob/master/tests/kw_test.sh

Thanks
Siqueira

terceiro:
I have just pushed a new version that should address your comments. the testing part was tricky, as `test/utils` already sets `KW_LIB_DIR` to be able to run tests, so `kw_test.sh` needs to unset it before sourcing `kw`. otherwise we would be testing `test/utils` and not this feature.

rodrigosiqueira:
Hi @terceiro  , thanks a lot for the new version and your effort to add tests for it.

I already applied your change to the unstable branch, see: https://github.com/kworkflow/kworkflow/commit/91e2527583e6147fa17b30b211da08c9aa3c4315

PR #274, opened 2021-05-20
src: etc: update network setup for QEMU

### LINKED ISSUE #257: Default network options for QEMU vm uses legacy option -net
The current default network configuration options for the QEMU virtual machine utilizes the `-net` option, which from the official documentation has been obsolete since version `0.12` (released in 2010). We should update the relevant files to use the new options `-devices`, `-netdev` and `-nic` options.

kwy95:
The previous default network options for QEMU used a legacy option `-net`.
This commit updates the default to use the newer `-nic`.
More specific network configurations could be set using `-netdev`, but this
is somewhat dependent on the individual setup, and I believe the solution used
is more general.

Closes #257

### DIFF
--- a/documentation/man/kw.rst
+++ b/documentation/man/kw.rst
@@ -424,7 +424,7 @@ Sets QEMU options. By default, **kw** uses
 qemu_net_options=OPTIONS
 ------------------------
 Defines the network configuration. By default, **kw** uses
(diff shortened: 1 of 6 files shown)


rodrigosiqueira:
Thanks @kwy95 :)
Merged, see https://github.com/kworkflow/kworkflow/commit/f6661c3ff647386f8b35ce2bde1ac0e74c30c4b4

PR #277, opened 2021-05-22
setup: Make setup show usage when incorrect number of args given

kwy95:
I felt it would be useful especially to new users to give this information
if they try to use setup incorrectly.
Alternatively we could show a different custom message.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -408,6 +408,7 @@ case "$1" in
     ;;
   *)
     complain "Invalid number of arguments"
+    usage
     exit 1
     ;;
 esac


kwy95:
As an alternative message I suggest:
`To show usage run: ./setup.sh -i`

kwy95:
Actually this command: `./setup.sh -h`

rodrigosiqueira:
Nice patch! Thanks @kwy95 
Merged to unstable branch, see:
https://github.com/kworkflow/kworkflow/commit/5e77902345e393838ea64717b721b64bc9183edd

Issue #270, opened 2021-05-20
Add signal handling system 


joaoseckler:
Currently, kw does not handle signals. If you pass it a SIGINT, for example, the subprocess that happens to be executing at that particular moment is the one who will receive and treat it. This results in a poor user interface, where the standard Ctrl-C doesn’t have a meaningful outcome. Worse than that, a simple SIGINT could leave the environment polluted by unremoved temporary files.

Possible breakdown:
 - Implementing a signal manager system for kw. It’s interface should allow the creation of handlers for arbitrary signals and to define at what point in the code execution those handlers are active or not;
 - Implementing default signal manager;
 - Defining a set of directives for kw developers on how to use the signal manager.

rodrigosiqueira:
@joaoseckler could you create a ticket for adding signal handling for ssh and close this ticket after that? Do not forget to assign the new ticket to this milestone.

Issue #271, opened 2021-05-20
Implement per command signal handlers


joaoseckler:
Depends on: #270 

Some commands need specific signal handlers to be able to elegantly be interrupted or halted (e.g. a command that uses temporary files needs to remove those files when receiving a `SIGINT`). Commands that will need specific handlers (but others may need it too):

`mount`, `umount`, `up`, `deploy`, `build`

rodrigosiqueira:
Two important scenarios:

1. When ssh to a machine, and if it suddenly dies, the ssh hangs the terminal, and we have two options: wait until the pc boot up again or close the terminal. Let's handle signals in the `kw s` feature.
2. If we cancel kw deploy, we will leave some old files in the cache directory. Let's clean it.

rodrigosiqueira:
Done! Thanks

PR #282, opened 2021-05-27
src: signal_manager: Implement signal manager

joaoseckler:
Currently, kw does not handle signals. If you pass it a SIGINT, for
example, the subprocess that happens to be executing at that particular
moment is the one who will receive and treat it. This results in a poor
user interface, where the standard Ctrl-C doesn’t have a meaningful
outcome. Worse than that, a simple SIGINT could leave the environment
polluted by unremoved temporary files. This patch:

 - Implements functions for adding and resetting handlers
 - Implements default handler for SIGINT and SIGTERM
 - Describes a use policy of signal handling in
   documentation/content/howtocontribute.rst

closes: #270

### DIFF
--- a/documentation/content/codingstyle.rst
+++ b/documentation/content/codingstyle.rst
@@ -263,6 +263,17 @@ ideally, we want it in the same file. For example, you should find details on
 using the `build` option in the mk.sh, and for `configm` in the file
 config_manager.sh.
 
+Handling Signals
+----------------
+It is natural for commands to set global variables or to create temporary files
+during their execution. However, all commands should expect to receive signals
+and be able to properly handle them. If you implement a new feature, take some
+time to check if it pollutes the environment. If it does, make sure to handle
+it's de-pollution upon receiving a SIGINT or a SIGTERM: an interrupted command
+should always leave the environment in the same state as it was prior to its
+invocation. Convenience functions for this purpose (setting and resetting
+handlers for arbitrary signals) are implemented in `src/signal_manager`.
+
 Conclusion
 ----------
 
--- a/kw
+++ b/kw
@@ -33,10 +33,14 @@ export KWORKFLOW
 # The include function (sourced from this file) should always be used for file sourcing.
 . "$KW_LIB_DIR/kw_include.sh" --source-only
 
+include "$KW_LIB_DIR"/signal_manager.sh
+
 function kw()
 {
   action="$1"
   shift
+
+  signal_manager || warning 'Was not able to set signal handler'
 
   mkdir -p "$KW_CACHE_DIR"
 
--- a/src/mk.sh
+++ b/src/mk.sh
@@ -19,6 +19,7 @@
 
 include "$KW_LIB_DIR/vm.sh" # It includes kw_config_loader.sh and kwlib.sh
 include "$KW_LIB_DIR/remote.sh"
+include "$KW_LIB_DIR/signal_manager.sh"
 
 # Hash containing user options
 declare -gA options_values
@@ -85,8 +86,19 @@ function get_kernel_version()
 # specialized cleanup functions.
 function cleanup()
 {
-  say "Cleanup deploy files"
+  say 'Cleanup deploy files'
   cleanup_after_deploy
+}
+
+function interrupt_cleanup()
+{
+  say 'Cleaning up...'
+  if [[ -v options_values['REMOTE'] ]]; then
+    cleanup_after_deploy 'SILENT'
+  fi
+
+  say 'Exiting...'
+  exit 0
 }
 
 # When kw deploy a new kernel it creates temporary files to be used for moving
@@ -487,6 +499,8 @@ function kernel_deploy()
     complain "Execute this command in a kernel tree."
     exit 125 # ECANCELED
   fi
+
+  signal_manager 'interrupt_cleanup' || warning 'Was not able to set signal handler'
 
   if [[ "$target" == "$VM_TARGET" ]]; then
     vm_mount
--- a/src/signal_manager.sh
+++ b/src/signal_manager.sh
@@ -1,0 +1,71 @@
+include "$KW_LIB_DIR/kwio.sh"
+
+function get_valid_signals()
+{
+  local -r signal_list="$(trap -l | grep -oE 'SIG\S+')"
+  declare -gA SIGNAL_MANAGER_VALID_SIGNALS
+
+  for s in $signal_list; do
+    SIGNAL_MANAGER_VALID_SIGNALS["$s"]='1'
+  done
+}
+
+get_valid_signals
+
+# This functions prints the default message when interrupting kw
+function default_interrupt_handler()
+{
+  say $'\n'Oh no! An interruption! See ya...
+}
+
+# This function adds a new signal handler to an arbitrary signal
+#
+# @command The handler to be called when a signal is catched. This
+#          must be a single command
+# @signals The list of signals to attach to this handler
+#
+# If command is empty, use the default handler
+# If the signal is empty, use the default signals
+
+# shellcheck disable=2120
+function signal_manager()
+{
+  local command="$1"
+  shift
+  local -a signals=("$@")
+
+  if [[ -z "$command" ]]; then
+    command='default_interrupt_handler'
+  elif ! type "$command" > /dev/null 2>&1; then
+    return 22 # EINVAL
+  fi
+
+  if [[ "${#signals}" == 0 ]]; then
+    signals=(SIGINT SIGTERM)
+  else
+    for s in "${signals[@]}"; do
+      if [[ ! -v SIGNAL_MANAGER_VALID_SIGNALS["$s"] &&
+        ! -v SIGNAL_MANAGER_VALID_SIGNALS[SIG"$s"] ]]; then
+        return 22 # EINVAL
+      fi
+    done
+  fi
+
+  # shellcheck disable=2064
+  trap "$command" "${signals[@]}" 2> /dev/null
+}
+
+# This function resets all signal handlers to their system defaults and
+# sets kw's default signal handler for SIGINT and SIGTERM
+function signal_manager_reset()
+{
+  local traps
+
+  traps=$(trap -p | grep -o '[^ ]*$')
+
+  while read -r sig; do
+    trap - "$sig"
+  done <<< "$traps"
+
+  signal_manager
+}
--- a/tests/signal_manager_test.sh
+++ b/tests/signal_manager_test.sh
@@ -1,0 +1,127 @@
+#!/bin/bash
+
+include './src/signal_manager.sh'
+include './tests/utils.sh'
+
+function oneTimeSetUp()
+{
+  declare -gr test_phrase="This is a phrase."
+  declare -gr file_name="out.tmp"
+  declare -gr original_dir="$PWD"
+}
+
+function setUp()
+{
+  cd "$SHUNIT_TMPDIR" || fail "($LINENO): Was not able to cd into temporary directory"
+}
+
+function write_to_file()
+{
+  echo "$test_phrase" > "$file_name"
+}
+
+function test_default_handler()
+{
+  local -r expected=$'\n'"Oh no! An interruption! See ya..."
+
+  trap 'kill -s SIGTERM $!' SIGUSR1
+  {
+    signal_manager default_interrupt_handler SIGINT SIGTERM
+    sleep infinity &
+    echo $! > sleep_pid
+    kill -s SIGUSR1 $$ # Send SIGUSR1 to parent process
+    wait $!            # wait for SIGTERM
+  } > out.tmp &
+
+  wait $! # wait for SIGUSR1
+  kill "$(cat sleep_pid)"
+  assertEquals "($LINENO)" "$expected" "$(cat "$file_name")"
+
+  rm -f out.tmp
+
+  trap 'kill -s SIGINT $!' SIGUSR1
+  {
+    signal_manager default_interrupt_handler SIGINT SIGTERM
+    sleep infinity &
+    echo $! > sleep_pid
+    kill -s SIGUSR1 $$ # Send SIGUSR1 to parent process
+    wait $!            # wait for SIGINT
+  } > out.tmp &
+
+  wait $!
+  kill "$(cat sleep_pid)"
+  assertEquals "($LINENO)" "$expected" "$(cat "$file_name")"
+}
+
+function test_non_signal()
+{
+  local ret
+
+  signal_manager echo SIGNONEXISTENT
+  ret="$?"
+  assertFalse "($LINENO) Should have received an error." "$ret"
+
+  signal_manager echo 'HUP SIG'
+  ret="$?"
+  assertFalse "($LINENO) Should have received an error." "$ret"
+}
+
+function test_new_signal()
+{
+  # Enable job control. This allows us to send a SIGINT signal to this
+  # very process without it being interrupted. The special parameter $$,
+  # used below, is used to get the current process's PID
+  set -m
+
+  signal_manager write_to_file SIGINT
+  kill -s SIGINT $$
+  assertEquals "($LINENO)" "$test_phrase" "$(cat "$file_name")"
+
+  signal_manager write_to_file SIGTERM
+  kill -s SIGTERM $$
+  assertEquals "($LINENO)" "$test_phrase" "$(cat "$file_name")"
+
+  set +m
+}
+
+function test_signal_reset()
+{
+  local ret
+  local output
+
+  signal_manager write_to_file SIGINT SIGTERM
+  signal_manager_reset
+  output="$(trap -p)"
+
+  echo "$output" | grep -q "trap -- 'default_interrupt_handler' SIGINT"
+  ret="$?"
+  assertTrue "($LINENO) Default handler not set." "$ret"
+
+  echo "$output" | grep -q "trap -- 'default_interrupt_handler' SIGTERM"
+  ret="$?"
+  assertTrue "($LINENO) Default handler not set." "$ret"
+}
+
+function test_add_two_signals()
+{
+  signal_manager write_to_file SIGINT SIGTERM
+  local -r output="$(trap)"
+  local ret
+
+  echo "$output" | grep -q "trap -- 'write_to_file' SIGINT"
+  ret="$?"
+  assertTrue "($LINENO) Trap not set" "$ret"
+
+  echo "$output" | grep -q "trap -- 'write_to_file' SIGTERM"
+  ret="$?"
+  assertTrue "($LINENO) Trap not set" "$ret"
+}
+
+function test_compound_signal()
+{
+  signal_manager 'echo something; ls'
+  ret="$?"
+  assertFalse "($LINENO) Compound command accepted" "$ret"
+}
+
+invoke_shunit


src/signal_manager.sh
```diff
@@ -0,0 +1,30 @@
...
+function signal_manager() {
+  local command="$1"
+  shift
+  local -a signals="$@"
+
+  if [ -z "$command" -o -z "${signals[@]}" ]; then
+    return 22 #EINVAL
+  fi
+
+  trap "$command" "${signals[@]}" 2>/dev/null
```
rodrigosiqueira, line 22: Let me check If I understood your strategy here. Basically, you want that specific files to add this `signal_manager()` that points to the correct function for handling it, something like this:

```diff
diff --git a/src/mk.sh b/src/mk.sh
index 43dc291..62bbbcf 100755
--- a/src/mk.sh
+++ b/src/mk.sh
@@ -19,6 +19,7 @@
 
 . "$KW_LIB_DIR/vm.sh" --source-only # It includes kw_config_loader.sh and kwlib.sh
 . "$KW_LIB_DIR/remote.sh" --source-only
+. "$KW_LIB_DIR/signal_manager.sh"
 
 # Hash containing user options
 declare -A options_values
@@ -576,6 +577,8 @@ function mk_build()
     exit 0
   fi
 
+  signal_manager mk_signal_handler SIGINT SIGTERM
+
   menu_config=${menu_config:-"nconfig"}
   arch=${arch:-"x86_64"}
 
@@ -810,3 +813,9 @@ function deploy_parser_options()
       ;;
   esac
 }
+
+function mk_signal_handler()
+{
+  echo "+++++++++++++++++++++++++++++++"
+  warning "====> Yoyoyo... yoyoyo... it's me! Do something really nice here!"
+}
```

Am I correct?
If so, I think the way you call this trap will not work; you need to eval an expression. In other words, you need something like this:

```bash
eval "trap "$command" "${signals[@]}" 2>/dev/null"
```

Maybe you can use cmd_manager to invoke this trap (it might work as well). Actually, if you use `cmd_manager` you can use the TEST mode for testing this function.

src/signal_manager.sh
```diff
@@ -0,0 +1,30 @@
...
+# This function adds a new signal handler to an arbitrary signal
+#
+# @command The handler to be called when a signal is catched
+# @signals The signals to attach to this handler
+function signal_manager() {
+  local command="$1"
+  shift
+  local -a signals="$@"
+
+  if [ -z "$command" -o -z "${signals[@]}" ]; then
```
rodrigosiqueira, line 18: Hmmm... I think this `-z "${signals[@]}"`reduces the value of having a `signal_manager` function. How about this:
1. Check if `"${signals[@]}"` is empty.
1.2. Empty: Define the default signals.
1.3. Not empty: Use the parameter value

Also, it might be a good idea to validate if we have a valid signal.

rodrigosiqueira (changes requested):
Hi @joaoseckler ,
Thanks a lot for this PR; it is a super nice feature! I'm really glad that you're working on it.

Most of my review can be seen in the inline comments. Here, I just have some suggestions:
1. Add unit tests: I see two types of tests
1.1. You can use `cmd_manager` and check the command sequence.
1.2. You can enable the signal handling during the test and send signals to it (Idk how to do it right now, you'll need to search for more details)
2. Add signal handler to some specific: Create two or three commits for handling specific cases. My suggestions are: kw build, kw deploy, and kw sh.

Thanks.

src/signal_manager.sh
```diff
@@ -0,0 +1,30 @@
+. "$KW_LIB_DIR/kwio.sh" --source-only
```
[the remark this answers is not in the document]
alan-barzilay: @rodrigosiqueira yes, I can confirm this will be fixed by #240. Actually this error is what originally motivated me to implement the include guards haha

src/signal_manager.sh
joaoseckler: Why is it that we need to eval? It seems to be working in my machine this way.

src/signal_manager.sh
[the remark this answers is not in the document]
joaoseckler: Updated in new version.

src/signal_manager.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# This functions prints the default message when interrupting kw
+function default_interrupt_handler() {
+  say $'\n'$SEPARATOR
+  say Exiting...
+}
+
+# This function adds a new signal handler to an arbitrary signal
+#
+# @command The handler to be called when a signal is catched
```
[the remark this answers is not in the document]
joaoseckler: Not necessarily... it should be a string to be passed to eval. The following is allowed:

`signal_manager 'echo goodbye!' SIGINT SIGTERM`

src/signal_manager.sh
joaoseckler: Done in the last update.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
...
+  kill -s SIGTERM $!
+  assertEquals "$expected" "$(cat out.tmp)"
+
+  {
+    signal_manager default_interrupt_handler SIGTERM SIGINT
+    sleep infinity &
+    wait $!
+  } > out.tmp &
+
+  $(echo :)
```
joaoseckler, line 37: Could somebody help me understand this behaviour? It seems taht for some reason the output of the grouped command (lines 31-35) only gets redirected to `out.tmp` if a subshell is created between that and the `kill` command.

joaoseckler:
@rodrigosiqueira About this:

> 2. Add signal handler to some specific: Create two or three commits for handling specific cases. My suggestions are: kw build, kw deploy, and kw sh.

I have opened issue #271 for this purpose. Should we merge it with this one?

src/signal_manager.sh
```diff
@@ -0,0 +1,30 @@
+. "$KW_LIB_DIR/kwio.sh" --source-only
+
+# This functions prints the default message when interrupting kw
+function default_interrupt_handler() {
+  say $'\n'$SEPARATOR
+  say Exiting...
```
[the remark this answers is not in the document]
joaoseckler: Haha! Nice. Done in new version

src/signal_manager.sh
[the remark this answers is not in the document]
rodrigosiqueira: I see, but in this case, your documentation is a little bit wrong because users must use quotes if they want to pass a function with parameters. If you do not use it, you will lose the values.

documentation/content/howtocontribute.rst
```diff
@@ -196,4 +196,15 @@ We have a simple rule for making a release: It'll be Done when it's done. If
 you want to follow our release plans, just take a look at the milestones_ since
 we use it for self-organization.
 
+Handling Signals
```
rodrigosiqueira, line 199: Now I'm not sure if we should keep this documentation in the `howtocontribute.rst` file or in the `codestyle.rst` file.

src/signal_manager.sh
```diff
@@ -0,0 +1,48 @@
...
+{
+  local command="$1"
+  shift
+  local -a signals=( "$@" )
+
+  if [[ -z "$command" ]]; then
+    command='default_interrupt_handler'
+  fi
+
+  if [[ "${#signals}" = 0 ]]; then
```
rodrigosiqueira, line 29: Use `==` for consistency.

src/signal_manager.sh
```diff
@@ -0,0 +1,48 @@
...
+
+function signal_manager()
+{
+  local command="$1"
+  shift
+  local -a signals=( "$@" )
+
+  if [[ -z "$command" ]]; then
+    command='default_interrupt_handler'
+  fi
```
rodrigosiqueira, line 27: You can simplify this to `[[ -z "$command" ]] && command='default_interrupt_handler'`

src/signal_manager.sh
```diff
@@ -0,0 +1,48 @@
...
+
+  if [[ "${#signals}" = 0 ]]; then
+    signals=( SIGINT SIGTERM )
+  else
+    for s in "${signals[@]}"; do
```
rodrigosiqueira, line 32: Is this loop working as you expect? I mean, from `"${signals[@]}"` and the lack of `IF=...` I suspect that this loop only runs a single time no matter how many parameters you have. You can check it by adding a simple `echo "---> $s" inside this loop and ensure that you pass multiple signals.

src/signal_manager.sh
rodrigosiqueira: When something behaves weirdly in bash, you can check the return value by looking at `$?`; I basically added `echo "$?"` right after your trap, and I catch `1`, which means that something was wrong. Next, I removed `2>/dev/null` and I could see this error:
```
line 24: trap: SIGINT SIGTERM: invalid signal specification  
```
After a few minutes, I realize that we were using the trap in the wrong way since the syntax says:
``
trap [COMMANDS] [SIGNALS ...]
``
Since you double-quote `"${signals[@]}"`, it became a single parameter; i.e., trap did not find the signal `SIGINT SIGTERM`. If you just remove the double-quote, it will work because each array's entry will be interpreted as a different parameter (`SIGINT` `SIGTERM`). Finally, if you use eval, it will expand everything for you and the rest you probably understand at this point :)

I suggested you to use cmd_manager because you can easy test this part.

src/signal_manager.sh
```diff
@@ -0,0 +1,48 @@
...
+  fi
+
+  trap "$command" "${signals[@]}" 2>/dev/null
+}
+
+# This function resets all signal handlers to their system defaults and
+# sets kw's default signal handler for SIGINT and SIGTERM
+function signal_manager_reset()
+{
+  trap | grep -o '[^ ]*$' | while read sig; do trap - $sig; done
```
rodrigosiqueira, line 46: That's a very clever line, but it is a little bit hard to understand... maybe you can expand it in small steps to make it easy to read.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
+
+. ./src/signal_manager.sh --source-only
+. ./tests/utils --source-only
+
+function suite() {
```
rodrigosiqueira, line 6: put `{` in a new line. Same for all this file.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
...
+  suite_addTest "default_handler_Test"
+  suite_addTest "non_signal_Test"
+  suite_addTest "new_signal_Test"
+  suite_addTest "signal_reset_Test"
+  suite_addTest "add_two_signals_Test"
```
rodrigosiqueira, line 11: Prefer a single quote if you don't have a variable expansion.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function default_handler_Test() {
+  local -r original_dir="$PWD"
+  local -r expected=$'\n'"Oh no! An interruption! See ya..."
+  local output
+
+  cd "$SHUNIT_TMPDIR"
```
rodrigosiqueira, line 19: What is this `$SHUNIT_TMPDIR`? That's something new to me; it looks interesting.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
...
+
+  {
+    signal_manager default_interrupt_handler SIGTERM SIGINT
+    sleep infinity &
```
rodrigosiqueira, line 23: Uouuu... that's something new, weirdly it does not have any entry to `infinity` in the man doc. Maybe this is a good contribution to the upstream :)

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
...
+    wait $!
```
rodrigosiqueira, line 24: For readability sake, add a comment like `# Get the process ID`.

rodrigosiqueira: btw, are you sure that we reach this line?

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
...
+  } > out.tmp &
+
+  $(echo :)
```
rodrigosiqueira, line 27: Why do you need to execute this `echo`? Why inside `$(..)`?

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
...
+  kill -s SIGTERM $!
```
rodrigosiqueira, line 28: I'm not sure what you are trying to achieve here... are you trying to kill what you created inside `{}`? Are you sure that the PID is the same?

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
...
+  kill -s SIGINT $!
+  assertEquals "$expected" "$(cat out.tmp)"
+
+  cd "$original_dir"
+}
+
+
+function non_signal_Test() {
+  signal_manager 'echo hello!' SIGNONEXISTENT
+  local ret=$?
```
rodrigosiqueira, line 47: Declare `local ret` at the beginning of this test (C style). Also, double quote "$?" when you capture it later.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
...
+  assertFalse "Should have received an error." "$ret"
+}
+
+function new_signal_Test() {
+  local test_phrase="This is a phrase."
+  local original_dir="$PWD"
+  cd "$SHUNIT_TMPDIR"
+
+  set -m
```
rodrigosiqueira, line 56: `set` is a super powerful command, and it is hard to know its options. Add a simple comment explaining what `-m` mean and explain why you need it if it is not clear.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,97 @@
...
+
+  signal_manager "echo $test_phrase > out.tmp" SIGINT
+  kill -s SIGINT $$
```
rodrigosiqueira, line 59: Add a comment here about `$$`; iirc, you're getting the script PID here.

rodrigosiqueira (changes requested):
@joaoseckler  This patch was super hard to review hahaha
Keep in mind that signals can be really confusing in any language. I strongly advise you to rewrite parts of your code with readability in mind because your code is a little bit complex, and we don't want extra complications for dealing with signals.

Also, add at least one commit on top of this one that uses your signal manager system. This is important for showing examples of how to use it.

P.s.: Try to make your tests easy to understand, since it can work as documentation.

src/signal_manager.sh
alan-barzilay: I think the original is more readable for someone not that familiar with bash

src/signal_manager.sh
joaoseckler: About the double quotes around the array see [this](https://www.gnu.org/software/bash/manual/bash.html#Arrays):

> Any element of an array may be referenced using ${name[subscript]}. The braces are required to avoid conflicts with the shell’s filename expansion operators. If the subscript is ‘@’ or ‘\*’, the word expands to all members of the array name. These subscripts differ only when the word appears within double quotes. If the word is double-quoted, ${name[*]} expands to a single word with the value of each array member separated by the first character of the IFS variable, and ${name[@]} expands each element of name to a separate word.

tl;dr: `"${signals[@]}"` expands to as many words as there are in the array `signals`, and `"${signals[*]}` expands to only one word. The output of these two snippets make it very clear:

```
$ array=(a b c); for i in "${array[@]}"; do echo element: $i; done
element: a
element: b
element: c
```
```
$ array=(a b c); for i in "${array[*]}"; do echo element: $i; done
element: a b c
```
Anyway, you are getting an error that I am not. I included a test for exactly this purpose (`add_two_signals_Test()`). What is the output of `./run_tests.sh test signal_manager` for you?

src/signal_manager.sh
joaoseckler: my [other](https://github.com/kworkflow/kworkflow/pull/282#discussion_r643485049) comment addresses this issue!

tests/signal_manager_test.sh
joaoseckler: I really don't know... See my [other](https://github.com/kworkflow/kworkflow/pull/282#discussion_r642117501) comment.

tests/signal_manager_test.sh
joaoseckler: My current working hypothesis is that there is a synchronization issue between the setting of the trap (l. 34) and the sending of the signal (l. 40). Calling a subshell between those only seemed to work (actually, it didn't work every time) because it favors the setting of the trap in the race for execution.

tests/signal_manager_test.sh
joaoseckler: Yes, one could test it with:
```
...
} > out.tmp &
local pid=$!

$(echo :)
echo "$pid should be equal to $!..."
```

That doesn't seem to be the problem.

src/signal_manager.sh
rodrigosiqueira: Excellent analysis, make sense to me. Btw, I noticed that you added `local -a signals=( "$@" )`, could you elaborate a little bit?

tests/signal_manager_test.sh
rodrigosiqueira: Tbh, I did not get why we need this `$(echo :)`, is it some bash-magic?

Anyway, when I was working with `kw pomodoro` I noticed that the subprocess takes a while to start (make sense, since starting a new process is a heavyweight task). In this sense, this part:

```
   {
      signal_manager default_interrupt_handler SIGTERM SIGINT
      sleep infinity &
      wait $!
    } > out.tmp &
```

might take a while to start. You can check this hypothesis by adding a `sleep 1s` right after the above code.

tests/signal_manager_test.sh
rodrigosiqueira: to me, your PR looks incomplete without an example of its usage. Just add a new commit on top of this one that uses your signal manager; you don't need to make something fancy just show a real example.

tests/signal_manager_test.sh
joaoseckler: It's useful! See [shunit's readme](https://github.com/kward/shunit2/#-advanced-usage): 

> SHUNIT_TMPDIR | Path to temporary directory that will be automatically cleaned up upon exit of shUnit2.

tests/signal_manager_test.sh
joaoseckler: Apparently this is not itself a feature of sleep. The thing is sleep accepts a floating point number, and `infinity` of `inf` is accepted as so. It actually truncates to a very high number. It seems like implementing an actual infinit is already [being considered](https://lists.gnu.org/archive/html/bug-gnulib/2020-02/txtd_O21xhDVT.txt)!

documentation/content/howtocontribute.rst
joaoseckler: I think it should be in codestyle. I think `howtocontribute.rst` should describe aspects of the development workflow and codestyle should describe how the code is supposed to look like and behave. With this reasoning in mind, I changed this to `codestyle.rst` in the last update

src/signal_manager.sh
joaoseckler: I think that is where the error was coming from! Suppose we receive the following arguments:
`'echo goodbye!' SIGINT SIGTERM`

After the `shift` command, `"$@"` will look like:
`SIGINT SIGTERM`

Thus, to create an array with all the signals, we use
`local -a signals=( "$@" )`

Which will expand to:

`local -a signals=( "SIGINT" "SIGTERM" )`

Which is exactly what we want. If we ommited the parenthesis, we would have

`local -a signals=$@`

which would expand to

`local -a signals=SIGINT SIGTERM`

Which would discard the second signal.

tests/signal_manager_test.sh
joaoseckler: Indeed, using `sleep 1` instead of `$(echo :)` works. In the last update of this PR I came up with a system to make sure the subprocess was already created before sending it the signal, using more signals. I think I was able to remove all the race conditions without having to use `sleep`. 

By the way, `$(echo :)` makes no sense at all. I thought of using the builtin `:` but instead passed it as an argument to echo.

joaoseckler:
The last update to this PR includes:

 - Removing the race condition in the test `default_handler_Test`
 - Adjusting formatting (use braces after a new line, using single quotes when no expansion is needed)
 - Adding a new commit to exemplify the use of the signal manager
 - Moving the signal manger's use policy from `howtocontribute.rst` to `codingstyle.rst`

src/mk.sh
```diff
@@ -482,6 +483,8 @@ function kernel_deploy
     exit 125 # ECANCELED
   fi
 
+  signal_manager 'cleanup; exit 0'
```
rodrigosiqueira, line 486: Hi @joaoseckler ,
In our previous conversation, I pointed out that this could be a problem. Maybe you can create a simple signal manager in the mk file that will detect which option the user used (you just need to take a look at `options_values`) and run a specific action.

Btw, remove the `Draft` from your PR. Your PR is excellent.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,117 @@
...
+
+  signal_manager 'kill -s SIGINT $!' SIGUSR1
+  {
+    signal_manager default_interrupt_handler SIGINT SIGTERM
+    sleep infinity &
+    kill -s SIGUSR1 $$ # Send SIGUSR1 to parent process
+    wait $! # wait for SIGINT
+  } > out.tmp &
+
+  wait $!
```
rodrigosiqueira, line 43: Consider adding a helper to make this part easy to understand.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,117 @@
...
+  echo $output | grep -q "trap -- 'default_interrupt_handler' SIGINT"
+  ret="$?"
+  assertTrue "($LINENO) Default handler not set." "$ret"
+
+  echo $output | grep -q "trap -- 'default_interrupt_handler' SIGTERM"
+  ret="$?"
+  assertTrue "($LINENO) Default handler not set." "$ret"
+}
+
+
```
rodrigosiqueira, line 101: Drop this extra line

src/signal_manager.sh
```diff
@@ -0,0 +1,53 @@
+. "$KW_LIB_DIR/kwio.sh" --source-only
+
+declare -ga SIGNAL_MANAGER_VALID_SIGNALS=( $(trap -l | sed -E 's/[0-9]+\)/ /g') )
```
rodrigosiqueira, line 3: Let's make this easy to read.

src/signal_manager.sh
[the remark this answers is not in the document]
rodrigosiqueira: @joaoseckler, I finally applied the include feature. Could you update your PR based on that? Also, could you check if this error gets fixed?

rodrigosiqueira (changes requested):
Hi @joaoseckler ,
Could you rebase and adapt your change with the latest code in the unstable branch? I merged the include feature. Also, see my inline comments.

documentation/content/codingstyle.rst
```diff
@@ -209,6 +209,16 @@ ideally, we want it in the same file. For example, you should find details on
...
 
+Handling Signals
+----------------
+When we execute some specific command, it is natural that we set global
+variables or create temporary files. However, all commands should expect
+receiving signals and should deal with each of them properly. If you implement a
+new feature, take some time to check if it pollutes the environment. If it does,
+handle it's depollution upon receiving SIGINT or SIGTERM. Convenience functions
+for this purpose (setting and resetting handlers for arbitrary signals) are
+implemented in `src/signal_manager`.
```
alan-barzilay, line 220: how about:
" It is natural for commands to set global variables or to create temporary files during their execution. However, all commands should expect to receive signals and be able to properly handle them. If you implement a
new feature, take some time to check if it pollutes the environment. If it does,
make sure to handle it's de-pollution upon receiving a SIGINT or a SIGTERM, an interrupted command should always leave the environment in the same state as it was prior to its invocation. (...)"

alan-barzilay (changes requested):

joaoseckler:
Changelog for the last update of this PR: 
- update documentation for handling signals in `codingstyle.rst` following a suggestion from @alan-barzilay 
- update test function names with the new convention
- use function to make declaration of `SIGNAL_MANAGER_VALID_SIGNALS` more readable
 - in `mk.sh`, check if current command is deploy to call `cleanup_deploy` upon receiving a signal

src/signal_manager.sh
```diff
@@ -0,0 +1,64 @@
...
+  say $'\n'Oh no! An interruption! See ya...
+}
+
+# This function adds a new signal handler to an arbitrary signal
+#
+# @command The handler to be called when a signal is catched. This can
+#          can be a compound command (e.g. a function with parameters)
+#          but it must be passed as a single word (i.e. between quotes)
+# @signals The list of signals to attach to this handler
+#
```
rodrigosiqueira, line 27: Add:
`Return:`

src/signal_manager.sh
```diff
@@ -0,0 +1,64 @@
...
+# If command is empty, use the default handler
+# If the signal is empty, use the default signals
+
```
rodrigosiqueira, line 30: Drop this extra line

src/signal_manager.sh
rodrigosiqueira: Nice! I really want to see a blog post on the Flusp website about this :)

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,111 @@
...
+  local -r expected=$'\n'"Oh no! An interruption! See ya..."
+
+  cd "$SHUNIT_TMPDIR"
+
+  signal_manager 'kill -s SIGTERM $!' SIGUSR1
+  {
+    signal_manager default_interrupt_handler SIGINT SIGTERM
+    sleep infinity &
+    kill -s SIGUSR1 $$ # Send SIGUSR1 to parent process
+    wait $! # wait for SIGTERM
```
rodrigosiqueira, line 18: just for curiosity, should you double-quote these variables? What shellcheck says?

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,111 @@
...
+    signal_manager default_interrupt_handler SIGINT SIGTERM
+    sleep infinity &
+    kill -s SIGUSR1 $$ # Send SIGUSR1 to parent process
+    wait $! # wait for SIGINT
+  } > out.tmp &
+
+  wait $!
+  assertEquals "($LINENO)" "$expected" "$(cat out.tmp)"
+
+  killall sleep
```
rodrigosiqueira, line 37: What happens if I have a sleep command running while I execute this test? I'm wondering if `killall` will not kill other `sleep` in the system. Maybe we should use the PID instead of killing all `sleep`.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,111 @@
...
+  cd "$original_dir"
+}
+
```
rodrigosiqueira, line 40: drop this extra line

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,111 @@
...
+function test_signal_reset()
+{
+  local ret
+  local output
+
+  signal_manager 'echo Some command' SIGINT SIGTERM
+  signal_manager_reset
+  output="$(trap -p)"
+
+  echo $output | grep -q "trap -- 'default_interrupt_handler' SIGINT"
```
rodrigosiqueira, line 87: s/$output/"$output"/

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,111 @@
...
+  ret="$?"
+  assertTrue "($LINENO) Default handler not set." "$ret"
+
+  echo $output | grep -q "trap -- 'default_interrupt_handler' SIGTERM"
```
rodrigosiqueira, line 91: "$output"

rodrigosiqueira (changes requested):
@joaoseckler, your PR is really good! I just had some tiny comments.

Also, rebase your change with the latest code from `unstable`.

src/signal_manager.sh
```diff
@@ -0,0 +1,66 @@
...
+function signal_manager()
```
rodrigosiqueira, line 31: I think the below shellcheck is a false positive. Just add `# shellcheck disable=SC2120` before this function declaration.

src/signal_manager.sh
```diff
@@ -0,0 +1,66 @@
...
+  else
+    for s in "${signals[@]}"; do
+      if [[ ! -v SIGNAL_MANAGER_VALID_SIGNALS["$s"] &&
+        ! -v SIGNAL_MANAGER_VALID_SIGNALS[SIG"$s"] ]]; then
+        return 22 # EINVAL
+      fi
+    done
+  fi
+
+  trap "$command" "${signals[@]}" 2> /dev/null
```
rodrigosiqueira, line 50: I think shellcheck has a good point here... Maybe you should add a check if the "$command" is something that exists, and after that, you can add `# shellcheck disable=SC2064` since we know that command is valid and it is a string that points to a function.
P.s.: Explain why you added this in the commit message.

src/signal_manager.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# This function resets all signal handlers to their system defaults and
+# sets kw's default signal handler for SIGINT and SIGTERM
+function signal_manager_reset()
+{
+  local traps
+
+  traps=$(trap -p | grep -o '[^ ]*$')
+
+  while read sig; do
```
rodrigosiqueira, line 61: Why not add `-r` here? It looks fine to me, and I don't think it changes the behavior.

tests/signal_manager_test.sh
```diff
@@ -0,0 +1,110 @@
...
+
+include './src/signal_manager.sh'
+include './tests/utils'
+
+function test_default_handler()
+{
+  local -r original_dir="$PWD"
+  local -r expected=$'\n'"Oh no! An interruption! See ya..."
+
+  cd "$SHUNIT_TMPDIR" || exit
```
rodrigosiqueira, line 11: Hmmm... I'm not too fond of this... if we exit in case of an invalid directory, we will not run the other tests. I prefer something like:
```
cd "$SHUNIT_TMPDIR" || fail "($LINENO): Any good message"
```

src/mk.sh
```diff
@@ -85,10 +86,21 @@ function get_kernel_version()
...
 {
-  say "Cleanup deploy files"
+  say 'Cleanup deploy files'
   cleanup_after_deploy
 }
 
+function interrupt_cleanup()
+{
+  say 'Cleaning up...'
+  if [[ -v options_values["REMOTE"] ]]; then
```
rodrigosiqueira, line 96: s/"REMOTE"/'REMOTE'/

rodrigosiqueira (changes requested):
Let's document in every commit, why we are ignoring something from shellcheck.
For example:
* shellcheck disable=SC2064: Explanation...
* shellcheck disable=...

src/signal_manager.sh
```diff
@@ -0,0 +1,71 @@
...
+{
+  local command="$1"
+  shift
+  local -a signals=("$@")
+
+  if [[ -z "$command" ]]; then
+    command='default_interrupt_handler'
+  elif ! type "$command" > /dev/null 2>&1; then
+    return 22 # EINVAL
+  fi
```
rodrigosiqueira, line 41: I think you missed a unit test to cover this.

kw
```diff
@@ -33,11 +33,15 @@ export KWORKFLOW
...
 . "$KW_LIB_DIR/kw_include.sh" --source-only
 
+include "$KW_LIB_DIR"/signal_manager.sh
+
 function kw()
 {
   action="$1"
   shift
 
+  signal_manager
```
rodrigosiqueira, line 43: Since `signal_manager` can return 22, you need to handle it here and in the `mk.sh` file. Just add a warning, in this case, we don't anything else.

joaoseckler:
The last update to this PR includes:

 - Keep sleep PID instead of using `killall sleep`
 - add unit test for compound commands
 - add warning if `signal_manager` returns an error

rodrigosiqueira:
Merged!
Thanks a lot for all your effort here.

Issue #267, opened 2021-05-19
Fix kw statistics output spacing


magalilemes:
If we run `kw statistics --year`, we might encounter an output like this:

```
Total          Max  Min       Average
Uninstall      3    00:00:03  00:00:02  00:00:02
Build_failure  3    00:00:01  00:00:00  00:00:00
Build          82   00:06:25  00:00:00  00:00:09
Deploy         6    00:00:21  00:00:03  00:00:11
List           3    00:00:08  00:00:00  00:00:02
```

When it should look like this

```
               Total  Max       Min       Average
Uninstall      3      00:00:03  00:00:02  00:00:02
Build_failure  3      00:00:01  00:00:00  00:00:00
Build          82     00:06:25  00:00:00  00:00:09
Deploy         6      00:00:21  00:00:03  00:00:11
List           3      00:00:08  00:00:00  00:00:02
```

What's probably causing this is the way we're printing the columns in the `print_basic_data` function in the src/statistics.sh file using the `column` command.

Issue #280, opened 2021-05-26
Statistics tests broken


joaoseckler:
I found that after the patch `https://github.com/kworkflow/kworkflow/commit/7b7b8e44d5f65e7f29188958423b044edfafebf7` the statistics test didn't work anymore in my machine. I'm getting this error:

```
calculate_average_Test
calculate_total_of_data_Test
max_value_Test
min_value_Test
sec_to_formatted_date_Test
basic_data_process_Test
day_statistics_Test
week_statistics_Test
month_statistics_Test
year_statistics_Test
ASSERT:(212) 0 - Expected cmd "Build   19   00:03:40  00:00:06  00:00:23" to be "Build   19     00:03:40  00:00:06  00:00:23"
ASSERT:(212) 1 - Expected cmd "Deploy  8    00:01:13  00:00:31  00:01:04" to be "Deploy  8      00:01:13  00:00:31  00:01:04"
ASSERT:(223) 0 - Expected cmd "Uninstall      3    00:00:03  00:00:02  00:00:02" to be "Build_failure  1      00:00:01  00:00:01  00:00:01"
ASSERT:(223) 1 - Expected cmd "Build_failure  1    00:00:01  00:00:01  00:00:01" to be "Build          3      00:00:55  00:00:01  00:00:19"
ASSERT:(223) 2 - Expected cmd "Build          3    00:00:55  00:00:01  00:00:19" to be "Deploy         1      00:00:21  00:00:21  00:00:21"
ASSERT:(223) 3 - Expected cmd "Deploy         1    00:00:21  00:00:21  00:00:21" to be "Uninstall      3      00:00:03  00:00:02  00:00:02"
ASSERT:(223) 4 - Expected cmd "List           2    00:00:08  00:00:00  00:00:04" to be "List           2      00:00:08  00:00:00  00:00:04"

Ran 10 tests.

FAILED (failures=7)
```

Cleaning the output a little, for better visualization (contiguous lines should be the same):

```
Build   19   00:03:40  00:00:06  00:00:23
Build   19     00:03:40  00:00:06  00:00:23

Deploy  8    00:01:13  00:00:31  00:01:04
Deploy  8      00:01:13  00:00:31  00:01:04

Uninstall      3    00:00:03  00:00:02  00:00:02
Build_failure  1      00:00:01  00:00:01  00:00:01

Build_failure  1    00:00:01  00:00:01  00:00:01
Build          3      00:00:55  00:00:01  00:00:19

[...]

Issue #276, opened 2021-05-22
Refactor statistics


rodrigosiqueira:
Before we try to expand `kw statistics`, we need to increase the test coverage and, after that, start to refactor the current version. See:

https://codecov.io/gh/kworkflow/kworkflow/src/unstable/src/statistics.sh

After we achieve at least 80% of coverage of this file, we can refactor it.

P.s.: You might encounter some bugs in the process to increase the test coverage; just fix them as you find them.

PR #283, opened 2021-05-28
Fix 'kw statistics' output

magalilemes:
The output of kw statistics was presented differently depending on the versions of Bash and the column command.
Instead of using the column utility, which can have different formatting options depending on where it comes from (util-linux, or bsdmanutils, or bsdextrautils), I chose to use `printf`, instead.
Another change was the type of variable of `statistics_opt`: it was an associative array, and I chose to let it be a normal array, since this type keeps the ordering.

This PR closes issues #267 and #280, which are pretty much relatable, and also addresses issue #276.

### DIFF
--- a/src/statistics.sh
+++ b/src/statistics.sh
@@ -6,7 +6,7 @@
 
 # This is a data struct that describes the main type of data collected. We use
 # this in some internal loops.
-declare -A statistics_opt=( ["deploy"]=1 ["build"]=2 ["list"]=3 ["uninstall"]=4 ["build_failure"]=5  ["Modules_deploy"]=5 )
+declare -a statistics_opt=( "deploy" "build" "list" "uninstall" "build_failure" "Modules_deploy" )
 
 # ATTENTION:
 # This variable is shared between function, for this reason, it is NOT SAFE to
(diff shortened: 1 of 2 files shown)


joaoseckler (approved):
Nice fix!

src/statistics.sh
```diff
@@ -195,14 +195,14 @@ function sec_to_formatted_date()
...
-  for option in "${!statistics_opt[@]}"; do
-    [[ -z "${shared_data[$option]}" ]] && continue
+  local header_format="%20s %4s %8s %12s\n"
+  local row_format="%-14s %5d %s %s %s\n"
 
-    line="${line}${option^} ${shared_data[$option]}\n"
+  printf "$header_format" Total Max Min Average
+  for option in "${statistics_opt[@]}"; do
+    [[ -z "${shared_data[$option]}" ]] && continue
+    printf "$row_format" ${option^} ${shared_data[$option]}
```
rodrigosiqueira, line 204: Any particular reason for not add double-quotes around `${option^}` and `${shared_data[$option]}`?

rodrigosiqueira (changes requested):
Hi @magalilemes ,

I just have these comments for this PR:
1. You can break this PR into two parts:
1.2. Fix the column issue: Keep this commit focused on the column bug. Keep the code and test related to that in this commit; if you introduce new tests, make sure that you use `$LINENO` instead of `$ID`.
1.3. Fix the `$ID` vs `$LINENO` in the test: This should be a separate commit since it is not related to the fix that you're working on.
2. Update the commit message.

src/statistics.sh
magalilemes: I could add it to `${option^}`, it slipped my mind. Thanks for pointing that out.
As for `${shared_data[$option]}`, if I do double-quote it, then it won't work. `printf` is expecting a string, a number, and then three other strings (that is the row format I specified above). From my understanding, if I double-quote `${shared_data[$option]}`, then it will be passed as one whole thing. If I don't, then it will pass an element of the array at a time. Does it make sense?

rodrigosiqueira: Yes, make sense. Thanks

rodrigosiqueira:
Hi @magalilemes , Thanks a lot for your effort.

I already applied your change to the unstable branch.  See https://github.com/kworkflow/kworkflow/commit/eff0103b67e22785c6e53d74b5a8912f63399912

PR #286, opened 2021-05-29
tests: kwlib_test: use help builtin to test pwd

joaoseckler:
Testing kw's command manager with `pwd --help` would break in older
versions of bash, since this option was not available. Using `help pwd`
instead gives consistency to the test.

Closes: #126

### DIFF
--- a/tests/kwlib_test.sh
+++ b/tests/kwlib_test.sh
@@ -101,7 +101,7 @@ function cmd_manager_check_silent_option_Test
   assertTrue "We expected to find scripts" '[[ $ret =~ scripts ]]'
 
   # Test command with parameters
-  ret=$(cmd_manager SILENT pwd --help)
+  ret=$(cmd_manager SILENT help pwd)
   assertTrue "We expected to find -P" '[[ $ret =~ -P ]]'
   assertTrue "We expected to find -L" '[[ $ret =~ -L ]]'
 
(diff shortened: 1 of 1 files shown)


joaoseckler:
Testing this feature requires an older version of bash. I recommend doing that creating a `chroot`. For Debian distributions, one option is to use `sbuild` which wrap a lot of the work involved. Inspired by (these)[https://wiki.debian.org/sbuild] steps, the following commands will allow testing this patch with an older version of bash:

`sudo sbuild-createchroot --include=eatmydata,ccache jessie /srv/chroot/bash-test http://127.0.0.1:3142/ftp.us.debian.org/debian`
`sudo cp -r <path-to-kw>/ /srv/chroot/bash-test/tmp/`
`sudo sbuild-shell jessie`
`cd tmp/kworkflow`
`./run_tests test kwlib_test.sh`

alan-barzilay:
>  break in older versions of bash, since this option was not available. 

how old are we talking about? Because declare -g also didn't exist in older versions and will lead to even bigger issues. Maybe we should decide on what versions of bash we want to support before merging this and #240

rodrigosiqueira:
Hi @joaoseckler ,
Thanks a lot for your PR! I already applied it to https://github.com/kworkflow/kworkflow/commit/5332f5c4100b57710b8fe194d7c520062f61ee76

@alan-barzilay 
About the bash version, do you know which version of `bash` accepts `-g` for the first time? Anyway, let's have this discussion in your PR.

PR #295, opened 2021-06-01
src: Change permissions

joaoseckler:
Included files don't need to be executable. This patch removes the
executable flag of files in `src/` which are not directly executed.

Warned by Lintian.


rodrigosiqueira:
Nice!!

Merged, see: https://github.com/kworkflow/kworkflow/commit/024e822232cd96d46191c8e7322cc9e99cfa6b3c

Issue #290, opened 2021-05-30
Utilize a code formatter to enforce codingstyle


alan-barzilay:
Hello everyone, I would like to implement a code formatter to enforce our codestyle and to make our lives easier.
I have found only 2 options that seem reasonable: [beautysh](https://github.com/lovesegfault/beautysh) and [shfmt](https://github.com/mvdan/sh).

They are both under the active development and seem to have strong communities around them, with many plugins for most common IDEs. So I'm not sure which one would be better, shfmt seems to be older and more consolidated, being a multipurpose tool and with many more stars on github.

My suggestion would be to use shfmt, it has a few interesting configurations that can be made to easily match parts of our codestyle and already provides an official pre-commmit hook configuration

```
Printer flags
       -i <uint>
           Indent: 0 for tabs (default), >0 for number of spaces.

       -bn
           Binary ops like && and | may start a line.

       -ci
           Switch cases will be indented.

       -sr
           Redirect operators will be followed by a space.

       -kp
           Keep column alignment paddings.

       -fn
           Function opening braces are placed on a separate line.
```

It also has a simplify method that could be interesting but I haven´t tried it out.
```
       -s
           Simplify the code.

       -mn
           Minify the code to reduce its size (implies -s).
```

Lastly, I figured I should mention this even though I'm not sure if this could be useful.
shfmt can also replace bash -n to check shell scripts for syntax errors. It is more exhaustive, as it parses all syntax statically and requires valid UTF-8:
```
           $ echo '${foo:1 2}' | bash -n
           $ echo '${foo:1 2}' | shfmt >/dev/null
           1:9: not a valid arithmetic operator: 2
           $ echo 'foo=(1 2)' | bash --posix -n
           $ echo 'foo=(1 2)' | shfmt -p >/dev/null
           1:5: arrays are a bash feature

```

Even if we can't configure it to perfectly fit our expectations and our codestyle, if it does a "good enough" job I think we should adopt it. This would help us ensure code quality and style adherence, and more importantly eliminate the back and forth on PRs about formatting.

alan-barzilay:
@rodrigosiqueira @magalilemes @joaoseckler @kwy95 
If you could try shfmt out and let me kwow what is your opinion, I would love to hear your thoughts on this.
`shfmt -i 2 -fn -ln bash` should be pretty similar to our current codestyle, but I haven´t tested all options yet

rodrigosiqueira:
[...]

PR #299, opened 2021-06-04
Improve code style compliance

kwy95:
This commit changes all function declarations to contain the parentheses
at the end.

Signed-off-by: Rubens Gomes Neto <rubens.gomes.neto@usp.br>

PR #300, opened 2021-06-04
codestyle: fix formatting of our codebase

alan-barzilay:
Make all function adhere to codestyle of `function name()`
Fix indentation to enforce our 2 space identation
Format codebase with `shfmt -w -i 2 -fn -ci -sr -ln bash`

This is the first step into implementing shfmt as a formatter, next step will be to add [this github action](https://github.com/luizm/action-sh-checker) to check PRs for code formatting errors. This is a step towards #290

### DIFF
--- a/.github/workflows/shfmt.yml
+++ b/.github/workflows/shfmt.yml
@@ -1,0 +1,19 @@
+name: formatter-shfmt
+on:
+  [push, pull_request]
+jobs:
+  sh-checker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run the sh-checker
+        uses: luizm/action-sh-checker@v0.3.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SHFMT_OPTS: -i=2 -ln=bash -fn -ci -sr # arguments to shfmt.
+        with:
+          sh_checker_comment: true
+          sh_checker_exclude: "documentation images sounds etc"
+          sh_checker_shfmt_disable: false
+          sh_checker_shellcheck_disable: true
+          sh_checker_checkbashisms_enable: false
--- a/documentation/content/codingstyle.rst
+++ b/documentation/content/codingstyle.rst
@@ -19,6 +19,23 @@ style, for this reason, we copied-and-pasted many pieces from both projects.
 .. _Git: https://github.com/git/git/blob/master/Documentation/CodingGuidelines#L41
 .. _Linux: https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst
 
+
+shfmt
+_____
+
+To help us enforce our codestyle decisions we utilize the
+`shfmt <https://github.com/mvdan/sh>`_ code formatter in our CI pipeline.
+Please refer to the shfmt repository for instructions on how to install it
+on your specific linux distribution, it is available as a package for most
+distributions and as a plugin for most IDEs and text editors.
+To format a file using shfmt::
+
+  shfmt -w -i=2 -ln=bash -fn -ci -sr FILE
+
+To check a file and error with a diff when the formatting differs::
+
+  shfmt -d -i=2 -ln=bash -fn -ci -sr FILE
+
 Indentation
 -----------
 
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -6,7 +6,7 @@ include './src/kwio.sh'
 
 declare -r PATH_TO_TESTS_EXTERNALS="tests/external"
 
-function show_help
+function show_help()
 {
   echo "Usage: $0 [help] [list] [test tfile1 tfile2 ... tfilen] [prepare [-f|--force-update]]"
   echo "Run tests for kworkflow."
@@ -24,9 +24,9 @@ function download_stuff()
   local OVERWRITE="$3"
 
   if "$OVERWRITE"; then
-      ret=$(wget -N "$URL" -P "$PATH_TO")
+    ret=$(wget -N "$URL" -P "$PATH_TO")
   else
-      ret=$(wget -nc "$URL" -P "$PATH_TO")
+    ret=$(wget -nc "$URL" -P "$PATH_TO")
   fi
 
   if [[ "$?" != 0 ]]; then
@@ -42,11 +42,11 @@ function get_external_scripts()
   local -r MAINTAINER_URL="https://raw.githubusercontent.com/torvalds/linux/master/scripts/get_maintainer.pl"
   local -r CHECKPATCH_CONST_STRUCTS="https://raw.githubusercontent.com/torvalds/linux/master/scripts/const_structs.checkpatch"
   local -r CHECKPATCH_SPELLING="https://raw.githubusercontent.com/torvalds/linux/master/scripts/spelling.txt"
-  local DOWNLOAD_URLS=( \
-        CHECKPATCH_URL \
-        CHECKPATCH_CONST_STRUCTS \
-        CHECKPATCH_SPELLING \
-        MAINTAINER_URL )
+  local DOWNLOAD_URLS=(
+    CHECKPATCH_URL
+    CHECKPATCH_CONST_STRUCTS
+    CHECKPATCH_SPELLING
+    MAINTAINER_URL)
 
   say "Downloading external scripts..."
   echo
@@ -55,7 +55,7 @@ function get_external_scripts()
   for url in ${DOWNLOAD_URLS[@]}; do
     download_stuff "${!url}" "$PATH_TO_TESTS_EXTERNALS" "$OVERWRITE"
     if [[ "$?" -eq 113 ]]; then
-     return 113 # Host unreachable errno
+      return 113 # Host unreachable errno
     fi
   done
 }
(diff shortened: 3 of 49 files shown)


alan-barzilay:
if we decide to merge this, we should probably merge #299 before so that I can rebase this and drop the function parenthesis parts

run_tests.sh
```diff
@@ -69,63 +69,63 @@ function check_required_files()
...
-           -f "$PATH_TO_TESTS_EXTERNALS/const_structs.checkpatch" &&
-           -f "$PATH_TO_TESTS_EXTERNALS/spelling.txt" &&
-           -f "$PATH_TO_TESTS_EXTERNALS/get_maintainer.pl" ]]; then
-       # Errno code for File exist
-       return 17
+  if [[ "$force_update" == 'false' && -f \
+    "$PATH_TO_TESTS_EXTERNALS/checkpatch.pl" && -f \
+    "$PATH_TO_TESTS_EXTERNALS/const_structs.checkpatch" && -f \
+    "$PATH_TO_TESTS_EXTERNALS/spelling.txt" && -f \
+    "$PATH_TO_TESTS_EXTERNALS/get_maintainer.pl" ]]; then
```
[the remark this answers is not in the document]
kwy95: It's especially confusing considering you removed the `\` from lines 44 to 48.

src/kwlib.sh
```diff
@@ -107,22 +107,22 @@ function is_kernel_root()
...
+    "${DIR}/Documentation" && -d \
+    "${DIR}/arch" && -d \
+    "${DIR}/include" && -d \
+    "${DIR}/drivers" && -d \
+    "${DIR}/fs" && -d \
+    "${DIR}/init" && -d \
+    "${DIR}/ipc" && -d \
+    "${DIR}/kernel" && -d \
+    "${DIR}/lib" && -d \
+    "${DIR}/scripts" ]]; then
```
kwy95, line 125: Again, I couldn't find anything about this style of listing conditionals in our codestyle guide, any specific reason for this change?

src/mk.sh
```diff
@@ -507,13 +507,13 @@ function kernel_deploy
 
     kernel_install "$reboot" "$name" "" "$target" "$remote"
     end=$(date +%s)
-    runtime=$(( runtime + (end - start) ))
+    runtime=$((runtime + (end - start)))
```
kwy95, line 510: I think the space in this expression adds to readability due to the ending token being `))`.
So it's not immediately obvious if it is a single `)` followed by a double `))` or three singles, or a double followed by a single.
Not sure what to prioritize here.

src/mk.sh
```diff
@@ -777,8 +777,8 @@ function deploy_parser_options()
           ;;
       esac
     else # Handle potential parameters
-      if [[ "$uninstall" != 1 &&
-            ${options_values["TARGET"]} == "$REMOTE_TARGET" ]]; then
+      if [[ "$uninstall" != 1 && \
```
kwy95, line 780: The `\` here is optional, are we choosing to adopt it into our codestyle?

kwy95 (changes requested):
Apart from the comments I laid out I like this a lot, it does everything I did in #299 and more. I propose merging this in place of my changes, and am therefore closing my PR in favor of this one.

alan-barzilay:
@kwy95 thank you for your comments! Instead of replying to your comments one by one I will give here a blanket response since they follow a common thread.

These PR has a mix of formatting changes made by beautysh and shfmt. 

I was testing beautysh and realized that it could fix our problem with missing parenthesis in function declarations, apart from that pretty much everything else was automatically done by shfmt. In particular, all the points you raised came from changes done by shfmt and honestly I had the same reaction.

I think shfmt did a pretty good job and I agree with almost everything it changed, except for a few points like the changes where we have a series of && together, as you pointed out in the kwlib.sh and run_tests.sh.
The only other flag that changes the behaviour of shfmt that I didn´t use was ` -bn       binary ops like && and | may start a line`, but it doesn't seem to affect this particular case of the rearranged &&s.
Currently shfmt has no option for disabling formatting on a part of the code, so I think we have 2 options, drop the parts of this PR formatting that we dont agree with and give up on having a formatter for the project. Or we accept the changes that we dont 100% agree in the name of consistency and the piece of mind that we have a codestyle that's being followed and enforced by shfmt.

Personally I believe it is fundamental to enforce the codestyle for our project and shfmt seems to be the best tool for the job.
I think this PR is a testament as to how flawed our codestyle enforcing is right now. I understand that the idea of a codestyle was not present since the projects conception but the lack of codestyle adherence is by no means a thing of the past, when I first had to rebase this PR there were a series of new formatting changes that had to be done. I'm not saying this to complain about anyone or lay blame, every single one of us had PRs accepted with code style infringements, myself included haha We are only human and things slip by us, we are generally more interested in what the changes being introduced do than ; and () placement. But if 5 people reviewing PRs can´t catch everything, we probably need help.

I believe it is essential for the health of the project that we introduce an enforcer of our codestyle. shfmt is not perfect but it will alleviate our PR revision work and ensure that the bare minimum is being done. By incorporating [a github action](https://github.com/luizm/action-sh-checker) we will have an exact feedback of what is wrong in a given line and what should be done to fix it (their example image focuses on  shellcheck but you get the idea). 
image
This would also help lower the barrier of entry for newcomers to contribute to kw, as there is nothing more frustrating than making a contribution and spend more time discussing nitpicky formatting issues than your real contribution.

If the developer desires to use shfmt locally, it is available as a plugin for most IDEs and text editors, it can also be used as a prehookcommit. From their readme page:
```
Related projects

The following editor integrations wrap shfmt:

    format-shell - Atom plugin
    intellij-shellcript - Intellij Jetbrains shell script plugin
    micro - Editor with a built-in plugin
    shell-format - VS Code plugin
    shfmt.el - Emacs package
    Sublime-Pretty-Shell - Sublime Text 3 plugin
    vim-shfmt - Vim plugin

Other noteworthy integrations include:

    Alternative docker image by PeterDaveHello
    modd - A developer tool that responds to filesystem changes
    prettier-plugin-sh - Prettier plugin using mvdan-sh
    sh-checker - A GitHub Action that performs static analysis for shell scripts
    mdformat-shfmt - mdformat plugin to format shell scripts embedded in Markdown with shfmt
```

rodrigosiqueira (changes requested):
Although I agree with Rubens and even with my dislike to some of the recommendations from shfmt, I think you have strong arguments in favor of shfmt. I'm ok with the trade-offs, but we need to be really careful with the adoption of this tool for the following reasons:

1. If we use shfmt we will rely on their definitions of what is the best, which we may not agree with sometimes. Additionally, this definition might change a lot during the years; we might mitigate this problem by enforcing a version.
2. We definitely need a CI for that; in particular, I don't want to force developers to install packages that came from snap.

After applying this patch, do we get a complete pass in the shfmt? Also, is there any overlapping between this shfmt and shellcheck?

Finally, I think we need to merge this change as a series with the following commits:
1. The patch that makes kw complaint with shfmt;
2. A patch that documents shfmt inside kw with details (how to install shfmt, how to run it, etc);
3. A patch that introduces shfmt to our CI.

Btw, great job here! Thanks a lot for keeping push our code quality to the next level.

Ps.: Rebase your change.

build-drm-misc
```diff
@@ -3,15 +3,16 @@
...
+function build_arch()
+{
+  echo "building $1..."
+  export KBUILD_OUTPUT=$BUILD_DIR/drm-misc/$1
+  mkdir -p $KBUILD_OUTPUT
+  cp $BASE/drm-rerere/drm-misc-$1_defconfig $KBUILD_OUTPUT/.config
+  make oldconfig
+  make $MAKE_OPTS > /dev/null
+  echo "Done."
+  echo "=============="
```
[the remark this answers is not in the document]
alan-barzilay: should I drop this file on this PR or should I open an issue and do it separately?

alan-barzilay:
@rodrigosiqueira yeah, I think I see shfmt as a necessary evil haha

> this definition might change a lot during the years; we might mitigate this problem by enforcing a version.

Currently shfmt is on version 3 and they seem to be discussing the possibility of a version 4, my understanding is that they would only introduce new options for formatting which could be useful to fine-tune shfmt to be closer to what we want. For example, we use the flag -fn to enforce that brackets in function declarations be in a new line (by default they are on the same line as the `function foo()` ).


> After applying this patch, do we get a complete pass in the shfmt? 

yes

> Also, is there any overlapping between this shfmt and shellcheck?

from the shellcheck readme:
image

setup.sh
```diff
@@ -285,7 +285,7 @@ function synchronize_files()
 
   if command_exists "bash"; then
     # Add tabcompletion to bashrc
-    if [[ -f "$HOME/.bashrc"  ||  -h "$HOME/.bashrc" ]]; then
+    if [[ -f "$HOME/.bashrc" || -L "$HOME/.bashrc" ]]; then
```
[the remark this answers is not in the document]
alan-barzilay: No idea, I'm actually more confused as to why these 2 identical options exist haha From the bash documentation: `-h file True if file exists and is a symbolic link.` and `-L file   True if file exists and is a symbolic link.`

build-drm-misc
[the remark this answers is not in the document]
alan-barzilay: I created #304

alan-barzilay:
@rodrigosiqueira  sorry for the mess with all these commits, it was the only way to test if the action was working properly (it was also my first time working with github actions). 

I've finished implementing the series as you described and this pr is ready to be reviewed

kwy95:
I understand and agree with your points about using a formatter. I am annoyed that `shfmt` and the other formatters don't have a little more customization on their codestyles, but the proposed changes that I have the biggest problem with are it's treatment of `&&`  conditionals, as it separates the test flag from the object being tested.
With this in mind I would like to propose a compromise that should be acceptable to `shfmt` and add to the readability, which is to use the `-bn` flag in `shfmt`, and move the `&&` to the start of the next line. For example lines 72 to 76 on the `run_test.sh` file would look something like this:
```
if [[ "$force_update" == 'false' \
    && -f "$PATH_TO_TESTS_EXTERNALS/checkpatch.pl" \
    && -f $PATH_TO_TESTS_EXTERNALS/const_structs.checkpatch" \
    && -f $PATH_TO_TESTS_EXTERNALS/spelling.txt" \
    && -f $PATH_TO_TESTS_EXTERNALS/get_maintainer.pl" ]]; then
```

alan-barzilay:
@kwy95 can't say I disagree with you haha 

> With this in mind I would like to propose a compromise

Have you tried this? Because I had the same idea and couldn´t make it work. I'm unsure if I did something wrong, if this is a bug that should be reported or if we misunderstood the flag. So let me know if you managed to make it work because I hate this particular point as well haha

kwy95:
> Have you tried this?

Can't say I have. I'll take a look at it later and let you know.
@alan-barzilay

.github/workflows/shfmt.yml
```diff
@@ -0,0 +1,19 @@
...
+        uses: luizm/action-sh-checker@v0.2.2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SHFMT_OPTS: -i=2 -ln=bash -fn -ci -sr # arguments to shfmt.
+        with:
+          sh_checker_comment: true
+          sh_checker_exclude: "documentation images sounds etc"
+          sh_checker_shfmt_disable: false
+          sh_checker_shellcheck_disable: true
+          sh_checker_checkbashisms_enable: false
```
rodrigosiqueira, line 19: Hmmm... could you explain how this thing works? I did not get where we installed shfmt. There is some magic behind it? Also, do you have an example in your fork?

documentation/content/codingstyle.rst
```diff
@@ -19,6 +19,16 @@ style, for this reason, we copied-and-pasted many pieces from both projects.
...
 
+
+shfmt
+_____
+
+To help us enforce our codestyle decisions we utilize the `shfmt <https://github.com/mvdan/sh>`_ code formatter in our CI pipeline.
+Please refer to the shfmt repository for instructions on how to install it on your specific linux distribution,
+it is available as a package for most distributions and as a plugin for most IDEs and text editors.
+To format a file using shfmt: `shfmt -w -i=2 -ln=bash -fn -ci -sr FILE`
+To check if a file is properly formatted and error with a diff when the formatting differs: `shfmt -d -i=2 -ln=bash -fn -ci -sr FILE`
```
rodrigosiqueira, line 30: Add each command per line to highlight it.

documentation/content/codingstyle.rst
```diff
@@ -19,6 +19,16 @@ style, for this reason, we copied-and-pasted many pieces from both projects.
 .. _Git: https://github.com/git/git/blob/master/Documentation/CodingGuidelines#L41
 .. _Linux: https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst
 
+
+shfmt
+_____
+
+To help us enforce our codestyle decisions we utilize the `shfmt <https://github.com/mvdan/sh>`_ code formatter in our CI pipeline.
```
rodrigosiqueira, line 26: Is this line under 80 characters?

kwy95:
@alan-barzilay I tried running shfmt on the codebase and got a result that is slightly different than yours and more like what we expected. Take a look at #311

.github/workflows/shfmt.yml
alan-barzilay: > Also, do you have an example in your fork?

The example I do have is for pushes, which would be different than when we have a PR (with PRs a bot will make a post similar to what codecov does)
https://github.com/alan-barzilay/kworkflow/actions

alan-barzilay:
@kwy95 @rodrigosiqueira, I have updated this series with your feedback.
It seems that beautysh had some unexpected side effects and using it as a shortcut made me have to take an even longer route, sorry about that.
Most of our problems with shfmt should be fixed, let  me know if there is anything else missing

alan-barzilay:
I figured it out, this never had anything to do with beautysh. The github action is using an older version of shfmt, I created a pull request to update it.

Specifically [version 3.3.0](https://github.com/mvdan/sh/releases/tag/v3.3.0) fixes our problem with the backlashes in `[[ foo && bar ]]` situations: 

> Test operators like && and || no longer escape newlines

incidentally, I was also using an older version and accidentally upgraded it, this is why I could replicate @kwy95

kwy95 (approved):

magalilemes (approved):
Awesome patch, really good job!

rodrigosiqueira:
Thanks a lot for your effort here!
Merged: https://github.com/kworkflow/kworkflow/commit/97a16252954b75f42b027668fe4185c90346b264

PR #301, opened 2021-06-05
Increase statistics test coverage

magalilemes:
This PR is part of #276. Test coverage was increased to ~93%, according to [Codecov](https://codecov.io/gh/magalilemes/kworkflow/src/increase_statistics_test/src/statistics.sh).

### DIFF
--- a/tests/statistics_test.sh
+++ b/tests/statistics_test.sh
@@ -24,10 +24,10 @@ function suite()
 
 function setUp()
 {
-  export KW_DATA_DIR="tests/samples"
+  export KW_DATA_DIR='tests/samples'
 
   # Samples file data
-  base_statistics="tests/samples/statistics/2020"
+  base_statistics='tests/samples/statistics/2020'
   sample_total='19'
   sample_build_avg='00:00:23' #23
   sample_build_min='00:00:06' #6
@@ -47,6 +47,11 @@ function setUp()
   pre_total='14'
   pre_total_sec='1846'
   pre_formated_sec='00:30:46'
+
+  declare -ga may_27_2020=(
+    'Deploy             8 00:01:13 00:00:31 00:01:04'
+    'Build             19 00:03:40 00:00:06 00:00:23'
+  )
 }
 
 function statistics_Test()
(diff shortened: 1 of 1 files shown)


tests/statistics_test.sh
```diff
@@ -67,6 +72,22 @@ function statistics_Test
   output=$(statistics --invalid-option)
   msg="Invalid parameter: --invalid-option"
   assertEquals "($LINENO)" "$msg" "$output"
+
+  output=$(statistics --day not_a_day 2> /dev/null)
```
rodrigosiqueira, line 76: Double-quotes

tests/statistics_test.sh
```diff
@@ -67,6 +72,22 @@ function statistics_Test
...
+  msg="Invalid parameter: not_a_day"
```
rodrigosiqueira, line 77: Single-quote

tests/statistics_test.sh
```diff
@@ -67,6 +72,22 @@ function statistics_Test
...
+  assertEquals "($LINENO)" "$msg" "$output"
+
+  output=$(statistics --month not_a_month 2> /dev/null)
```
rodrigosiqueira, line 80: Fwiu here, you're testing an empty value which is nice. However, how about adding another test with an invalid month?

tests/statistics_test.sh
```diff
@@ -67,6 +72,22 @@ function statistics_Test
...
+  msg="Invalid parameter: not_a_month"
+  assertEquals "($LINENO)" "$msg" "$output"
+
+  output=$(statistics --year not_a_year 2> /dev/null)
```
rodrigosiqueira, line 84: Try an invalid year.

tests/statistics_test.sh
```diff
@@ -159,27 +179,43 @@ function day_statistics_Test
 
   day_data=$(day_statistics "$base_statistics/05/28")
   assertEquals "($LINENO)" "$msg2" "$day_data"
+
+  day_data=$(day_statistics "$base_statistics/05/27" | tail -n 2)
```
rodrigosiqueira, line 183: Nice use of `tail` :)

rodrigosiqueira (changes requested):
Hi @magalilemes ,

First of all, thanks a lot for your patch. It is always good to see more tests :)

I added a few inline comments; please take a look at them. Also, review the single- vs. double-quote usage in your patch.

tests/statistics_test.sh
magalilemes: Where in `output=$(statistics --day not_a_day 2> /dev/null)` exactly?

tests/statistics_test.sh
magalilemes: not_a_month isn't an empty value. I meant it to be an invalid month, somehow. Like someone was trying to invoke `kw statistics --month not_a_month`.

tests/statistics_test.sh
rodrigosiqueira: My bad... ignore my comment here.

tests/statistics_test.sh
rodrigosiqueira: Hmm... correct if I'm wrong, but `not_a_month` and `not_a_year` are just a random string that you're passing to `statistics`.  The test that I suggested in my previous comment was using something like `13` or `-3` in the month parameter.

tests/statistics_test.sh
rodrigosiqueira: Ignore my previous comment.

rodrigosiqueira:
@magalilemes rebase your code.

rodrigosiqueira:
Thanks a lot for this patch!
Merged, see: https://github.com/kworkflow/kworkflow/commit/f98c639e44da47a8ae8bc5bdd36ce569c0fc59df

PR #312, opened 2021-06-08
Improve tests with shunit2's features

joaoseckler:
This patch improves kw's tests by using two features of shunit2:
 - The fact that function names starting with `test` are automatically queued up for execution, sparing the need of a `suite` function;
 - The offering of a temporary directory which will be cleaned up by shunit2 upon it's exiting.

Each of these features gets a commit implementing it in kw's tests. When implementing these changes to `remote_test.sh`, I noticed it needed a bit of refactoring, so there is a commit for that too.

Addresses:  #152

PR #315, opened 2021-06-10
Fix problems with `kw vars`

### LINKED ISSUE #142: `kw v` doesn't show all configurations
Possible solution: just iterate through `$configurations` and print its entries.
Downside: will also possibly print the configurations not used by kw (if user defines an unknown setting).

joaoseckler:
This PR includes two commits regarding `kw vars`:
 - `574037c` The command was not showing every configuration option. Now it shows all configuration options which are defined. Closes #142.
 - `580bf26` The abbreviation `kw v` was removed in the code when merging #169, but the documentation was never updated. This commit does that.

@matheustavares suggested as a possible solution for #142 to iterate over configurations and print it's entries. Besides the downside he described himself for this approach (printing mistakenly user-defined options) I would add that it doesn't allow a meaningful description of the options. The downside of hard-coding options in `src/kw_config_loader.sh` (our approach here) is that every new option added to `kworkflow_template.config` needs to be implemented in `src/kw_config_loader.sh` too. For developers to remember to do so, this PR also introduces a test to check whether the options in `etc/kworkflow_template.config` match the options defined in `src/kw_config_loader.sh`.

### DIFF
--- a/src/kw_config_loader.sh
+++ b/src/kw_config_loader.sh
@@ -19,6 +19,7 @@ declare -gA deploy_target_opt=(['vm']=1 ['local']=2 ['remote']=3)
 # This function is used to show the current set up used by kworkflow.
 function show_variables()
 {
+  local test_mode=0
   local has_local_config_path="No"
 
   if [ -f "$PWD/kworkflow.config" ]; then
@@ -27,14 +28,70 @@ function show_variables()
     has_local_config_path="No"
   fi
 
-  say "Variables:"
-  echo -e "\tLocal config file: $has_local_config_path"
-  echo -e "\tTarget arch: ${configurations[arch]}"
-  echo -e "\tMount point: ${configurations[mount_point]}"
-  echo -e "\tVirtualization tool: ${configurations[virtualizer]}"
-  echo -e "\tQEMU options: ${configurations[qemu_hw_options]}"
-  echo -e "\tQEMU Net options: ${configurations[qemu_net_options]}"
-  echo -e "\tVdisk: ${configurations[qemu_path_image]}"
+  if [[ "$1" == 'TEST_MODE' ]]; then
+    test_mode='1'
+  fi
+
+  local -Ar ssh=(
+    [ssh_ip]='SSH address'
+    [ssh_port]='SSH port'
+  )
+
+  local -Ar build=(
+    [mount_point]='Mount point'
+    [arch]='Target arch'
+    [kernel_img_name]='Kernel image name'
+    [cross_compile]='Cross-compile name'
+    [menu_config]='Kernel menu config'
+    [doc_type]='Command to generate kernel-doc'
+  )
+
+  local -Ar qemu=(
+    [virtualizer]='Virtualisation tool'
+    [qemu_hw_options]='QEMU hardware setup'
+    [qemu_net_options]='QEMU Net options'
+    [qemu_path_image]='Path for QEMU image'
+  )
+
+  local -Ar notification=(
+    [alert]='Default alert options'
+    [sound_alert_command]='Command for sound notification'
+    [visual_alert_command]='Command for visual notification'
+  )
+
+  local -Ar deploy=(
+    [default_deploy_target]='Deploy target'
+    [reboot_after_deploy]='Reboot after deploy'
+  )
+
+  local -Ar misc=(
+    [disable_statistics_data_track]='Disable tracking of statistical data'
+    [gui_on]='Command to activate GUI'
+    [gui_off]='Command to deactivate GUI'
+  )
+
+  local -Ar group_descriptions=(
+    [build]='Kernel build options'
+    [deploy]='kernel deploy options'
+    [ssh]='SSH options'
+    [qemu]='QEMU options'
+    [notification]='Notification options'
+    [misc]='Miscellaneous options'
+  )
+
+  say "kw configuration variables:"
+  echo -e "  Local config file: $has_local_config_path"
+
+  for group in 'build' 'deploy' 'qemu' 'ssh' 'notification' 'misc'; do
+    echo "  ${group_descriptions["$group"]}:"
+    local -n descriptions="$group"
+
+    for option in "${!descriptions[@]}"; do
+      if [[ -v configurations["$option"] || "$test_mode" == '1' ]]; then
+        echo "    ${descriptions[$option]} ($option): ${configurations[$option]}"
+      fi
+    done
+  done
 }
 
 # This function read the configuration file and make the parser of the data on
--- a/tests/kw_config_loader_test.sh
+++ b/tests/kw_config_loader_test.sh
@@ -131,4 +131,84 @@ $PWD/$CONFIG_FILENAME"
   assertTrue "Wrong config file reading order.\n$expected_vs_got" $?
 }
 
+function test_show_variables_completeness()
+{
+  local -A shown_options
+  local -A possible_options
+  local output
+
+  # get all assigned options, including commented ones
+  # remove #'s and ='s to get option names
+  output="$(cat 'etc/kworkflow_template.config')"
+  output="$(echo "$output" | grep -oE '^(#?\w+=?)' | sed -E 's/[#=]//g')"
+
+  for option in $output; do
+    possible_options["$option"]='1'
+  done
+
+  output="$(show_variables 'TEST_MODE' | grep -E '^    ')"
+  # shellcheck disable=2001
+  output="$(echo "$output" | sed 's/.*(\(\S*\)).*/\1/')"
+
+  for option in $output; do
+    shown_options["$option"]=1
+  done
+
+  for option in "${!possible_options[@]}"; do
+    if [[ ! -v shown_options["$option"] ]]; then
+      fail "show_variables is missing option $option"
+    fi
+  done
+
+  for option in "${!shown_options[@]}"; do
+    if [[ ! -v possible_options["$option"] ]]; then
+      fail "show_variable is showing $option not present in kworkflow_template.config"
+    fi
+  done
+}
+
+function test_show_variables_correctness()
+{
+  local output
+  local option
+  local value
+  local message
+
+  unset configurations
+  declare -gA configurations=(
+    [ssh_ip]=1
+    [ssh_port]=2
+    [mount_point]=3
+    [arch]=4
+    [kernel_img_name]=5
+    [cross_compile]=6
+    [menu_config]=7
+    [doc_type]=8
+    [virtualizer]=9
+    [qemu_hw_options]=10
+    [qemu_net_options]=11
+    [qemu_path_image]=12
+    [alert]=13
+    [sound_alert_command]=14
+    [visual_alert_command]=15
+    [default_deploy_target]=16
+    [reboot_after_deploy]=17
+    [disable_statistics_data_track]=18
+    [gui_on]=19
+    [gui_off]=20
+  )
+
+  output="$(show_variables | grep -E '^    ')"
+
+  while read -r line; do
+    option="$(echo "$line" | sed -E 's/.*\((\S*)\).*/\1/')"
+    value=$(echo "$line" | sed -E 's/.*: (.*)/\1/')
+    if [[ "${configurations["$option"]}" != "$value" ]]; then
+      message="Value of option $option should be "
+      message+="${configurations["$option"]} but is $value"
+      fail "$message"
+    fi
+  done <<< "$output"
+}
+
 invoke_shunit
--- a/documentation/man/kw.rst
+++ b/documentation/man/kw.rst
@@ -284,8 +284,8 @@ the current .config file. The user can suppress this warning by using -f flag.
 warning message because it will remove the config file from kw management. The
 user can suppress this warning by using -f.
 
-v, vars
-~~~~~~~
+vars
+~~~~
 Shows configurations being used by **kw** in the current working directory. To
 do that, it examines both global and local *kworkflow.config* files.
 
--- a/src/help.sh
+++ b/src/help.sh
@@ -29,7 +29,7 @@ function kworkflow_help()
     "\tstatistics - Provide basic statistics related to daily development\n" \
     "\tumount,um - Umount partition created with qemu-nbd\n" \
     "\tup,u - Wake up vm\n" \
-    "\tvars,v - Show variables\n" \
+    "\tvars - Show variables\n" \
     "\tversion,--version,-v - show kw version\n"
 }
 


src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
...
+    test_mode='1'
+  fi
+
+  local -A descriptions=(
+    [arch]='Target arch'
+    [mount_point]='Mount point'
+    [virtualizer]='Virtualization tool'
+    [qemu_hw_options]='QEMU options'
+    [qemu_net_options]='QEMU Net options'
+    [ssh_ip]='Default ssh ip'
```
rodrigosiqueira, line 41: how about "SSH address".

src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
...
+    [ssh_port]='Default ssh port'
```
rodrigosiqueira, line 42: SSH Port

src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
...
+    [kernel_img_name]='Kernel image name'
+    [cross_compile]='Cross-compile name'
+    [menu_config]='Default kernel menu config'
```
rodrigosiqueira, line 45: Kernel menu config

src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
...
+    [qemu_path_image]='Default virtualization tool'
```
rodrigosiqueira, line 46: That's the QEMU image path, not a "virtualization tool"

src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
...
+    [alert]='Default alert options'
+    [sound_alert_command]='Command for sound completion alert'
+    [visual_alert_command]='Command for visual notification of completion'
+    [doc_type]='Command to generate kernel-doc'
+    [default_deploy_target]='Deploy default target'
```
rodrigosiqueira, line 51: Target deploy

src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
...
   echo -e "\tLocal config file: $has_local_config_path"
-  echo -e "\tTarget arch: ${configurations[arch]}"
-  echo -e "\tMount point: ${configurations[mount_point]}"
-  echo -e "\tVirtualization tool: ${configurations[virtualizer]}"
-  echo -e "\tQEMU options: ${configurations[qemu_hw_options]}"
-  echo -e "\tQEMU Net options: ${configurations[qemu_net_options]}"
-  echo -e "\tVdisk: ${configurations[qemu_path_image]}"
+
+  for option in "${!descriptions[@]}"; do
+    if [[ -v configurations[$option] || "$test_mode" == '1' ]]; then
```
rodrigosiqueira, line 62: Prefer "${configurations[$option]}"

src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
...
+    [visual_alert_command]='Command for visual notification of completion'
+    [doc_type]='Command to generate kernel-doc'
+    [default_deploy_target]='Deploy default target'
+    [reboot_after_deploy]='Reboot after deploy'
+    [disable_statistics_data_track]='Disable tracking of statistical data'
+    [gui_on]='Command to activate GUI'
+    [gui_off]='Command to deactivate GUI'
+  )
+
   say "Variables:"
```
rodrigosiqueira, line 58: Maybe we can say something like "kw configuration values"

src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
...
-  echo -e "\tTarget arch: ${configurations[arch]}"
-  echo -e "\tMount point: ${configurations[mount_point]}"
-  echo -e "\tVirtualization tool: ${configurations[virtualizer]}"
-  echo -e "\tQEMU options: ${configurations[qemu_hw_options]}"
-  echo -e "\tQEMU Net options: ${configurations[qemu_net_options]}"
-  echo -e "\tVdisk: ${configurations[qemu_path_image]}"
+
+  for option in "${!descriptions[@]}"; do
+    if [[ -v configurations[$option] || "$test_mode" == '1' ]]; then
+      echo -e "\t${descriptions[$option]} ("$option"): ${configurations[$option]}"
```
rodrigosiqueira, line 63: IMO, replace `\t` for two spaces will look better.

tests/kw_config_loader_test.sh
```diff
@@ -12,6 +12,7 @@ function suite
   suite_addTest "parse_configuration_output_Test"
   suite_addTest "parse_configuration_standard_config_Test"
   suite_addTest "parse_configuration_files_loading_order_Test"
+  suite_addTest "test_show_variables"
```
rodrigosiqueira, line 15: I thought we gonna drop `suite_addTest`...

tests/kw_config_loader_test.sh
```diff
@@ -138,4 +139,34 @@ $PWD/$CONFIG_FILENAME"
   assertTrue "Wrong config file reading order.\n$expected_vs_got" $?
 }
 
+function test_show_variables() {
```
rodrigosiqueira, line 142: `{` should be in a new line

tests/kw_config_loader_test.sh
```diff
@@ -138,4 +139,34 @@ $PWD/$CONFIG_FILENAME"
...
+  local showed_options
+
+  # get all assigned options, including commented ones
+  # remove #'s and ='s to get option names
+  local -A possible_options
+  for o in $(cat "etc/kworkflow_template.config" \
```
rodrigosiqueira, line 148: For the test, use files from `tests/samples/`. Feel free to update it if you need.

tests/kw_config_loader_test.sh
```diff
@@ -138,4 +139,34 @@ $PWD/$CONFIG_FILENAME"
...
+    | grep -oE '^(#?\w+=?)' \
+    | sed  -E 's/[#=]//g'); do
```
rodrigosiqueira, line 150: I think this code will be easy to read if you put it before this `for` loop.

tests/kw_config_loader_test.sh
```diff
@@ -138,4 +139,34 @@ $PWD/$CONFIG_FILENAME"
...
+    possible_options[$o]='1'
```
rodrigosiqueira, line 151: Double-quote around `$o`. Also, `o` is a bad name...

tests/kw_config_loader_test.sh
```diff
@@ -138,4 +139,34 @@ $PWD/$CONFIG_FILENAME"
...
+  done
+
+  local -A shown_options
+  for o in $(show_variables 'TEST_MODE' | tail +3 | sed 's/.*(\(\S*\)).*/\1/'); do
```
rodrigosiqueira, line 155: Move the code after `in` before this `for`.

src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
     has_local_config_path="No"
   fi
 
+  if [[ "$1" == 'TEST_MODE' ]]; then
```
rodrigosiqueira, line 31: I did not understand the value of TEST_MODE in this function.

rodrigosiqueira (changes requested):
Hi @joaoseckler ,

In general, this PR looks really good, I left a couple of comments in the code, but I also have the following comments:

* When you print the information, you just display it in a random order, making it hard to understand the output. Could you group the information? For example, show ssh info, followed by build info, qemu...
* Can we have a test that checks the displayed option? I was thinking about a test that will call show_variables, grep the value after `:` and compare with an expected sequence.

src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
...
+    test_mode='1'
```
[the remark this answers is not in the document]
joaoseckler: I was following the convention of single-quoting strings without expansions... That doesn't apply here?

src/kw_config_loader.sh
```diff
@@ -27,14 +28,41 @@ function show_variables()
...
+  fi
+
+  local -A descriptions=(
+    [arch]='Target arch'
+    [mount_point]='Mount point'
+    [virtualizer]='Virtualization tool'
```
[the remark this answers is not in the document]
joaoseckler: Hm, `virtualizer` doesn't seem to be the notification tool, check this line from `kworkflow_template.config` out:

`virtualizer=qemu-system-x86_64`

Maybe `VM tool` instead?

src/kw_config_loader.sh
joaoseckler: Oops, that was a copy-paste I forgot to alter later...

src/kw_config_loader.sh
joaoseckler: There is a semantic difference here:

`[[ -v "${configurations[$option]}" ]]` checks whether the variable whose name is defined by the expansion of `"${configurations[$option]}"` is set;

`[[ -v configurations["$option"] ]]` check wheter the variable itself if set.

But the quotes around `$option` were indeed missing

src/kw_config_loader.sh
joaoseckler: agree!

tests/kw_config_loader_test.sh
joaoseckler: I think so too! But only after #312 is merged... The problem is that if a `suite` function is defined, shunit2 will not automatically run functions which start with `test`. I thought it was better to leave the refactoring of the entire test code to #312.

tests/kw_config_loader_test.sh
joaoseckler: Makes sense... but it seems like this test is made for developers not to forget to update `kw vars` when they add a new option to `kworkflow_template.config`. Making a copy of it in `tests/samples` would defeat this purpose... What do you think?

src/kw_config_loader.sh
joaoseckler: Test mode assures all possible configurations are printed, regardless if they are defined or not.

src/kw_config_loader.sh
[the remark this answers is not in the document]
rodrigosiqueira: I think we don't need quotes around numbers, we can use the plain number. But you're are right, this is not clear; we probably need to update our codestyle for that.

src/kw_config_loader.sh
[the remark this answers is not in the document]
rodrigosiqueira: You're right, I missed that. VM tool sounds good.

tests/kw_config_loader_test.sh
rodrigosiqueira: That's a good point, you're right.

src/kw_config_loader.sh
rodrigosiqueira: Usually, I use the test mode for validating command sequences or other things that might be hard to test. Here, I'm struggling to see the value of having this TEST_MODE. For example, take a look at some of the tests associated with mk.

tests/kw_config_loader_test.sh
joaoseckler: For example, right now there is an old version of `kworkflow_template.config` in the `tests/samples` folder.

src/kw_config_loader.sh
joaoseckler: Our first approach for this test was to define a custom `configurations` array, including every possible valid configuration option for kw. This way we can test if all the valid options are being printed by `show_variables`, by grepping it's output. Now we check that the set of configuration options that `show_variables` is capable of printing is not smaller than the set of valid configuration options.

Consider the case where a configuration option is removed from kw, and is no longer valid. Consider that the developer who removed that option forgot to remove it from `show_variables`. Now if a user defines this obsolete option manually in his `kworkflow.config`, `kw vars` will still print it as a valid option! We want to avoid that and we want our test to cover this case. (keep in mind that if an option is not defined, it will not be printed by `show_variables`). But how can we check if some undefined option could potentially be printed by `show_variables`? That is, how can we guarantee that the set of configuration option that `show_variables` is capable of printing is not bigger, i.e., does not contain more elements than the set of valid configuration options for kw? We would have to test every possible option, but that is not feasible. This is the purpose of TEST_MODE: it alters the default behavior of not printing options if they are undefined: in TEST_MODE, `show_variables` will print every configuration option that it is capable of printing. That way we can test this edge case.

joaoseckler:
The latest update to this PR includes:
 - Grouping of configuration options by type (ssh, build, deploy, etc)
 - Implementation of a "correctness" test, in addition to the previous "completeness" test
 - formatting and clearer code

tests/kw_config_loader_test.sh
```diff
@@ -141,4 +143,84 @@ $PWD/$CONFIG_FILENAME"
...
+  # remove #'s and ='s to get option names
+  output="$(cat 'etc/kworkflow_template.config')"
+  output="$(echo "$output" | grep -oE '^(#?\w+=?)' | sed -E 's/[#=]//g')"
+
+  for option in $output; do
+    possible_options["$option"]='1'
+  done
+
+  output="$(show_variables 'TEST_MODE' | grep -E '^    ')"
+  output="$(echo "$output" | sed 's/.*(\(\S*\)).*/\1/')"
```
rodrigosiqueira, line 162: I think we can ignore this shellcheck warning. However, it is worth checking if you can simplify...

rodrigosiqueira (changes requested):
After reviewing this patch, I noticed that your approach would require a code update for every new option in the config file. In this sense, what do you think about this approach:

Let's update kworkflow.config in a way that we enforce a basic syntax. Something like this:

# Section: Kernel build options
# Summary: Target arch
arch=x86_64
# Summary: Kernel image name
kernel_img_name=bzImage
...

You will also need to reorganize all kworkflow.config section a logical way. After that, instead of creating arrays, you can parser the file.

What do you think?

joaoseckler:
> After reviewing this patch, I noticed that your approach would require a code update for every new option in the config file. In this sense, what do you think about this approach:
> 
> Let's update kworkflow.config in a way that we enforce a basic syntax. Something like this:
> 
> ```
> # Section: Kernel build options
> # Summary: Target arch
> arch=x86_64
> # Summary: Kernel image name
> kernel_img_name=bzImage
> ...
> ```
> 
> You will also need to reorganize all kworkflow.config section a logical way. After that, instead of creating arrays, you can parser the file.
> 
> What do you think?

I think it is a good idea, but I have a few remarks:

- Should we impose another syntax to a configuration file which could be as as simple as possible? 
- Wouldn't it make `kworkflow.config` less user friendly?
- Your suggestion would imply a bigger feature and demand a bit of time, which I think is not our priority right now;
- The test is there to help developers remember to update the function whenever they update `kworkflow_template.config` (which shouldn't occur too often, as well)

What do you think?

rodrigosiqueira:
> > After reviewing this patch, I noticed that your approach would require a code update for every new option in the config file. In this sense, what do you think about this approach:
> > Let's update kworkflow.config in a way that we enforce a basic syntax. Something like this:
> > ```
> > # Section: Kernel build options
> > # Summary: Target arch
> > arch=x86_64
> > # Summary: Kernel image name
> > kernel_img_name=bzImage
> > ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You will also need to reorganize all kworkflow.config section a logical way. After that, instead of creating arrays, you can parser the file.
> > What do you think?
> 
> I think it is a good idea, but I have a few remarks:
> 
> * Should we impose another syntax to a configuration file which could be as as simple as possible?
> * Wouldn't it make `kworkflow.config` less user friendly?
> * Your suggestion would imply a bigger feature and demand a bit of time, which I think is not our priority right now;
> * The test is there to help developers remember to update the function whenever they update `kworkflow_template.config` (which shouldn't occur too often, as well)
> 
> What do you think?

I agree, let's not waste time on this feature. Please, tackle the other problems and make a new version of this PR.

tests/kw_config_loader_test.sh
joaoseckler: Well, the task here is to get what is between parenthesis in the current line. I think it needs a bit of regex anyway...

rodrigosiqueira:
Merged! Thanks

Issue #320, opened 2021-06-11
SC2034 -  foo appears unused. Verify it or export it.


alan-barzilay:
This issue will be used to keep track of an error/warning type that is common on our codebase. For more information about it, take a look at the official shellcheck wiki entry:
https://github.com/koalaman/shellcheck/wiki/SC2034

PR #324, opened 2021-06-11
code: removing unused variables

alan-barzilay:
Closes kworkflow/kworkflow#320
There are more variables that are flagged as unused but they are false
positves caused by indirection. The shellcheck wiki entry for SC2034 has
a good discussion on how and why this happens: https://github.com/koalaman/shellcheck/wiki/SC2034

We can let shellcheck know specific cases are false positives by adding a directive of the form ```shellcheck disable=SC2034 ``` right before the variable it thinks is unused. But this is such a common occurrence that we would be littering the code base with this directive. I think this particular error should be ignored in our CI and we could easlly disable it if we so desire.

for more information on ignoring errors in shellscheck and how directives work:
https://github.com/koalaman/shellcheck/wiki/Directive
https://github.com/koalaman/shellcheck/wiki/Ignore


alan-barzilay:
@rodrigosiqueira  yeah, I forgot to format it but decided to not fix it yet because the bot hasn't been triggered. I [opened an issue](https://github.com/luizm/action-sh-checker/issues/49) on the repo of the action and the owner said our configuration looked right and he will try to look into it in the near future

kw
```diff
@@ -19,9 +19,6 @@ fi
 
 # Global paths
 KW_LIB_DIR=${KW_LIB_DIR:-"$KW_INSTALL_PREFIX/lib/kw"}
-KW_SHARE_DOC_DIR=${KW_SHARE_DIR:-"$KW_INSTALL_PREFIX/share/doc/kw/"}
-KW_SHARE_MAN_DIR=${KW_SHARE_DIR:-"$KW_INSTALL_PREFIX/share/man"}
-KW_SHARE_SOUND_DIR=${KW_SHARE_DIR:-"$KW_INSTALL_PREFIX/share/sound/kw"}
```
rodrigosiqueira, line 24: We use these variables.

rodrigosiqueira (changes requested):
In general, looks good, but I have to admit that I really miss the integration test to make sure that we do not introduce any regression... Unit tests are not enough anymore...

alan-barzilay:
@joaoseckler since you are looking into the shunit temporary directories, could you give us your input? There are a few variables that I deleted that are related to that, I think you are more qualified than me to decide if we keep them or not haha

kw
alan-barzilay: I found it weird that no tests failed when I removed this, I will be adding them back

joaoseckler:
@alan-barzilay LGTM!

joaoseckler (approved):

rodrigosiqueira:
Hi @alan-barzilay , thanks a lot for your patch.
I merged it with the unstable branch.

PR #325, opened 2021-06-11
Fix typo

alan-barzilay:
There is a typo on one of the arch packages listed as a dependency in setup.sh


rodrigosiqueira:
@alan-barzilay 
Nice catch! Merged to unstable branch. Thanks

Issue #323, opened 2021-06-11
SC2199 -  Arrays implicitly concatenate in [[ ]]. Use a loop (or explicit * instead of @).


alan-barzilay:
This issue will be used to keep track of an error/warning type that is common on our codebase. For more information about it, take a look at the official shellcheck wiki entry:
https://github.com/koalaman/shellcheck/wiki/SC2199

alan-barzilay:
@magalilemes this issue seems to happen in the statistics and pomodoro files (maybe in other files too), since you have been working on them would you mind helping me with this?
I would also like to know your opinion, do you agree this is an issue that should be fixed?
@rodrigosiqueira I would like your opinion too, sorry I missed this error in our meeting

magalilemes:
> @magalilemes this issue seems to happen in the statistics and pomodoro files (maybe in other files too), since you have been working on them would you mind helping me with this?
> I would also like to know your opinion, do you agree this is an issue that should be fixed?

@alan-barzilay Sure thing. I think the rationale behind this warning makes total sense. Will look into it.

PR #326, opened 2021-06-11
code: fixing minor shellcheck issues

alan-barzilay:
This commit fixes all occurrences of the following shellcheck errors:
[SC2059](https://github.com/koalaman/shellcheck/wiki/SC2059) - Don't use variables in the printf format string. Use printf "..%s.." "$foo".
[SC2162](https://github.com/koalaman/shellcheck/wiki/SC2162) - Read without -r will mangle backslashes
[SC2236](https://github.com/koalaman/shellcheck/wiki/SC2236) - Use -n instead of ! -z.

There are 2 occurrences of SC2059 that I'm not really sure what to do with, I don't know if the logic of this error apply to these cases. We can change them, manually disable this error by marking them as false positives or ignore this error as whole in the entire project. 
Here are the 2 occurrences:
on statistics.sh
image

on kwio.sh
image

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -10,7 +10,7 @@ function show_help()
 {
   echo "Usage: $0 [help] [list] [test tfile1 tfile2 ... tfilen] [prepare [-f|--force-update]]"
   echo "Run tests for kworkflow."
-  printf "Example: $0 test kw_test\n\n"
+  printf "Example: %s test kw_test\n\n" "$0"
   echo "  help - displays this help message"
   echo "  list - lists all test files under tests/"
   echo "  test - runs the given test files"
--- a/setup.sh
+++ b/setup.sh
@@ -69,7 +69,7 @@ function check_dependencies()
     return 0
   fi
 
-  if [[ ! -z "$package_list" ]]; then
+  if [[ -n "$package_list" ]]; then
     if [[ "$FORCE" == 0 ]]; then
       if [[ $(ask_yN "Can we install the following dependencies $package_list ?") =~ "0" ]]; then
         return 0
--- a/src/get_maintainer_wrapper.sh
+++ b/src/get_maintainer_wrapper.sh
@@ -24,7 +24,7 @@ function print_files_authors()
     authors=$(grep -oE "MODULE_AUTHOR *\(.*\)" $file |
       sed -E "s/(MODULE_AUTHOR *\( *\"|\" *\))//g" |
       sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/, /g')
-    if [[ ! -z $authors ]]; then
+    if [[ -n $authors ]]; then
       if [[ $printed_authors_separator = false ]]; then
         say $SEPARATOR
         say "MODULE AUTHORS:"
--- a/src/kw_config_loader.sh
+++ b/src/kw_config_loader.sh
@@ -49,7 +49,7 @@ function parse_configuration()
   if [ ! -f "$config_path" ] || [ "$filename" != kworkflow.config ]; then
     return 22 # 22 means Invalid argument - EINVAL
   fi
-
+  # shellcheck disable=SC2162
   while read line; do
     # Line started with # should be ignored
     [[ "$line" =~ ^# ]] && continue
--- a/src/kwio.sh
+++ b/src/kwio.sh
@@ -55,21 +55,22 @@ function alert_completion()
 # the color to be used and two optional params:
 #   - the option '-n', to not output the trailing newline
 #   - text message to be printed
-#
+#shellcheck disable=SC2059
 function colored_print()
 {
   local message="${@:2}"
+  local colored_format="${!1}"
 
   if [[ $# -ge 2 && $2 = "-n" ]]; then
     message="${@:3}"
     if [ -t 1 ]; then
-      printf ${!1} "$message"
+      printf "$colored_format" "$message"
     else
       echo -n "$message"
     fi
   else
     if [ -t 1 ]; then
-      printf "${!1}\n" "$message"
+      printf "$colored_format\n" "$message"
     else
       echo "$message"
     fi
--- a/src/mk.sh
+++ b/src/mk.sh
@@ -703,7 +703,7 @@ function deploy_parser_options()
   options_values["MENU_CONFIG"]="nconfig"
 
   # Set basic default values
-  if [[ ! -z ${configurations[default_deploy_target]} ]]; then
+  if [[ -n ${configurations[default_deploy_target]} ]]; then
     local config_file_deploy_target=${configurations[default_deploy_target]}
     options_values["TARGET"]=${deploy_target_opt[$config_file_deploy_target]}
   else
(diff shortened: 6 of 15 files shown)


rodrigosiqueira:
@alan-barzilay could you split this PR into three commits? One per error. This will make it easy to bisect and revert changes if we need.

alan-barzilay:
@rodrigosiqueira do you have an idea on how to do this efficiently? manually splitting the commit seems pretty time consuming, I'm open to suggestions haha
also, I'm not sure if this is really necessary, only the `read -r` changes could change the code behavior so I think it should still be easy to bisect and revert.

About those 2 printf cases I raised in the PR message, do you think we should mark them as exceptions with an exclude directive or should we refactor them?

magalilemes:
> About those 2 printf cases I raised in the PR message, do you think we should mark them as exceptions with an exclude directive or should we refactor them?

@alan-barzilay As for the statistics.sh one, I think we should mark it as an exception. I had those two variables (`header_format` and `row_format`) assigned to format placeholders to make the code more clear. 😃

rodrigosiqueira:
Maybe you can try something like this:
1. Go to your branch;
2. Generate a patch file with: `git format-patch -1 -o /tmp`;
3. Search for the commit before your change;
4. Next use: `git rebase -i SHA_BEFORE`. Git will open your default editor;
5. Replace `pick` by `e` in front of your commit;
6. Choose one of the shellcheck failures to fix, and delete the other code (I recommend you to keep the largest patch);
7. Add all your changes, and type `git rebase --continue`. Now you have a single commit with one change.
8. Next, use `git am /tmp/PATCH_FROM_STEP_2` and you will see multiple failures related to the part that you already have in your branch; Fix it by only adding the next set of code.

Keep in mind that break this commit into three parts enables us to revert a single patch in case of problems, make it easy to use git bisect, and keep the kw tree healthy. Finally, as a GSoC mentor, I want to make sure that all of you have good knowledge in git; try to do the exercise of break your changes using git. This learning process will be beneficial for your entire life as a software developer :)

alan-barzilay:
@rodrigosiqueira, I ended up doing it a bit different, I used `git reset HEAD^` followed by `git add -p .`. It was faster than I expected haha There is still the problem with printf in the kwio file that I don't know if I mark it as an exception or if I refactor it, honestly it looks like a bit of an esoteric use of printf haha I'm not 100% sure I understand what it is doing.
image


Also, the errors found by shellcheck are the ones I described in #323 that @magalilemes will be working on, so they don't really mater this time, they were already present

rodrigosiqueira:
> @rodrigosiqueira, I ended up doing it a bit different, I used `git reset HEAD^` followed by `git add -p .`. It was faster than I expected haha There is still the problem with printf in the kwio file that I don't know if I mark it as an exception or if I refactor it, honestly it looks like a bit of an esoteric use of printf haha I'm not 100% sure I understand what it is doing.

I agree with you, and I think this color function is not so friendly to read. I think the correct fix for this warning, is to refactor this function. At a first glance, you can have something like this:

```
diff --git a/src/kwio.sh b/src/kwio.sh
index 44ea77a..7cdb7b1 100644
--- a/src/kwio.sh
+++ b/src/kwio.sh
@@ -60,17 +60,18 @@ function alert_completion()
 function colored_print()
 {
   local message="${@:2}"
+  local color="${!1}"
 
   if [[ $# -ge 2 && $2 = "-n" ]]; then
     message="${@:3}"
     if [ -t 1 ]; then
-      printf ${!1} "$message"
+      printf "$color%s" "$message"
     else
       echo -n "$message"
     fi
   else
     if [ -t 1 ]; then
-      printf "${!1}\n" "$message"
+      printf "$color%s\n" "$message"
     else
       echo "$message"
     fi
```
Notice that `$color` is a color code, and for this reason, we need to pass it in the print (I think this is the exception described in the shellcheck doc). However, the message can be passed using `%s`. Maybe create one commit that refactors this function and also get a clean pass to shellcheck.

joaoseckler:
> Notice that $color is a color code...

Actually, it is more than that! `${!1}` is a pattern to be used with printf, check out the `%s` inside of it:

`BLUECOLOR='\033[1;34;49m%s\033[m'`

This makes sense, since we have to set and unset the color in the terminal every time we colour-print: 

But the conclusion is the same: if falls under the exception cited at the [wiki](https://github.com/koalaman/shellcheck/wiki/SC2059):

>Other times, you might have a pattern in a variable:
>
> filepattern="file-%d.jpg"
> printf -v filename "$filepattern" "$number"
>
> This has no good rewrite. Please ignore the warning with a directive.

A welcome refactor, but maybe not pertinent to this patch, would be:
```diff
diff --git a/src/kwio.sh b/src/kwio.sh
index 44ea77a..7cdb7b1 100644
--- a/src/kwio.sh
+++ b/src/kwio.sh
@@ -60,17 +60,18 @@ function alert_completion()
 function colored_print()
 {
   local message="${@:2}"
+  local color_format="${!1}"
 
   if [[ $# -ge 2 && $2 = "-n" ]]; then
     message="${@:3}"
     if [ -t 1 ]; then
-      printf ${!1} "$message"
+      printf "$color_format" "$message"
     else
       echo -n "$message"
     fi
   else
     if [ -t 1 ]; then
-      printf "${!1}\n" "$message"
+      printf "$color_format\n" "$message"
     else
       echo "$message"
     fi
```

alan-barzilay:
> A welcome refactor, but maybe not pertinent to this patch, would be:

@rodrigosiqueira  suggestion for some reason passes shellcheck, even presenting the same issue. @joaoseckler I believe your suggestion wouldn't pass but I can just add a directive marking it as an exception.

Which one of the 2 suggestions should we pick?should we call the variable color or color_format? To me they look pretty much the same, I could go either way

rodrigosiqueira:
> > Notice that $color is a color code...
> 
> Actually, it is more than that! `${!1}` is a pattern to be used with printf, check out the `%s` inside of it:
> 
> `BLUECOLOR='\033[1;34;49m%s\033[m'`
> 
> This makes sense, since we have to set and unset the color in the terminal every time we colour-print:
> 
> But the conclusion is the same: if falls under the exception cited at the [wiki](https://github.com/koalaman/shellcheck/wiki/SC2059):
> 
> > Other times, you might have a pattern in a variable:
> > filepattern="file-%d.jpg"
> > printf -v filename "$filepattern" "$number"
> > This has no good rewrite. Please ignore the warning with a directive.
> 
> A welcome refactor, but maybe not pertinent to this patch, would be:
> 
> ```diff
> diff --git a/src/kwio.sh b/src/kwio.sh
> index 44ea77a..7cdb7b1 100644
> --- a/src/kwio.sh
> +++ b/src/kwio.sh
> @@ -60,17 +60,18 @@ function alert_completion()
>  function colored_print()
>  {
>    local message="${@:2}"
> +  local color_format="${!1}"
>  
>    if [[ $# -ge 2 && $2 = "-n" ]]; then
>      message="${@:3}"
>      if [ -t 1 ]; then
> -      printf ${!1} "$message"
> +      printf "$color_format" "$message"
>      else
>        echo -n "$message"
>      fi
>    else
>      if [ -t 1 ]; then
> -      printf "${!1}\n" "$message"
> +      printf "$color_format\n" "$message"
>      else
>        echo "$message"
>      fi
> ```

@joaoseckler ,
I did not get why we cannot use:
```
printf "$color_format%s\n" "$message"
```
What is the problem to use `%s`?

alan-barzilay:
@rodrigosiqueira everything ready to be merged

rodrigosiqueira:
Merged! Thanks a lot!

Issue #319, opened 2021-06-11
SC2164 - Use cd ... || exit in case cd fails.


alan-barzilay:
This issue will be used to keep track of an error/warning type that is common on our codebase. For more information about it, take a look at the official shellcheck wiki entry:
https://github.com/koalaman/shellcheck/wiki/SC2164

alan-barzilay:
@kwy95 you mentioned you were dealing with something related to this, may I assign you to it?

alan-barzilay:
@kwy95 fyi, this error also seems to be triggered on pushd and popd operations, such as in the kw_config_loader_test.sh file

PR #327, opened 2021-06-11
code: separate declarations from assignements

alan-barzilay:
When a declaration and an assignment happen at the same time, the
return code may be masked, so in this commit we separate them to avoid
this.

This commit fixes all occurrences of the [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155) shellcheck error

### DIFF
--- a/src/diff.sh
+++ b/src/diff.sh
@@ -103,9 +103,10 @@ function diff_side_by_side()
   local file_2="$2"
   local interactive="$3"
   local flag="$4"
-  local columns=$(tput cols)
   local diff_cmd="diff -y --color=always --width=$columns"
+  local columns
 
+  columns=$(tput cols)
   flag=${flag:-""}
 
   if [[ ! -f "$file_1" || ! -f "$file_2" ]]; then
--- a/src/plugins/subsystems/drm/drm.sh
+++ b/src/plugins/subsystems/drm/drm.sh
@@ -91,6 +91,8 @@ function module_control()
   local parameters="$4"
   local flag="$5"
   local module_cmd=""
+  local remote
+  local port
 
   module_cmd=$(convert_module_info "$operation" "$parameters")
   if [[ "$?" != 0 ]]; then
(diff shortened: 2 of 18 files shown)


src/plugins/subsystems/drm/drm.sh
```diff
@@ -255,8 +259,8 @@ function get_available_connectors()
       ;;
     3) # REMOTE TARGET
       local find_conn_cmd="find $SYSFS_CLASS_DRM -name 'card*'"
```
rodrigosiqueira, line 261: Why not move this local to the top of the function as well?

rodrigosiqueira (changes requested):

src/plugins/subsystems/drm/drm.sh
alan-barzilay: @rodrigosiqueira  because there is no return value to be masked, I only touched the declarations that could mask errors

alan-barzilay: @rodrigosiqueira should I split them too?

rodrigosiqueira: > @rodrigosiqueira should I split them too?

By "split them" do you mean the line that I commented? If so, I think it does not hurt to split it.

alan-barzilay:
once again, the  errors reported by shellcheck were already present and are out of scope for this pull request

src/diff.sh
```diff
@@ -103,9 +103,10 @@ function diff_side_by_side()
   local file_2="$2"
   local interactive="$3"
   local flag="$4"
-  local columns=$(tput cols)
   local diff_cmd="diff -y --color=always --width=$columns"
+  local columns
```
joaoseckler, line 107: For the next time, I would suggest not inverting the order of declaration of variables, to make these large refactoring patches as imperceptible as possible. There is no execution difference, but sometimes the order has some meaning. There doesn't seem to be the case anywhere in this patch, so I don't think it justifies pushing another version :)

joaoseckler (approved):

rodrigosiqueira:
> once again, the errors reported by shellcheck were already present and are out of scope for this pull request

@alan-barzilay Why not fix the other shellchecks? Is this PR depending on other PR?

alan-barzilay:
@rodrigosiqueira because they are not lone occurrences, they happen a lot in our code base and need their one commits. Some of them have their own issues to track them (#323 #319) and people assigned to them. This week I dealt with 6 different shellcheck errors, I could make another PR or expand this one but I would rather work on the raspberrypi haha so for now I think we should just merge this single commit as is

PS: rebased and travis seems fine

alan-barzilay:
@rodrigosiqueira everything ready to be merged

rodrigosiqueira:
Merged! thanks a lot

PR #328, opened 2021-06-11
CI/CD: add shellcheck as a test in our pipeline

### LINKED ISSUE #228: Add shellcheck support
There is a fascinating lintian tool dedicated to shell script named shellcheck [1]. It would be nice to make kw complaint to many (but not everything) of the shellcheck validations. However, we probably need to conduct this work in three steps:

1. Setup shellcheck in the project;
2. Update the code to follow shellcheck guideline;
3. Integrate shellcheck in our CI.

1. https://github.com/koalaman/shellcheck

alan-barzilay:
After studying the reviewdog documentation I realized it is possible to run shellcheck only on our changed files, so I have configured a new action to do this. Hopefully this will help us avoid introducing more problems into our code base while we cleanup the ones we already have.

Reviewdog can create a report in 3 levels: error, warning and info. I configured it to fail if any of these are raised, and excluded a few shellcheck errors:

 ## [SC2016](https://github.com/koalaman/shellcheck/wiki/SC2016) - Expressions don't expand in single quotes, use double quotes for that.

This is not so much an error as it is more a warning. The problem is how we use the assertTrue and assertFalse assertions of shunit, we pass commands delimited by single quotes that are seen as errors by shellcheck. We could insert a directive next to each one of these, but they are so bountiful in our code that I don´t think its worth it. I believe this particular error would just introduce noise. An example from configm_test.sh:
image

## [SC2181](https://github.com/koalaman/shellcheck/wiki/SC2181) - Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.

As we discussed in our last meeting, we often opt to do this to improve legibility or because we need the output of the command as well as checking its return value. To avoid littering our codebase with exclude directives I think we should just exclude it from the analysis.

## [SC2034](https://github.com/koalaman/shellcheck/wiki/SC2034) - foo appears unused. Verify it or export it.

This error is an important one but it will always comes with a lot of noise from false positives. I would rather avoid littering our codebase with exclude directives. From the shellcheck wiki:

>  #### Exceptions
> 
> This warning may be falsely emitted when a variable is only referenced indirectly, and for variables that are intentionally unused.
>  ##### Indirection
> 
> It's ShellCheck's intended behavior to emit this warning for any variable that is only referenced though indirection:
> 
> `# foo generates a warning, even though it has five indirect references`
>` foo=42     `
> `name=foo`
> `echo "${!name} $((name))"`
> `export "$name"; eval "echo $name"`
> `declare -n name; echo "$name"`
> 
> This is an intentional design decision and not a bug. If you have variables that will not have direct references, consider using an associative array in bash, or just Ignore the warning.
> 
> Tracking indirect references is a common problem for compilers and static analysis tool, and it is known to be unsolvable in the most general case. There are two ways to handle unresolved indirections (which in a realistic program will essentially be all of them):
> 
>     
> - Avoid false positives by assuming all variables may potentially be used (disable all unused variable warnings)
> - Keep true positives by allowing some false positives (emit unused variable warnings that may be incorrect)
> 
> Compilers are forced to do the former, but static analysis tools generally do the latter. This includes pylint, jshint and shellcheck itself. This is a design decision meant to make the tools more helpful at the expense of some noise. For consistency and to avoid giving the impression that it should work more generally, ShellCheck does not attempt to resolve even trivial indirections.
> 

----------------------

To illustrate how this action works and to see it working in practice, I prepared [this dummy PR](https://github.com/alan-barzilay/kworkflow/pull/2) in my fork.
There may be other shellcheck errors that we disagree with, but looking at the most prevalent ones in our code I think we already decided that we agree with them and thus I think we should deal with new errors on a case by case basis, after all there are more then 400 different shellcheck errors

Closes #228

### DIFF
--- a/.github/workflows/shellcheck_reviewdog.yml
+++ b/.github/workflows/shellcheck_reviewdog.yml
@@ -1,0 +1,19 @@
+name: shellcheck-reviewdog
+on: [pull_request]
+jobs:
+  shellcheck:
+    name: runner / shellcheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: shellcheck
+        uses: reviewdog/action-shellcheck@v1
+        with:
+          github_token: ${{ secrets.github_token }}
+          reporter: github-pr-review
+          level: info
+          filter_mode: added # This determines the scope of what reviewdog will report
+          fail_on_error: true
+          path: "."
+          pattern: "*.sh"
+          shellcheck_flags: "--external-sources --shell=bash --exclude=SC2016,SC2181,SC2034"


.github/workflows/shellcheck_reviewdog.yml
```diff
@@ -0,0 +1,19 @@
...
+      - name: shellcheck
+        uses: reviewdog/action-shellcheck@v1
+        with:
+          github_token: ${{ secrets.github_token }}
+          reporter: github-pr-review
+          level: info
+          filter_mode: added # This determines the scope of what reviewdog will report
+          fail_on_error: true
+          path: "."
+          pattern: "*.sh"
```
alan-barzilay, line 18: I have a problem with this line, the way it is the `tests/utils` file is being ignored. From the [action documenation](https://github.com/reviewdog/action-shellcheck#pattern):

> ### `pattern`
> Optional. File patterns of target files. Same as -name [pattern] of find command. Default: *.sh

Maybe I just don't understand `find` enough, but I can't seem to create multiple patterns. If anyone could help me with this I would appreciate it. A somewhat dirty workaround would be to create another reviewdog step just for utils, but this seems a bit "gambiarra" to me

alan-barzilay: we could also just rename utils to utils.sh haha it always bugged me that it is the only one without a file extension

kwy95: Apparently `find` takes in a glob, and I think you can match multiple patterns like this:
```
{utils, *.sh}
```

alan-barzilay: @kwy95, [here](https://github.com/alan-barzilay/kworkflow/pull/1/checks?check_run_id=2810073860) I tried using the line `           pattern: "{utils,*.sh}" `, but it didn't work. Also, I tried using that with find on the command line and couldn't make it work haha  how exactly should be the find command?

kwy95: Is there a way to activate extended globbing in the test? It's usually done with `shopt -s extglob`

kwy95: With `extglob` you can try this:
```
@(utils|*.sh)
```

alan-barzilay: @kwy95 not sure, I never used glob that much so I'm kinda lost. Btw, i found [where this is evaluated](https://github.com/reviewdog/action-shellcheck/blob/master/entrypoint.sh):
 `$(find "${INPUT_PATH:-'.'}" -not -path "${INPUT_EXCLUDE}" -type f -name "${INPUT_PATTERN:-'*.sh'}")`
what we are providing is that `INPUT_PATTERN` variable

alan-barzilay: `-name "utils" -or -name "*.sh"` works perfectly, but I don't think we can escape from the variable in a way to input more flags

rodrigosiqueira: I'm a little bit late to the discussion, but I also agree that we can add `.sh` to the end of `utils`.

rodrigosiqueira: I'm going to merge your change first, and later we can update the `utils` file.

rodrigosiqueira:
Merged! Let's see shellcheck in action :)
https://github.com/kworkflow/kworkflow/commit/1e46b8f383264b532474e4c0a26fb7843cc7cf63

PR #330, opened 2021-06-13
Refactor statistics

magalilemes:
This is related to #276 and tries to improve the current structure of the `src/statistics`.sh file.

### DIFF
--- a/src/statistics.sh
+++ b/src/statistics.sh
@@ -17,10 +17,11 @@ declare -gA shared_data=(["deploy"]='' ["build"]='' ["list"]='' ["uninstall"]=''
 
 function statistics()
 {
-  local info_request="$1"
-  local day=$(get_today_info '+%d')
-  local year_month_dir=$(get_today_info '+%Y/%m')
-  local date_param
+  local target="$1"
+  local day
+  local week
+  local month
+  local year
 
   if [[ "$1" == -h ]]; then
     statistics_help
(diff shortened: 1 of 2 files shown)


alan-barzilay:
@magalilemes, the problem shellcheck found is that it cannot understand where configurations came from. If I'm not mistaken this is because configurations is defined elsewhere and then imported here with include, right? I think that by using the include function shellcheck may not be able to follow the imports and understand what is being imported. Maybe we should add this error to the the list of things to be ignored, or rethink (again) how we deal with imports (I think import guards wouldn't have this problem)

src/statistics.sh
```diff
@@ -17,82 +17,83 @@ declare -gA shared_data=(["deploy"]='' ["build"]='' ["list"]='' ["uninstall"]=''
...
-  fi
-
-  case "$info_request" in
+  case "$target" in
     --day)
-      if [[ -z "$date_param" ]]; then
-        target_day="$KW_DATA_DIR/statistics/$year_month_dir/$day"
-      else
-        target_day="$KW_DATA_DIR/statistics/$(date_to_format "$date_param")"
+      local day
```
alan-barzilay, line 42: @rodrigosiqueira, do we want to keep the `local` declarations all together in the beginning of the function? you mentioned something like this

rodrigosiqueira: Yeah... let's do that since this can avoid shellcheck problems.

rodrigosiqueira:
> @magalilemes, the problem shellcheck found is that it cannot understand where `configurations` came from. If I'm not mistaken this is because configurations is defined elsewhere and then imported here with include, right? I think that by using the include function shellcheck may not be able to follow the imports and understand what is being imported. This could also be seen as a counter part for the[ SC2034](https://github.com/koalaman/shellcheck/wiki/SC2034) error
> Maybe we should add this error to the the list of things to be ignored, or rethink (again) how we deal with imports (I think import guards wouldn't have this problem)

Btw, using `configurations` in the way that Magali did, is something common around kw and harmless in this specific case. So let's ignore this check.

alan-barzilay:
> So let's ignore this check.

@rodrigosiqueira I have created PR #331 for this

src/statistics.sh
```diff
@@ -291,11 +287,13 @@ function week_statistics()
 
 function month_statistics()
 {
-  local month_path="$KW_DATA_DIR/statistics/$1"
+  local month="$1"
+  local month_path="$KW_DATA_DIR/statistics/$month"
   local all_data=""
+  local pretty _month
```
joaoseckler, line 293: I think there is a left over space!

joaoseckler (changes requested):
Other than the one little change I commented, LGTM!

rodrigosiqueira:
Merged, thanks a lot.

PR #331, opened 2021-06-13
shellcheck: add error SC2154 to ignore list

alan-barzilay:
SC2154 - foo is referenced but not assigned

As discussed in PR #330, this is a common false positive in our codebase
caused by our include function that renders shellcheck incapable of
seeing imports, so we will start to ignore it occurrence.


rodrigosiqueira:
Thanks @alan-barzilay ,

PR #334, opened 2021-06-15
Add development workflow page

magalilemes:
This PR adds a new documentation page that briefly explains what should be considered when developing for kw, pointing out specific pages related to some of the topics.

### DIFF
--- a/documentation/content/developmentworkflow.rst
+++ b/documentation/content/developmentworkflow.rst
@@ -1,0 +1,122 @@
+========================
+  Development workflow
+========================
+
+.. contents::
+   :depth: 1
+   :local:
+   :backlinks: none
+
+.. highlight:: console
+
+kw is developed using Git and is hosted at Github. If you want to understand
+our development workflow better, clone the repository to have a local copy::
+
+  git clone https://github.com/kworkflow/kworkflow.git
+
+Install
+-------
+Installing kw is one of the first steps to understanding its development
+workflow. After cloning the repository, read the page related to
+:ref:`installing and removing<install-and-remove-kw>`.
+
+Tests
+-----
+To make sure our code is working as it is supposed to, we add
+:ref:`tests<tests>` to check that.
+
+All our tests can be found in the `tests` folder. Notice that every test file
+name ends with test before the .sh extension (i.e. `mk_test.sh` and
+`statistics_test.sh`). This is done so that the `run_tests.sh` file can detect
+those tests and run or list them.
+
+Unit tests
+==========
+Unit testing is an essential part of our development workflow and we use the
+`shUnit2 <https://github.com/kward/shunit2/>`_ framework to write our tests.
+Ideally, every new function should have a test related to it. So, if you're
+working on a new feature, added a new function, or changed the behavior of an
+existing one, it's fundamental to make the necessary ajustments related to
+testing.
+
+To see how to run individual test files, check our :ref:`page on that<tests>`.
+
+Run kw without installing it
+----------------------------
+Apart from using the installed version of kw, you can also use your local
+version by going into it and running `./kw` with any other option you want,
+just like you would do when you run kw that is installed in your machine.
+
+For instance, suppose you changed the `help.sh` file and want to see if
+everything's working as expected. Inside the kw directory you're working on,
+you can just invoke::
+
+  ./kw help
+
+This is a convenient option when you're developing for kw and don't want to
+install it every time you make a new change to see the result. Running `./kw b`
+or `./kw d` won't work, though, since these two options require you to be inside
+a repository.
+
+Documentation
+-------------
+As important as writing tests is documenting things. Our documentation is
+located in the `documentation` folder, and we use
+`Sphinx <https://www.sphinx-doc.org/en/master/>`_ to create it.
+
+Whether you changed the way a command behaves or added a new feature, modify
+the documentation accordingly and preferably with examples. Also, in the source
+code, we tend to comment on top of each function, explaining its usage.
+
+ShellCheck and shfmt
+--------------------
+kw follows its own :ref:`code style guide<coding-style>`, and makes use of
+tools, such as ShellCheck and :ref:`shfmt<shfmt-label>`, to make sure this code
+style is being followed.
+
+Git pre-commit hook
+===================
+A very handy way to ensure that your code is well-formatted and follows our
+standards is to have a pre-commit hook configured to run the aforementioned
+tools before committing. This may, for example, prevent code with bad syntax and
+wrong formatting from even being committed before having these things fixed.
+
+To configure this Git pre-commit hook:
+
+0. Have shellcheck installed on your computer.
+
+1. Install `pre-commit <https://pre-commit.com/>`_.
+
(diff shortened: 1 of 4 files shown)


documentation/content/developmentworkflow.rst
```diff
@@ -0,0 +1,110 @@
...
+
+.. contents::
+   :depth: 1
+   :local:
+   :backlinks: none
+
+.. highlight:: console
+
+kw is developed using Git and is hosted at Github. If you want to understand
+our development workflow better, clone our repository to have your local copy::
```
alan-barzilay, line 13: maybe: "(...) to have your own local copy" or "(...) to have a local copy"

alan-barzilay: also, maybe "clone the repository" instead of "our"?

documentation/content/developmentworkflow.rst
```diff
@@ -0,0 +1,110 @@
...
+:ref:`tests<tests>` to check that.
+
+All our tests can be found in the `tests` folder. Notice that every test file
+name ends with test before the .sh extension (i.e. `mk_test.sh` and
+`statistics_test.sh`). This is done so that the `run_tests.sh` file can detect
+those tests and run or list them.
+
+Unit tests
+==========
+Unit testing is an essential part of our development workflow. Ideally, every
```
alan-barzilay, line 35: maybe mention that we use shunit2?

documentation/content/developmentworkflow.rst
```diff
@@ -0,0 +1,110 @@
...
+new function should have a test related to it. So, if you're working on a new
+feature, added a new function, or changed the behavior of an existing one, it's
+fundamental to make the necessary ajustments related to testing.
+
+To see how to run individual test files, check our :ref:`page on that<tests>`.
+
+Run kw without installing it
```
alan-barzilay, line 42: It may be a good idea to explicit the development benefits of this, I think the way it is written may lead a new user to believe this is another way of using kw while it is primarily focused on making development more agile

documentation/content/developmentworkflow.rst
```diff
@@ -0,0 +1,110 @@
...
+  .. code-block:: yaml
+
+    repos:
+    - repo: local
+      hooks:
+        - id: shfmt
+          name: shfmt
+          minimum_pre_commit_version: 2.4.0
+          language: golang
+          additional_dependencies: [mvdan.cc/sh/v3/cmd/shfmt@v3.3.0]
```
alan-barzilay, line 94: not sure if this needs to be mentioned on the documentation, but by changing what comes after  the @ we can control the shfmt version so we should always use the same one that is being used in the github action. I don´t know enough about pre-commit to say if there is a way to set this to use always the latest version, but this may be something that will need constant updates otherwise. Every time shmft has a new release and we want to update it we need to remember to also update this part of the docs

documentation/content/developmentworkflow.rst
```diff
@@ -0,0 +1,110 @@
...
+Git pre-commit hook
+===================
+A very handy way to ensure that our code is well-formatted and follows our
+standards is to have a pre-commit hook configured to run the aforementioned
+tools before committing. This may, for example, prevent code with bad syntax and
+wrong formatting from even being committed before having these things fixed.
+
+To configure this Git pre-commit hook:
+
+0. Have shfmt and shellcheck installed on your computer.
```
alan-barzilay, line 78: I'm not sure if shfmt needs to be installed, I have a strong suspicion that there is docker black magic happening that configures shfmt automatically when you install the hook.
(shellcheck I think really needs to be installed, their hooks seem to be configured differently)

documentation/index.rst
```diff
@@ -10,6 +10,7 @@ available in the command line interface.
    :maxdepth: 1
    :caption: Content:
 
+   content/developmentworkflow
```
joaoseckler, line 13: I think this line makes "Development workflow" become the first item in kw's documentation list. It think it makes sense for the development workflow to come in the last place.

joaoseckler (changes requested):

documentation/content/developmentworkflow.rst
```diff
@@ -0,0 +1,110 @@
...
+----------------------------
+Apart from using the installed version of kw, you can also use your local
+version by going into it and running `./kw` with any other option you want,
+just like you would do when you run kw that is installed in your machine.
+
+For instance, suppose you changed the `help.sh` file and want to see if
+everything's working as expected. Inside the kw directory you're working on,
+you can just invoke::
+
+  ./kw --help
```
joaoseckler, line 52: Unintuitively, `./kw --help` is an invalid option (although it gets you what you want, since invalid options trigger the printing of the help message). Maybe it shouldn't be like this, but since it is this line should read `./kw help`

joaoseckler (changes requested):

documentation/content/developmentworkflow.rst
```diff
@@ -0,0 +1,115 @@
...
+
+For instance, suppose you changed the `help.sh` file and want to see if
+everything's working as expected. Inside the kw directory you're working on,
+you can just invoke::
+
+  ./kw help
+
+This a convenient option when you're developing for kw and don't want to install
+it every time you make a new change to see the result.
+
```
rodrigosiqueira, line 58: It is important to highlight that some options may not work, such as `kw b` or `kw d`, because it requires a kernel. Maybe you can talk about a limitation around this approach.

documentation/content/developmentworkflow.rst
```diff
@@ -0,0 +1,115 @@
...
+Documentation
+-------------
+As important as writing tests is documenting things. Our documentation is
+located in the `documentation` folder.
+
+Whether you changed the way a command behaves or added a new feature, modify
+the documentation accordingly and preferably with examples. Also, in the source
+code, we tend to comment on top of each function, explaining its usage.
```
rodrigosiqueira, line 66: Highlight that we try to use the Sphinx way to document code.

documentation/content/developmentworkflow.rst
```diff
@@ -0,0 +1,115 @@
...
+
+ShellCheck and shfmt
+--------------------
+kw follows its own :ref:`code style guide<coding-style>`, and makes use of
+tools, such as ShellCheck and :ref:`shfmt<shfmt-label>`, to make sure this code
+style is being followed.
+
+Git pre-commit hook
+===================
+A very handy way to ensure that our code is well-formatted and follows our
```
rodrigosiqueira, line 76: s/ensure that our code is/ensure that your code is/

hooks are something on the developer side, replace 'our' by 'your'.

documentation/content/developmentworkflow.rst
rodrigosiqueira: It does not hurt to add a note about that. On the other hand, if we update shfmt version, we probably need to update the documentation.

rodrigosiqueira (changes requested):

rodrigosiqueira:
Merged! Thanks a lot.

See: https://github.com/kworkflow/kworkflow/commit/423408b1eab01bb94105d6b9ae44337eea2302a5

Issue #317, opened 2021-06-10
Add more notification sounds to kw


joaoseckler:
kworkflow currently only has one option for sound completion notification. We should have more options. The licenses for these sounds should be compatible with kw.

alan-barzilay:
Have you seen these resources?
https://freesound.org/
https://soundbible.com/royalty-free-sounds-1.html

Issue #333, opened 2021-06-15
Update kw sound alert


rodrigosiqueira:
When Joao was packaging kw for Debian, we realize that the `complete.wav` does not have the correct attribution to the original author (https://github.com/kworkflow/kworkflow/tree/master/sounds). Unfortunately, I couldn't find where I downloaded it, and for this reason, we need to replace this sound. In this sense, I'm looking for suggestions for a new set of sound alerts for kw; I want to add five options, but one of them will be the default one. The audio should follow these requirements:

1. It should be under creative commons license;
2. We should be able to hear it even if we are listening to music;
3. It should use something that does not give a fright;

This website has many samples:
https://freesound.org/

@kwy95 @magalilemes @joaoseckler @alan-barzilay @melissawen I'm open for suggestions.

joaoseckler:
Is this a duplicate of #317?

rodrigosiqueira:
Done! Thanks

PR #335, opened 2021-06-15
sounds: Add new notification sounds to kw

joaoseckler:
This commit introduces six new sounds to kw, and removes the old one,
for it's lack of licensing. It also adds a page to the documentation
with credits to the authors of the sounds, changes the default sound
notification in etc/kworkflow_template.config and adjust tests
accordingly.

Closes: #317 and #333


joaoseckler:
I would like feedback specifically for these aspects of this PR:

- Where should we put the `sound_attribution.rst` file? I thought maybe a new section "credits" could exist.
- Is that the best name for it? "Sound attributions?" I was inspired by minecraft's analogous page: [https://www.minecraft.net/en-us/attribution/sound](https://www.minecraft.net/en-us/attribution/sound)
- Most importantly: are the sounds good? Do they notify? Do they overcome some music that a user might be listening? Is it clipping?

Thanks a lot!

alan-barzilay:
> Where should we put the sound_attribution.rst file? I thought maybe a new section "credits" could exist.

Since sounds are the only thing we need to attribute credit for at the moment I think it would make sense to just leave it inside the `sounds` directory, this way it is also closer to the file being attributed

joaoseckler:
> > Where should we put the sound_attribution.rst file? I thought maybe a new section "credits" could exist.
> 
> Since sounds are the only thing we need to attribute credit for at the moment I think it would make sense to just leave it inside the `sounds` directory, this way it is also closer to the file being attributed

And then keep it as a plain text file instead of rst, right? Makes sense to me. At the same time, shouldn't this be inside of kw's documentation?

rodrigosiqueira:
> > > Where should we put the sound_attribution.rst file? I thought maybe a new section "credits" could exist.
> > 
> > 
> > Since sounds are the only thing we need to attribute credit for at the moment I think it would make sense to just leave it inside the `sounds` directory, this way it is also closer to the file being attributed
> 
> And then keep it as a plain text file instead of rst, right? Makes sense to me. At the same time, shouldn't this be inside of kw's documentation?

I think the plain text should be enough.

rodrigosiqueira:
Most of the sounds are hard to hear when we are playing some music. For example, try to hear them while playing this music:

https://youtu.be/ohJxNSn9aL0?t=1526

Also, `ovo.wav` is really hard to hear. I think we can remove it.

Can you try to make these sounds a little bit louder?
P.s.: `bell.wav` is the best one in terms of volume.

joaoseckler:
Felipe Kouznets has helped me louden these sounds more. At this point, if my volume is at 100%, my ears hurt a little bit when listening to these sounds. I don't know if that should be a concern.

I also removed "ovo.wav"

magalilemes:
I like all the sounds. They're at a good volume level and not clipping, but I believe the caixa one is a bit too short and abrupt.

joaoseckler:
> I like all the sounds. They're at a good volume level and not clipping, but I believe the caixa one is a bit too short and abrupt.

Do you think we should delete it? I think I can see myself using it

alan-barzilay:
>    > I like all the sounds. They're at a good volume level and not clipping, but I believe the caixa one is a bit too short and abrupt.
> 
> Do you think we should delete it? I think I can see myself using it


I think we should keep it, since its not the default I see no problem

rodrigosiqueira:
bell.wav and berimbau.wav are my favorite :)
I would make `bell.wav` the default one.

@melissawen any comment here?

rodrigosiqueira:
@joaoseckler did you try to add the fade-in effect at the begging of each sound?

joaoseckler:
@rodrigosiqueira Fade-in effects added!

rodrigosiqueira (changes requested):
Hi @joaoseckler ,

Final change:
Add a rule in the setup that creates a symbolic link from complete.wav to bell.wav. Users with the old `kworkflow.config` still have `complete.wav` in their config file, which does not exist after your patch; the symbolic link will ensure that old config files will still work.

kwy95 (approved):
LGTM

rodrigosiqueira:
@joaoseckler did you see my previous comment?

rodrigosiqueira:
Merged, thanks

PR #339, opened 2021-06-17
src: mk: run make olddefconfig before building

kwy95:
Runs `make olddefconfig` before building kernel

### DIFF
--- a/documentation/tutorials/buildlinux.rst
+++ b/documentation/tutorials/buildlinux.rst
@@ -82,7 +82,10 @@ And it can be even simpler by just running::
 
   kw b
 
+.. note:
+    kw runs `make olddefconfig` before building.
+
 Well, that's it. kw will automatically infer the number of job slots for
-compiling based on the number of cores of your machine (i.e. when running make 
+compiling based on the number of cores of your machine (i.e. when running make
 -jNUMBER, NUMBER is an integer that specifies the number of processes that will
 run at the same time), and compilation will begin!
--- a/src/build.sh
+++ b/src/build.sh
@@ -90,6 +90,11 @@ function kernel_build()
     esac
   done
 
+  if [ -f "$(join_path "$PWD" '.config')" ]; then
+    local mkflag=${flag:-'SILENT'}
+    cmd_manager "$mkflag" 'make olddefconfig &> /dev/null'
+  fi
+
   if [ -x "$(command -v nproc)" ]; then
     PARALLEL_CORES=$(nproc --all)
   else
--- a/tests/build_test.sh
+++ b/tests/build_test.sh
@@ -92,6 +92,14 @@ function test_kernel_build()
   expected_result="make -j$PARALLEL_CORES ARCH=x86_64 "
   assertEquals "($LINENO)" "$expected_result" "${output}"
 
+  cp "$original_dir/$SAMPLES_DIR/.config" .config
+  output=$(kernel_build 'TEST_MODE' | head -2) # Remove statistics output
+  declare -a expected_result=(
+    'make olddefconfig &> /dev/null'
+    "make -j$PARALLEL_CORES ARCH=x86_64"
+  )
+  compare_command_sequence expected_result[@] "$output" "$LINENO"
+  rm .config
 }
 
 function test_build_info()


documentation/tutorials/buildlinux.rst
```diff
@@ -81,7 +81,11 @@ And it can be even simpler by just running::
 
   kw b
 
+.. note:
+    kw will also run `make olddefconfig` for you before building, so you can
```
rodrigosiqueira, line 85: Maybe:
kw runs `...` before start building...

Also, you can drop the other part that says "so you..."

src/mk.sh
```diff
@@ -610,6 +610,9 @@ function mk_build()
     esac
   done
 
+  # sets default values for any missing configurations in .config
+  cmd_manager "$flag" "make olddefconfig"
```
rodrigosiqueira, line 614: Single quote around the command.

Also, `cmd_manager` will show this command; I think we can set the default flag to SILENT here. Finally, you need to check if there is a `.config` file before running this command; if there is no config file, do not run this command.

src/mk.sh
```diff
@@ -610,6 +610,9 @@ function mk_build()
     esac
   done
 
+  # sets default values for any missing configurations in .config
```
rodrigosiqueira, line 613: I think we don't need this comment.

tests/mk_test.sh
```diff
@@ -156,8 +156,9 @@ function mk_build_Test()
   assertEquals "($ID)" "$ret" "22"
 
   ID=5
-  output=$(mk_build 'TEST_MODE' | head -1) # Remove statistics output
-  expected_result="make -j$PARALLEL_CORES ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu-"
+  output=$(mk_build 'TEST_MODE' | head -2) # Remove statistics output
+  expected_result="make olddefconfig
+make -j$PARALLEL_CORES ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu-"
```
rodrigosiqueira, line 161: I would prefer that you create an array with the command sequence. See `mk_list_remote_kernels_Test` as a reference.

rodrigosiqueira (changes requested):

rodrigosiqueira:
@kwy95, you forgot to add tests.

src/mk.sh
```diff
@@ -639,6 +639,10 @@ function mk_build()
     esac
   done
 
+  if [ -f "$(join_path "$PWD" '/.config')" ]; then
```
rodrigosiqueira, line 642: Why did you add this `/` before `.config`?

src/mk.sh
```diff
@@ -639,6 +639,10 @@ function mk_build()
...
+    cmd_manager "${flag:-"SILENT"}" 'make olddefconfig'
```
rodrigosiqueira, line 643: Use single-quote around 'SILENT'

tests/mk_test.sh
```diff
@@ -155,6 +155,22 @@ function test_mk_build()
   expected_result="make -j$PARALLEL_CORES ARCH=x86_64 "
   assertEquals "($ID)" "$expected_result" "${output}"
 
+  ID=7
+  local count=0
+  cp "$original/tests/samples/.config" .config
+  output=$(mk_build 'TEST_MODE' | head -2) # Remove statistics output
```
rodrigosiqueira, line 161: Remove statistics output? Do we have statistics output here?

tests/mk_test.sh
```diff
@@ -155,6 +155,22 @@ function test_mk_build()
   expected_result="make -j$PARALLEL_CORES ARCH=x86_64 "
   assertEquals "($ID)" "$expected_result" "${output}"
 
+  ID=7
```
rodrigosiqueira, line 158: We replaced `ID` with `LINENO`.

tests/mk_test.sh
```diff
@@ -155,6 +155,22 @@ function test_mk_build()
...
+  local count=0
+  cp "$original/tests/samples/.config" .config
+  output=$(mk_build 'TEST_MODE' | head -2) # Remove statistics output
+  expected_result=(
+    "make olddefconfig"
```
rodrigosiqueira, line 163: single-quotes

tests/mk_test.sh
```diff
@@ -155,6 +155,22 @@ function test_mk_build()
...
+    "make -j$PARALLEL_CORES ARCH=x86_64"
+  )
+  while read -r f; do
+    if [[ ${expected_result[$count]} != "${f}" ]]; then
+      assertEquals "($ID)" "${expected_result[$count]}" "${f}"
+    fi
+    ((count++))
+  done <<< "$output"
```
rodrigosiqueira, line 171: We have a function for that; take a look at `compare_command_sequence`.

rodrigosiqueira (changes requested):

tests/mk_test.sh
kwy95: Yes, `mk_build` always tries to update the statistics folder when run. Most other tests do something similar in this section, maybe we could make it so the statistics part of the code doesn't trigger while in `TEST_MODE`.

src/mk.sh
kwy95: Oversight, thanks.

tests/mk_test.sh
kwy95: I left the `ID` as a comment, a lot of the tests have not been updated to use `LINENO` and I think a bigger refactor should be done to update these tests.

rodrigosiqueira:
@kwy95 Recently, Joao break the file mk in the build.sh and deploy.sh. Could you update your PR based on the latest code in the unstable branch?

rodrigosiqueira:
@kwy95 there is a bug in your implementation. Try `kw b`, you will see this error:
```
/home/siqueira/.local//lib/kw/kwlib.sh: line 93: SILENT: command not found
```

rodrigosiqueira (changes requested):
Let's suppress this output:
```
#
# configuration written to .config
#
```

rodrigosiqueira:
Also, if we don't have `.config` in the local kernel, we should not run olddefconfig.

kwy95:
OK. I'll suppress the output, but I'm pretty sure The command checks if there is a `.config` file before executing.

rodrigosiqueira:
Merged

PR #341, opened 2021-06-19
shellcheck: address SC2164

### LINKED ISSUE #319: SC2164 - Use cd ... || exit in case cd fails.

kwy95:
Make kw exit if a `cd` command fails, in fixing shellcheck's SC2164
warnings.

Closes #319

### DIFF
--- a/src/checkpatch_wrapper.sh
+++ b/src/checkpatch_wrapper.sh
@@ -76,12 +76,12 @@ function execute_checkpatch()
       continue
     fi
 
-    cd "$kernel_root"
+    cd "$kernel_root" || exit_msg 'It was not possible to move to kernel root dir'
 
     cmd_manager "$flag" "$cmd_script $file"
     [[ "$?" != 0 ]] && say "$SEPARATOR"
 
-    cd "$original_working_dir"
+    cd "$original_working_dir" || exit_msg 'It was not possible to move back from kernel dir'
   done
 }
 
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -356,3 +356,17 @@ function command_exists()
   fi
   return 22 # EINVAL
 }
+
+# This function exits with a custom error message
+#
+# @err The error code to be used on exit, it takes the return code of the
+#        last command executed as default
+# @msg The custom message to be displayed
+function exit_msg()
+{
+  local err=${2:-"$?"}
+  local msg=${1:-'Something went wrong!'}
+
+  complain "$msg"
+  exit "$err"
+}
--- a/src/statistics.sh
+++ b/src/statistics.sh
@@ -300,7 +300,7 @@ function month_statistics()
   fi
 
   current_path=$(pwd)
-  cd "$month_path" || exit
+  cd "$month_path" || exit_msg 'It was not possible to move to month dir'
   shopt -s nullglob
   for day in *; do
     all_file_data=$(cat "$day")
@@ -308,7 +308,7 @@ function month_statistics()
 
     all_data="${all_data}${all_file_data}\n"
   done
-  cd "$current_path" || exit
+  cd "$current_path" || exit_msg 'It was not possible to move back from month dir'
 
   if [[ -z "$all_data" ]]; then
     say "Sorry, kw does not have any record for $month"
--- a/tests/configm_test.sh
+++ b/tests/configm_test.sh
@@ -22,10 +22,16 @@ function setUp()
 {
   local -r current_path="$PWD"
 
-  cd "$SHUNIT_TMPDIR" || fail 'Was not able to move to temporary directory'
+  cd "$SHUNIT_TMPDIR" || {
+    fail "($LINENO) It was not possible to move to temporary directory"
+    return
+  }
   touch .config
   echo "$CONTENT" > .config
-  cd "$current_path"
+  cd "$current_path" || {
+    fail "($LINENO) It was not possible to move back from temp directory"
+    return
+  }
 
   mkdir "$SHUNIT_TMPDIR/configs"
   KW_DATA_DIR="$SHUNIT_TMPDIR"
@@ -76,7 +82,10 @@ function test_save_config_file_check_save_failures()
   local ret=0
 
   # Test without config file -> should fail
-  cd "$SHUNIT_TMPDIR" || fail 'Was not able to move to temporary directory'
+  cd "$SHUNIT_TMPDIR" || {
+    fail "($LINENO) It was not possible to move to temporary directory"
+    return
+  }
   rm -f .config
   ret=$(save_config_file "$NO_FORCE $NAME_1" "$DESCRIPTION_1")
   assert_equals_helper 'No .config file should return ENOENT' "$LINENO" "$?" "2"
@@ -87,7 +96,10 @@ function test_save_config_file_check_save_failures()
   assert_equals_helper "Should return ENOENT, because '.config' != '.configuration'" "$LINENO" "$?" "2"
   rm .configuration
 
-  cd "$current_path"
+  cd "$current_path" || {
+    fail "($LINENO) It was not possible to move back from temp directory"
+    return
+  }
 }
 
 function test_save_config_file_check_directories_creation()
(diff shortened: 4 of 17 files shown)


src/checkpatch_wrapper.sh
```diff
@@ -76,12 +76,12 @@ function execute_checkpatch()
       continue
     fi
 
-    cd "$kernel_root"
+    cd "$kernel_root" || exit
```
rodrigosiqueira, line 79: Add a message that says why it fails, and also exit with a proper exit code.

tests/configm_test.sh
```diff
@@ -43,10 +43,10 @@ function setUp()
   rm -rf "$TMP_TEST_DIR"
 
   mkdir -p "$TMP_TEST_DIR"
-  cd "$TMP_TEST_DIR"
+  cd "$TMP_TEST_DIR" || exit
```
rodrigosiqueira, line 46: In the case of tests, use `fail`. See https://github.com/kworkflow/kworkflow/pull/312/files for a reference

rodrigosiqueira (changes requested):

src/statistics.sh
```diff
@@ -300,15 +300,15 @@ function month_statistics()
   fi
 
   current_path=$(pwd)
-  cd "$month_path" || exit
+  cd "$month_path" || exit || exit
```
magalilemes, line 303: I think one `exit` is enough hehe

alan-barzilay:
@kwy95 the SC2164 error also seems to be triggered on pushd and popd operations, such as in the kw_config_loader_test.sh file. Please amend your commit to deal with this too

src/checkpatch_wrapper.sh
kwy95: Running `exit` this way utilizes the last commands exit code, therefore therefore propagating the error code from the failed `cd` command. The `exit` command doesn't take any arguments apart from a custom exit code, shellcheck suggests the following option for displaying custom messages:
```cd foo || { echo "Failure"; exit 1; }```
I could change to it if that's preferred.

File: tests/configm_test.sh (hunk 2 above)
Comment (joaoseckler) [reply tests/configm_test.sh:46]: @rodrigosiqueira thinking about it again, maybe we should fail and return. If we are `cd`ing into a temporary directory, it means we want to pollute that folder. If we fail to `cd`, we might end up polluting our main directory (or worse, deleting something from it). Exiting is bad since it cancels all other test functions in this file, but canceling one function seems reasonable. What do you think?

Comment (rodrigosiqueira) [reply tests/configm_test.sh:46]: @kwy95

File: src/checkpatch_wrapper.sh (hunk 1 above)
Comment (alan-barzilay) [reply src/checkpatch_wrapper.sh:79]: @rodrigosiqueira, dont know if you missed this last reply from rubens

File: tests/configm_test.sh (hunk 2 above)
Comment (alan-barzilay) [reply tests/configm_test.sh]: failing doesnt lead to automatically returning? @joaoseckler

File: src/kwlib.sh [hunk 4]
```diff
@@ -322,3 +322,19 @@ function command_exists()
...
+# This function exits with a custom error message
+#
+# @err The error code to be used on exit
+# @msg The custom message to be displayed
+function exit_msg()
+{
+  local err="$1"
+  local msg="$2"
+
+  err=${err:-1}
```
alan-barzilay, line 335: why are you taking 1 as the default error code?

tests/configm_test.sh
```diff
@@ -87,17 +96,26 @@ function test_save_config_file_check_save_failures()
   assert_equals_helper "Should return ENOENT, because '.config' != '.configuration'" "$LINENO" "$?" "2"
   rm .configuration
 
-  cd "$current_path"
+  cd "$current_path" || {
+    fail 'Was not able to move back from temp directory'
+    return 2 #ENOENT
```
alan-barzilay, line 101: @kwy95 Isn't there a situation where cd would fail with error code 1 instead of 2?

kwy95: Yes; in fact, it usually does. From the bash reference the builtin functions return exit code of 2 to "indicate incorrect usage", and I haven't been able to produce any error code that isn't 1 from `cd`. I used 2 here based on the output to `errno -l`, that defines 1 and 2 as:
```
EPERM 1 Operation not permitted
ENOENT 2 No such file or directory
```
I think that maybe 1 would be a better option as in bash it's used as a catch-all error, this would also make it consistent with my choice to use 1 as the default of `exit_msg`.
I am open to suggestions of a more appropriate error code if anyone has suggestions.

Further reading:
https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html
https://tldp.org/LDP/abs/html/exitcodes.html

alan-barzilay: wait, the builtin functions do not follow the errno convention? if you cd to a folder that does not exist you don't get the error 2?

>  I am open to suggestions of a more appropriate error code if anyone has suggestions.

I think the best way would be to just propagate the error code, without hardcoding it. Is there some sort of technical challenge for doing this? because that's what you did in the exit_msg function

src/checkpatch_wrapper.sh
```diff
@@ -76,12 +76,12 @@ function execute_checkpatch()
       continue
     fi
 
-    cd "$kernel_root"
+    cd "$kernel_root" || exit_msg "$?" 'Was not able to move to kernel root dir'
```
rodrigosiqueira, line 79: Let's add "It" before "was"

src/kwlib.sh
```diff
@@ -322,3 +322,19 @@ function command_exists()
...
+  msg=${msg:-"Something went wrong!"}
```
rodrigosiqueira, line 336: Sigle-quotes

src/kwlib.sh
rodrigosiqueira: 1 means 'Operation not permitted", which does not make sense here. Maybe 22 is a better exit value.

rodrigosiqueira (changes requested):

tests/configm_test.sh
kwy95: They do not, trying to cd to a folder that doesn't exist just gives you error code 1. The errno convention is apparently target more at C and C++ files, and bash functions don't usually conform. As I said previously for the built-ins error code 2 has a strictly different meaning from the one defined in errno.

I do propagate from outside of this function, I will set the default to `$?` which should give the same result.

kwy95: I overlooked the fact this refers to the tests section. In the case of tests we don't want to exit as discussed above.
Here we can't simply `return $?` as at that point it will be referencing the exit code of the `fail` command. I could add a variable, but it will add a lot of clutter.
Any suggestions are welcome!

joaoseckler (approved):

src/kwlib.sh
```diff
@@ -322,3 +322,16 @@ function command_exists()
...
   return 22 # EINVAL
 }
+
+# This function exits with a custom error message
+#
+# @err The error code to be used on exit
+# @msg The custom message to be displayed
+function exit_msg()
+{
+  local err=${2:-"$?"}
```
rodrigosiqueira, line 332: Why `$?`? Add some comments that make clear your intention here.

tests/configm_test.sh
rodrigosiqueira: @kwy95 maybe we can have something similar to `exit_msg` in the `tests/utils.sh`, but dedicated to tests. Maybe `fail_and_cancel_test`

rodrigosiqueira:
Hi @kwy95 , ignore my previous comment. From our discussion, there is no simple way to address my comment. Please, let's finish this.

alan-barzilay (approved):

tests/signal_manager_test.sh
```diff
@@ -12,7 +12,10 @@ function oneTimeSetUp()
 
 function setUp()
 {
-  cd "$SHUNIT_TMPDIR" || fail "($LINENO): Was not able to cd into temporary directory"
+  cd "$SHUNIT_TMPDIR" || {
+    fail "($LINENO) ($LINENO): It was not able to cd into temporary directory"
```
alan-barzilay, line 16: you used LINENO twice here

alan-barzilay (changes requested):

alan-barzilay (approved):

rodrigosiqueira:
Merged, thanks.

PR #345, opened 2021-06-25
New option to show hardware information

magalilemes:
Here's a draft of `kw device`, which, for now, only supports the `--local` option.

For storing the data, I got inspired by src/statistics.sh, and chose to have an associative array for each pieace of information. It's also called `shared_data`.

Each function for retrieving information behaves mostly the same: their main goal is to update the associative array with their corresponding value.

Two things that I see that could be improved:
1) The way the CPU frequency is stored. As for now, it's a whole string with tabs, letters and numbers:
```
		CPU MHz:                         799.951
		CPU max MHz:                     4900,0000
		CPU min MHz:                     400,0000
```
I think this should be just numbers. Maybe I could add more keys to `shared_data`, and have `cpu_currently`, `cpu_max`, `cpu_min`, and then associate it with the numbers above.

2) Same thing with how the storage devices information is stored. It's also a string, with the device name, and, after a few spaces, the information telling whether it's a rotational disk or not:
```
	nvme0n1   not rotational
```
I haven't thought about another way to handle this, keeping in mind that a machine could have more than one storage disk.

And I also haven't written any tests for this new file.

Any tips on what to change, improve and add?

### DIFF
--- a/src/device_info.sh
+++ b/src/device_info.sh
@@ -1,0 +1,369 @@
+# This file deals with functions related to hardware information
+
+include "$KW_LIB_DIR/remote.sh"
+include "$KW_LIB_DIR/kwlib.sh"
+
+declare -gA device_info_data=(['ram']='' # RAM memory in KB
+  ['cpu_model']=''                       # CPU model vendor
+  ['cpu_currently']=''                   # Current frequency of CPU in MHz
+  ['cpu_max']=''                         # Maximum frequency of CPU in MHz
+  ['cpu_min']=''                         # Minimum frequency of CPU in MHz
+  ['desktop_environment']=''             # Desktop environment
+  ['disk_size']=''                       # Disk size in KB
+  ['root_path']=''                       # Root directory path
+  ['fs_mount']=''                        # Path where root is mounted
+  ['os']=''                              # Operating system name
+  ['motherboard_name']=''                # Motherboard name
+  ['motherboard_vendor']=''              # Motherboard vendor
+  ['chassis']='')                        # Chassis type
+
+declare -gA gpus
+
+declare -gA device_options
+
+# This function calls other functions to process and display the hardware
+# information of a target machine.
+function device_info()
+{
+  local ret
+  device_info_parser "$@"
+
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    return "$ret"
+  fi
+
+  learn_device "${device_options['target']}"
+  show_data
+}
+
+# This function populates the ram element from the device_info_data global
+# variable with the total RAM memory from the target machine in kB.
+#
+# @target Target machine
+function get_ram()
+{
+  local target="$1"
+  local flag="$2"
+  local ram
+  local cmd
+
+  flag=${flag:-'SILENT'}
+  cmd="[ -f '/proc/meminfo' ] && cat /proc/meminfo | grep 'MemTotal' | grep -o '[0-9]*'"
+  case "$target" in
+    2) # LOCAL_TARGET
+      ram=$(cmd_manager "$flag" "$cmd")
+      ;;
+  esac
+
+  if [[ "$flag" == 'TEST_MODE' ]]; then
+    echo "$ram"
+    return 0
+  fi
+
+  device_info_data['ram']="$ram"
+}
+
+# This function provides the model and frequency of the CPU from a machine
+#
+# @target Target machine
+function get_cpu()
+{
+  local target="$1"
+  local flag="$2"
+  local cpu_model
+  local cpu_frequency
+  local cpu_currently
+  local cpu_max
+  local cpu_min
+
+  flag=${flag:-'SILENT'}
+  cmd_frequency="lscpu | grep MHz | sed -r 's/(CPU.*)/\t\t\1/'"
+  cmd_model="lscpu | grep 'Model name:' | sed -r 's/Model name:\s+//g' | cut -d' ' -f1"
+  case "$target" in
+    2) # LOCAL_TARGET
+      cpu_model=$(cmd_manager "$flag" "$cmd_model")
+      cpu_frequency=$(cmd_manager "$flag" "$cmd_frequency")
+      ;;
+  esac
+
+  device_info_data['cpu_model']="$cpu_model"
+
+  if [[ "$flag" == 'TEST_MODE' ]]; then
+    echo "$cpu_model"
+    echo "$cpu_frequency"
+    return 0
+  fi
+
+  cpu_currently=$(echo "$cpu_frequency" | grep 'CPU MHz')
+  cpu_max=$(echo "$cpu_frequency" | grep 'CPU max MHz')
+  cpu_min=$(echo "$cpu_frequency" | grep 'CPU min MHz')
+
+  cpu_currently=${cpu_currently//[!0-9,.]/}
+  cpu_max=${cpu_max//[!0-9,.]/}
+  cpu_min=${cpu_min//[!0-9,.]/}
+
+  device_info_data['cpu_currently']="$cpu_currently"
+  device_info_data['cpu_max']="$cpu_max"
+  device_info_data['cpu_min']="$cpu_min"
+}
+
+# This function populates the values from the size and fs (filesystem) key of
+# the device_info_data variable.
+#
+# @target Target machine
+function get_disk()
+{
+  local target="$1"
+  local flag="$2"
+  local cmd
+  local info
+  local size
+  local mount
+  local fs
+
+  flag=${flag:-'SILENT'}
+  cmd="df -h / | tail -n 1 | tr -s ' '"
+  case "$target" in
+    2) # LOCAL_TARGET
+      info=$(cmd_manager "$flag" "$cmd")
+      ;;
+  esac
+
+  if [[ "$flag" == 'TEST_MODE' ]]; then
+    echo "$info"
+    return 0
+  fi
+
+  fs="$(echo "$info" | cut -d' ' -f1)"
+  size="$(echo "$info" | cut -d' ' -f2)"
+  mount="$(echo "$info" | cut -d' ' -f6)"
+
+  device_info_data['disk_size']="$size"
+  device_info_data['root_path']="$fs"
+  device_info_data['fs_mount']="$mount"
+}
+
+# This function populates the os and desktop environment variables from the
+# device_info_data variable.
+#
+# @target Target machine
+function get_os()
+{
+  local target="$1"
+  local flag="$2"
+  local cmd
+  local os
+  local desktop_env
+
+  flag=${flag:-'SILENT'}
+  cmd="find /usr/share/xsessions -type f -printf '%f ' | sed -r 's/\.desktop//g'"
+  case "$target" in
+    2) # LOCAL_TARGET
+      os=$(detect_distro '/')
+      desktop_env=$(cmd_manager "$flag" "$cmd")
+      ;;
+  esac
+
+  device_info_data['os']="$os"
+  device_info_data['desktop_environment']="$desktop_env"
+}
+
+# This function populates the gpu associative array with the vendor and
+# fetchable memory from each GPU found in the target machine.
+function get_gpu()
+{
+  local target="$1"
+  local flag="$2"
+  local pci_addresses
+  local gpu_info
+  local cmd_pci_address
+
+  flag=${flag:-'SILENT'}
+  # The first thing we want to do is retrieve all PCI addresses from any GPU in
+  # the target machine. After that, we will get, for each GPU, the desired
+  # information.
+  cmd_pci_address="lspci | grep -e VGA -e 3D | cut -d' ' -f1"
+  case "$target" in
+    2) # LOCAL_TARGET
+      pci_addresses=$(cmd_manager "$flag" "$cmd_pci_address")
+      for g in $pci_addresses; do
+        gpu_info=$(cmd_manager "$flag" "lspci -v -s $g")
+        gpu_name=$(echo "$gpu_info" | sed -nr '/Subsystem/s/\s*.*:\s+(.*)/\1/p')
+        gpu_provider=$(echo "$gpu_info" | sed -nr '/controller/s/.+controller: *([^\[\(]+).+/\1/p')
+        gpus["$g"]="$gpu_name;$gpu_provider"
+      done
+      ;;
+  esac
+}
+
+# This function retrieves both the name and vendor from the motherboard of a
+# target machine.
+#
+# @target Target machine
+function get_motherboard()
+{
+  local target="$1"
+  local flag="$2"
+  local mb_name
+  local mb_vendor
+  local cmd_name
+  local cmd_vendor
+
+  flag=${flag:-'SILENT'}
+  cmd_name='[ -f /sys/devices/virtual/dmi/id/board_name ] && cat /sys/devices/virtual/dmi/id/board_name'
+  cmd_vendor='[ -f /sys/devices/virtual/dmi/id/board_vendor ] && cat /sys/devices/virtual/dmi/id/board_vendor'
+  case "$target" in
+    2) # LOCAL_TARGET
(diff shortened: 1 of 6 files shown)


src/device_info.sh
```diff
@@ -0,0 +1,173 @@
...
+  local ram
+  local cmd
+
+  case "$target" in
+    1) # VM_TARGET
+      echo TODO
+      ;;
+    2) # LOCAL_TARGET
+      cmd="cat /proc/meminfo | grep 'MemTotal' | grep -o '[0-9]*'"
+      ram=$(cmd_manager "SILENT" "$cmd")
```
rodrigosiqueira, line 24: How about converting the memory size to G? Maybe you can have both information, kB and G.

src/device_info.sh
```diff
@@ -0,0 +1,173 @@
...
+  local cmd
+
+  case "$target" in
+    1) # VM_TARGET
+      echo TODO
+      ;;
+    2) # LOCAL_TARGET
+      cmd="lscpu | grep 'Model name:' | sed -r 's/Model name:\s+//g'"
+      cpu_model=$(cmd_manager 'SILENT' "$cmd")
+      cmd="lscpu | grep MHz | sed -r 's/(CPU.*)/\t\t\1/'"
```
rodrigosiqueira, line 50: Hmmm... I'm not too fond of the idea of storing all of these CPU data in a single variable. It will be better if you have a key per option and only save the raw value.

src/device_info.sh
```diff
@@ -0,0 +1,173 @@
...
+      ;;
+    3) # REMOTE
+      echo TODO
+      ;;
+  esac
+
+  shared_data['cpu_model']="$cpu_model"
+  shared_data['cpu_frequency']="$cpu_frequency"
+}
+
```
rodrigosiqueira, line 61: Add documentation

src/device_info.sh
```diff
@@ -0,0 +1,173 @@
...
+  local disks
+  local cmd
+
+  case "$target" in
+    1) # VM_TARGET
+      echo TODO
+      ;;
+    2) # LOCAL_TARGET
+      cmd="lsblk -d -o name,rota | grep -v loop | tail -n +2 |"
+      cmd+="sed -r 's/(.*)\s+1/\t\1rotational/;s/(.*)\s+0/\t\1not rotational/'"
```
rodrigosiqueira, line 74: Maybe for the disk, we just want to know the type (sda, nvme, etc), the size, and maybe the vendor.

src/device_info.sh
```diff
@@ -0,0 +1,173 @@
...
+  local target="$1"
+  local os
+  local cmd
+
+  case "$target" in
+    1) # VM_TARGET
+      echo TODO
+      ;;
+    2) # LOCAL_TARGET
+      os=$(hostnamectl | grep 'Operating System' | sed -r 's/.*:\s(.*)/\1/')
```
rodrigosiqueira, line 96: That's a nice approach, but we can get the OS information without external tools. Actually, we already have it :)
See `which_distro` and `detect_distro`, maybe we can move some of this function to your new file.

src/device_info.sh
```diff
@@ -0,0 +1,173 @@
...
+  get_cpu "$target"
+  get_disks "$target"
+  get_os "$target"
+}
+
+# This function shows the information stored in the shared_data variable.
+function display_data()
+{
+  say 'CPU:'
+  printf '\tName: %s\n' "${shared_data['cpu_model']}"
```
rodrigosiqueira, line 124: I prefer one or two spaces instead of tabs.

rodrigosiqueira (changes requested):
Hi Magali,

First of all, thanks a lot for your patch; it looks really nice, and you are on the right path. You can see some of my comments inline. Some generic comments:

1. I noticed that you added code to deal with local, remote, and VM. Check if you can do it in a way that you do not duplicate code in multiple parts.
2. Remember that we want to register some of the hardware data and associate it with the config information. Try to keep this in mind in your design.
3. Maybe you can consider local as the default option. You can also rely on the default_deploy_target in the kworfklow.config file. I think the command line option should have precedence over the config file.

alan-barzilay:
what sort of information do you wish to obtain and what for? 
I think this is a really useful feature but it is not clear to me if this is just a convenient function to automate system description for the user or if this will be used in our pipeline to auto configure some functions. If this is the latter, maybe information such as CPU frequency is not very useful.
Also, how about adding the system architecture?

alan-barzilay:
Ah! I don't know if this is relevant right now, but it may be useful to determine a raspberry model with this command. One way of doing this is with `/sys/firmware/devicetree/base/model`:

> pi@raspberrypi:~ $ cat /sys/firmware/devicetree/base/model
> Raspberry Pi 3 Model B Rev 1.2

magalilemes:
@alan-barzilay 
> what sort of information do you wish to obtain and what for?

Well, pretty much everything that is mentioned in the files and conversation above (type of disk, cpu info, operating system, ram...). We want to use this information to somehow associate it with the time that it takes the kernel to compile and install (still something to think about, though). You could also use this tool to quickly check information of a machine you don't know (for example, imagine you use kw device --remote)

src/device_info.sh
joaoseckler: Maybe [`numfmt`](https://www.gnu.org/software/coreutils/manual/html_node/numfmt-invocation.html) can help here!

rodrigosiqueira: uouuu... numfmt looks really nice!

rodrigosiqueira:
@joaoseckler @alan-barzilay, we can also use this file as a central point for handling device information. This can be useful for deploy to a different set of devices.

magalilemes:
@rodrigosiqueira kw device now has remote and vm option. One remark: haven't added tests yet. When you have time, could you please review this new series?

src/device_info.sh
```diff
@@ -0,0 +1,186 @@
...
+      exit 0
+      ;;
+    *)
+      complain "Invalid option: $option"
+      return 22 # EINVAL
+      ;;
+  esac
+
+  process_all_data "$target"
+  display_data
```
rodrigosiqueira: I would prefer that we separate the parser from the function that manages the action. In other words, create a new function dedicated to parser command-line options (e.g., `device_info_parser`) and keep this function as an entry point for `device_info`. You probably want to put this as the first function in this file.

src/device_info.sh
```diff
@@ -0,0 +1,186 @@
+# This file deals with functions related to hardware information
+
+include "$KW_LIB_DIR/remote.sh"
+include "$KW_LIB_DIR/kwlib.sh"
+
+declare -gA shared_data=(['ram']='' ['cpu_model']='' ['cpu_currently']=''
```
rodrigosiqueira, line 7: We will want to use the functions provided here in the deploy and maybe other features, for this reason, rename this to something like `device_info_data`.

src/device_info.sh
```diff
@@ -0,0 +1,186 @@
...
+  esac
+
+  shared_data['os']="$os"
+}
+
+# This function calls other functions to populate the shared_data variable with
+# the data related to the hardware from the target machine.
+#
+# @target Target machine
+function process_all_data()
```
rodrigosiqueira, line 168: Inspired by ChromeOS scripts, let call this function `learn_device()` or similar.
See: https://chromium.googlesource.com/chromiumos/platform/crosutils/+/master/update_kernel.sh#67

rodrigosiqueira (changes requested):
Great PR! I just had a few comments. Please, add tests and documentation, and let's finish this first part :)
Next:
- Save part of this data in a way that we can associate with statistics.
- Could you take a look at this script [1] and check if we can add some extra information to your series?
https://chromium.googlesource.com/chromiumos/platform/crosutils/+/master/update_kernel.sh

magalilemes:
@rodrigosiqueira I now added the possibility to retrieve GPU information, containing the vendor and the total of fetchable memory.

src/device_info.sh
```diff
@@ -0,0 +1,261 @@
...
+      ;;
+    3) # REMOTE_TARGET
+      info=$(cmd_remotely "$cmd" "$flag" "$ip" "$port")
+      ;;
+  esac
+
+  for g in $info; do
+    gpu_info=$(lspci -v -s "$g")
+    vendor=$(echo "$gpu_info" | head -n 1 | sed -r 's/.+controller: ([^ ]+).+/\1/')
+    memory=$(echo "$gpu_info" | sed -n -r '/Memory.+, prefetchable/s/.+\[size=(.+)\]/\1/p' | tr '\n' ',' | sed 's/.$//')
```
rodrigosiqueira, line 214: When I use `kw device` to my test machine, I get:
```
Vendor: 03:00.0 PCI bridge: Advanced Micro Devices
Prefetchable memory:  Inc. [AMD] 300 Series Chipset PCIe Port (rev 02) (prog-if 00 [Normal decode]),
```
I think the above information is not super useful. When you run the command:
```
lspci  -v -s $(lspci | grep ' VGA ' | cut -d" " -f 1)
```
You get a lot of interesting info, one of them is the "Subsystem" since it provides the GPU name and "VGA compatible controller" which gives the GPU provider. Let's collect this info.

src/device_info.sh
```diff
@@ -0,0 +1,261 @@
...
+    gpus["$g"]="$vendor,$memory"
+  done
+}
+
+# This function calls other functions to populate the device_info_data variable
+# with the data related to the hardware from the target machine.
+#
+# @target Target machine
+function learn_device()
```
rodrigosiqueira, line 241: This function depends on `device_info`, which is not ideal. I want to use this function in the deploy context, rewrite it in a way that I can use `include devices.sh` in the `deploy.sh`, and just call this function to collect that hardware data that I need.

src/device_info.sh
```diff
@@ -0,0 +1,261 @@
...
+{
+  get_ram 'SILENT'
+  get_cpu 'SILENT'
+  get_disk 'SILENT'
+  get_os 'SILENT'
+  get_gpu 'SILENT'
+}
+
+# This function shows the information stored in the device_info_data variable.
+function display_data()
```
rodrigosiqueira, line 274: Let's call this function `show_data`

src/device_info.sh
```diff
@@ -0,0 +1,261 @@
...
+{
+  say 'CPU:'
+  printf '  Model: %s\n' "${device_info_data['cpu_model']}"
```
rodrigosiqueira, line 292: new line

src/device_info.sh
```diff
@@ -0,0 +1,261 @@
...
+  if [[ -n "${device_info_data['cpu_currently']}" ]]; then
+    printf '  Current frequency (MHz): %s\n' "${device_info_data['cpu_currently']}"
+  fi
```
rodrigosiqueira, line 295: new line between these `ifs`

rodrigosiqueira (changes requested):
Nice series! I added a few comments.

src/device_info.sh
```diff
@@ -0,0 +1,261 @@
+# This file deals with functions related to hardware information
+
+include "$KW_LIB_DIR/remote.sh"
+include "$KW_LIB_DIR/kwlib.sh"
+
+declare -gA device_info_data=(['ram']='' ['cpu_model']='' ['cpu_currently']=''
```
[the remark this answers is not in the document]
magalilemes: Just to make it clear that this isn't a static value, as it stores the `current` frequency of the CPU at the time the command was invoked.

src/device_info.sh
magalilemes: Oops, I may have messed something up. The output was supposed to be:
```
GPU:
  Vendor: Intel
  Prefetchable memory: 256M
  Vendor: NVIDIA
  Prefetchable memory: 256M,32M
```
I'll make the changes to show GPU name and provider 😄

src/device_info.sh
```diff
@@ -0,0 +1,413 @@
...
+  port=${port:-"${device_options['port']}"}
+  flag=${flag:-'SILENT'}
+  cmd="lspci | grep -e VGA -e 3D | cut -d' ' -f1"
+  case "$target" in
+    2) # LOCAL_TARGET
+      pci_addresses=$(cmd_manager "$flag" "$cmd")
+      for g in $pci_addresses; do
+        gpu_info=$(cmd_manager "$flag" "lspci -v -s $g")
+        gpu_name=$(echo "$gpu_info" | sed -nr '/Subsystem/s/\s*.*:\s+(.*)/\1/p')
+        gpu_provider=$(echo "$gpu_info" | sed -nr '/controller/s/.+controller: *([^ ]+).+/\1/p')
```
rodrigosiqueira, line 235: You need to update your regex, in my case, I'm your function return `Advanced instead of ` Advanced Micro Devices`.

src/device_info.sh
```diff
@@ -0,0 +1,261 @@
...
+# fetchable memory from each GPU found in the target machine.
+function get_gpu()
+{
+  local target="${device_options['target']}"
+  local flag="$1"
+  local cmd
+  local info
+  local gpu_info
+
+  cmd="lspci | grep -e VGA -e 3D | cut -d' ' -f1"
```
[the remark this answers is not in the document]
rodrigosiqueira: I think you can use `cmd_pci_address` or similar to make this part of the code easy to read. Maybe add two or three phrases that explain your intention here.

rodrigosiqueira (changes requested):
Hi @magalilemes ,
You can see a few comments in the PR. I also have one final request, can we take the chassis and motherboard info as well?

For the motherboard, you can use:
`cat /sys/devices/virtual/dmi/id/board_{vendor,name}`
And for the chassis:
`cat /sys/devices/virtual/dmi/id/chassis_type`
However, you will need to parse this number to a string, here is the meaning for each one:

```
1 Other
2 Unknown
3 Desktop
4 Low Profile Desktop
5 Pizza Box
6 Mini Tower
7 Tower
8 Portable
9 Laptop
10 Notebook
11 Hand Held
12 Docking Station
13 All in One
14 Sub Notebook
15 Space-Saving
16 Lunch Box
17 Main System Chassis
18 Expansion Chassis
19 SubChassis
20 Bus Expansion Chassis
21 Peripheral Chassis
22 Storage Chassis
23 Rack Mount Chassis
24 Sealed-Case PC
```
Reference:
https://superuser.com/questions/877677/programatically-determine-if-an-script-is-being-executed-on-laptop-or-desktop/877796

rodrigosiqueira:
@alan-barzilay could you test it in your raspberry pi?

rodrigosiqueira:
Actually, after the discussion in https://github.com/kworkflow/kworkflow/pull/355 , how about add the Desktop Environment info? For example, under the `Operating System` section we could have:

```
Operating System:
Distribution: ubuntu
Desktop Environment: Gnome (KDE, etc)
Compositor: Wayland (or X)
```

The challenge here is to find a generic way to discover these info.

magalilemes:
> Actually, after the discussion in #355 , how about add the Desktop Environment info? For example, under the `Operating System` section we could have:
> 
> ```
> Operating System:
> Distribution: ubuntu
> Desktop Environment: Gnome (KDE, etc)
> Compositor: Wayland (or X)
> ```
> 
> The challenge here is to find a generic way to discover these info.

@rodrigosiqueira 
For a local machine, I believe `echo $XDG_CURRENT_DESKTOP` and `echo $XDG_SESSION_TYPE` would be good options to finding the desktop environment and compositor, but I still don't know what commands to use to get those remotely.

rodrigosiqueira:
> For a local machine, I believe echo $XDG_CURRENT_DESKTOP and echo $XDG_SESSION_TYPE would be good options to finding the desktop environment and compositor, but I still don't know what commands to use to get those remotely.

@magalilemes afaik, this will not work if you ssh to the machine (try it). You might need to find a more generic way...

rodrigosiqueira:
Hi @magalilemes ,
After thinking a lot about this PR, I realize that we already have more than enough. You did an amazing working here, and I decided to merge this change. We can keep upating it.

Thanks a lot

PR #346, opened 2021-06-25
Fix all occurrences of SC2086

magalilemes:
This PR pretty much adds double quotes around variables. The biggest change happened in the `src/vm.sh` file, more specifically, in the `vm_up` function, where I chose to call `cmd_manager`.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -165,13 +165,14 @@ function strip_path()
   local base
   TESTS=()
   for file in "$@"; do
-    base=$(basename ${file})
+    base=$(basename "${file}")
     TESTS+=("${base%.*}")
   done
 }
 
 check_required_files
 check_files="$?"
+#shellcheck disable=SC2086
 if [[ "$#" -eq 0 ]]; then
   check_required_files
   files_list=$(find ./tests -name "*_test.sh" | grep --invert-match "/shunit2/")
--- a/src/bash_autocomplete.sh
+++ b/src/bash_autocomplete.sh
@@ -11,7 +11,7 @@ function _kw_autocomplete()
 
   # By default, autocomplete with kw_options
   if [[ ${previous_command} == kw ]]; then
-    COMPREPLY=($(compgen -W "${kw_options}" -- ${current_command}))
+    COMPREPLY=($(compgen -W "${kw_options}" -- "${current_command}"))
     return 0
   fi
 
--- a/src/checkpatch_wrapper.sh
+++ b/src/checkpatch_wrapper.sh
@@ -37,13 +37,13 @@ function execute_checkpatch()
   fi
 
   # Get realpath for using inside checkpatch
-  FILE_OR_DIR_CHECK="$(realpath $FILE_OR_DIR_CHECK)"
+  FILE_OR_DIR_CHECK="$(realpath "$FILE_OR_DIR_CHECK")"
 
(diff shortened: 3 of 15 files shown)


alan-barzilay (approved):
LGTM!
just a quick question, we have 3 unrelated shellcheck failures in this PR, since you have been looking into these particular lines of code do you have any insights on them? I have been thinking about this one regarding echo for a while, it's true that we don't need sed but I think it is more readable, also I feel like no one knows that you can do this with echo. Maybe we should add it to the global ignore list

alan-barzilay:
@magalilemes there is a related error:
 [SC2231](https://github.com/koalaman/shellcheck/wiki/SC2231) - `Quote expansions in this for loop glob to prevent wordsplitting, e.g. "$dir"/*.txt .`
Maybe this could be a second commit in this PR? I think there are only a few occurrences of this.
Or we can open another issue for this and make another PR in the future

magalilemes:
> LGTM!
> just a quick question, we have 3 unrelated shellcheck failures in this PR, since you have been looking into these particular lines of code do you have any insights on them?

1) In the **command_exists** function from `./src/kwlib.sh`, there are two errors. We could slightly improve this function, by replacing the local variables with
```
  local command=( $1 )
  local package="${command[0]}"
```
instead. I think this makes it more readable and also gets us rid of the two errors. On the other hand, we do get a new false positive error, because there are no double quotes around `$1` and we need it so. We want word expansion here. 
We could disable SC2206 for this function, and we're done.

2) In `./src/bash_autocomplete.sh`, we could replace
`    COMPREPLY=($(compgen -W "${kw_options}" -- "${current_command}"))`
with
`    COMPREPLY=( "$(compgen -W "${kw_options}" -- "${current_command}")" )`
I could do this in this PR, since it's related to double quotes and stuff.

> I have been thinking about this one regarding echo for a while, it's true that we don't need sed but I think it is more readable, also I feel like no one knows that you can do this with echo. Maybe we should add it to the global ignore list

I agree we should ignore this error globally. For this specific error in `./src/kwio.sh`, I think shellcheck gives a nice - although small - tip, actually. We could, indeed, replace
`opts="$(echo "$ALERT_OPT" | sed s/--alert=//)"`
with
`opts="$(echo "${ALERT_OPT//--alert=/}")"`
and save 6 characters 🙃

magalilemes:
> @magalilemes there is a related error:
> [SC2231](https://github.com/koalaman/shellcheck/wiki/SC2231) - `Quote expansions in this for loop glob to prevent wordsplitting, e.g. "$dir"/*.txt .`
> Maybe this could be a second commit in this PR? I think there are only a few occurrences of this.
> Or we can open another issue for this and make another PR in the future

I ended up fixing it in this PR. I found two occurrences of this error, and they just required me to use double quotes around the variables. It seemed similar enough with the rest of the PR, so I fixed these two errors as well.

alan-barzilay:
> I ended up fixing it in this PR. I found two occurrences of this error, and they just required me to use double quotes around the variables. It seemed similar enough with the rest of the PR, so I fixed these two errors as well.

Nice! Maybe you should add to your commit message the shellcheck error code for both of them, just for completeness sake and future reference

rodrigosiqueira:
Since Alan gave his R-o-b, and everything looks ok to me, I decided to merge it to the unstable branch. Thanks

Issue #245, opened 2021-04-25
Add kw man per feature


rodrigosiqueira:
We have `kw [feature] -h`, which provides a command summary. We need to add more details per feature, we can expand this feature to `kw [feature] --help` and display a detailed man per feature.

magalilemes:
I don't know if this relates to this issue, but could we try to standardize the output from `kw [feature] -h`? Compare, for instance, the output from `kw drm -h`, `kw c -h`, and `kw b -h`; they all have different styles.

rodrigosiqueira:
> I don't know if this relates to this issue, but could we try to standardize the output from `kw [feature] -h`? Compare, for instance, the output from `kw drm -h`, `kw c -h`, and `kw b -h`; they all have different styles.

I agree, but I think this should be another issue. Could you create it?

rodrigosiqueira:
Done! Thanks

PR #348, opened 2021-06-27
Man per feature

kwy95:
Man pages per feature, still a WIP but I need help reviewing what I've already done.
Right now it's possible to to call `kw man feature` or `kw feature --help` to get a man page describing the feature and its usage. I'll soon change the main man page to reference the specific man pages instead of explaining all of them.


rodrigosiqueira (changes requested):
Hi @kwy95 ,

I saw your attempt, and that's not what I meant about man per feature. From https://github.com/kworkflow/kworkflow/issues/245:

We have kw [feature] -h, which provides a command summary. We need to add more details per feature. We can expand this feature to kw [feature] --help and display a detailed man per feature.

Let's take a look at `kw b` as a reference:

```
kw b -h
```
The above command shows a simple summary, which is fine. Your goal is implement:
```
kw b --help
```
And after this command, we should open a "man" doc for "kw b". Take a look at the function `kworkflow_man` as a reference. Anyway, you will need:

1. Make --help per feature open a specific man doc
2. Break documentation/man/kw in different files

rodrigosiqueira:
@kwy95 any update here?

src/vm.sh
```diff
@@ -3,6 +3,10 @@ include "$KW_LIB_DIR/kwlib.sh"
 
 function vm_mount()
 {
+  if [[ "$1" =~ -h|--help ]]; then
+    vm_help "$1"
+    exit 0
+  fi
```
rodrigosiqueira, line 9: Why did you add this here? This does not look nice... the top of the function should be used to declare variables. If necessary, refactor this function to suit your goals better.
Ps.: Same for the parts that you have this.

src/vm.sh
```diff
@@ -75,3 +87,16 @@ function vm_up()
...
+{
+  if [[ "$1" == --help ]]; then
+    include "$KW_LIB_DIR/help.sh"
+    kworkflow_man 'features/vm'
+  else
+    echo -e "kw <mount|umount|up>:\n" \
+      "  mo | mount - Mount VM\n" \
+      "  um | umount - Unmount VM\n" \
+      "  u | up - Start VM"
+  fi
```
rodrigosiqueira, line 101: Please, follow the same pattern used in the build feature.

documentation/man/features/build.rst
```diff
@@ -0,0 +1,38 @@
...
+    The menu option invokes the kernel menuconfig. Notice that the default menu
+    config can be changed in the `kworkflow.config` file by setting a different
+    option in *menu_config*. If the user is working in a *cross-compile*
+    environment, it is recommended to use this option to avoid messing with the
+    config file manually.
+
+-d, \--doc:
+    The doc option provides a mechanism for build kernel-doc; by default, it
+    will build htmldocs. Users can change the default documentation output by
+    changing the parameter *doc_type* in the `kworkflow.config` file.
```
rodrigosiqueira, line 38: In the kw.rst, we have a few examples of this feature. Please, add an example section here.

documentation/man/features/configm.rst
```diff
@@ -0,0 +1,46 @@
...
+
+\--get <name> [-f]:
+    Get a config file based on the <name> and paste it in the current
+    directory. It pop-up a warning message because this operation override the
+    current .config file. The user can suppress this warning by using **-f** flag.
+
+-rm, \--remove <name> [-f]:
+    Remove config labeled with <name>. It pop-up a warning message because it
+    will remove the config file from kw management. The user can suppress this
+    warning by using **-f**.
```
rodrigosiqueira, line 46: configm has some examples, please, add them here.

rodrigosiqueira (changes requested):
Hi @kwy95 ,
I think you are on the right path. You can see some of my comments in each commit, but I also have these generic points:

* Use the same pattern for each man page. Every man you handle it differently.
* You forgot these options:
```
kw diff --help
kw c --help
kw drm --help
kw explore --help
kw maintainers --help
kw Pomodoro --help
kw ssh --help
kw statistics --help
kw vars --help
```
* Add unit test for `kworkflow_man`

Thanks

rodrigosiqueira (changes requested):
@kwy95 this new set looks great!
It looks like that you forgot to create the main page, i.e., kw man.

@magalilemes @joaoseckler @alan-barzilay Please, take some time to review this PR.

documentation/man/features/deploy.rst
```diff
@@ -0,0 +1,105 @@
...
+-----------------
+deploy the kernel
+-----------------
+
+SYNOPSIS
+========
+*kw* (*d* | *deploy*) [\--remote [<remote>:<port>] | \--local | \--vm]
+                      [-r | \--reboot] [-m | \--modules] [-s | \--ls-line]
+                      [-l | \--list] [(-u | \--uninstall) <kernel-name>[,...]]
+
```
rodrigosiqueira, line 16: Why are we using this pattern here? I prefer the one that you use in the configm.

documentation/man/features/deploy.rst
```diff
@@ -0,0 +1,105 @@
...
+2. Send all the files to the target machine; and
+
+3. Execute the operations that will update the target machine.
+
+You can specify the deploy target via command line by using the flag
+**\--remote [<remote>:<port>]** (e.g., **\--remote 172.16.254.1:22**); however,
+if you do it frequently you probably will prefer to add this information in
+your local **kworkflow.config**. See the example below::
+
+  default_deploy_target=remote
```
rodrigosiqueira, line 35: Add ssh_user as well, we update this documentation recently.

rodrigosiqueira (changes requested):
Hi @kwy95 ,
* You need to validate invalid parameters; for example, `kw man lalala` should not work (add tests for that).
* Create a specific page for `kw man`, something similar to the git-man. See https://git-scm.com/docs/git and notice this section https://git-scm.com/docs/git#_main_porcelain_commands
* Let's use the same Synopsis pattern used in the configm

documentation/man/features/deploy.rst
kwy95: So the idea is to show the different ways to use the commands, and in the case of `configm` the options are not supposed to used at the same time, and here it is. I agree it looks kind of messy, and looking at some of the man pages for git it sometimes uses the `git <feature> [<options>]` pattern to simplify things, I'm thinking we should probably use it for some of our commands as well.

rodrigosiqueira:
@kwy95 let me know when you have a new version.

src/help.sh
```diff
@@ -38,21 +39,16 @@ function kworkflow_help()
 # installed to the system
 function kworkflow_man()
 {
+  feature="$1"
```
rodrigosiqueira, line 42: fwiu, `feature` is a path and you need to check it. Otherwise, we will get errors like this one:

```
kw man lala
man: /home/siqueira/.local/share/man/lala.1: No such file or directory
```

documentation/man/kw.rst
```diff
@@ -17,14 +17,14 @@ Inglorious kernel developer workflow scripts
 
 SYNOPSIS
 ========
-**kw** *COMMAND* [*OPTION* ...]
+*kw* *<command>* [<option> ...]
 
 DESCRIPTION
 ===========
 **kw** is a set of scripts that have a single mission: reduce the overhead
```
rodrigosiqueira, line 24: Since we are here, let replace `**kw** is a set of scripts that have a single mission` to `**kw** mission is`

documentation/man/kw.rst
```diff
@@ -70,136 +57,22 @@ provide the *remote* option, which is much more flexible since it uses network;
 notice that this approach is the most generic one because you can use it for
 *vm* and *local*.
 
+  :ref:`kw-deploy<deploy-doc>`
+
 .. note::
   **Currently, we don't support the Kernel image update in the --vm option.
   However, you can use the remote option for a workaround this issue**.
```
rodrigosiqueira, line 64: That's not true anymore. Could you drop this note?

rodrigosiqueira (changes requested):
Hi @kwy95 ,
I just added a few comments; you're very close to complete this PR.

documentation/man/features/deploy.rst
```diff
@@ -0,0 +1,112 @@
...
+already configured.
+
+First, if you are working in a specific kernel module, and if you want to
+install your recent changes in your VM you can use::
+
+  cd </kernel/path>
+  kw d --vm --modules
+
+.. note::
+  Turn off your VM before use the **intall** command.
```
magalilemes, line 106: install

rodrigosiqueira:
Hi @kwy95 , there is one test failure :(

./tests/build_test.sh

Try to run all tests on your side.

documentation/man/features/deploy.rst
kwy95: nice, thanks!

rodrigosiqueira:
Hi @kwy95 , thanks a lot for this change! That's a great improvement for kw!

PR #350, opened 2021-06-27
Add scripts/kwreview.sh

joaoseckler:
This script aids kw's developers in running shfmt and shellcheck. It
uses reviewdog in order to check only for warnings and errors in the
curent patch (compared to branch unstable by default). It also updates
the "Development workflow" documentation to show the usage of this
script.

### DIFF
--- a/documentation/content/developmentworkflow.rst
+++ b/documentation/content/developmentworkflow.rst
@@ -120,3 +120,31 @@ To configure this Git pre-commit hook:
 
 The next time you try to commit your work, your pre-commit hook will run both
 shellcheck and shfmt, and warn you of any errors you may have made.
+
+Kwreview
+========
+Another way to have your code checked against ``shfmt`` and ``shellcheck`` is to
+use ``scripts/kwreview.sh``. By default, it will check your current patch (i.e.
+what has changed since the branch unstable) and print ``shellcheck``'s warnings
+and ``shfmt``'s warnings as diffs. Use ``kwreview -w`` to apply ``shfmt``'s
+changes to the files. Make sure you have ``shfmt``, ``shellcheck`` and
+``reviewdog`` installed.
+
+For vim users, it is possible to use ``kwreview.sh`` to populate the
+`quickfix <http://vimdoc.sourceforge.net/htmldoc/quickfix.html>`_ (or locations)
+window, making it easy to navigate the warnings and errors. As an example,
+consider adding the following lines to your ``.vimrc``::
+
+  function Kwreview(...)
+    let arg = get(a:, 1, "")
+    let &l:makeprg="scripts/kwreview.sh " . arg
+    make
+    execute ":e"
+  endfunction
+
+  command -nargs=* Kwreview :call Kwreview(<q-args>)
+
+Now you can call ``kwreview.sh`` with the vim command ``:Kwreview``
+(notice the upper case initial) and navigate the errors/warnings with
+``:copen``, ``:cn``, ``:cp``, etc.
+
--- a/documentation/content/developmentworkflow.rst
+++ b/documentation/content/developmentworkflow.rst
@@ -123,12 +123,14 @@ shellcheck and shfmt, and warn you of any errors you may have made.
 
 Kwreview
 ========
+
 Another way to have your code checked against ``shfmt`` and ``shellcheck`` is to
 use ``scripts/kwreview.sh``. By default, it will check your current patch (i.e.
 what has changed since the branch unstable) and print ``shellcheck``'s warnings
 and ``shfmt``'s warnings as diffs. Use ``kwreview -w`` to apply ``shfmt``'s
-changes to the files. Make sure you have ``shfmt``, ``shellcheck`` and
-``reviewdog`` installed.
+changes to the files. If you supply it with path to shell files, it will analyse
+those. Make sure you have ``shfmt``, ``shellcheck`` and ``reviewdog`` installed
+(see :ref:`dependencies_label`).
 
 For vim users, it is possible to use ``kwreview.sh`` to populate the
 `quickfix <http://vimdoc.sourceforge.net/htmldoc/quickfix.html>`_ (or locations)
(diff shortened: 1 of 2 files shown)


documentation/content/developmentworkflow.rst
```diff
@@ -120,3 +120,29 @@ To configure this Git pre-commit hook:
 
 The next time you try to commit your work, your pre-commit hook will run both
 shellcheck and shfmt, and warn you of any errors you may have made.
+
+Kwreview
+========
+Another way to have your code checked against ``shfmt`` and ``shellcheck`` is to
+use ``scripts/kwreview.sh``. By default, it will check you current patch (i.e. what
```
alan-barzilay, line 127: your* current patch

alan-barzilay (changes requested):

rodrigosiqueira:
Hi @joaoseckler ,
This PR looks great!
By the way, I would like to see an explanation of how to add this to git hooks, e.g., `.git/hooks/pre-commit.sh`. Also, could you organize your code in a way that uses functions and better describes each part's intention?

joaoseckler:
The last update of this PR organizes the code into functions.

joaoseckler:
> This PR looks great!

Thanks!

> By the way, I would like to see an explanation of how to add this to git hooks, e.g., `.git/hooks/pre-commit.sh.`

I've come up with this hook. Do you think it's worth to adding it to the documentation?
```bash
#!/bin/bash

output="$(./scripts/kwreview.sh)"
if [ "$output" != "" ]; then
  echo "$output"
  echo
  echo "Please fix the errors above before committing."
  exit 1
else
  exit 0
fi
```

joaoseckler:
Future work on this PR includes:
 - Documenting functions
 - Using appropriate returnr values (instead of only 1 and 0)

documentation/content/developmentworkflow.rst
```diff
@@ -120,3 +120,29 @@ To configure this Git pre-commit hook:
...
+    let &l:makeprg="scripts/kwreview.sh " . arg
+    make
+    execute ":e"
+  endfunction
+
+  command -nargs=* Kwreview :call Kwreview(<q-args>)
+
+Now you can call ``kwreview.sh`` with the vim command ``:Kwreview``
+(notice the upper case initial) and navigate the errors/warnings with
+``:copen``, ``:cn``, ``:cp``, etc.
```
rodrigosiqueira, line 148: It looks like that your script depends on reviewdog. You need to highlight this information and tell how to install it on Debian and Arch.

rodrigosiqueira: Maybe we can add a new option under `setup -i` where we install development dependencies.

scripts/kwreview.sh
```diff
@@ -0,0 +1,214 @@
...
+
+declare -gA opt
+declare -ga changed_files
+
+init_options()
+{
+  opt['branch']='unstable'
+  opt['path']='.'
+  opt['exclude']=''
+  opt['shellcheck_exclude_default']='SC2016,SC2181,SC2034,SC2154'
```
rodrigosiqueira, line 11: how about reading this from the `.travis` file?

scripts/kwreview.sh
```diff
@@ -0,0 +1,214 @@
...
+  long_options+='all,shfmt-inplace,list,help,'
+  long_options+='shfmt-only,shellcheck-only'
+
+  options=$(getopt --name "$(basename "$0")" \
+    --options "$short_options" \
+    --longoptions "$long_options" \
+    -- "$@")
+
+  # shellcheck disable=2086
+  # see man getopt
```
rodrigosiqueira, line 33: which section?

scripts/kwreview.sh
```diff
@@ -0,0 +1,214 @@
...
+  eval set --$options
+
+  while [[ $# -gt 0 ]]; do
```
rodrigosiqueira, line 36: **"**#**"**

scripts/kwreview.sh
```diff
@@ -0,0 +1,214 @@
...
+          opt['exclude']=':^'"$2"
+        fi
+        shift 2
+        ;;
+      --shellcheck-exclude | -s)
+        opt['shellcheck_exclude']="$2"
+        shift 2
+        ;;
+      --all | -a)
+        opt['shellcheck_exclude']=""
```
rodrigosiqueira, line 57: '' instead of ""

scripts/kwreview.sh
```diff
@@ -0,0 +1,214 @@
...
+  fi
+}
+
+function run_shfmt()
+{
+  local shellcheck_only="${opt['shellcheck_only']}"
+  local shfmt_inplace="${opt['shfmt_inplace']}"
+  local branch="${opt['branch']}"
+
+  if [ "$shellcheck_only" = 'FALSE' ]; then
```
rodrigosiqueira, line 172: can we usee `[[` and `==` here?

rodrigosiqueira (changes requested):
In general, it looks really good. However, it lacks some information on how to make it work, and I also felt that you used some different code styles here (especially in the `if` conditions).

joaoseckler:
The last update to this PR improves the documentation, accepts siqueira's suggestions and makes use of better exit codes.

documentation/content/developmentworkflow.rst
```diff
@@ -120,3 +120,74 @@ To configure this Git pre-commit hook:
...
+With go installed, run the following command to install shfmt::
+
+  GO111MODULE=on go get mvdan.cc/sh/v3/cmd/shfmt
+
+More information can be found in their `github page <https://github.com/mvdan/sh>`__.
+
+Installing pre-commit
+~~~~~~~~~~~~~~~~~~~~~
+``pre-commit`` is available as a package for most distributions. On debian based
+distros, use ``sudo apt install pre-commit``. On Arch based distros, use ``sudo
```
rodrigosiqueira, line 182: It looks like that we don't have this pre-commit on Ubuntu yet :(
Do you know how I can work around this?

documentation/content/developmentworkflow.rst
```diff
@@ -120,3 +120,74 @@ To configure this Git pre-commit hook:
...
+pacman -S python-pre-commit``
+
+Installing reviewdog
+~~~~~~~~~~~~~~~~~~~~
+To install reviewdog, run the following command::
+
+  curl -sfL https://raw.githubusercontent.com/reviewdog/reviewdog/master/install.sh | sh -s
```
rodrigosiqueira, line 189: fwiu, if you use this method, it will install reviewdog in `$PWD/bin/`. It looks like that your script does not handle this situation. Also, even after changing your code, it does not give me any update; I tried in the `src/plugins/subsystems/drm/drm.sh` which has some shellcheck problems, but I did not get anything (:Kwreview)

joaoseckler: > fwiu, if you use this method, it will install reviewdog in $PWD/bin/. It looks like that your script does not handle this situation.

I changed the documentation a bit, see if it is better. 

> Also, even after changing your code, it does not give me any update; I tried in the src/plugins/subsystems/drm/drm.sh which has some shellcheck problems, but I did not get anything (:Kwreview)

By default, `kwreview` will compare the current patch with unstable, and will ignore the staging area and the working tree. I just added an option to compare against working tree (`-t`) and agains the staging area (`-g`). Tell me if that was not the problem.

rodrigosiqueira:
@joaoseckler did you update this PR?

joaoseckler:
> @joaoseckler did you update this PR?

I had unpushed changes in my local repository... Now they are here. They add the `-g` and `-t` flag and the suggestion to install pre-commit with pip instead of apt or pacman.

joaoseckler:
The last update to this PR adds the `-f` / `--file` option, which allows for a direct checking of files, without checking the VCS.

joaoseckler:
> The last update to this PR adds the `-f` / `--file` option, which allows for a direct checking of files, without checking the VCS.

On a second thought, this is better implemented with an `--ignore-vcs`/`-i` option, which considers all files passed by the user in the command line. We also now make use of the `-filter-mode` flag of reviewdog, making the code cleaner. Since we were dealing with it, we also implemented a `--filter-mode` option to kwreview. We also fixed a problem with `--shfmt-only`, which was not working.

magalilemes (approved):
Reaaally cool work.

Only thing I'd check is that you wrote in developmentworkflow.rst that pre-commit is available on debian-based distros, but I think it's still in the experimental phase. Besides that, everything looks great and easy to use.

documentation/content/developmentworkflow.rst
alan-barzilay: @rodrigosiqueira `pip install pre-commit`

alan-barzilay (approved):
lgtm, nice job!

rodrigosiqueira:
Hi @joaoseckler ,
This new change is great; I really like the `-i` option. However, I don't think the name `--ignore-vcs` is intuitive (actually, what vcs stands for?). I have to admit that I prefer something like `--file`... how about make `./script/kreview.sh PATH` do the same work as `-i` does? Also, add the instruction for this option in the `developmentworkflow.rst` file.

alan-barzilay:
@rodrigosiqueira vcs stands for version control system

joaoseckler:
The last update to this PR accepts the suggestion of using the -i option by default, changes the documentation about that, and change the documentation regarding the installation of pre-commit.

rodrigosiqueira:
Hi @joaoseckler ,
Thanks a lot for all your effort here, merged to the unstable branch.

PR #351, opened 2021-06-29
src: mk.sh: Break into deploy.sh and build.sh

joaoseckler:
Commands related to kernel compiling and installation were in one single
file, src/mk.sh, and it has gotten too big. This commit splits this file
into src/deploy.sh (installation) and src/build.sh (compilation). It
also moves some common functions to src/kwlib.sh and changes the name of
some functions (e.g. mk_build becomes kernel_build) since there is no
reference to "mk" anymore.

The test file tests/mk_test.sh is also split accordingly. The resulting
file tests/build_test.sh is refactored.

### DIFF
--- a/src/build.sh
+++ b/src/build.sh
@@ -1,0 +1,124 @@
+include "$KW_LIB_DIR/kwlib.sh"
+include "$KW_LIB_DIR/kwio.sh"
+include "$KW_LIB_DIR/kw_config_loader.sh"
+
+# This function retrieves and prints information related to the kernel that
+# will be compiled.
+# shellcheck disable=2120
+function build_info()
+{
+  local flag
+  local kernel_name
+  local kernel_version
+  local compiled_modules
+
+  flag="$1"
+  kernel_name=$(get_kernel_release "$flag")
+  kernel_version=$(get_kernel_version "$flag")
+
+  say "Kernel source information"
+  echo -e "\tName: $kernel_name"
+  echo -e "\tVersion: $kernel_version"
+
+  if [[ -f '.config' ]]; then
+    compiled_modules=$(grep -c '=m' .config)
+    echo -e "\tTotal modules to be compiled: $compiled_modules"
+  fi
+}
+
+# This function is responsible for manipulating kernel build operations such as
+# compile/cross-compile and menuconfig.
+#
+# @flag How to display a command, see `src/kwlib.sh` function `cmd_manager`
+# @raw_options String with all user options
+#
+# Return:
+# In case of successful return 0, otherwise, return 22 or 125.
+function kernel_build()
+{
+  local flag="$1"
+  shift 1
+  local raw_options="$*"
+  local PARALLEL_CORES=1
+  local CROSS_COMPILE=""
+  local command=""
+  local start
+  local end
+  local arch="${configurations[arch]}"
+  local menu_config="${configurations[menu_config]}"
+
+  if [[ "$1" == -h ]]; then
+    build_help
+    exit 0
+  fi
+
+  menu_config=${menu_config:-"nconfig"}
+  arch=${arch:-"x86_64"}
+
+  if ! is_kernel_root "$PWD"; then
+    complain "Execute this command in a kernel tree."
+    exit 125 # ECANCELED
+  fi
+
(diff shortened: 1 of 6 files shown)


src/build.sh
```diff
@@ -0,0 +1,124 @@
+include "$KW_LIB_DIR/kwlib.sh"
+include "$KW_LIB_DIR/kwio.sh"
+include "$KW_LIB_DIR/kw_config_loader.sh"
+
+# This function retrieves and prints information related to the kernel that
+# will be compiled.
+# shellcheck disable=2120
+function build_info()
```
alan-barzilay, line 8: we should probably document this function parameters at some point

src/deploy.sh
```diff
@@ -11,10 +10,9 @@
 #    important to highlight that a VM in the localhost can be treated as a
 #    remote machine.
 #
-# Usually, install modules and update the kernel image requires root
-# permission, with this idea in mind we rely on the `/root` in the remote
-# machine. Additionally, for local deploy, you will be asked to enter your
-# root password.
+# Usually, install modules and update the kernel image requires root permission,
```
alan-barzilay, line 13: how about: Usually, **to** install modules and update the kernel image **root permission is required**

src/deploy.sh
```diff
@@ -11,10 +10,9 @@
...
+# with this idea in mind we rely on the `/root` in the remote machine.
+# Additionally, for local deploy, you will be asked to enter your root password.
```
alan-barzilay, line 15: not sure if we need this second comma after deploy

src/kwlib.sh
```diff
@@ -147,6 +147,40 @@ function find_kernel_root()
   echo "$kernel_root"
 }
 
+# Get the kernel release based on the command kernel release.
```
alan-barzilay, line 150: `kernelrelease` shouldn't have this space, right?

src/kwlib.sh
```diff
@@ -147,6 +147,40 @@ function find_kernel_root()
...
+#
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/kwlib.sh` function `cmd_manager`
+#
+# Note: Make sure that you called is_kernel_root before try to execute this
```
alan-barzilay, line 155: trying instead of try

alan-barzilay (commented):
I made a few remarks but they are mostly about things that were already like this before the PR so I'm not even sure if this is the best place for me to bring them up haha

alan-barzilay:
@joaoseckler there are a few variables and functions in deploy.sh that still reference `mk`, should we rename them as well?

mk_list_installed_kernels
mk_kernel_uninstall
mkinitcpio_name
template_mkinit

joaoseckler:
> @joaoseckler there are a few variables and functions in deploy.sh that still reference `mk`, should we rename them as well?
> 
> mk_list_installed_kernels
> mk_kernel_uninstall
> mkinitcpio_name
> template_mkinit

Thanks for the review! I did some of the english revision. mk_kernel_uninstall and mk_list_installed_kernels were updated. I think mkinitcpio_name and template_mkinit refer to the command `mkinitcpio` and should remain with the `mk` prefix.

alan-barzilay (approved):
LGTM!

tests/build_test.sh
```diff
@@ -0,0 +1,122 @@
...
+{
+  echo '5.4.0-rc7-test'
+}
+
+function get_kernel_version_mock()
+{
+  echo '5.4.0-rc7'
+}
+
+oneTimeSetUp()
```
rodrigosiqueira, line 14: Why you did not add a function? In the original function, we had it. Same for setUp() and tearDown()

joaoseckler: @rodrigosiqueira, I didn't understand your question. What function was not added?

alan-barzilay: I think he means in the fucntion definitions, we usually do something like `function foo(){ stuff}`, but the keyword `function`is missing here (nice catch)

rodrigosiqueira: Yes, I was talking about the keyword `function`. Anyway, I reviewed this patch again, and I decided to not waste reviewing cycles and fix this and other minor issues by myself.

rodrigosiqueira:

PR #352, opened 2021-06-30
Improve bash_autocomplete file

magalilemes:
This PR fixes two minor things in src/bash_autocomplete.sh: it removes options from `kw_options` that are not supported by kw, and adds the ones that are supported and were missing. It also fixes [SC2207](https://github.com/koalaman/shellcheck/wiki/SC2207).

### DIFF
--- a/src/bash_autocomplete.sh
+++ b/src/bash_autocomplete.sh
@@ -4,10 +4,10 @@ function _kw_autocomplete()
   COMPREPLY=()
   current_command="${COMP_WORDS[COMP_CWORD]}"
   previous_command="${COMP_WORDS[COMP_CWORD - 1]}"
-  kw_options="explore e build b bi init new n ssh s clear-cache
+  kw_options="explore e bd build b init ssh s clear-cache
                 mount mo umount um vars up u codestyle c configm g
                 maintainers m deploy d help h version statistics
-                pomodoro p drm diff --version -v"
+                pomodoro p drm diff df --version -v man"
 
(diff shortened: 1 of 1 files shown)


alan-barzilay (approved):

src/bash_autocomplete.sh
```diff
@@ -4,14 +4,14 @@ function _kw_autocomplete()
   COMPREPLY=()
   current_command="${COMP_WORDS[COMP_CWORD]}"
   previous_command="${COMP_WORDS[COMP_CWORD - 1]}"
-  kw_options="explore e build b bi init new n ssh s clear-cache
-                mount mo umount um vars up u codestyle c configm g
+  kw_options="explore e bd build b init ssh s clear-cache
+                mount mo umount um vars v up u codestyle c configm g
```
joaoseckler, line 8: `kw v` was dropped since the introduction of `kw -v`! See commit `f8af2abeae4265cc0d55fd9f3544427b02ffadd1` and pull request #169.

joaoseckler (changes requested):

joaoseckler:
For another PR, I think it would be nice if kw had a more complete autocompletion feature. For example, it would be nice if

`kw d --rem<TAB>`

would complete to

`kw d --remote`

rodrigosiqueira:
Hi @magalilemes , thanks for your patch. Merged to unstable branch.

@alan-barzilay I noticed that our code coverage is looking at the autocomplete file, could you make a patch to make codecov ignore this?

Issue #318, opened 2021-06-10
SC2124 - Assigning an array to a string! Assign as array, or use * instead of @ to concatenate


alan-barzilay:
This issue will be used to keep track of an error/warning type that is common on our codebase. For more information about it, take a look at the official shellcheck wiki entry:
https://github.com/koalaman/shellcheck/wiki/SC2124

alan-barzilay:
Related error: 
https://github.com/koalaman/shellcheck/wiki/SC2178

alan-barzilay:
@joaoseckler It seems like we missed a few files:

In src/plugins/kernel_install/kernel_install_utils.sh line 39:
  local message=$@
                ^-- SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.


--
In src/plugins/kernel_install/utils.sh line 39:
  local message=$@
                ^-- SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.


--
In src/plugins/subsystems/drm/drm.sh line 129:
  local raw_modules_str="$@"
                        ^--^ SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.


In src/plugins/subsystems/drm/drm.sh line 346:
  local raw_options="$@"
                    ^--^ SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.

alan-barzilay:
also one for SC2178

In tests/utils.sh line 243:
  local expected="$3"
        ^------^ SC2178: Variable was used as an array but is now assigned a string.

alan-barzilay:
Maybe this 2 error codes are also related? dont know, what do you think joao?

In tests/utils.sh line 243:
  local expected="$3"
        ^------^ SC2178: Variable was used as an array but is now assigned a string.

In tests/utils.sh line 248:
  assertEquals "line $line: $msg" "$target" "$expected"
                                             ^-------^ SC2128: Expanding an array without an index only gives the first element.

rodrigosiqueira:
Done! Thanks

PR #353, opened 2021-07-01
Refactor deploy

alan-barzilay:
This PR introduces a few refactorings to deploy.sh 

With the entry-point as the root of the file, I re-ordered the functions to follow the order upon which they are called in the entry-point, I believe this should make the file more readable and organized.
I also fixed a few minor issues such as few instances of file sourcing that weren't using the include function.

There were 3 functions for cleanup, they had a lot of overlap and were needlessly complicated so I merged them into one.

I also removed some unnecessary checks, the `is_kernel_root` check is done once in the entry-point function kernel_deploy so there is no need to remake this check in other functions later on

### DIFF
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -303,13 +303,13 @@ function list_installed_kernels()
         return 125 # ECANCELED
       fi
 
-      . "$KW_PLUGINS_DIR/kernel_install/utils.sh" --source-only
+      include "$KW_PLUGINS_DIR/kernel_install/utils.sh"
       list_installed_kernels "$single_line" "${configurations[mount_point]}"
 
       vm_umount
       ;;
     2) # LOCAL_TARGET
-      . "$KW_PLUGINS_DIR/kernel_install/utils.sh" --source-only
+      include "$KW_PLUGINS_DIR/kernel_install/utils.sh"
       list_installed_kernels "$single_line"
       ;;
     3) # REMOTE_TARGET
@@ -362,8 +362,8 @@ function kernel_uninstall()
 
       # Local Deploy
       # We need to update grub, for this reason we to load specific scripts.
-      . "$KW_PLUGINS_DIR/kernel_install/$distro.sh" --source-only
-      . "$KW_PLUGINS_DIR/kernel_install/utils.sh" --source-only
+      include "$KW_PLUGINS_DIR/kernel_install/$distro.sh"
+      include "$KW_PLUGINS_DIR/kernel_install/utils.sh"
       # TODO: Rename kernel_uninstall in the plugin, this name is super
       # confusing
       kernel_uninstall "$reboot" 'local' "$kernels_target" "$flag"
@@ -574,8 +574,8 @@ function kernel_install()
         exit 95 # ENOTSUP
       fi
 
-      . "$KW_PLUGINS_DIR/kernel_install/utils.sh" --source-only
-      . "$KW_PLUGINS_DIR/kernel_install/$distro.sh" --source-only
+      include "$KW_PLUGINS_DIR/kernel_install/utils.sh"
+      include "$KW_PLUGINS_DIR/kernel_install/$distro.sh"
       install_kernel "$name" "$distro" "$kernel_img_name" "$reboot" "$arch_target" 'vm' "$flag"
       return "$?"
       ;;
@@ -593,8 +593,8 @@ function kernel_install()
         exit 1 # EPERM
       fi
 
-      . "$KW_PLUGINS_DIR/kernel_install/utils.sh" --source-only
-      . "$KW_PLUGINS_DIR/kernel_install/$distro.sh" --source-only
+      include "$KW_PLUGINS_DIR/kernel_install/utils.sh"
+      include "$KW_PLUGINS_DIR/kernel_install/$distro.sh"
       install_kernel "$name" "$distro" "$kernel_img_name" "$reboot" "$arch_target" 'local' "$flag"
       return "$?"
       ;;
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -102,7 +102,7 @@ function kernel_deploy()
     exit 125 # ECANCELED
   fi
 
-  signal_manager 'interrupt_cleanup' || warning 'Was not able to set signal handler'
+  signal_manager 'cleanup' || warning 'Was not able to set signal handler'
 
   if [[ "$target" == "$VM_TARGET" ]]; then
     vm_mount
@@ -140,10 +140,8 @@ function kernel_deploy()
     vm_umount
   fi
 
-  if [[ "$target" == "$REMOTE_TARGET" ]]; then
-    say "Cleanup temporary files"
-    cleanup_after_deploy
-  fi
+  #shellcheck disable=SC2119
+  cleanup
 }
 
 # This function gets raw data and based on that fill out the options values to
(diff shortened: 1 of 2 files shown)


alan-barzilay:
the shellcheck failures were already present and are being worked on by  @joaoseckler  and tracked in issue #323

rodrigosiqueira (changes requested):
Hi @alan-barzilay ,
This series looks really good. Can you include one extra commit that fixes the SC2124 issue in this file?

alan-barzilay:
> Can you include one extra commit that fixes the SC2124 issue in this file?

Are you sure? @joaoseckler is working on resolving all cases of SC2124 in issue #318. Wouldn't a single commit fixing all occurrences be better? Also, if we end up needing to follow a certain specific pattern (or implement a helper function like in #319) it would be easier to make things consistent by doing a single commit

src/deploy.sh
```diff
@@ -23,169 +23,336 @@ include "$KW_LIB_DIR/signal_manager.sh"
...
+    cleanup_after_deploy
+  fi
+}
 
-      # 3. Deploy: Execute script
-      local cmd="bash $REMOTE_KW_DEPLOY/deploy.sh --modules $release.tar"
-      cmd_remotely "$cmd" "$flag" "$remote" "$port"
-      ;;
-  esac
+function deploy_help()
```
rodrigosiqueira, line 145: How about putting this `deploy_help` at the end of this file?

rodrigosiqueira:
> > Can you include one extra commit that fixes the SC2124 issue in this file?
> 
> @rodrigosiqueira Are you sure? @joaoseckler is working on resolving all cases of SC2124 in issue #318. Wouldn't a single commit fixing all occurrences be better? Also, if we end up needing to follow a certain specific pattern (or implement a helper function like in #319) it would be easier to make things consistent by doing a single commit

Yes, I'm sure:
1. This series provides a set of basic refactors, and you actually fixed one of the shellcheck problems in this series.
2. If we have a single commit to fixing this issue, we can easily revert this change without affecting others. When we have a single commit that changes everything, it is terrible for revert because we will revert everywhere instead of a single point.

Honestly, I don't like these huge shellcheck fixes for the following reasons:
1. They are super hard to review.
2. We can easily miss something during the review and merge a bad code (see `is_kernel_root`).
3. It is terrible for bisect.
4. We can end up change multiple parts of the code, which messes with the git log.

Even though I don't like our current approach, I accepted because:
1. I don't want to make this shellcheck work endless.
2. We have five people paying attention to these types of changes.
3. I don't want to block other people's work since I have a limited amount of time with all of you (technically, after August, I'm on my own again).

alan-barzilay:
> Yes, I'm sure

ok

src/deploy.sh
alan-barzilay: the whole point of reordering was to keep the functions in order of appearance, I would rather not move `deploy_help`

rodrigosiqueira: What is the value of seeing such a function in the middle of the code? Technically, this function does not have any value from the development perspective, and if we keep it in the middle, it just looks like a "noise" when I'm looking for something more substantial. That was the exact situation that I have in other functions that have this helper function in the middle of the code.

alan-barzilay: ok, I will move it

alan-barzilay:
apparently pr #351 introduced a bug when naming the deploy test, the extra "s" in the name made it undiscoverable by run_tests.sh.

There are 2 tests failing, one of the cleanup function that I changed and needs to be adapted and one that I believe to me reminiscent from pr #351 as well. The kernel_install function cant be found because the plugins/kernel_install/utils.sh was never included. But including it makes everything  break because invoke shunit cant be found anymore.
I believe this is happening because now we have 2 file called utils.sh and this is breaking the code.

rodrigosiqueira:
@alan-barzilay Could you be more specific about this bug? Because I just tested `kw d [-l|-u]`, and I just sent a PR that everything passes.

alan-barzilay:
> @alan-barzilay Could you be more specific about this bug? Because I just tested kw d [-l|-u], and I just sent a PR that everything passes.

deploy is working, its tests are broken

rodrigosiqueira:
Yeah... but what is exactly break? Because I just submitted this PR https://github.com/kworkflow/kworkflow/pull/358, and everything passes.

alan-barzilay:
> Yeah... but what is exactly break? Because I just submitted this PR #358, and everything passes.

the test is not being run because of the file name

rodrigosiqueira:
@alan-barzilay 
See https://github.com/kworkflow/kworkflow/pull/359  and https://app.travis-ci.com/github/kworkflow/kworkflow/builds/231853501#L1099 . I was running deploy_test locally, and everything was working as expected.

joaoseckler:
Hey @alan-barzilay, @rodrigosiqueira! I just made a PR which will hopefully fix the issue with the inclusion of `utils.sh`, see #361. Maybe this could be rebased off that branch.

tests/deploy_test.sh
```diff
@@ -0,0 +1,600 @@
...
+  cp -f "$MAINTAINERS_SAMPLE" "$FAKE_KERNEL"
+  cp -f "$CHECKPATH_EXT" "$FAKE_KERNEL/scripts/"
+
+  export preset_name="template_mkinitcpio.preset"
+  export test_path="$PWD/$FAKE_KERNEL"
+  export KW_PLUGINS_DIR="$test_path"
+  export KW_CACHE_DIR="$test_path"
+  export KW_ETC_DIR="$PWD/$SAMPLES_DIR/etc"
+  export DEPLOY_SCRIPT="$test_path/$kernel_install_path/deploy.sh"
+  export KW_PLUGINS_DIR="$PWD/src/plugins/"
```
magalilemes, line 52: Maybe have just one `KW_PLUGINS_DIR` export? (see line 48)

alan-barzilay:
>  I was running deploy_test locally, and everything was working as expected.

@rodrigosiqueira  ok let me try to explain again in more detail because this involves at least 3 mistakes we made in the past, this test should really be working in unstable. It only broke here because I refactored deploy to use `include` instead of `. blabla.sh --sources-only.

There were 3 problems: first when implementing include we missed the imports that happened inside functions in `mk.sh`. Then we renamed utils to util.sh but we forgot there was already another utils.sh file and this is not something allowed, if  you try to import 2 files with the same name one of them won't be imported. Lastly when breaking down mk.sh we gave a bad name to the deploy tests, and it wasn't being recognized by the ./run_tests script anymore.

What happened is that I started using include inside the deploy.sh file, but only when I realized the tests weren't being run that this bug emerged. The install_kernel function  (not to be confused with the kernel_install function) needed for one of the tests couldn't be found because of the double import of utils.sh, this test file is the only occasion where both files are included at the same time.

I started fixing this issue in my last commit but #359 and now #361 seem to be addressing this.

TLDR: for this PR to pass the tests  #361 or something equivalent needs to be merged into unstable

tests/deploy_test.sh
alan-barzilay: nice catch, I guess we can remove the first export

rodrigosiqueira:
@alan-barzilay let me know when you have a new version.

alan-barzilay:
> @alan-barzilay let me know when you have a new version.

@rodrigosiqueira this pr is ready for review

The only shellcheck error comes from a code that in #368  I argue that it should be disabled

rodrigosiqueira:
@alan-barzilay could you fix this shellcheck issue?

alan-barzilay:
> @alan-barzilay could you fix this shellcheck issue?

you fixed it by merging #386, I could rebase it into unstable to make it pass but I'm not at my computer right now, I would only be able to do it tomorrow night

alan-barzilay:
@rodrigosiqueira everything ready

rodrigosiqueira:
Merged, thanks.

PR #355, opened 2021-07-03
refactor and expand ssh

alan-barzilay:
the current ssh configuration options in kw are really limited, to
expand them this commit introduces the option for the user to input a
ssh configuration file. This way we can allow the user to configure ssh
however it wants with minimal code added to kw.

These limitation were blocking some of my work so I had to tackle them.

Another change this commit introduces is the refactoring of the functions `cmd_remotely` and `cphost2remote` to use sudo instead of the root user. Doing a ssh operation as root is a major security risk, one that our end-users may not want or even be able to take. 

maybe the last 3 commits should be squashed as one, I initially made them separate commits mostly to be able to partially revert them more easily while I was testing

### DIFF
--- a/etc/kworkflow_template.config
+++ b/etc/kworkflow_template.config
@@ -15,6 +15,11 @@ ssh_ip=localhost
 # Default ssh port
 ssh_port=22
 
+# Optional ssh configuration file to be used
+#ssh_configfile=~/.ssh/config
+# Hostname of the target in ssh_configfile
+#hostname=
+
 # Specify the default architecture used by KW
 arch=x86_64
 
--- a/src/remote.sh
+++ b/src/remote.sh
@@ -35,10 +35,10 @@ declare -gA remote_parameters
 function cmd_remotely()
 {
   local command="$1"
-  local flag="$2"
-  local remote="$3"
-  local port="$4"
-  local user="$5"
+  local flag=${2:-"HIGHLIGHT_CMD"}
+  local remote=${3:-${configurations[ssh_ip]}}
+  local port=${4:-${configurations[ssh_port]}}
+  local user=${5:-${configurations[ssh_user]}}
   local bash_code="$6"
   local composed_cmd=""
 
@@ -46,12 +46,6 @@ function cmd_remotely()
     warning "No command specified"
     exit 22
   fi
-
-  # Set default values if not specified
-  remote=${remote:-"localhost"}
-  port=${port:-"22"}
-  user=${user:-"root"}
-  flag=${flag:-"HIGHLIGHT_CMD"}
 
   composed_cmd="ssh -p $port $user@$remote"
   if [[ -v configurations['ssh_configfile'] && -v configurations['hostname'] ]]; then
@@ -81,20 +75,12 @@ function cmd_remotely()
 #   more options see `src/kwlib.sh` function `cmd_manager`
 function cp_host2remote()
 {
-  local src="$1"
-  local dst="$2"
-  local remote="$3"
-  local port="$4"
-  local user="$5"
-  local flag="$6"
-
-  src=${src:-"$KW_CACHE_DIR/$LOCAL_TO_DEPLOY_DIR/*"}
-
-  dst=${dst:-"$REMOTE_KW_DEPLOY"}
-  remote=${remote:-"localhost"}
-  port=${port:-"22"}
-  user=${user:-"root"}
-  flag=${flag:-"HIGHLIGHT_CMD"}
+  local src=${1:-"$KW_CACHE_DIR/$LOCAL_TO_DEPLOY_DIR/*"}
+  local dst=${2:-"$REMOTE_KW_DEPLOY"}
+  local remote=${3:-${configurations[ssh_ip]}}
+  local port=${4:-${configurations[ssh_port]}}
+  local user=${5:-${configurations[ssh_user]}}
+  local flag=${6:-"HIGHLIGHT_CMD"}
 
   cmd_manager "$flag" "rsync -e 'ssh -p $port' -La $src $user@$remote:$dst"
   if [[ -v configurations['ssh_configfile'] && -v configurations['hostname'] ]]; then
@@ -114,17 +100,12 @@ function cp_host2remote()
 # subshell and save it to a variable.
 function which_distro()
 {
-  local remote="$1"
-  local port="$2"
-  local user="$3"
-  local flag="$4"
+  local remote=${1:-${configurations[ssh_ip]}}
+  local port=${2:-${configurations[ssh_port]}}
+  local user=${3:-${configurations[ssh_user]}}
+  local flag=${4:-"SILENT"}
 
   cmd="cat /etc/os-release | grep -w ID | cut -d = -f 2"
-  remote=${remote:-"localhost"}
-  port=${port:-"22"}
-  user=${user:-"root"}
-  flag=${flag:-"SILENT"}
-
   cmd_remotely "$cmd" "$flag" "$remote" "$port" "$user"
 }
 
--- a/src/remote.sh
+++ b/src/remote.sh
@@ -53,10 +53,17 @@ function cmd_remotely()
   user=${user:-"root"}
   flag=${flag:-"HIGHLIGHT_CMD"}
 
-  composed_cmd="ssh -p $port $user@$remote \"$command\""
-  if [[ "$bash_code" == 1 ]]; then
-    composed_cmd="ssh -p $port $user@$remote '$command'"
+  composed_cmd="ssh -p $port $user@$remote"
+  if [[ -v configurations['ssh_configfile'] && -v configurations['hostname'] ]]; then
+    composed_cmd="ssh -F ${configurations['ssh_configfile']} ${configurations['hostname']}"
   fi
+
+  if [[ "$bash_code" == 1 ]]; then
+    composed_cmd="$composed_cmd '$command'"
+  else
+    composed_cmd="$composed_cmd \"$command\""
+  fi
+
   cmd_manager "$flag" "$composed_cmd"
 }
 
@@ -90,6 +97,9 @@ function cp_host2remote()
   flag=${flag:-"HIGHLIGHT_CMD"}
 
   cmd_manager "$flag" "rsync -e 'ssh -p $port' -La $src $user@$remote:$dst"
+  if [[ -v configurations['ssh_configfile'] && -v configurations['hostname'] ]]; then
+    cmd_manager "$flag" "rsync -e 'ssh -F ${configurations['ssh_configfile']}' -La $src ${configurations['hostname']}:$dst --rsync-path='sudo rsync'"
+  fi
 }
 
 # Access the target device and query the distro name.
--- a/src/remote.sh
+++ b/src/remote.sh
@@ -53,9 +53,9 @@ function cmd_remotely()
   fi
 
   if [[ "$bash_code" == 1 ]]; then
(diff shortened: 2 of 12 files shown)


src/remote.sh
```diff
@@ -48,13 +48,21 @@ function cmd_remotely()
...
+    composed_cmd="ssh -F ${configurations['ssh_configfile']} ${configurations['hostname']}"
+  else
+    composed_cmd="ssh -p $port $user@$remote"
+  fi
+
   if [[ "$bash_code" == 1 ]]; then
-    composed_cmd="ssh -p $port $user@$remote '$command'"
+    composed_cmd="$composed_cmd sudo '$command'"
+  else
+    composed_cmd="$composed_cmd sudo \"$command\""
```
magalilemes, line 63: Suppose I don't assign any value to `ssh_configfile`, then cmd_remotely will run all commands as sudo*, right?

src/remote.sh
```diff
@@ -84,10 +92,17 @@ function cp_host2remote()
...
   flag=${flag:-"HIGHLIGHT_CMD"}
 
-  cmd_manager "$flag" "rsync -e 'ssh -p $port' -La $src $user@$remote:$dst"
+  if [[ -v configurations['ssh_configfile'] ]]; then
+    cmd_manager "$flag" "rsync -e 'ssh -F ${configurations['ssh_configfile']}' -La $src ${configurations['hostname']}:$dst --rsync-path='sudo rsync'"
+  else
+    cmd_manager "$flag" "rsync -e 'ssh -p $port' -La $src $user@$remote:$dst --rsync-path='sudo rsync'"
+  fi
+
+  cmd_remotely "sudo chown -R root:root $dst" "$flag" "$remote" "$port" "$user"
```
magalilemes, line 104: \* Then we would have something as `sudo sudo` here?

src/remote.sh
```diff
@@ -48,13 +48,21 @@ function cmd_remotely()
   # Set default values if not specified
   remote=${remote:-"localhost"}
   port=${port:-"22"}
-  user=${user:-"root"}
+  user=${user:-$(whoami)}
```
magalilemes, line 51: This assigns the username to a user from the local machine, which may not be the same from the remote machine, I guess.

src/remote.sh
alan-barzilay: yes!

src/remote.sh
alan-barzilay: Correct!
but this assignment happens if and only if no other user name has been provided. 
I only changed the default value from root to the username of the current user which I think is the best guess we could make. If such a user does not exist then an username would have to be provided in the config file anyway.

Currently the only way to provide an user in kw is through the ssh_ip field in the config file, which is a whole other can of worms that I didn't want to mess with right now haha

src/remote.sh
alan-barzilay: oops, you are right. I though I had fixed that, thank you! I will update it

alan-barzilay: done

src/remote.sh
joaoseckler: #358 addresses this user issue! Maybe this could be rebased of that...

src/remote.sh
```diff
@@ -48,13 +48,21 @@ function cmd_remotely()
...
   flag=${flag:-"HIGHLIGHT_CMD"}
 
-  composed_cmd="ssh -p $port $user@$remote \"$command\""
+  if [[ -v configurations['ssh_configfile'] ]]; then
+    composed_cmd="ssh -F ${configurations['ssh_configfile']} ${configurations['hostname']}"
```
joaoseckler, line 55: The configuration variable `hostname` is used without checking if it exists! This could be checked and warned or solved at `kw_config_loader`. I think not handling this error might be user-unfriendly.

joaoseckler (changes requested):
I think it is OK to use sudo instead of root in kw's functioning, but I wouldn't say it is a serious security risk to do otherwise. First, I think that systematically using sudo is not very different from logging in as root. Also, typically there is no sensitive information in the remote machine. I've recently come across a [video](https://www.youtube.com/watch?v=fKuqYQdqRIs) which talks a bit about this, if anyone is curious. I'll not approve this right now since I've not actually tested it, but it LGTM.

etc/kworkflow_template.config
```diff
@@ -12,6 +12,11 @@ ssh_ip=localhost
 # Default ssh port
 ssh_port=22
 
+# Optional ssh configuration file to be used
```
rodrigosiqueira, line 15: Make it clear that you're talking about the config file from `ssh` (e.g., `~/.ssh/config`) because the first time I read it, I thought that you're adding a new config file kw.

src/remote.sh
```diff
@@ -124,7 +124,7 @@ function which_distro()
   cmd="cat /etc/os-release | grep -w ID | cut -d = -f 2"
   remote=${remote:-"localhost"}
   port=${port:-"22"}
-  user=${user:-"root"}
+  user=${user:-$(whoami)}
```
rodrigosiqueira, line 127: Is this a leftover of the previous commit?

tests/remote_test.sh
```diff
@@ -135,41 +135,47 @@ function test_cp_host2remote()
...
+  expected_command="rsync -e 'ssh -p $port' -La $src $user@$remote:$dst --rsync-path='sudo rsync'
+ssh -p $port ${user}@${remote} sudo \"chown -R root:root $dst\""
   output=$(cp_host2remote "$src" "$dst" "$remote" "$port" "$user" "$flag")
   assertEquals "Command did not match ($ID)" "$expected_command" "$output"
 
   ID=2
   src="$KW_CACHE_DIR/$LOCAL_TO_DEPLOY_DIR/*"
-  expected_command="rsync -e 'ssh -p $port' -La $src $user@$remote:$dst"
+  expected_command="rsync -e 'ssh -p $port' -La $src $user@$remote:$dst --rsync-path='sudo rsync'
+ssh -p $port ${user}@${remote} sudo \"chown -R root:root $dst\""
```
rodrigosiqueira, line 146: No space?

tests/remote_test.sh
```diff
@@ -240,17 +246,21 @@ function test_prepare_remote_dir()
...
-    "rsync -e 'ssh -p 2222' -La $FAKE_KW/kernel_install/deploy.sh root@172.16.224.1:/root/kw_deploy/"
-    "rsync -e 'ssh -p 2222' -La $FAKE_KW/kernel_install/utils.sh root@172.16.224.1:/root/kw_deploy/"
+    "ssh -p 2222 root@172.16.224.1 sudo \"mkdir -p /root/kw_deploy\""
+    "rsync -e 'ssh -p 2222' -La $FAKE_KW/kernel_install/debian.sh root@172.16.224.1:/root/kw_deploy/distro_deploy.sh --rsync-path='sudo rsync'"
+    "ssh -p $port ${user}@${remote} sudo \"chown -R root:root /root/kw_deploy/distro_deploy.sh\""
+    "rsync -e 'ssh -p 2222' -La $FAKE_KW/kernel_install/deploy.sh root@172.16.224.1:/root/kw_deploy/ --rsync-path='sudo rsync'"
+    "ssh -p $port ${user}@${remote} sudo \"chown -R root:root /root/kw_deploy/\""
+    "rsync -e 'ssh -p 2222' -La $FAKE_KW/kernel_install/utils.sh root@172.16.224.1:/root/kw_deploy/ --rsync-path='sudo rsync'"
+    "ssh -p $port ${user}@${remote} sudo \"chown -R root:root /root/kw_deploy/\""
+
```
rodrigosiqueira, line 256: extra space

rodrigosiqueira (changes requested):
I added a couple of comments, but we already discussed it last week. I'm looking forward to the new version.

alan-barzilay:
The last version of this PR rebases into unstable and addresses all points previously raised. This version also alters how we handle default values in ssh as discussed in our last meeting.

rodrigosiqueira (changes requested):

src/remote.sh
```diff
@@ -53,10 +53,18 @@ function cmd_remotely()
   user=${user:-"root"}
   flag=${flag:-"HIGHLIGHT_CMD"}
 
-  composed_cmd="ssh -p $port $user@$remote \"$command\""
+  if [[ -v configurations['ssh_configfile'] && -v configurations['hostname'] ]]; then
+    composed_cmd="ssh -F ${configurations['ssh_configfile']} ${configurations['hostname']}"
+  else
+    composed_cmd="ssh -p $port $user@$remote"
```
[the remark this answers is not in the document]
alan-barzilay: ok

src/remote.sh
```diff
@@ -53,10 +53,18 @@ function cmd_remotely()
...
+  fi
+
   if [[ "$bash_code" == 1 ]]; then
-    composed_cmd="ssh -p $port $user@$remote '$command'"
+    composed_cmd="$composed_cmd '$command'"
+  else
```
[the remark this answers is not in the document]
alan-barzilay: @rodrigosiqueira I can´t, if I remove the if else structure the `$command` will be appended twice

src/remote.sh
```diff
@@ -47,16 +47,18 @@ function cmd_remotely()
...
+  if [[ -v configurations['ssh_configfile'] && -v configurations['hostname'] ]]; then
+    composed_cmd="ssh -F ${configurations['ssh_configfile']} ${configurations['hostname']}"
+  else
+    composed_cmd="ssh -p $port $user@$remote"
+  fi
 
-  composed_cmd="ssh -p $port $user@$remote \"$command\""
   if [[ "$bash_code" == 1 ]]; then
-    composed_cmd="ssh -p $port $user@$remote '$command'"
+    composed_cmd="$composed_cmd sudo '$command'"
```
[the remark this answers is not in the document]
alan-barzilay: @rodrigosiqueira please let me know if this problem persists in this last revision

alan-barzilay:
@rodrigosiqueira this series is ready to be reviewed again

alan-barzilay:
@rodrigosiqueira this series is ready to be reviewed again

rodrigosiqueira (changes requested):
@alan-barzilay 
I'm still getting:
```
ash: -c: line 0: syntax error near unexpected token `do'
bash: -c: line 0: `sudo bash -c for i in /sys/class/vtconsole/*/bind; do echo 1 > $i; done; sleep 0.5'
```

You can try it on your side. Setup a remote machine with Gnome, and try `kw drm --gui-off`

alan-barzilay:
> Setup a remote machine with Gnome, and try kw drm --gui-off

@rodrigosiqueira does it has to be gnome or any DE will work?

rodrigosiqueira:
In theory, any DE that uses `systemctl isolate multi-user.target` should work; however, I never tested it. Which DE do you have?

alan-barzilay:
> Which DE do you have?

KDE, but if its untested I will get a new image with gnome just to be safe

rodrigosiqueira:
> > Which DE do you have?
> 
> KDE, but if its untested I will get a new image with gnome just to be safe

Try on KDE, I'm curious :)

alan-barzilay:
@rodrigosiqueira done!

rodrigosiqueira:
I'm still not sure if I like the "sudo" approach and the way that we are dealing with the ssh config... but I decided to go ahead with it and start to use kw with this change in my daily work. Since this is in the unstable branch, we can change it if need.

Merged, thanks.

PR #356, opened 2021-07-04
tests: fix message on compare_command_sequence

alan-barzilay:
I spent a few hours pulling my hair trying to understand why my tests were failing. Turns out the command and expected command are reversed in the  `compare_command_sequence`, this PR fixes this


rodrigosiqueira:
Nice, merged.

PR #358, opened 2021-07-05
Deploy small refactor

rodrigosiqueira:
I was working on adding the basic support for ChromeOS, and I started to hit multiple limitations in our deploy. In summary, we need to refactor many things around deploy, and one of them is how we handle IP, port, and users. This PR starts by refactoring how we handle IPs and ports. Next, it adds basic support for reading the user name from the configuration file. For the user part, most of this work was triggered by https://github.com/kworkflow/kworkflow/pull/355, which highlighted that we do not enable users to set users that they want.

### DIFF
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -95,7 +95,6 @@ function modules_install()
 {
   local flag="$1"
   local target="$2"
-  local formatted_remote="$3"
   local remote
   local port
   local distro
@@ -161,13 +160,11 @@ function modules_install()
 #   available kernels in a single line separated by commas. If it gets 0 it
 #   will display each kernel name by line.
 # @target Target can be 1 (VM_TARGET), 2 (LOCAL_TARGET), and 3 (REMOTE_TARGET)
-# @unformatted_remote We expect the REMOTE:PORT string
 function list_installed_kernels()
 {
   local flag="$1"
   local single_line="$2"
   local target="$3"
-  local unformatted_remote="$4"
   local remote
   local port
 
@@ -328,7 +325,6 @@ function kernel_install()
 # @target Target machine Target machine Target machine Target machine
 # @reboot If this value is equal 1, it means reboot machine after kernel
 #         installation.
-# @formatted_remote Remote formatted as IP:PORT or USE@MACHINE:PORT
 # @kernels_target List containing kernels to be uninstalled
 # @flag How to display a command, see `src/kwlib.sh` function `cmd_manager`
 #
@@ -338,9 +334,8 @@ function kernel_uninstall()
 {
   local target="$1"
   local reboot="$2"
-  local formatted_remote="$3"
-  local kernels_target="$4"
-  local flag="$5"
+  local kernels_target="$3"
+  local flag="$4"
   local distro
   local remote
   local port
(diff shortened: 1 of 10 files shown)


src/remote.sh
```diff
@@ -222,40 +224,51 @@ function generate_tarball()
...
-# @parameters String to be parsed
+# @parameters: Command line parameter to be parsed
 #
 # Returns:
-# This function has two returns, and we make the second return by using
-# capturing the "echo" output. The standard return ("$?") can be 22 if
-# something is wrong or 0 if everything finished as expected; the second
-# output is the remote info as IP:PORT
-function get_remote_info()
+# This function populates the variables REMOTE_IP and REMOTE_PORT based on the
```
joaoseckler, line 232: Maybe specifying that they are entries of the remote_parameters array would clear things a bit. I propose:

`This function populates the entries REMOTE_IP and REMOTE_PORT of the remote_parameters array...`

src/remote.sh
```diff
@@ -222,40 +224,51 @@ function generate_tarball()
...
+    # Use the specific port
+    [[ "$?" != 22 ]] && remote_parameters['REMOTE_PORT']="$port"
+
+    # Handling user
+    remote_parameters['REMOTE_USER']="$user"
+    user=$(get_based_on_delimiter "$ip" '@' 1)
+    # There is no specific user, let's assume root
+    [[ "$?" != 22 ]] && remote_parameters['REMOTE_USER']="$user"
+
+    # Handling IP by making sure that we elimite user and port
```
joaoseckler, line 258: Eliminate

joaoseckler (approved):
Besides the two small documentation suggestions, look good to me! Nice improvement

src/deploy.sh
```diff
@@ -163,13 +160,11 @@ function modules_install()
 #   available kernels in a single line separated by commas. If it gets 0 it
 #   will display each kernel name by line.
 # @target Target can be 1 (VM_TARGET), 2 (LOCAL_TARGET), and 3 (REMOTE_TARGET)
-# @unformatted_remote We expect the REMOTE:PORT string
 function list_installed_kernels()
 {
   local flag="$1"
   local single_line="$2"
   local target="$3"
-  local unformatted_remote="$4"
```
magalilemes, line 172: At line 437, `list_installed_kernels` is still called with this fourth parameter.

src/deploy.sh
```diff
@@ -224,7 +218,6 @@ function kernel_install()
   local name="$2"
   local flag="$3"
   local target="$4"
-  local formatted_remote="$5"
```
magalilemes, line 227: At line 485, 5th parameter of `kernel_install` is still being used.

magalilemes (changes requested):
Really good, I just noticed two things that need to be changed.

src/deploy.sh
```diff
@@ -363,6 +358,8 @@ function kernel_uninstall()
       # We need to update grub, for this reason we to load specific scripts.
       . "$KW_PLUGINS_DIR/kernel_install/$distro.sh" --source-only
       . "$KW_PLUGINS_DIR/kernel_install/utils.sh" --source-only
+      # XXX: Rename kernel_uninstall in the plugin, this name is super
```
alan-barzilay, line 361: I think instead of XXX it would be better to just write TODO, so we can easilly find all TODOS in the code (or open an issue in github and keep the code clean)

alan-barzilay (approved):

rodrigosiqueira:
Merged

PR #361, opened 2021-07-05
src: kw_include: Disambiguate files with same name

joaoseckler:
Previously, files with the same name but in different folders couldn't
be included concomitantly, because the include manager would use the
same variable to keep track of which file was already included. This
commit introduces a method for creating such variable that is based on
the path of the file, not only its name.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -250,7 +250,7 @@ function synchronize_files()
   cmd_output_manager "cp $app_name $binpath" "$verbose"
   ASSERT_IF_NOT_EQ_ZERO "The command 'cp $app_name $binpath' failed" "$?"
 
-  sed -i -e "s,##KW_INSTALL_PREFIX_TOKEN##,$PREFIX/,g" "$binpath/$app_name"
+  sed -i -e "s,##KW_INSTALL_PREFIX_TOKEN##,$PREFIX,g" "$binpath/$app_name"
   sed -i -e "/##BEGIN-DEV-MODE##/,/##END-DEV-MODE##/ d" "$binpath/$app_name"
 
   # Lib files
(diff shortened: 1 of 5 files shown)


setup.sh
```diff
@@ -250,7 +250,7 @@ function synchronize_files()
   cmd_output_manager "cp $app_name $binpath" "$verbose"
   ASSERT_IF_NOT_EQ_ZERO "The command 'cp $app_name $binpath' failed" "$?"
 
-  sed -i -e "s,##KW_INSTALL_PREFIX_TOKEN##,$PREFIX/,g" "$binpath/$app_name"
+  sed -i -e "s,##KW_INSTALL_PREFIX_TOKEN##,$PREFIX,g" "$binpath/$app_name"
```
alan-barzilay, line 253: was this intentional? i dont see how it relates to the rest of the commit

alan-barzilay:
I don´t see anything wrong with this PR, LGTM.
But I'm not sure if it is the best approach. It is for sure easier and quicker way to fix this but maybe we shouldn't allow 2 files to have the same name, personally I find it confusing having multiple utils files. A more tedious solution would be to rename them with some sort of qualifier such as testing_utils or kernel_install_utils. 

I'm good either way, whatever you prefer

setup.sh
joaoseckler: This was causing KW_LIB_DIR to have two slashes in it's name, making the code in `include_manager` not function properly

joaoseckler:
The last update to this PR adds a test to the new feature.

alan-barzilay (approved):
LGTM!
just a quick question, this include_test files used in the new test won't be picked up by the run_tests as tests of their own?

rodrigosiqueira:
Merged!

Issue #189, opened 2020-07-06
Add a mechanism for change kernel name via kw


rodrigosiqueira:
A kw user requested a mechanism for adding a kernel name. It could be nice if we can handle a kernel name via kw, for example, in the `kworkflow.config`:

```
kernel_name=AN-COOL-NAME
```
we could also support it via command line:
```
kw deploy --name AN-COOL-NAME
```

Issue #187, opened 2020-06-07
kw d --ls  shows ".old" and kernels not installed via kw


rodrigosiqueira:
**Expected behavior:**
1. kw d --ls
2. Only show kernels under kw management and any other ".old" kernel

**Current Behavior:**
1.kw d --ls
2. Show all kernels find in `/boot`

Kernels with ".old" extensions only work as a fallback option and it is associated with a specific kernel; users should not be able to uninstall a kernel based on a ".old" version. Additionally, it is dangerous to use kw to remove a kernel installed via package manager such as apt or pacman.

rodrigosiqueira:
https://github.com/kworkflow/kworkflow/pull/472

PR #362, opened 2021-07-08
Implement package building and deploy

joaoseckler:
This PR introduces to kw the ability to handle packages. This is still work in progress. The following steps include:

 - Include configuration options for package building and deploying (`build_as_package`, etc)
 - In `kw b --pkg`, parse the command line and accept types of packages (debian, rpm, etc)
 - In `kw d --pkg`, parse the command line and accept a path to a package
 - Handle uninstallation of package
 - Check if `kw d --list` is listing package-installed kernels properly
 - A tutorial on using packages

### DIFF
--- a/src/build.sh
+++ b/src/build.sh
@@ -48,6 +48,8 @@ function kernel_build()
   local menu_config
   local parallel_cores
   local doc_type
+  local pkg
+  local pkg_rule
 
   parse_build_options "$@"
   if [[ "$?" != 0 ]]; then
@@ -59,6 +61,7 @@ function kernel_build()
   menu_config=${options_values['MENU_CONFIG']}
   parallel_cores=${options_values['PARALLEL_CORES']}
   doc_type=${options_values['DOC_TYPE']}
+  pkg=${options_values['PKG']}
 
   if [[ -n "${options_values['INFO']}" ]]; then
     build_info
@@ -86,13 +89,22 @@ function kernel_build()
     exit 125 # ECANCELED
   fi
 
-  if [ -x "$(command -v nproc)" ]; then
+  if [[ -x "$(command -v nproc)" ]]; then
     parallel_cores=$(nproc --all)
   else
     parallel_cores=$(grep -c ^processor /proc/cpuinfo)
   fi
 
   command="make -j$parallel_cores ARCH=$arch $cross_compile"
+
+  if [[ -n "$pkg" ]]; then
+    pkg_rule="$(pkg_get_rule "$pkg")"
+    if [[ "$?" == 95 ]]; then
+      complain "Package type '$pkg' not supported yet. Consider sending a patch for that!"
+      return 95 # ENOTSUP
+    fi
+    command="$command $pkg_rule"
+  fi
 
   start=$(date +%s)
   cmd_manager "$flag" "$command"
@@ -107,13 +119,15 @@ function kernel_build()
     statistics_manager "build" "$runtime"
   fi
 
+  pkg_info "$pkg"
+
   return "$ret"
 }
 
 function parse_build_options()
 {
-  local long_options='help,info,menu,doc'
-  local short_options='h,i,n,d'
+  local long_options='help,info,menu,doc,pkg:'
+  local short_options='h,i,n,d,p'
   local doc_type
 
   options="$(getopt \
@@ -133,6 +147,7 @@ function parse_build_options()
   options_values['PARALLEL_CORES']=1
   options_values['INFO']=''
   options_values['DOC_TYPE']=''
+  options_values['PKG']="${configurations[use_pkg]}"
 
   eval "set -- $options"
 
@@ -147,23 +162,90 @@ function parse_build_options()
         shift
         ;;
       --menu | -n)
-        options_values['MENU_CONFIG']="${configurations[menu_config]:-nconfig}"
+        options_values['MENU_CONFIG']="${configurations[menu_config]:-"nconfig"}"
         shift
         ;;
       --doc | -d)
         options_values['DOC_TYPE']="${configurations[doc_type]:-htmldocs}"
         shift
         ;;
+      --pkg | -p)
+        options_values['PKG']="$2" # Default value is debian
+        shift 2
+        ;;
       --)
         shift
         ;;
       *)
-        complain "Invalid option: $option"
-        exit 22 # EINVAL
+        complain "Invalid option: $1"
+        return 22 # EINVAL
+        shift
         ;;
     esac
   done
 
+  if [[ "${options_values['PKG']}" == 'no' ]]; then
+    options_values['PKG']=''
+  fi
+}
+
+# This function returns the makefile rule for creating a package, according to
+# the type of package required
+#
+# @pkg_type The type of package created
+#
+# Returns:
+# The command necessary to create a package
+function pkg_get_rule()
+{
+  local pkg_type="$1"
+
+  case "$pkg_type" in
+    debian)
+      echo 'bindeb-pkg'
+      return
+      ;;
+    *)
+      return 95 # ENOTSUP
+      ;;
+  esac
+}
+
+# This function prints information about recently created debian packages
+#
+# @pkg_type The type of package created. If empty, nothing is done.
+#
+function pkg_info()
+{
+  local pkg_type="$1"
+
+  case "$pkg_type" in
+    debian)
+      pkg_info_deb
+      return
+      ;;
+  esac
+}
+
+# This function prints information about recently created debian packages
+#
+function pkg_info_deb()
+{
+  local revision
+  local version
+  local release
+  local created_packages
+
+  revision="$(($(cat .version) - 1))"
+  version="$(get_kernel_version 2> /dev/null)"
+  release="$(get_kernel_release 2> /dev/null)"
+
+  created_packages=(../*"${release}-${revision}"*'.deb')
+
+  say 'The following packages have been created:'
+  for p in "${created_packages[@]}"; do
+    say "  $(basename "$p")"
+  done
 }
 
 function build_help()
@@ -177,5 +259,6 @@ function build_help()
     '  build - Build kernel' \
     '  build (-n | --menu) - Open kernel menu config' \
     '  build (-i | --info) - Display build information' \
-    '  build (-d | --doc) - Build kernel documentation'
+    '  build (-d | --doc) - Build kernel documentation' \
+    '  build (-p | --pkg ) - Build kernel as a package'
 }
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -48,6 +48,7 @@ function kernel_deploy()
   local runtime=0
   local ret=0
   local list_all
+  local pkg=''
 
   if [[ "$1" =~ -h|--help ]]; then
     deploy_help "$1"
@@ -69,6 +70,7 @@ function kernel_deploy()
   uninstall="${options_values['UNINSTALL']}"
   uninstall_force="${options_values['UNINSTALL_FORCE']}"
   remote="${remote_parameters['REMOTE']}"
+  pkg="${options_values['PKG']}"
 
   if [[ "$list" == 1 || "$single_line" == 1 || "$list_all" == 1 ]]; then
     say "Available kernels:"
@@ -99,6 +101,11 @@ function kernel_deploy()
   signal_manager 'cleanup' || warning 'Was not able to set signal handler'
 
   if [[ "$target" == "$VM_TARGET" ]]; then
+    if [[ -n "$pkg" ]]; then
+      complain "Package deployment not supported for VM"
+      return 95 # ENOTSUP
+    fi
+
     vm_mount
     ret="$?"
     if [[ "$ret" != 0 ]]; then
@@ -111,17 +118,19 @@ function kernel_deploy()
   # we can break the boot. For security reason, every time we want to deploy a
   # new kernel version we also update all modules; maybe one day we can change
   # it, but for now this looks the safe option.
-  start=$(date +%s)
-  modules_install '' "$target"
-  end=$(date +%s)
-  runtime=$((end - start))
+  if [[ -z "$pkg" ]]; then
+    start=$(date +%s)
+    modules_install '' "$target"
+    end=$(date +%s)
+    runtime=$((end - start))
+  fi
 
   if [[ "$modules" == 0 ]]; then
     start=$(date +%s)
     # Update name: release + alias
     name=$(make kernelrelease)
 
-    kernel_install "$reboot" "$name" "" "$target"
+    kernel_install "$reboot" "$name" "" "$target" "$pkg"
     end=$(date +%s)
     runtime=$((runtime + (end - start)))
     statistics_manager "deploy" "$runtime"
@@ -152,8 +161,8 @@ function parse_deploy_options()
   local remote
   local options
   local long_options='remote:,local,vm,reboot,modules,list,ls-line,uninstall:'
-  long_options+=',list-all,force'
-  local short_options='r,m,l,s,u:,a,f'
+  long_options+=',list-all,force,pkg:,no-pkg'
+  local short_options='r,m,l,s,u:,a,f,p'
 
   options="$(kw_parse "$short_options" "$long_options" "$@")"
 
@@ -171,6 +180,7 @@ function parse_deploy_options()
   options_values['REBOOT']=0
   options_values['MENU_CONFIG']='nconfig'
   options_values['LS_ALL']=''
+  options_values['PKG']="${configurations['use_pkg']}"
 
   remote_parameters['REMOTE']=''
 
@@ -233,6 +243,14 @@ function parse_deploy_options()
         options_values['LS_LINE']=1
         shift
         ;;
+      --pkg)
+        options_values['PKG']="$2"
+        shift 2
+        ;;
+      --no-pkg)
+        options_values['PKG']='no'
+        shift
+        ;;
       --uninstall | -u)
         if [[ "$2" =~ ^-- ]]; then
           options_values['ERROR']='Uninstall requires a kernel name'
@@ -255,6 +273,10 @@ function parse_deploy_options()
         ;;
     esac
   done
+
+  if [[ "${options_values['PKG']}" == 'no' ]]; then
+    options_values['PKG']=''
+  fi
 
   case "${options_values['TARGET']}" in
     1 | 2 | 3) ;;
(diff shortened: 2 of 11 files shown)


src/build.sh
```diff
@@ -83,6 +87,9 @@ function kernel_build()
         cmd_manager "$flag" "$command"
         return
         ;;
+      --pkg | -p)
+        pkg='debian' # todo: read which type of package from config file/command line
```
rodrigosiqueira, line 91: Yeah... let's read this information from the file/command line in this series.

src/build.sh
```diff
@@ -98,6 +105,12 @@ function kernel_build()
 
   command="make -j$PARALLEL_CORES ARCH=$arch $CROSS_COMPILE"
 
+  if [ -n "$pkg" ]; then
```
rodrigosiqueira, line 108: let's use `[[`

src/build.sh
```diff
@@ -98,6 +105,12 @@ function kernel_build()
...
+    pkg_rule="$(pkg_get_rule "$pkg")"
+    [ -n "$pkg_rule" ] && command="$command $pkg_rule"
```
rodrigosiqueira, line 110: `[[`

src/build.sh
```diff
@@ -111,14 +124,87 @@ function kernel_build()
...
+  pkg_info "$pkg"
+
   return "$ret"
 }
 
+# This function returns the makefile rule for creating a package, according to
+# the type of package required
+#
+# @pkg_type The type of package created
+#
```
rodrigosiqueira, line 136: You have a return here (`echo`), please, document it.

src/build.sh
```diff
@@ -111,14 +124,87 @@ function kernel_build()
...
 
+  pkg_info "$pkg"
+
   return "$ret"
 }
 
+# This function returns the makefile rule for creating a package, according to
+# the type of package required
+#
+# @pkg_type The type of package created
```
rodrigosiqueira, line 135: @pkg_type **:**

src/build.sh
```diff
@@ -111,14 +124,87 @@ function kernel_build()
...
+#
+function pkg_get_rule()
+{
+  local pkg_type="$1"
+
+  case "$pkg_type" in
+    debian)
+      echo 'bindeb-pkg'
+      return
+      ;;
```
rodrigosiqueira, line 145: Let's handle an unsupported distro by adding a message and return 22.

src/build.sh
```diff
@@ -111,14 +124,87 @@ function kernel_build()
...
+  local dbg_package
+  local pkg_info_path="$1"
+
+  version="$(get_kernel_version 2> /dev/null)"
+  release="$(get_kernel_release 2> /dev/null)"
+
+  headers_package=("../linux-headers-${release}_${release}-${revision}"*".deb")
+  libc_package=("../linux-libc-dev_${release}-${revision}"*".deb")
+  image_package=("../linux-image-${release}_${release}-${revision}"*".deb")
+  dbg_package=("../linux-image-${release}-dbg_${release}-${revision}"*".deb")
```
rodrigosiqueira, line 187: Hmmm... maybe we should check if these files still exist before showing them.

src/build.sh
```diff
@@ -111,14 +124,87 @@ function kernel_build()
...
+
+  say 'The following packages have been created:'
```
rodrigosiqueira, line 189: Let's add date and hour information here.

src/build.sh
```diff
@@ -111,14 +124,87 @@ function kernel_build()
...
+  echo "  ${headers_package[0]}"
+  echo "  ${libc_package[0]}"
+  echo "  ${image_package[0]}"
+  echo "  ${dbg_package[0]}"
+
+  [ -d "$(dirname "$pkg_info_path")" ] || return 2 # ENOENT
```
rodrigosiqueira, line 195: `[[`

tests/build_test.sh
```diff
@@ -119,4 +120,11 @@ function test_build_info()
   rm .config
 }
 
+function test_kernel_build_pkg()
+{
+  output=$(kernel_build 'TEST_MODE' '--pkg' | head -1)
+  expected_result="make -j$PARALLEL_CORES ARCH=x86_64  bindeb-pkg"
+  assertEquals "($LINENO)" "$expected_result" "$output"
+}
```
rodrigosiqueira, line 128: I suppose that you'll add more tests, right? I mean, you added a lot of new functions and new behavior... we need to capture them in the test.

documentation/man/kw.rst
```diff
@@ -126,6 +126,9 @@ need root access. 'kw deploy –local' should not be executed with sudo or root.
 8. --uninstall|-u: Remove a single kernel or multiple kernels; for removing
    multiple kernels it is necessary to separate them with comma.
 
+9. --pkg|-p: Deploy a kernel package. By default, kw will deploy the
+   last package it built.
```
rodrigosiqueira, line 130: Expand the documentation; consider adding a paragraph about that.

src/deploy.sh
```diff
@@ -433,6 +434,7 @@ function kernel_deploy()
   test_mode="${options_values['TEST_MODE']}"
   remote="${options_values['REMOTE']}"
   uninstall="${options_values["UNINSTALL"]}"
+  pkg="${options_values["PKG"]}"
```
rodrigosiqueira, line 437: 'PKG' instead of "PKG"

src/deploy.sh
```diff
@@ -476,6 +478,11 @@ function kernel_deploy()
     fi
   fi
 
+  if [ -n "$pkg" ]; then
```
rodrigosiqueira, line 481: `[[`

src/deploy.sh
```diff
@@ -476,6 +478,11 @@ function kernel_deploy()
...
+    pkg_deploy "" "$pkg" "$target" "$remote"
```
rodrigosiqueira, line 482: '' instead of ""

src/deploy.sh
```diff
@@ -509,10 +516,70 @@ function kernel_deploy()
...
+  local target="$3"
+  local remote_raw="$4"
+  local package_path
+  local target_distro
+  local host
+  local port
+
+  package_path="$(tail -1 < "$KW_LIB_DIR/PKG_INFO")"
+  package_type="$(head -1 | "$KW_LIB_DIR/PKG_INFO")"
+  package_name="$(basename "$package_path")"
```
rodrigosiqueira, line 539: Could you add a few lines explaining what's going on here? I think you can also document it a little bit in the function doc and in the commit message.

src/deploy.sh
```diff
@@ -509,10 +516,70 @@ function kernel_deploy()
...
+
+  install_cmd="$(pkg_get_install_command "$package_type")"
+
+  host=$(get_based_on_delimiter "$remote_raw" ":" 1)
+  port=$(get_based_on_delimiter "$remote_raw" ":" 2)
+  host=$(get_based_on_delimiter "$host" "@" 2)
```
rodrigosiqueira, line 545: Update to use the new version of this.

src/deploy.sh
```diff
@@ -509,10 +516,70 @@ function kernel_deploy()
...
+      ;;
+  esac
+}
+
 function deploy_help()
 {
   echo -e "kw deploy|d installs kernel and modules:\n" \
-    "\tdeploy,d [--remote [REMOTE:PORT]|--local|--vm] [--reboot|-r] [--modules|-m]\n" \
+    "\tdeploy,d [--remote [REMOTE:PORT]|--local|--vm] [--reboot|-r]\n" \
+    "\t         [--modules|-m] [--pkg|-p]\n" \
```
rodrigosiqueira, line 582: you forgot to add `deploy`

rodrigosiqueira: Actually, why did you add this together with modules? Kw provides the option to only deploy modules, the way that you put here sounds like your pkg feature will also work for modules-update.

src/deploy.sh
rodrigosiqueira: Notice that you are skipping the statistic data. It would be nice if you also consider keeping track of the build time using pkg and differentiating it from the normal build, enabling us to compare performance.
@magalilemes might be interested in this part.

rodrigosiqueira: Hmmm... I also think that you should keep this part close to the kernel_install code. Did you check if we can only build the Debian package for modules? I doubt we have this option, but I think it is worth checking.

src/deploy.sh
```diff
@@ -509,10 +516,70 @@ function kernel_deploy()
...
 }
 
+# Deploy a package
+#
+# @flag How to display commands. See `cmd_manager` inside `src/kwlib.sh`
+# @pkg_type The type of package
+# @target VM, local or remote?
+# @remote_raw Address of remote target
+#
+function pkg_deploy()
```
rodrigosiqueira, line 526: I'm not sure if I like this approach because we are duplicating code from `kernel_install` and complicating how we read the code. I agree that we can have a specific function for deploying packages in a remote machine but just for it.

rodrigosiqueira (changes requested):
Hi,

First of all, thanks for keeping working on this feature. I think you made great progress in this area.

I added a lot of comments, but I have these generic points:

1. I expect tests for this part;
2. Check if we need to add something else in the `./setup.sh` to manage any new dependence.
3. When I play with custom kernels, I like to add meaningful suffix such "-FEATURE-X". This helps me to quickly check the kernel version that I want to boot, but the Debian package option does not allow it, as you can see from the below output:
```
dpkg-buildpackage: info: host architecture amd64
 dpkg-source --before-build .
dpkg-source: error: source package name 'linux-5.11.0-TEST+' is illegal: character 'T' not allowed
dpkg-buildpackage: error: dpkg-source --before-build . subprocess returned exit status 255
make[1]: *** [scripts/Makefile.package:83: bindeb-pkg] Error 255
```
I see two alternatives here:
- Check if we can make the Debian package more relaxed about this naming rule.
- Inform the user that they need to change the kernel name to something like "blablab-version"
Also, I strongly recommend that you also work on this issue:
https://github.com/kworkflow/kworkflow/issues/189
4. Do you know if kernel has any support for generating Arch packages? I'm asking because we officially support Arch.
5. Since you're working on the deploy, you will need to handle the kernel list and uninstall it soon. Consider refactoring the list option (https://github.com/kworkflow/kworkflow/issues/187)

rodrigosiqueira:
@joaoseckler any updates?

joaoseckler:
The last update to this PR updates the building of packages.

We also propose that, unlike previous version, when building, kw will not remember it's last built package, but rather search for it based on the current directory or accept it on the command line. What do you think @rodrigosiqueira

rodrigosiqueira:
Hi @joaoseckler , I agree with you.

Could you rebase your change? I'll review it after we get the CI result.

rodrigosiqueira:
Hi @joaoseckler ,
I already merged many of the commits that you have in this PR. To make things easier to review, could you rebase your change?

src/build.sh
rodrigosiqueira: Let's use `[[` to keep the consistency with the code.

src/build.sh
```diff
@@ -94,6 +98,16 @@ function kernel_build()
 
   command="make -j$parallel_cores ARCH=$arch $cross_compile"
 
+  if [ -n "$pkg" ]; then
+    revision=$(cat .version) # Has to be gathered before building
```
rodrigosiqueira, line 102: What does this version mean? Does the kernel generate this? Maybe add a top comment that describes this version file and why we need it.

src/build.sh
```diff
@@ -147,23 +164,91 @@ function parse_build_options()
         shift
         ;;
       --menu | -n)
-        options_values['MENU_CONFIG']="${configurations[menu_config]:-nconfig}"
+        options_values['MENU_CONFIG']="${configurations[menu_config]:-"nconfig"}"
         shift
         ;;
       --doc | -d)
-        options_values['DOC_TYPE']="${configurations[doc_type]:-htmldocs}"
+        options_values['DOC_TYPE']="${configurations[doc_type]:-"htmldocs"}"
```
rodrigosiqueira, line 171: This is a great change, but can we create a separate commit for this change and the one in the `--menu`? I'm asking because this is not part of this pkg change.

src/build.sh
rodrigosiqueira: Hi @joaoseckler , do not forget to document the return value here (in this case, the echo value).

src/build.sh
```diff
@@ -94,6 +98,16 @@ function kernel_build()
 
   command="make -j$parallel_cores ARCH=$arch $cross_compile"
 
+  if [ -n "$pkg" ]; then
+    revision=$(cat .version) # Has to be gathered before building
+    pkg_rule="$(pkg_get_rule "$pkg")"
+    if [[ "$?" == 95 ]]; then
+      complain "Package type '$pkg' not supported"
```
rodrigosiqueira, line 105: How about something like this:
"[..] not supported yet. Consider sending a patch for that."

src/build.sh
```diff
@@ -147,23 +164,91 @@ function parse_build_options()
...
+# @revision The revision number of the package
+#
+function pkg_info()
+{
+  local pkg_type="$1"
+  local revision="$2"
+
+  case "$pkg_type" in
+    debian)
+      pkg_info_debian "$revision"
```
rodrigosiqueira, line 226: I think you can make things more isolated per distro type if you move `revision=$(cat .version) # Has to be gathered before building` to the specific function. For example, I think this `revision` should be inside `pkg_info_debian()`. If RPM packages use something different, they can encapsulate it inside `pkg_info_rpm`.

src/build.sh
```diff
@@ -147,23 +164,91 @@ function parse_build_options()
...
+      return
+      ;;
+  esac
+}
+
+# This function prints information about recently created debian packages
+#
+# @revision The revision number of the package
+#
+function pkg_info_debian()
```
rodrigosiqueira, line 236: how about use `pkg_info_deb`? I think this is a better name because we can use the deb package for different OSes based on Debian.

src/build.sh
```diff
@@ -94,6 +98,16 @@ function kernel_build()
...
   command="make -j$parallel_cores ARCH=$arch $cross_compile"
 
+  if [ -n "$pkg" ]; then
+    revision=$(cat .version) # Has to be gathered before building
+    pkg_rule="$(pkg_get_rule "$pkg")"
+    if [[ "$?" == 95 ]]; then
+      complain "Package type '$pkg' not supported"
+      return 95 # ENOTSUP
+    fi
+    command="$command $pkg_rule"
```
rodrigosiqueira, line 108: I suppose that the pkg magic happens here, right? How about creating a function named `create_kernel_pkg`? When I was reading your code, I felt that this deb package feature was a bit hidden; a function dedicated to that will make the code easier to read.

rodrigosiqueira: Additionally, you need something to check if we can create the target package in the host system. For example, If I try to use this command in ArchLinux, it will fail due to the lack of a deb package. For now, you can add a simple check or warning that lets the user know that this is only available for Debian-based systems. Finally, if the host is Debian-based, we need to check if the user has the build package.

rodrigosiqueira (changes requested):
Hi @joaoseckler ,
Thanks a lot for your PR; I added a couple of comments.

src/build.sh
joaoseckler: The ideia of using a command like this and not a dedicated function was to use the same code for choosing the number of threads and statistics manager as the main build function. If we use a function for that, there will be code duplication. What do you think?

joaoseckler: > Additionally, you need something to check if we can create the target package in the host system. For example, If I try to use this command in ArchLinux, it will fail due to the lack of a deb package.

you mean for the lack of the dpkg command, right?

> For now, you can add a simple check or warning that lets the user know that this is only available for Debian-based systems. Finally, if the host is Debian-based, we need to check if the user has the build package.

Hmmm what build package? For the build part, I think it'd suffice to check whether dpkg is a valid command, right? For the deploy part, there are more complications...

rodrigosiqueira: > The ideia of using a command like this and not a dedicated function was to use the same code for choosing the number of threads and statistics manager as the main build function. If we use a function for that, there will be code duplication. What do you think?

Are you sure that we can use the same command? For example, for generating an RPM package, do you know if we need something else?
In general, I prefer to create a new function to make things easier to read, test, and document.

rodrigosiqueira: > > Additionally, you need something to check if we can create the target package in the host system. For example, If I try to use this command in ArchLinux, it will fail due to the lack of a deb package.
> 
> you mean for the lack of the dpkg command, right?

yes

> > For now, you can add a simple check or warning that lets the user know that this is only available for Debian-based systems. Finally, if the host is Debian-based, we need to check if the user has the build package.
> 
> Hmmm what build package? For the build part, I think it'd suffice to check whether dpkg is a valid command, right? For the deploy part, there are more complications...

Actually, the deploy part should be easier because we always identify the target distro. If the target is not a Debian-based system, and the user is trying to deploy a deb file, you can block the process.

joaoseckler: > Are you sure that we can use the same command? For example, for generating an RPM package, do you know if we need something else?

The thing is it's not an actual command, it's an argument to the make command already being built by kw build. This approach allows for the creation of the following packages:

```
  rpm-pkg             - Build both source and binary RPM kernel packages
  binrpm-pkg          - Build only the binary kernel RPM package
  deb-pkg             - Build both source and binary deb kernel packages
  bindeb-pkg          - Build only the binary kernel deb package
  snap-pkg            - Build only the binary kernel snap package
                        (will connect to external hosts)
  dir-pkg             - Build the kernel as a plain directory structure
  tar-pkg             - Build the kernel as an uncompressed tarball
  targz-pkg           - Build the kernel as a gzip compressed tarball
  tarbz2-pkg          - Build the kernel as a bzip2 compressed tarball
  tarxz-pkg           - Build the kernel as a xz compressed tarball
  perf-tar-src-pkg    - Build perf-5.13.0-rc6.tar source tarball
  perf-targz-src-pkg  - Build perf-5.13.0-rc6.tar.gz source tarball
  perf-tarbz2-src-pkg - Build perf-5.13.0-rc6.tar.bz2 source tarball
  perf-tarxz-src-pkg  - Build perf-5.13.0-rc6.tar.xz source tarball
```

(taken from kernel's `make help`)

joaoseckler:
Future work on this PR includes:

 - Checking if target machine supports the installation command (e.g. check if machine supports the dpkg command)
 - Working VM deploy

rodrigosiqueira:
I tried to rebase this change, but it is too old. I'll close this PR.
Thanks.

PR #365, opened 2021-07-11
Improve `zsh` compatibility and setup

### LINKED ISSUE #309: Installation always appends to `.zshrc`
Whenever I run `./setup.sh -i` it appends the same lines to my `.zshrc` file, when I checked my file it had well over 20 repetitions of the same lines at the end. It was slowing down my environment loading
We should do a check to see if the lines need to be added, or maybe just source a file for the settings. Or some other solution as to not clutter the users settings.

kwy95:
This PR adds `safe_append` which is then used to avoid polluting files during setup.
It also updates path for zsh users.

This fixes #309

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -109,14 +109,14 @@ function remove_kw_from_PATH_variable()
 
 function update_path()
 {
-  local new_path=""
+  local shellrc=${1:-".bashrc"}
 
   IFS=':' read -ra ALL_PATHS <<< "$PATH"
   for path in "${ALL_PATHS[@]}"; do
     [[ "$path" -ef "$binpath" ]] && return
   done
 
-  echo "PATH=$HOME/.local/bin:\$PATH # kw" >> "$HOME/.bashrc"
+  safe_append "PATH=$HOME/.local/bin:\$PATH # kw" "$HOME/$shellrc"
 }
 
 function update_current_bash()
@@ -305,6 +305,7 @@ function synchronize_files()
 
       safe_append "$zshcomp" "$HOME/.zshrc"
       append_bashcompletion ".zshrc"
+      update_path ".zshrc"
     else
       warning "Unable to find a .zshrc file."
     fi
--- a/setup.sh
+++ b/setup.sh
@@ -300,8 +300,10 @@ function synchronize_files()
   if command_exists "zsh"; then
     # Add tabcompletion to zshrc
     if [[ -f "$HOME/.zshrc" || -L "$HOME/.zshrc" ]]; then
-      echo "# Enable bash completion for zsh" >> "$HOME/.zshrc"
-      echo "autoload bashcompinit && bashcompinit" >> "$HOME/.zshrc"
+      local zshcomp=$'# Enable bash completion for zsh\n'
+      zshcomp+='autoload bashcompinit && bashcompinit'
+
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -300,9 +300,10 @@ function synchronize_files()
   if command_exists "zsh"; then
     # Add tabcompletion to zshrc
     if [[ -f "$HOME/.zshrc" || -L "$HOME/.zshrc" ]]; then
-      echo "# Enable bash completion for zsh" >> "$HOME/.zshrc"
-      echo "autoload bashcompinit && bashcompinit" >> "$HOME/.zshrc"
+      safe_append '# Enable bash completion for zsh' "$HOME/.zshrc"
```
alan-barzilay, line 303: couldnt this 2 lines be merged into one?

setup.sh
```diff
@@ -319,8 +320,15 @@ function append_bashcompletion()
 {
   local shellrc="$1"
 
-  echo "# $app_name" >> "$HOME/$shellrc"
-  echo "source $libdir/$BASH_AUTOCOMPLETE.sh" >> "$HOME/$shellrc"
+  safe_append "# $app_name" "$HOME/$shellrc"
```
alan-barzilay, line 323: same here

alan-barzilay (changes requested):

setup.sh
kwy95: Apparently yes, I was afraid that the text would be split into 2 arguments and that would mess up the function, but I changed it and it seems to work.

setup.sh
```diff
@@ -300,9 +300,12 @@ function synchronize_files()
   if command_exists "zsh"; then
     # Add tabcompletion to zshrc
     if [[ -f "$HOME/.zshrc" || -L "$HOME/.zshrc" ]]; then
-      echo "# Enable bash completion for zsh" >> "$HOME/.zshrc"
-      echo "autoload bashcompinit && bashcompinit" >> "$HOME/.zshrc"
+      local zshcomp=$'# Enable bash completion for zsh\n'
```
alan-barzilay, line 303: couldnt this be just `local zshcomp=$'# Enable bash completion for zsh\n autoload bashcompinit && bashcompinit'` ?

kwy95: Yes, but it takes up more than 80 characters, so I split it.

alan-barzilay: I see, good call, but I wonder if there is a better way of splitting this to be less than 80 characters. Does it work if you introduce a real break line instead of \n?

kwy95: I'm not sure what you mean with "real line break", can you elaborate?

alan-barzilay: literally pressing enter instead of using \n

kwy95: I think it's possible, but it would mess with the indentation, unless I'm missing something. It would look something like:
```
if command_exists "zsh"; then
  if [[ -f "$HOME/.zshrc" || -L "$HOME/.zshrc" ]]; then
    local zshcomp='# Enable bash completion for zsh
autoload bashcompinit && bashcompinit'
    safe_append "$zshcomp" "$HOME/.zshrc"
  fi
fi
```
If you have a better way of doing it let me know.

alan-barzilay (approved):

rodrigosiqueira:
Nice PR! Merged, thanks

PR #376, opened 2021-07-11
shellcheck: SC2091: use eval

alan-barzilay:
SC2091: Remove surrounding $() to avoid executing output
(or use eval if intentional).

### DIFF
--- a/src/kwio.sh
+++ b/src/kwio.sh
@@ -25,7 +25,7 @@ function alert_completion()
     opts="${configurations[alert]}"
   fi
 
-  grep -o . <<< "$opts" | while read -r option; do
+  while read -rN 1 option; do
     if [ "$option" == "v" ]; then
       if command_exists "${configurations[visual_alert_command]} &"; then
         eval "${configurations[visual_alert_command]} &"
@@ -45,7 +45,7 @@ function alert_completion()
         warning "Check if the necessary packages are installed."
       fi
     fi
-  done
+  done <<< "$opts"
 }
 
 # Print colored message. This function verifies if stdout
--- a/tests/kwio_test.sh
+++ b/tests/kwio_test.sh
@@ -30,27 +30,32 @@ function test_alert_completion_options()
   configurations["alert"]="n"
 
   rm -f "$sound_file" "$visual_file"
-  $(alert_completion "" "--alert=vs")
+  alert_completion "" "--alert=vs"
+  wait "$!"
   [[ -f "$sound_file" && -f "$visual_file" ]]
   assertTrue "Alert's vs option didn't work." $?
 
   rm -f "$sound_file" "$visual_file"
-  $(alert_completion "" "--alert=sv")
+  alert_completion "" "--alert=sv"
+  wait "$!"
   [[ -f "$sound_file" && -f "$visual_file" ]]
   assertTrue "Alert's sv option didn't work." $?
 
   rm -f "$sound_file" "$visual_file"
-  $(alert_completion "" "--alert=s")
+  alert_completion "" "--alert=s"
+  wait "$!"
   [[ -f "$sound_file" && ! -f "$visual_file" ]]
   assertTrue "Alert's s option didn't work." $?
 
   rm -f "$sound_file" "$visual_file"
-  $(alert_completion "" "--alert=v")
+  alert_completion "" "--alert=v"
+  wait "$!"
   [[ ! -f "$sound_file" && -f "$visual_file" ]]
   assertTrue "Alert's v option didn't work." $?
 
   rm -f "$sound_file" "$visual_file"
-  $(alert_completion "" "--alert=n")
+  alert_completion "" "--alert=n"
+  wait "$!"
   [[ ! -f "$sound_file" && ! -f "$visual_file" ]]
   assertTrue "Alert's n option didn't work." $?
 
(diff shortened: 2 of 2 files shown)


joaoseckler (commented):
Consider shellcheck's [wiki page](https://github.com/koalaman/shellcheck/wiki/SC2091#rationale) about this warning:

`If you do have a command that outputs a second command, similar to how ssh-agent outputs export commands to run, then you should do this via eval.`

But this is not the case here! Actually alert_completion does not output nothing. The `$(...)` should just be abandoned. Well, that is what I thought. But if you do that it doesn't work! Any ideas why?

rodrigosiqueira:
Hi @alan-barzilay and @joaoseckler ,

As @joaoseckler described, we don't need `eval` nor `$()`. The reason that you saw failures is related to synchronization. See the below example that solver your issue:

```
diff --git a/tests/kwio_test.sh b/tests/kwio_test.sh
index ac065b5..7a3915e 100755
--- a/tests/kwio_test.sh
+++ b/tests/kwio_test.sh
@@ -30,7 +30,8 @@ function test_alert_completion_options()
   configurations["alert"]="n"
 
   rm -f "$sound_file" "$visual_file"
-  eval "$(alert_completion "" "--alert=vs")"
+  alert_completion "" "--alert=vs"
+  sleep 0.1
   [[ -f "$sound_file" && -f "$visual_file" ]]
   assertTrue "Alert's vs option didn't work." $?
```

@joaoseckler, any better idea on how we can handle this concurrency case? I know that you had a similar issue with the signal handler.

joaoseckler:
Since siqueira discovered it is a concurrency problem, the ideal would be to `wait` the completion of the command, with

```
alert_completion "" "--alert=vs"
wait "$!"
```

But inside the alert_completion function the `&` is called inside a piped command and thus the father process does not now about it's children subprocesses to wait for. I think the easiest way of dealing with this is a small and welcome refactor of the `alert_completion` function, using the `-N` option of `read` instead of relying on the `grep -o .` (which is also more difficult to read). Both are used to get one character at each iteration. A patch with all these changes is reproduced below (apply with `$ git apply patch`):

```diff
diff --git a/src/kwio.sh b/src/kwio.sh
index 3124dac..6e89e3c 100644
--- a/src/kwio.sh
+++ b/src/kwio.sh
@@ -25,7 +25,7 @@ function alert_completion()
     opts="${configurations[alert]}"
   fi
 
-  grep -o . <<< "$opts" | while read option; do
+  while read -N 1 option; do
     if [ "$option" == "v" ]; then
       if command_exists "${configurations[visual_alert_command]} &"; then
         eval "${configurations[visual_alert_command]} &"
@@ -45,7 +45,7 @@ function alert_completion()
         warning "Check if the necessary packages are installed."
       fi
     fi
-  done
+  done <<< "$opts"
 }
 
 # Print colored message. This function verifies if stdout
diff --git a/tests/kwio_test.sh b/tests/kwio_test.sh
index 74ab2d7..5dfb122 100755
--- a/tests/kwio_test.sh
+++ b/tests/kwio_test.sh
@@ -30,27 +30,32 @@ function test_alert_completion_options()
   configurations["alert"]="n"
 
   rm -f "$sound_file" "$visual_file"
-  eval "$(alert_completion "" "--alert=vs")"
+  alert_completion "" "--alert=vs"
+  wait "$!"
   [[ -f "$sound_file" && -f "$visual_file" ]]
   assertTrue "Alert's vs option didn't work." $?
 
   rm -f "$sound_file" "$visual_file"
-  eval "$(alert_completion "" "--alert=sv")"
+  alert_completion "" "--alert=sv"
+  wait "$!"
   [[ -f "$sound_file" && -f "$visual_file" ]]
   assertTrue "Alert's sv option didn't work." $?
 
   rm -f "$sound_file" "$visual_file"
-  eval "$(alert_completion "" "--alert=s")"
+  alert_completion "" "--alert=s"
+  wait "$!"
   [[ -f "$sound_file" && ! -f "$visual_file" ]]
   assertTrue "Alert's s option didn't work." $?
 
   rm -f "$sound_file" "$visual_file"
-  eval "$(alert_completion "" "--alert=v")"
+  alert_completion "" "--alert=v"
+  wait "$!"
   [[ ! -f "$sound_file" && -f "$visual_file" ]]
   assertTrue "Alert's v option didn't work." $?
 
   rm -f "$sound_file" "$visual_file"
-  eval "$(alert_completion "" "--alert=n")"
+  alert_completion "" "--alert=n"
+  wait "$!"
   [[ ! -f "$sound_file" && ! -f "$visual_file" ]]
   assertTrue "Alert's n option didn't work." $?
 ```

rodrigosiqueira:
> Since siqueira discovered it is a concurrency problem, the ideal would be to `wait` the completion of the command, with
> 
> ```
> alert_completion "" "--alert=vs"
> wait "$!"
> ```
> 
> But inside the alert_completion function the `&` is called inside a piped command and thus the father process (in the test) does not now about its children subprocesses to wait for. I think the easiest way of dealing with this is a small and welcome refactor of the `alert_completion` function, using the `-N` option of `read` instead of relying on the `grep -o .` (which is also more difficult to read). Both are used to get one character at each iteration. A patch with all these changes is reproduced below (apply with `$ git apply patch`):
> 
> ```diff
> diff --git a/src/kwio.sh b/src/kwio.sh
> index 3124dac..6e89e3c 100644
> --- a/src/kwio.sh
> +++ b/src/kwio.sh
> @@ -25,7 +25,7 @@ function alert_completion()
>      opts="${configurations[alert]}"
>    fi
>  
> -  grep -o . <<< "$opts" | while read option; do
> +  while read -N 1 option; do
>      if [ "$option" == "v" ]; then
>        if command_exists "${configurations[visual_alert_command]} &"; then
>          eval "${configurations[visual_alert_command]} &"
> @@ -45,7 +45,7 @@ function alert_completion()
>          warning "Check if the necessary packages are installed."
>        fi
>      fi
> -  done
> +  done <<< "$opts"
>  }
>  
>  # Print colored message. This function verifies if stdout
> diff --git a/tests/kwio_test.sh b/tests/kwio_test.sh
> index 74ab2d7..5dfb122 100755
> --- a/tests/kwio_test.sh
> +++ b/tests/kwio_test.sh
> @@ -30,27 +30,32 @@ function test_alert_completion_options()
>    configurations["alert"]="n"
>  
>    rm -f "$sound_file" "$visual_file"
> -  eval "$(alert_completion "" "--alert=vs")"
> +  alert_completion "" "--alert=vs"
> +  wait "$!"
>    [[ -f "$sound_file" && -f "$visual_file" ]]
>    assertTrue "Alert's vs option didn't work." $?
>  
>    rm -f "$sound_file" "$visual_file"
> -  eval "$(alert_completion "" "--alert=sv")"
> +  alert_completion "" "--alert=sv"
> +  wait "$!"
>    [[ -f "$sound_file" && -f "$visual_file" ]]
>    assertTrue "Alert's sv option didn't work." $?
>  
>    rm -f "$sound_file" "$visual_file"
> -  eval "$(alert_completion "" "--alert=s")"
> +  alert_completion "" "--alert=s"
> +  wait "$!"
>    [[ -f "$sound_file" && ! -f "$visual_file" ]]
>    assertTrue "Alert's s option didn't work." $?
>  
>    rm -f "$sound_file" "$visual_file"
> -  eval "$(alert_completion "" "--alert=v")"
> +  alert_completion "" "--alert=v"
> +  wait "$!"
>    [[ ! -f "$sound_file" && -f "$visual_file" ]]
>    assertTrue "Alert's v option didn't work." $?
>  
>    rm -f "$sound_file" "$visual_file"
> -  eval "$(alert_completion "" "--alert=n")"
> +  alert_completion "" "--alert=n"
> +  wait "$!"
>    [[ ! -f "$sound_file" && ! -f "$visual_file" ]]
>    assertTrue "Alert's n option didn't work." $?
> ```

Nice! Thanks a lot @joaoseckler.

@alan-barzilay could you update your PR?

alan-barzilay:
@rodrigosiqueira done

rodrigosiqueira:
Merged, thanks

Issue #375, opened 2021-07-11
shellcheck and quotes


alan-barzilay:
These 4 shellcheck errors are kinda similar  

https://www.shellcheck.net/wiki/SC2046 -- Quote this to prevent word splitting
https://www.shellcheck.net/wiki/SC2053 -- Quote the right-hand side of != in [[ ]] to prevent glob matching.
https://www.shellcheck.net/wiki/SC2076 -- Remove quotes from right-hand side of =~, it'll match literally rather than as a regex.
https://github.com/koalaman/shellcheck/wiki/SC2206 -- Quote to prevent word splitting/globbing, or split robustly with mapfile or read -a.

alan-barzilay:
#321 also dealt with 2 similar errors dealing with quotes

rodrigosiqueira:
Done! Thanks

PR #378, opened 2021-07-12
Shellcheck and Quote errors

alan-barzilay:
This PR fixes 2 different shellcheck errors. Fixes part of #375 

SC2046: Quote this to prevent word splitting

SC2076: Remove quotes from right-hand side of =~ to match as a regex rather than literally

### DIFF
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -208,7 +208,7 @@ function is_a_patch()
   )
 
   for expected_str in "${PATCH_EXPECTED_STRINGS[@]}"; do
-    if [[ ! "$file_content" =~ "$expected_str" ]]; then
+    if [[ ! "$file_content" =~ $expected_str ]]; then
       return 1
     fi
   done
--- a/src/pomodoro.sh
+++ b/src/pomodoro.sh
@@ -287,8 +287,8 @@ function show_active_pomodoro_timebox()
     timebox=$(calculate_missing_time "$timebox" "$diff_time")
 
     say "Started at: $timestamp_to_date"
-    say "- Elapsed time:" $(sec_to_format "$diff_time")
-    say "- You still have" $(sec_to_format "$timebox")
+    say "- Elapsed time:" "$(sec_to_format "$diff_time")"
+    say "- You still have" "$(sec_to_format "$timebox")"
   done < "$POMODORO_LOG_FILE"
 }
 


alan-barzilay:
ok, so the last commit that fixes  SC2206 has a tiny issue that may warrant some discussion.

The basic idea is to make more explicit the way we split things when passing them to an array,
originally in `kwlib.sh` in the  `command_exists()` function at line 352 we had: `local package=($command)` 
This implicitly separates `command` on every space tab and/or newline. To fix this warning and pass our tests we need something like this:
```
  local package
  IFS=" " read -r -a package <<< "$command"
```
which besides being confusing to read messes with the IFS variable which I'm not 100% comfortable with, I believe this could introduce some really hard to debug errors in the future. I opted to simply disable this case as an exception, but I'm not sure if this is the best approach, I would love to have a second opinion about this.

(Personally I would rather just use command -v directly instead of this function so I also don't have any strong opinions on how to fix this)

Messing with this line will make the kwio tests break (since we dont touch kwio in this PR it was really hard to uncover this, but at least now I know how to use git bisect):
```
❯ ./run_tests.sh test kwio
Running test [kwio]
test_alert_completion_options
./tests/kwio_test.sh: line 33: The: command not found
ASSERT:Alert's vs option didn't work.
./tests/kwio_test.sh: line 38: The: command not found
ASSERT:Alert's sv option didn't work.
./tests/kwio_test.sh: line 43: The: command not found
ASSERT:Alert's s option didn't work.
./tests/kwio_test.sh: line 48: The: command not found
ASSERT:Alert's v option didn't work.
test_alert_completition_validate_config_file_options
./tests/kwio_test.sh: line 66: The: command not found
ASSERT:Alert's vs option didn't work.
./tests/kwio_test.sh: line 72: The: command not found
ASSERT:Alert's sv option didn't work.
./tests/kwio_test.sh: line 78: The: command not found
ASSERT:Alert's s option didn't work.
./tests/kwio_test.sh: line 84: The: command not found
ASSERT:Alert's v option didn't work.
test_alert_completion_visual_alert
ASSERT:Variable v should exist. expected:<The following command set in the visual_alert_command variable couldn't be run:
/bin/echo $COMMAND
Check if the necessary packages are installed.> but was:<TESTING COMMAND>
shunit2:ERROR test_alert_completion_visual_alert() returned non-zero return code.
test_alert_completion_sound_alert
ASSERT:Variable s should exist. expected:<The following command set in the sound_alert_command variable couldn't be run:
/bin/echo $COMMAND
Check if the necessary packages are installed.> but was:<TESTING COMMAND>
shunit2:ERROR test_alert_completion_sound_alert() returned non-zero return code.

Ran 4 tests.

FAILED (failures=12)
```

PS: I will move this to a new PR and delete this comment once I'm done with that

rodrigosiqueira (changes requested):
Please, update the first commit tittle.

alan-barzilay:
> Please, update the first commit tittle.

Ok, what should I update it to?

rodrigosiqueira:
Merged,
I rewrite both commit messages.

Thanks.

PR #380, opened 2021-07-12
shellcheck: SC2206: quote to prevent word splitting

alan-barzilay:
SC2206 -- Quote to prevent word splitting/globbing, or split robustly with
mapfile or read -a.
Fixes part of #375 

### DIFF
--- a/src/get_maintainer_wrapper.sh
+++ b/src/get_maintainer_wrapper.sh
@@ -11,11 +11,11 @@ function print_files_authors()
   if [[ -d $FILE_OR_DIR ]]; then
     for file in "$FILE_OR_DIR"/*; do
       if [[ -f $file ]]; then
-        files+=($file)
+        files+=("$file")
       fi
     done
   elif [[ -f $FILE_OR_DIR ]]; then
-    files+=($FILE_OR_DIR)
+    files+=("$FILE_OR_DIR")
   fi
 
   local printed_authors_separator=false
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -349,6 +349,7 @@ function store_statistics_data()
 function command_exists()
 {
   local command="$1"
+  #shellcheck disable=SC2206 #FIXME: see issue #388
   local package=($command)
 
   if [[ -x "$(command -v "${package[@]}")" ]]; then


alan-barzilay:
ok, so this commit has a tiny issue that may warrant some discussion.

The basic idea is to make more explicit the way we split things when passing them to an array,
originally in `kwlib.sh` in the  `command_exists()` function at line 352 we had: `local package=($command)` 
This implicitly separates `command` on every space tab and/or newline. To fix this warning and pass our tests we need something like this:
```
  local package
  IFS=" " read -r -a package <<< "$command"
```
which besides being confusing to read messes with the IFS variable which I'm not 100% comfortable with, I believe this could introduce some really hard to debug errors in the future. I opted to simply disable this case as an exception, but I'm not sure if this is the best approach, I would love to have a second opinion about this.

(Personally I would rather just use command -v directly instead of this function so I also don't have any strong opinions on how to fix this)

Messing with this line will make the kwio tests break (since we dont touch kwio in this PR it was really hard to uncover this, but at least now I know how to use git bisect):
```
❯ ./run_tests.sh test kwio
Running test [kwio]
test_alert_completion_options
./tests/kwio_test.sh: line 33: The: command not found
ASSERT:Alert's vs option didn't work.
./tests/kwio_test.sh: line 38: The: command not found
ASSERT:Alert's sv option didn't work.
./tests/kwio_test.sh: line 43: The: command not found
ASSERT:Alert's s option didn't work.
./tests/kwio_test.sh: line 48: The: command not found
ASSERT:Alert's v option didn't work.
test_alert_completition_validate_config_file_options
./tests/kwio_test.sh: line 66: The: command not found
ASSERT:Alert's vs option didn't work.
./tests/kwio_test.sh: line 72: The: command not found
ASSERT:Alert's sv option didn't work.
./tests/kwio_test.sh: line 78: The: command not found
ASSERT:Alert's s option didn't work.
./tests/kwio_test.sh: line 84: The: command not found
ASSERT:Alert's v option didn't work.
test_alert_completion_visual_alert
ASSERT:Variable v should exist. expected:<The following command set in the visual_alert_command variable couldn't be run:
/bin/echo $COMMAND
Check if the necessary packages are installed.> but was:<TESTING COMMAND>
shunit2:ERROR test_alert_completion_visual_alert() returned non-zero return code.
test_alert_completion_sound_alert
ASSERT:Variable s should exist. expected:<The following command set in the sound_alert_command variable couldn't be run:
/bin/echo $COMMAND
Check if the necessary packages are installed.> but was:<TESTING COMMAND>
shunit2:ERROR test_alert_completion_sound_alert() returned non-zero return code.

Ran 4 tests.

FAILED (failures=12)
```

alan-barzilay:
@rodrigosiqueira done

rodrigosiqueira:
Merged

PR #389, opened 2021-07-19
Manage installed kernels

joaoseckler:
With this commit kw keeps track of kernels it installed, and by default
will only list and uninstall those. Two new options for kw deploy are
added: --force will remove kernels even if they were not installed by
kw, and --all will list all installed kernels, even if they were not
installed by kw.

Closes: #187

### DIFF
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -47,6 +47,8 @@ function kernel_deploy()
   local end=0
   local runtime=0
   local ret=0
+  local list_all
+  local flag
 
   if [[ "$1" =~ -h|--help ]]; then
     deploy_help "$1"
@@ -59,18 +61,21 @@ function kernel_deploy()
     return 22 # EINVAL
   fi
 
+  flag="${options_values['TEST_MODE']}"
   target="${options_values['TARGET']}"
   reboot="${options_values['REBOOT']}"
   modules="${options_values['MODULES']}"
   single_line="${options_values['LS_LINE']}"
+  list_all="${options_values['LS_ALL']}"
   list="${options_values['LS']}"
   uninstall="${options_values['UNINSTALL']}"
+  uninstall_force="${options_values['UNINSTALL_FORCE']}"
   remote="${remote_parameters['REMOTE']}"
 
-  if [[ "$list" == 1 || "$single_line" == 1 ]]; then
+  if [[ "$list" == 1 || "$single_line" == 1 || "$list_all" == 1 ]]; then
     say "Available kernels:"
     start=$(date +%s)
-    list_installed_kernels "" "$single_line" "$target"
+    list_installed_kernels "$flag" "$single_line" "$target" "$list_all"
     end=$(date +%s)
 
     runtime=$((end - start))
@@ -80,7 +85,7 @@ function kernel_deploy()
 
   if [[ -n "$uninstall" ]]; then
     start=$(date +%s)
-    kernel_uninstall "$target" "$reboot" "$uninstall" "$flag"
+    kernel_uninstall "$target" "$reboot" "$uninstall" "$flag" "$uninstall_force"
     end=$(date +%s)
 
     runtime=$((end - start))
@@ -149,7 +154,8 @@ function parse_deploy_options()
   local remote
   local options
   local long_options='remote:,local,vm,reboot,modules,list,ls-line,uninstall:'
-  local short_options='r,m,l,s,u:'
+  long_options+=',list-all,force'
+  local short_options='r,m,l,s,u:,a,f'
 
   options="$(kw_parse "$short_options" "$long_options" "$@")"
 
@@ -159,12 +165,15 @@ function parse_deploy_options()
     return 22 # EINVAL
   fi
 
+  options_values['TEST_MODE']='SILENT'
   options_values['UNINSTALL']=''
+  options_values['UNINSTALL_FORCE']=''
   options_values['MODULES']=0
   options_values['LS_LINE']=0
   options_values['LS']=0
   options_values['REBOOT']=0
   options_values['MENU_CONFIG']='nconfig'
+  options_values['LS_ALL']=''
 
   remote_parameters['REMOTE']=''
 
@@ -219,6 +228,10 @@ function parse_deploy_options()
         options_values['LS']=1
         shift
         ;;
+      --list-all | -a)
+        options_values['LS_ALL']=1
+        shift
+        ;;
       --ls-line | -s)
         options_values['LS_LINE']=1
         shift
@@ -231,7 +244,15 @@ function parse_deploy_options()
         options_values['UNINSTALL']+="$2"
         shift 2
         ;;
+      --force | -f)
+        options_values['UNINSTALL_FORCE']=1
+        shift
+        ;;
       --) # End of options, beginning of arguments
+        shift
+        ;;
+      TEST_MODE)
+        options_values['TEST_MODE']='TEST_MODE'
         shift
         ;;
       *)
@@ -263,13 +284,17 @@ function parse_deploy_options()
 #   available kernels in a single line separated by commas. If it gets 0 it
 #   will display each kernel name by line.
 # @target Target can be 1 (VM_TARGET), 2 (LOCAL_TARGET), and 3 (REMOTE_TARGET)
+# @all If this option is set to one, this will list all kernels
+#   availble. If not, will list only kernels that were installed by kw.
 function list_installed_kernels()
 {
   local flag="$1"
   local single_line="$2"
   local target="$3"
+  local all="$4"
   local remote
   local port
+  local cmd
 
   flag=${flag:-"SILENT"}
 
@@ -283,16 +308,16 @@ function list_installed_kernels()
       fi
 
       include "$KW_PLUGINS_DIR/kernel_install/utils.sh"
-      list_installed_kernels "$single_line" "${configurations[mount_point]}"
+      list_installed_kernels "$flag" "$single_line" "${configurations[mount_point]}" "$all"
 
       vm_umount
       ;;
     2) # LOCAL_TARGET
       include "$KW_PLUGINS_DIR/kernel_install/utils.sh"
-      list_installed_kernels "$single_line"
+      list_installed_kernels "$flag" "$single_line" '' "$all"
       ;;
     3) # REMOTE_TARGET
-      local cmd="bash $REMOTE_KW_DEPLOY/deploy.sh --list_kernels $single_line"
+      cmd="bash $REMOTE_KW_DEPLOY/deploy.sh --list_kernels '$flag' '$single_line' '' '$all'"
       remote="${remote_parameters['REMOTE_IP']}"
       port="${remote_parameters['REMOTE_PORT']}"
 
@@ -312,6 +337,8 @@ function list_installed_kernels()
 #         installation.
 # @kernels_target List containing kernels to be uninstalled
 # @flag How to display a command, see `src/kwlib.sh` function `cmd_manager`
+# @force If this value is equal to 1, try to uninstall kernels even if they are
+#        not managed by kw
 #
 # Return:
 # Return 0 if everything is correct or an error in case of failure
@@ -321,6 +348,7 @@ function kernel_uninstall()
   local reboot="$2"
   local kernels_target="$3"
   local flag="$4"
+  local force="$5"
   local distro
   local remote
   local port
@@ -345,7 +373,7 @@ function kernel_uninstall()
       include "$KW_PLUGINS_DIR/kernel_install/utils.sh"
       # TODO: Rename kernel_uninstall in the plugin, this name is super
       # confusing
-      kernel_uninstall "$reboot" 'local' "$kernels_target" "$flag"
+      kernel_uninstall '' "$reboot" 'local' "$kernels_target" "$flag" "$force"
       ;;
     3) # REMOTE_TARGET
       remote="${remote_parameters['REMOTE_IP']}"
@@ -357,7 +385,7 @@ function kernel_uninstall()
       # TODO
       # It would be better if `cmd_remotely` handle the extra space added by
       # line break with `\`; this may allow us to break a huge line like this.
-      local cmd="bash $REMOTE_KW_DEPLOY/deploy.sh --uninstall_kernel $reboot remote $kernels_target $flag"
+      local cmd="bash $REMOTE_KW_DEPLOY/deploy.sh --uninstall_kernel '' '$reboot' remote '$kernels_target' '$flag' '$force'"
       cmd_remotely "$cmd" "$flag" "$remote" "$port"
       ;;
   esac
@@ -586,7 +614,8 @@ function deploy_help()
     '  deploy (--remote <remote>:<port> | --local | --vm) - choose target' \
     '  deploy (--reboot | -r) - reboot machine after deploy' \
     '  deploy (--modules | -m) - install only modules' \
-    '  deploy (--uninstall | -u) <kernel-name>,... - uninstall given kernels' \
+    '  deploy (--uninstall | -u) [(--force | -f)] <kernel-name>,... - uninstall given kernels' \
     '  deploy (--list | -l) - list kernels' \
-    '  deploy (--ls-line | -s) - list kernels separeted by commas'
+    '  deploy (--ls-line | -s) - list kernels separeted by commas' \
+    '  deploy (--list-all | -a) - list all available kernels'
 }
--- a/src/plugins/kernel_install/utils.sh
+++ b/src/plugins/kernel_install/utils.sh
@@ -85,7 +85,7 @@ function list_installed_kernels()
   output=$(echo "$output" | grep recovery -v | grep with | awk -F" " '{print $NF}')
 
   while read -r kernel; do
-    if [[ -f "$prefix/boot/vmlinuz-$kernel" ]]; then
+    if [[ -f "$prefix/boot/vmlinuz-$kernel" && ! "$kernel" =~ .*\.old$ ]]; then
       available_kernels+=("$kernel")
     fi
   done <<< "$output"
--- a/src/plugins/kernel_install/utils.sh
+++ b/src/plugins/kernel_install/utils.sh
@@ -1,3 +1,6 @@
+declare -g REMOTE_KW_DEPLOY="/root/kw_deploy"
+declare -g INSTALLED_KERNELS_PATH="$REMOTE_KW_DEPLOY/INSTALLED_KERNELS"
+
 # ATTENTION:
 # This function follows the cmd_manager signature (src/kwlib.sh) because we
 # share the specific distro in the kw main code. However, when we deploy for a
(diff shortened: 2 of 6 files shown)


src/plugins/kernel_install/utils.sh
```diff
@@ -284,12 +292,18 @@ function kernel_uninstall()
...
   local flag="$4"
+  local managed_kernels="$5"
 
   if [[ -z "$kernel" ]]; then
     echo "Invalid argument"
     exit 22 #EINVAL
   fi
 
+  if ! grep -q "$kernel" <(echo "$managed_kernels"); then
+    echo "Kernel not managed by kw"
```
rodrigosiqueira, line 303: `''` instead of `""`

rodrigosiqueira (changes requested):
After installing your PR, `kw d --list` does not work as expected even though I have three kernels installed via kw (I tested using remote).

src/deploy.sh
```diff
@@ -287,9 +288,14 @@ function list_installed_kernels()
   local target="$3"
   local remote
   local port
+  local cmd
+  local managed_kernels
 
   flag=${flag:-"SILENT"}
 
+  touch "$KW_LIB_DIR/MANAGED_KERNELS"
```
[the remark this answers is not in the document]
joaoseckler: It shouldn't go into `.cache` because it is not temporary. But maybe `LIB_DIR` is really not the best place for it. Any suggestions?

joaoseckler:
> After installing your PR, `kw d --list` does not work as expected even though I have three kernels installed via kw (I tested using remote).

I am supposing, naturally, that these kernels were deployed before this patch. This is really a problem. How will `kw` know retroactively which kernels are managed by it (before this patch) and which are not? I can think of adding an `--all` flag to `kw d -l`, but that does not completely solve the problem.

joaoseckler:
This PR is currently based on #390

documentation/man/kw.rst
```diff
@@ -74,8 +74,8 @@ notice that this approach is the most generic one because you can use it for
   **Currently, we don't support the Kernel image update in the --vm option.
   However, you can use the remote option for a workaround this issue**.
 
-d, deploy [--remote [REMOTE:PORT]|--local|--vm] [--reboot|-r] [--modules|-m] [--ls-line|-s] [--list|-l] [--uninstall|-u KERNEL_NAME]
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+d, deploy [--remote [REMOTE:PORT]|--local|--vm] [--reboot|-r] [--modules|-m] [--ls-line|-s] [--list|-l] [--all|-a] [--uninstall|-u KERNEL_NAME] [--force|-f]
```
rodrigosiqueira, line 77: How about rename `--all` to something like `--list-all`? I think `--all`, is a little bit vague.

src/deploy.sh
```diff
@@ -63,14 +64,16 @@ function kernel_deploy()
...
   uninstall="${options_values['UNINSTALL']}"
+  uninstall_force="${options_values['UNINSTALL_FORCE']}"
   remote="${remote_parameters['REMOTE']}"
 
-  if [[ "$list" == 1 || "$single_line" == 1 ]]; then
+  if [[ "$list" == 1 || "$single_line" == 1 || "$list_all" == 1 ]]; then
     say "Available kernels:"
     start=$(date +%s)
-    list_installed_kernels "" "$single_line" "$target"
+    list_installed_kernels "" "$single_line" "$target" "$list_all"
```
rodrigosiqueira, line 76: Since we are here, can we pass `$flag` to the first parameter?

src/plugins/kernel_install/utils.sh
```diff
@@ -86,19 +95,29 @@ function list_installed_kernels()
...
+        managed_kernels+=("$kernel")
+      fi
       available_kernels+=("$kernel")
     fi
   done <<< "$output"
 
-  echo
+  if [[ -z "$all" ]]; then
+    if [[ "${#managed_kernels[@]}" == 0 ]]; then
+      echo 'No kernels installed are managed by kw.' \
```
rodrigosiqueira, line 107: Maybe "None of the kernels..."

src/plugins/kernel_install/utils.sh
```diff
@@ -86,19 +95,29 @@ function list_installed_kernels()
...
+  fi
 
   if [[ "$single_line" != 1 ]]; then
     printf '%s\n' "${available_kernels[@]}"
   else
     echo -n "${available_kernels[0]}"
-    available_kernels=("${available_kernels[@]:1}")
-    printf ',%s' "${available_kernels[@]}"
-    echo ""
+    printf ',%s' "${available_kernels[@]:1}"
```
rodrigosiqueira, line 119: When I tested your patch, I deploy one single test in the remote, and when I used:
```
kw d -s
```
I got
```
5.13.0-joao-kernel+,
```
It looks like that your patch is adding one extra `,` at the end. Can we remove that?

src/plugins/kernel_install/utils.sh
```diff
@@ -370,9 +402,12 @@ function install_kernel()
   # Each distro has their own way to update their bootloader
   eval "update_$distro""_boot_loader $name $target $flag"
 
+  echo "$name" | sudo tee -a "$managed_kernels_path" > /dev/null
```
rodrigosiqueira, line 405: Just add a simple comment like "Registering a new kernel."

src/plugins/kernel_install/utils.sh
```diff
@@ -370,9 +402,12 @@ function install_kernel()
...
+
   # Reboot
   if [[ "$target" != 'vm' && "$reboot" == "1" ]]; then
     cmd="$sudo_cmd reboot"
     cmd_manager "$flag" "$cmd"
   fi
+
```
rodrigosiqueira, line 412: Do we need this extra line?

tests/plugins/kernel_install/utils_test.sh
```diff
@@ -52,37 +62,46 @@ function test_ask_yN()
...
-
   declare -a expected_out=(
-    "" # Extra espace in the beginning
     "5.5.0-rc2-VKMS+"
     "5.6.0-rc2-AMDGPU+"
     "linux"
   )
 
+  printf "%s\n" "${expected_out[@]}" > "$REMOTE_KW_DEPLOY/MANAGED_KERNELS"
+
```
rodrigosiqueira, line 72: Something is wrong here. When I tried to run all tests locally, I got:
```
Running test [utils_test]
test_cmd_manager
test_ask_yN
test_human_list_installed_kernels
[sudo] password for siqueira:
```

Idk why our CI did not complain. Any idea? I suspect of `sudo grep -q "$kernel" "$managed_kernels_path"`

rodrigosiqueira (changes requested):
Hi @joaoseckler ,
Great job here! I added a couple of inline comments.

tests/plugins/kernel_install/utils_test.sh
joaoseckler: I think the CI runs a passwordless sudo, so there is no problem.

Either way, how to fix this? We need permissions to access the /root/kw_deploy folder

src/deploy.sh
joaoseckler: Isn't it already?

tests/plugins/kernel_install/utils_test.sh
rodrigosiqueira: Could you check the CI? Something is wrong; see my comment in the deploy parser refactor.

Btw, we can mock the `sudo` command in that test. Or you can use the `cmd_manager` to validate the command.

src/deploy.sh
rodrigosiqueira: Hi @joaoseckler ,

I think I did not explain really well what that I mean about using the flag here. Allow me to elaborate a little bit better... Basically, when I use your implementation, I can see the following output:
```
 sudo mkdir -p '/root/kw_deploy' 
 sudo touch '/root/kw_deploy/MANAGED_KERNELS'
```
I believe that we don't need to show that to users. Additionally, if we use the flag variable, we have more ways to test the command sequence. For trying to illustrate my change suggestion, take a look at the below diff:
```diff
diff --git a/src/deploy.sh b/src/deploy.sh
index ffc548d..9dd2915 100644
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -47,6 +47,7 @@ function kernel_deploy()
   local end=0
   local runtime=0
   local ret=0
+  local flag
   local list_all
 
   if [[ "$1" =~ -h|--help ]]; then
@@ -69,11 +70,12 @@ function kernel_deploy()
   uninstall="${options_values['UNINSTALL']}"
   uninstall_force="${options_values['UNINSTALL_FORCE']}"
   remote="${remote_parameters['REMOTE']}"
+  flag=${options_values['test_mode']:-'SILENT'}
 
   if [[ "$list" == 1 || "$single_line" == 1 || "$list_all" == 1 ]]; then
     say "Available kernels:"
     start=$(date +%s)
-    list_installed_kernels "" "$single_line" "$target" "$list_all"
+    list_installed_kernels "$flag" "$single_line" "$target" "$list_all"
     end=$(date +%s)
 
     runtime=$((end - start))
@@ -309,7 +311,7 @@ function list_installed_kernels()
       list_installed_kernels '' "$single_line" '' "$all"
       ;;
     3) # REMOTE_TARGET
-      local cmd="bash $REMOTE_KW_DEPLOY/deploy.sh --list_kernels '' '$single_line' '' '$all'"
+      local cmd="bash $REMOTE_KW_DEPLOY/deploy.sh --list_kernels '$flag' '$single_line' '' '$all'"
       remote="${remote_parameters['REMOTE_IP']}"
       port="${remote_parameters['REMOTE_PORT']}"
```

src/plugins/kernel_install/utils.sh
```diff
@@ -84,21 +95,27 @@ function list_installed_kernels()
 
   output=$(echo "$output" | grep recovery -v | grep with | awk -F" " '{print $NF}')
 
+  cmd="sudo grep -q '$kernel' '$managed_kernels_path'"
+
   while read -r kernel; do
-    if [[ -f "$prefix/boot/vmlinuz-$kernel" && ! "$kernel" =~ .*\.old$ ]]; then
+    if [[ -f "$prefix/boot/vmlinuz-$kernel" ]] &&
+      ([[ -n "$all" ]] ||
+        ([[ ! "$kernel" =~ .*\.old$ ]] && cmd_manager '' "$check_managed_cmd")); then
```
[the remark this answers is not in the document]
rodrigosiqueira: I also noticed one problem; when I tried `kw d -l`, I got this output:
```
Available kernels:
 sudo mkdir -p '/root/kw_deploy'
 sudo touch '/root/kw_deploy/MANAGED_KERNELS'
 
 
 
 
 
 
5.13.0-KERNEL-X+
5.13.0-KERNEL-Y+
5.13.0-KERNEL-Z+
5.11.0-25-generic
5.8.0-63-generic
```
All of these empty lines came from the command `cmd_manager '' "$check_managed_cmd"`; you can easily fix this by passing `cmd_manager 'SILENT' "$check_managed_cmd"` or use `$flag` where `flag == SILENT` by default. However, I did not get where `$check_managed_cmd` came from; it looks empty in my side.

src/plugins/kernel_install/utils.sh
```diff
@@ -51,15 +53,24 @@ function ask_yN()
 #   available kernels in a single line separated by commas. If it gets 0 it
 #   will display each kernel name by line.
 # @prefix Set a base prefix for searching for kernels.
+# @all List all available kernels, not only the ones installed by kw
```
rodrigosiqueira, line 56: I think your PR is the right direction for handling kernels list; for this reason, I would suggest that you keep your new code under this function but create a new one named `list_installed_kernels_based_on_grub` and move the old code to that function. The current implementation is heavily tied to grub which should be treated as a fallback after your change.
Ps.: Maybe create a new commit for that.

rodrigosiqueira (changes requested):
Hi @joaoseckler , thanks a lot for your PR; that's a nice improvement!

I added a couple of comments.

Thanks

tests/plugins/kernel_install/utils_test.sh
```diff
@@ -52,37 +52,26 @@ function test_ask_yN()
...
-  local count=0
-
   declare -a expected_out=(
     "" # Extra espace in the beginning
-    "5.5.0-rc2-VKMS+"
-    "5.6.0-rc2-AMDGPU+"
-    "linux"
+    '5.5.0-rc2-VKMS+'
+    '5.6.0-rc2-AMDGPU+'
+    'linux'
```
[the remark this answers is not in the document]
joaoseckler: Isn't that as intended? We don't consider .old kernels as managed or installed. They are only shown if you pass `--list-all`, which is covered by the test `test_list_unmanaged_kernels`

src/plugins/kernel_install/utils.sh
```diff
@@ -84,21 +95,27 @@ function list_installed_kernels()
 
   output=$(echo "$output" | grep recovery -v | grep with | awk -F" " '{print $NF}')
 
+  cmd="sudo grep -q '$kernel' '$managed_kernels_path'"
```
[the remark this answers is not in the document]
joaoseckler: Maybe this should be a topic of kw's codestyle. Here we actually need the quotes. If by any chance the variable kernel is empty and the quotes are not empty, grep will interpret the second argument as the pattern to be matched and will wait for input on the stdin, blocking the execution of kw. The same if managed_kernels_path is empty

src/plugins/kernel_install/utils.sh
```diff
@@ -84,21 +95,27 @@ function list_installed_kernels()
...
+  fi
 
   if [[ "$single_line" != 1 ]]; then
     printf '%s\n' "${available_kernels[@]}"
   else
-    echo -n "${available_kernels[0]}"
-    available_kernels=("${available_kernels[@]:1}")
-    printf ',%s' "${available_kernels[@]}"
-    echo ""
+    local IFS=','
```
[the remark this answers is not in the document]
joaoseckler: It seems that using local solves this issue! https://stackoverflow.com/questions/50728845/can-ifs-be-changed-locally-in-a-bash-function

src/plugins/kernel_install/utils.sh
```diff
@@ -280,16 +297,30 @@ function do_uninstall()
...
+  cmd_manager "$flag" "sudo touch '$managed_kernels_path'"
 
   if [[ -z "$kernel" ]]; then
     echo "Invalid argument"
     exit 22 #EINVAL
   fi
 
+  cmd="sudo grep -q '$kernel' '$managed_kernels_path'"
+
+  if ! cmd_manager '' "$cmd" && [[ -z "$force" ]]; then
```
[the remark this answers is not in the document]
joaoseckler: We cannot put commands directly into `[[` expressions. Do you think this is easier to read?

```bash
  local is_kernel_managed_by_kw=''
  ! cmd_manager '' "$cmd" && [[ -z "$force" ]] && is_kernel_managed_by_kw=1
  if [[ -n "$is_kernel_managed_by_kw" ]]; then
    echo 'Kernel not managed by kw. Use --force/-f to uninstall anyway.'
    exit 22 # EINVAL
  fi
```

src/plugins/kernel_install/utils.sh
```diff
@@ -370,6 +404,10 @@ function install_kernel()
   # Each distro has their own way to update their bootloader
   eval "update_$distro""_boot_loader $name $target $flag"
 
+  # Registering a new kernel
+  cmd="sudo tee -a '$managed_kernels_path' > /dev/null"
+  echo "$name" | cmd_manager "$flag" "$cmd"
```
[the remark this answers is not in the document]
joaoseckler: See shellcheck warning [SC2024](https://github.com/koalaman/shellcheck/wiki/SC2024): "sudo doesn't affect redirects. Use ..| sudo tee file". Indeed, `cmd_manager` passes the stdin to the eval

tests/plugins/kernel_install/utils_test.sh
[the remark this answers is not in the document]
rodrigosiqueira: I missed that, thanks.

src/plugins/kernel_install/utils.sh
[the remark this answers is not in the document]
rodrigosiqueira: If you want the return value, you can create something like this:
```
cmd_manager '' "$cmd"
ret="$?"
[[ -z "$force" && "$ret" != 0 ]] && is_kernel_managed_by_kw=1
```

src/plugins/kernel_install/utils.sh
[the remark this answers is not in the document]
rodrigosiqueira: Nice! Can we add a comment in this specific part?

src/plugins/kernel_install/utils.sh
```diff
@@ -370,6 +403,10 @@ function install_kernel()
   # Each distro has their own way to update their bootloader
   eval "update_$distro""_boot_loader $name $target $flag"
 
+  # Registering a new kernel
+  cmd="sudo tee -a '$INSTALLED_KERNELS_PATH' > /dev/null"
+  echo "$name" | cmd_manager "$flag" "$cmd"
+
```
rodrigosiqueira, line 418: After reading your explanation about these lines, I think it is a good idea to add a comment that explains these lines (add the shellcheck ref).

rodrigosiqueira (changes requested):
Thanks a lot for this new version; it looks really nice :)

joaoseckler:
The last update includes a warning in the documentation and adds a comment explaining why we are using `sudo tee` to do a simple redirection.

src/deploy.sh
```diff
@@ -231,9 +244,17 @@ function parse_deploy_options()
         options_values['UNINSTALL']+="$2"
         shift 2
         ;;
+      --force | -f)
+        options_values['UNINSTALL_FORCE']=1
+        shift
```
rodrigosiqueira, line 249: Did you add a test that checks this new option?

src/deploy.sh
```diff
@@ -231,9 +244,17 @@ function parse_deploy_options()
...
+        ;;
       --) # End of options, beginning of arguments
         shift
         ;;
+      TEST_MODE)
+        options_values['TEST_MODE']='TEST_MODE'
```
rodrigosiqueira, line 255: Same here. Did you add a test that passes through this code path?

src/plugins/kernel_install/utils.sh
```diff
@@ -85,20 +96,27 @@ function list_installed_kernels()
...
+      fi
       available_kernels+=("$kernel")
     fi
   done <<< "$output"
 
-  echo
+  if [[ "${#available_kernels[@]}" -eq 0 ]]; then
+    echo 'None of the installed kernels are managed by kw.' \
+      'Pass --list-all|-a to see all installed kernels'
+    return 0
```
rodrigosiqueira: Maybe I missed something, but did you add a test that passes through this code path?

rodrigosiqueira (changes requested):
Hi @joaoseckler ,

I think we are close to complete this, just update this PR and take a look at my latest comments.

src/plugins/kernel_install/utils.sh
joaoseckler: The execution of the code was tested, but not the parsing of the option! Added in new version.

rodrigosiqueira:
Hi @joaoseckler do you have plans to finish this PR? If you are busy, I can take over and finish it for you (I'll keep your authorship).

rodrigosiqueira:
PR closed due to the lack of activities. Since this is a substantial improvement, I made a new PR which can be seen here:

https://github.com/kworkflow/kworkflow/pull/472

Notice that I kept the original authorship.

Thanks

PR #390, opened 2021-07-22
src: deploy: Refactor option parsing

joaoseckler:
Use getopt to simplify the parsing of options for kw deploy. This
implies one change of behaviour: getopt accepts abbreviations of options,
as long as there is no ambiguity. For example, now --rem is accepted
and is equivalent to --remote.

### DIFF
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -40,7 +40,6 @@ function kernel_deploy()
   local reboot=0
   local modules=0
   local target=0
-  local test_mode=''
   local list=0
   local single_line=0
   local uninstall=''
@@ -54,10 +53,10 @@ function kernel_deploy()
     exit 0
   fi
 
-  deploy_parser_options "$@"
+  parse_deploy_options "$@"
   if [[ "$?" -gt 0 ]]; then
-    complain "Invalid option: ${options_values['ERROR']}"
-    exit 22 # EINVAL
+    complain "${options_values['ERROR']}"
+    return 22 # EINVAL
   fi
 
   target="${options_values['TARGET']}"
@@ -65,14 +64,8 @@ function kernel_deploy()
   modules="${options_values['MODULES']}"
   single_line="${options_values['LS_LINE']}"
   list="${options_values['LS']}"
-  test_mode="${options_values['TEST_MODE']}"
   uninstall="${options_values['UNINSTALL']}"
   remote="${remote_parameters['REMOTE']}"
-
-  if [[ "$test_mode" == 'TEST_MODE' ]]; then
-    echo "$reboot $modules $target ${remote_parameters['REMOTE_IP']} ${remote_parameters['REMOTE_PORT']} $single_line $list"
-    return 0
-  fi
 
   if [[ "$list" == 1 || "$single_line" == 1 ]]; then
     say "Available kernels:"
@@ -150,12 +143,21 @@ function kernel_deploy()
 # Return:
 # In case of successful return 0, otherwise, return 22.
 #
-function deploy_parser_options()
+function parse_deploy_options()
 {
-  local raw_options="$*"
-  local uninstall=0
   local enable_collect_param=0
   local remote
+  local options
+  local long_options='remote:,local,vm,reboot,modules,list,ls-line,uninstall:'
+  local short_options='r,m,l,s,u:'
+
+  options="$(kw_parse "$short_options" "$long_options" "$@")"
+
+  if [[ "$?" != 0 ]]; then
+    options_values['ERROR']="$(kw_parse_get_errors 'kw deploy' "$short_options" \
+      "$long_options" "$@")"
+    return 22 # EINVAL
+  fi
 
   options_values['UNINSTALL']=''
   options_values['MODULES']=0
@@ -176,7 +178,7 @@ function deploy_parser_options()
 
   populate_remote_info ''
   if [[ "$?" == 22 ]]; then
-    options_values['ERROR']="$remote"
+    options_values['ERROR']="Invalid remote: $remote"
     return 22 # EINVAL
   fi
 
@@ -184,87 +186,68 @@ function deploy_parser_options()
     options_values['REBOOT']=1
   fi
 
-  IFS=' ' read -r -a options <<< "$raw_options"
-  for option in "${options[@]}"; do
-    if [[ "$option" =~ ^(--.*|-.*|test_mode) ]]; then
-      if [[ "$enable_collect_param" == 1 ]]; then
-        options_values['ERROR']='expected paramater'
-        return 22
-      fi
+  eval "set -- $options"
 
-      case "$option" in
-        --remote)
-          options_values['TARGET']="$REMOTE_TARGET"
-          continue
-          ;;
-        --local)
-          options_values['TARGET']="$LOCAL_TARGET"
-          continue
-          ;;
-        --vm)
-          options_values['TARGET']="$VM_TARGET"
-          continue
-          ;;
-        --reboot | -r)
-          options_values['REBOOT']=1
-          continue
-          ;;
-        --modules | -m)
-          options_values['MODULES']=1
-          continue
-          ;;
-        --list | -l)
-          options_values['LS']=1
-          continue
-          ;;
-        --ls-line | -s)
-          options_values['LS_LINE']=1
-          continue
-          ;;
-        --uninstall | -u)
-          enable_collect_param=1
-          uninstall=1
-          continue
-          ;;
-        test_mode)
-          options_values['TEST_MODE']='TEST_MODE'
-          ;;
-        *)
-          options_values['ERROR']="$option"
-          return 22 # EINVAL
-          ;;
-      esac
-    else # Handle potential parameters
-      if [[ "$uninstall" != 1 &&
-        ${options_values['TARGET']} == "$REMOTE_TARGET" ]]; then
-        populate_remote_info "$option"
+  while [[ "$#" -gt 0 ]]; do
+    case "$1" in
+      --remote)
+        options_values['TARGET']="$REMOTE_TARGET"
+        populate_remote_info "$2"
         if [[ "$?" == 22 ]]; then
-          options_values['ERROR']="$option"
-          return 22
+          options_values['ERROR']="Invalid remote: $2"
+          return 22 # EINVAL
         fi
-      elif [[ "$uninstall" == 1 ]]; then
-        options_values['UNINSTALL']+="$option"
-        enable_collect_param=0
-      else
-        # Invalind option
-        options_values['ERROR']="$option"
-        return 22
-      fi
-    fi
+        shift 2
+        ;;
+      --local)
+        options_values['TARGET']="$LOCAL_TARGET"
+        shift
+        ;;
+      --vm)
+        options_values['TARGET']="$VM_TARGET"
+        shift
+        ;;
+      --reboot | -r)
+        options_values['REBOOT']=1
+        shift
+        ;;
+      --modules | -m)
+        options_values['MODULES']=1
+        shift
+        ;;
+      --list | -l)
+        options_values['LS']=1
+        shift
+        ;;
+      --ls-line | -s)
+        options_values['LS_LINE']=1
+        shift
+        ;;
+      --uninstall | -u)
+        if [[ "$2" =~ ^-- ]]; then
+          options_values['ERROR']='Uninstall requires a kernel name'
+          return 22 # EINVAL
+        fi
+        options_values['UNINSTALL']+="$2"
+        shift 2
+        ;;
+      --) # End of options, beginning of arguments
+        shift
+        ;;
+      *)
+        options_values['ERROR']="Unrecognized argument: $1"
+        return 22 # EINVAL
+        shift
+        ;;
+    esac
   done
-
-  # Uninstall requires an option
-  if [[ "$uninstall" == 1 && -z "${options_values['UNINSTALL']}" ]]; then
-    options_values['ERROR']='uninstall requires a kernel name'
-    return 22
-  fi
 
   case "${options_values['TARGET']}" in
     1 | 2 | 3) ;;
 
     *)
-      options_values['ERROR']='remote option'
-      return 22
+      options_values['ERROR']="Invalid target value: ${options_values['TARGET']}"
+      return 22 # EINVAL
       ;;
   esac
 }
@@ -599,7 +582,7 @@ function deploy_help()
   fi
   printf '%s\n' 'kw deploy:' \
     '  deploy - installs kernel and modules:' \
-    '  deploy (--remote [<remote>:<port>] | --local | --vm) - choose target' \
+    '  deploy (--remote <remote>:<port> | --local | --vm) - choose target' \
     '  deploy (--reboot | -r) - reboot machine after deploy' \
     '  deploy (--modules | -m) - install only modules' \
     '  deploy (--uninstall | -u) <kernel-name>,... - uninstall given kernels' \
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -371,3 +371,50 @@ function exit_msg()
   complain "$msg"
   exit "$err"
 }
+
+# This function parses command line arguments. Each option may be
+# followed by a one colon to indicate it has a required argument, and by
+# two colons to indicate it has an optional argument. If any errors are
+# found, this will print an error message to stderr indicating it.
+#
+# @short_options Short options to be accepted
+# @long_options Long options to be accepted
+# @{@:2} Arguments to be parsed
+#
+# Returns:
+# Parsed command line arguments.
+function kw_parse()
+{
+  local short_options="$1"
+  local long_options="$2"
+  shift 2
+
+  getopt -q --options "$short_options" \
+    --longoptions "$long_options" \
+    -- "$@"
+}
+
+# This function gets the error messages for a kw_parse call. The same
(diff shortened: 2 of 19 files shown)


alan-barzilay (approved):
LGTM, great idea!

src/deploy.sh
```diff
@@ -152,10 +152,29 @@ function kernel_deploy()
 #
 function deploy_parser_options()
```
alan-barzilay, line 153: we discussed this a while a go, but what do you think about taking this opportunity to rename this function to `parse_deploy_options`?

alan-barzilay (changes requested):

tests/deploy_test.sh
```diff
@@ -153,10 +153,10 @@ function test_kernel_deploy()
 
   # Test some invalid parameters
   ID=9
-  output=$(kernel_deploy test_mode --rem 192.168.0.10:1287)
+  output=$(kernel_deploy test_mode --remtoe 192.168.0.10:1287)
```
rodrigosiqueira, line 156: it looks like that we are matching "partial" strings, and I suppose this is why you changed this test (--rem does not fail as previously expected). Can we make getopt or the case condition only accept the exact match?

rodrigosiqueira (changes requested):

rodrigosiqueira:
Btw, thanks a lot for your PR. I really like it!

rodrigosiqueira:
@joaoseckler I also noticed that we don't have a proper test for `deploy_parser_options`, and in some way, this `test_mode` in `kernel_deploy` looks bad... tbh, I think it was a workaround to test the function. Could you also update the test to validate `deploy_parser_options` properly?

src/deploy.sh
```diff
@@ -152,10 +152,29 @@ function kernel_deploy()
...
   local remote
+  local options
+  local short_options
+  local long_options
+
+  long_options='remote,local,vm,reboot,modules,list,ls-line,uninstall:'
+  short_options='r,m,l,s,u:'
+  export error_msg
+
+  options="$(getopt -q \
```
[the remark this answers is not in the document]
joaoseckler: We don't want getopt to print errors to the screen! If there are any, they will be captured in the next if statement.

src/deploy.sh
```diff
@@ -184,73 +203,64 @@ function deploy_parser_options()
...
-          continue
-          ;;
-        test_mode)
-          options_values['TEST_MODE']='TEST_MODE'
-          ;;
-        *)
-          options_values['ERROR']="$option"
+  eval "set -- $options"
+
+  while [[ $# -gt 0 ]]; do
```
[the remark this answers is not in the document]
joaoseckler: Silly errors...

tests/deploy_test.sh
joaoseckler: Exactly! I had commented about it in the commit description. It seems there is no way to make getopt change this behavior, and we cannot easily change this in the case condition, since getopt detects `--rem` as being `--remote` and outputs `--remote` to be parsed, even if the user supplied `--rem`. Since there is no ambiguity (getopt makes sure of that) I propose that we accept this behavior.

src/deploy.sh
```diff
@@ -152,10 +152,29 @@ function kernel_deploy()
...
 {
-  local raw_options="$*"
   local uninstall=0
   local enable_collect_param=0
   local remote
+  local options
+  local short_options
+  local long_options
+
+  long_options='remote,local,vm,reboot,modules,list,ls-line,uninstall:'
```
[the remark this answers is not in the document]
joaoseckler: We are getting that as the only not dashed argument for the call. See the `*)` case

src/deploy.sh
```diff
@@ -184,73 +203,64 @@ function deploy_parser_options()
...
+        options_values['LS_LINE']=1
+        shift
+        ;;
+      --uninstall | -u)
+        uninstall=1
+        options_values['UNINSTALL']+="$2"
+        shift 2
+        ;;
+      --) # End of options, beginning of arguments
+        shift
```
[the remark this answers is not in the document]
joaoseckler: This is part of of getopt! It will separate dashed options and not dashed options with two dashes. For example

`kw  xpto1 --option --option2=xpto2 xpto3` will be converted to

`kw --option --option2 'xpto2' -- xpto1 xpto3`

So the two dashes must be handled in the case statement

src/deploy.sh
```diff
@@ -184,73 +203,64 @@ function deploy_parser_options()
...
+      --list | -l)
+        options_values['LS']=1
+        shift
+        ;;
+      --ls-line | -s)
+        options_values['LS_LINE']=1
+        shift
+        ;;
+      --uninstall | -u)
+        uninstall=1
```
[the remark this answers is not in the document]
joaoseckler: Nice catch

joaoseckler:
The last update to this PR:
 - Renames deploy_parser_options to parse_deploy_options
 - Removes the unnecessary uninstall variable. We also removed a test associated with that variable which made sure the uninstall arguments were supplied with an option (e.g. `--uninstall kernel_5.12`), but getopt checks that for us now. However, getopt will consider the next argument in the command line as an option for `--uninstall`. That is, to getopt the following are the same:

`kw d --uninstal --local`

`kw d --uninstall="--local"`

Thus, the test is changed to detect that.

 - Adds tests for `parse_deploy_options`

joaoseckler:
The last update to this patch abstracts the process of getting getopt's errors in a new function called `getopt_errors`

alan-barzilay (approved):

src/deploy.sh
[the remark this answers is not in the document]
rodrigosiqueira: btw, let's change the `--uninstall=KERNEL` synstax to `--uninstall "KERNEL"`. I think this makes more sense from the getopt perspective.

tests/deploy_test.sh
```diff
@@ -697,4 +684,126 @@ function test_cleanup()
   done <<< "$output"
 }
 
+function test_parse_deploy_options()
```
rodrigosiqueira, line 687: Now that you added this test function, I think we can remove all of the tests that rely on `test_mode`, i.e., we can implode `test_kernel_deploy`.

tests/deploy_test.sh
```diff
@@ -697,4 +684,126 @@ function test_cleanup()
...
+{
+  unset options_values
+  unset remote_parameters
+  declare -gA options_values
+  declare -gA remote_parameters
+
+  for i in "${!options_values[@]}"; do
+    echo " - $i: ${options_values["$i"]}"
```
rodrigosiqueira, line 695: Do we need to `echo` here?

rodrigosiqueira (changes requested):
Thanks a lot for the update; it looks really nice. After reviewing your PR, I wonder if we can make the remote code from the parser something generic because we use it in multiple commands; for this reason, we have a lot of code duplication related to that.

src/deploy.sh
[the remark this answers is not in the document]
joaoseckler: getopt handles both as valid input and meaning the same thing!

src/deploy.sh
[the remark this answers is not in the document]
joaoseckler: But what if my remote parameters are in my kworkflow.config? Then it makes sense to pass --remote without parameters, right?

joaoseckler:
> Thanks a lot for the update; it looks really nice. After reviewing your PR, I wonder if we can make the remote code from the parser something generic because we use it in multiple commands; for this reason, we have a lot of code duplication related to that.

I think it would be hard to abstract the parsing of options. Each default value is unique to each command and what to do with each option is also defined by each command. We could wrap around getopt, but that wouldn't change the code too much. It would go from

```
  options="$(getopt -q \
    --options "$short_options" \
    --longoptions "$long_options" \
    -- "$@")"
```
 to something like 
 ```
options="$(kw_parse "$short_options" " $long_options" "$@")"
```

What more could be abstracted? Maybe I'm missing something here.

File: src/deploy.sh (hunk 6 above)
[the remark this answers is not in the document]
Comment (joaoseckler) [reply src/deploy.sh:266]: To make it clearer. This case doesn't refer to the double dash between the argument name and option. It refers to the separation of arguments with options (which getopt has recognized) and arguments without options. Not this one:


```
kw --arg "option" -- xpto
   ^
   |
```

But rather this one:

```
kw --arg "option" -- xpto
                   ^
                   |
```

File: src/deploy.sh [hunk 10]
```diff
@@ -142,6 +142,29 @@ function kernel_deploy()
   cleanup
 }
 
+# This function return the errors outputed by getopt, formatted to be
+# printed by kw
+function getopt_errors()
```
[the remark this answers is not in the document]
joaoseckler: Actually, this only makes sense in the context of deploy, since there is this error managing system here. For example, we remove the prefix `unrecognized option` only because `kernel_deploy` will Add "Invalid option" later: See line 57:

```
  parse_deploy_options "$@"
  if [[ "$?" -gt 0 ]]; then
    complain "Invalid option: ${options_values['ERROR']}"
    exit 22 # EINVAL
  fi
```

joaoseckler:
The last update to this PR removes the `kernel_deploy` tests as well as the "test_mode" for `kernel_deploy` and `parse_deploy_options`

src/deploy.sh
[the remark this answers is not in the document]
alan-barzilay: we can still define the default target through the kworkflow.config file but it would be somewhat cumbersome to make this change all the time, it looks a bit error prone to me. A new user may get confused as to why it is not working, it may not realize that vm is the default target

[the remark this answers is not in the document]
alan-barzilay: in any case, if this change does occur we cant fortget to update the help message

rodrigosiqueira:
@joaoseckler ,

It lgtm, I still think we can abstract part of the getopt as we discussed in our last meeting; however, we can leave it for a future refactor.

About the `--remote` option, I still think we must enforce a parameter. If users want to use the value from the config file, they just need to drop `--remote`. We can consider adding another two options named `--vm` and `--local` to make things consistent with the config file.

src/kwlib.sh
```diff
@@ -371,3 +371,24 @@ function exit_msg()
...
+# This function parses command line arguments. Each option may be
+# followed by a one colon to indicate it has a required argument, and by
+# two colons to indicate it has an optional argument. If any errors are
+# found, this will print an error message to stderr indicating it.
+#
+# @short_options Short options to be accepted
+# @long_options Long options to be accepted
+# @{@:2} Arguments to be parsed
+#
+function kw_parse()
```
rodrigosiqueira, line 384: Maybe `kw_set_parse`?

src/kwlib.sh
```diff
@@ -371,3 +371,24 @@ function exit_msg()
...
+{
+  local short_options="$1"
+  local long_options="$2"
+  shift 2
+
+  getopt --name "kw" \
```
rodrigosiqueira, line 390: Maybe we should get `name` via a parameter to provide a better error message.

src/kwlib.sh
```diff
@@ -371,3 +371,24 @@ function exit_msg()
...
+    --options "$short_options" \
+    --longoptions "$long_options" \
+    -- "$@"
+}
```
rodrigosiqueira, line 394: Where is the for this function in this commit? :)

src/deploy.sh
```diff
@@ -184,87 +186,68 @@ function deploy_parser_options()
...
-      esac
-    else # Handle potential parameters
-      if [[ "$uninstall" != 1 &&
-        ${options_values['TARGET']} == "$REMOTE_TARGET" ]]; then
-        populate_remote_info "$option"
+  while [[ "$#" -gt 0 ]]; do
+    case "$1" in
+      --remote)
+        options_values['TARGET']="$REMOTE_TARGET"
+        populate_remote_info "$2"
```
rodrigosiqueira, line 195: Now that you changed this `--remote` behavior to enforce a parameter, we need to update the `help` and the documentation.

tests/deploy_test.sh
```diff
@@ -697,4 +592,109 @@ function test_cleanup()
...
+  assertEquals "($LINENO)" 'user' "${remote_parameters['REMOTE_USER']}"
+  assertEquals "($LINENO)" '127.0.2.1:8888' "${remote_parameters['REMOTE']}"
+  assertEquals "($LINENO)" '8888' "${remote_parameters['REMOTE_PORT']}"
+  assertEquals "($LINENO)" '127.0.2.1' "${remote_parameters['REMOTE_IP']}"
+
+  unset options_values
+  declare -gA options_values
+  parse_deploy_options --remote ':8888' > /dev/null
+  assertEquals "($LINENO)" '22' "$?"
+  assertEquals "($LINENO)" ':8888' "${options_values['ERROR']}"
```
rodrigosiqueira, line 625: Add one test that tries `parse_deploy_options --remote`.

rodrigosiqueira (changes requested):

rodrigosiqueira:
@joaoseckler do you have any updates here? If so, could you rebase your change?

tests/kwlib_test.sh
```diff
@@ -392,4 +392,30 @@ function test_exit_msg()
...
+  local so='x:,f,b'
+  local out
+  local expected
+
+  out="$(kw_parse_get_errors "$so" "$lo" --fee --bar biz --xpto)"
+  expected=(
+    "kw: unrecognized option '--fee'" "kw: option '--xpto' requires an argument"
+  )
+  compare_command_sequence expected[@] "$out" "$LINENO"
+}
```
rodrigosiqueira, line 419: Idk why, but when I tried this test locally, it failed, and based on the way that you're calling `kw_parse_get_errors`, I actually expect a failure... the weird thing is why CI did not complain about it. Is this something specific to my setup? Btw, I fixed it by using:

```diff
diff --git a/tests/kwlib_test.sh b/tests/kwlib_test.sh
index 2a82ec8..9d95a4c 100755
--- a/tests/kwlib_test.sh
+++ b/tests/kwlib_test.sh
@@ -406,16 +406,17 @@ function test_kw_parse()
 
 function test_kw_parse_get_errors()
 {
-  local lo='xpto:,foo,bar'
-  local so='x:,f,b'
-  local out
-  local expected
+  local long_option='xpto:,foo,bar'
+  local short_option='x:,f,b'
+  local output
 
-  out="$(kw_parse_get_errors "$so" "$lo" --fee --bar biz --xpto)"
-  expected=(
-    "kw: unrecognized option '--fee'" "kw: option '--xpto' requires an argument"
+  declare -a expected_errors=(
+    "kw: unrecognized option '--fee'"
+    "kw: option '--xpto' requires an argument"
   )
-  compare_command_sequence expected[@] "$out" "$LINENO"
+
+  output=$(kw_parse_get_errors 'kw' "$short_option" "$long_option" --fee --bar biz --xpto)
+  compare_command_sequence expected_errors[@] "$output" "$LINENO"
 }
 
 invoke_shunit
```

From Github check, I think something is wrong:
https://github.com/kworkflow/kworkflow/pull/390/checks?check_run_id=3302532172#step:5:320

Could you check that?

tests/kwlib_test.sh
```diff
@@ -392,4 +392,30 @@ function test_exit_msg()
   assertEquals "($LINENO) We expected 3 as a return" 3 "$ret"
 }
 
+function test_kw_parse()
+{
+  local lo='xpto:,foo,bar'
+  local so='x:,f,b'
+  local out
```
rodrigosiqueira, line 399: Let's use the full name to make the test easier to read.

src/kwlib.sh
```diff
@@ -371,3 +371,52 @@ function exit_msg()
...
+# arguments passed to kw_parse should be passed to this function.
+#
+# @name Name to be prefixed in error messages
+# @short_options Short options to be accepted
+# @long_options Long options to be accepted
+# @{@:2} Arguments to be parsed
+#
+# Returns:
+# Error messages separated by a newline and prefixed with @name
+#
```
rodrigosiqueira, line 408: Drop this extra `#`

src/kwlib.sh
```diff
@@ -371,3 +371,52 @@ function exit_msg()
...
+# two colons to indicate it has an optional argument. If any errors are
+# found, this will print an error message to stderr indicating it.
+#
+# @short_options Short options to be accepted
+# @long_options Long options to be accepted
+# @{@:2} Arguments to be parsed
+#
+# Returns:
+# Parsed command line arguments.
+#
```
rodrigosiqueira, line 386: Drop this extra `#`

rodrigosiqueira (changes requested):

tests/kwlib_test.sh
joaoseckler: I think I added the option to pass a name to kw_parse_get_errors but forgot to add it to the tests!

rodrigosiqueira:
@joaoseckler could you rebase this PR? It looks good to me, but we have conflicts; after you fix that, I'll be able to apply your change.

tests/vm_test.sh
```diff
@@ -42,38 +41,38 @@ function test_vm_mount()
...
 
-  ID=1
-  output=$(vm_mount "TEST_MODE")
+  output=$(
+    function findmnt()
+    {
+      echo "anything"
+    }
+    vm_mount "TEST_MODE"
+  )
```
rodrigosiqueira, line 50: Niceeeee!!! Could you make a doc patch under `test.rst` that explains this technique?

rodrigosiqueira (commented):
Very nice patch! Thanks a lot for your dedication to this task.

PR #391, opened 2021-07-24
remote: bugfix for the use of ssh config files

alan-barzilay:
cp_host2remote was working in such a way that if a ssh_configfile was
present, rsync would be used twice and fail since the first time it
would use the wrong parameters. By reintroducing the else clause this
bug is fixed


rodrigosiqueira:
Hi @alan-barzilay ,
Merged, thanks for your patch!

Issue #393, opened 2021-07-26
olddefconfig bug with raspberrypi


alan-barzilay:
**Describe the bug**
Since PR #339 the build process on raspberrypies is kinda broken. Usually we prepare the .config file by running `make foodefconfig` for the appropriate rpi model and then run make as normal. However, by also running olddefconfig, we get a restarting config message during the build process and the user gets prompted to decide a bunch of compiling options (which olddefconfig should take care for us).

**To Reproduce**
1. `git clone --depth=1 --branch <branch> https://github.com/raspberrypi/linux`
2. `cp  arch/arm/configs/bcm2711_defconfig . `
3. `make bcm2711_defconfig`
4. `kw b`

**Expected behavior**
The build process should run without the need for user interaction

**Screenshots**
![image](https://user-images.githubusercontent.com/28787373/127002902-236a7c77-4e2d-4777-b70e-2d04eeaa2722.png)

**Desktop (please complete the following information):**
 - OS: Manjaro
 - Version: unstable branch
 - Bash Version 5.1.8(1)-release

alan-barzilay:
I believe we might need to set the architecture to be used with the make olddefconfig command.
Interesting resources:
https://stackoverflow.com/questions/13636129/make-modules-install-restarts-configuration-process
https://stackoverflow.com/questions/41885015/what-exactly-does-linux-kernels-make-defconfig-do

Issue #188, opened 2020-06-23
Compress files before sending it to a remote machine


rodrigosiqueira:
A user reported that kw generates a large file when compared with kpackge approach. We need to improve it by adding a compression mechanism in the package that we want to send to the remote machine.

rodrigosiqueira:
Before implementing this feature, we probably want to solve the below issue first:

https://github.com/kworkflow/kworkflow/issues/243

There is a chance that we will need to install an app in the target machine to decompress the kernel image.

rodrigosiqueira:
@joaoseckler, when you have time, please, take a look at this ticket. It should be simple.

rodrigosiqueira:
Now that we have:
https://github.com/kworkflow/kworkflow/commit/30b0459c51f2ad58e595350f8e9bf45e62a2299c

This task should be easy.

Issue #161, opened 2020-02-21
Optimize kw deploy


rodrigosiqueira:
Recently we noticed that kw deploy can take a long time if the tarball file is big (e.g., 5.0 G) and users will not get any visual feedback about the transfer progress. Additionally, if we think a little bit more about this issue we will conclude that we are wasting time and bandwidth by sending a new tarball file time that we deploy a new Kernel; we use rsync in a better way. With this idea in mind, follows the required steps for fixing this issue:

**COMMIT 1: Stop to use tarball**
1. **Identify the correct functions**: In the `src/mk.sh` file we have `modules_install` function, which will call `generate_tarball()` for putting all drivers together and finally it will call `cp_host2remote` to all files to the remote.
2. **Next, try to find a way to remove the tarball creation**. Notice that `cp_host2remote` already utilizes rsync which means that we probably have to make tiny changes.
3. If everything is fine, **update all tests** related to mk.sh and remote.sh for making sure that everything is working well. Probably you want to take a look at:
* *tests/mk_test.sh*: See `kernel_modules_Test`
* *tests/remote.sh*

**COMMIT 2: Better use of rsync**
Now, we have to improve our use of rsync by providing a better output.
1. For adding progress information we probably want to change `cp_host2remote()` at `src/remote.sh` and make it show the progress by default.
2. Add an option to `cp_host2remote()` that hide the progress.
3. Update the required tests.

I think that is a good idea to use `pv`[1,2] to kw, however, if we use it we should think in a way to make this available for other kw functions.

1. http://www.ivarch.com/programs/pv.shtml
2. https://www.cyberciti.biz/faq/show-progress-during-file-transfer/

Ps.: If this is your first time contributing to `kw` take a few minutes to read https://siqueira.tech/doc/kw/content/howtocontribute.html

rodrigosiqueira:
@joaoseckler @alan-barzilay 

This is something important for V1 and in some way intersect with your work. Let's raise the priority for this issue and start to work on it soon.

rodrigosiqueira:
Also, when working on this issue, please consider this one as well:

https://github.com/kworkflow/kworkflow/issues/179

PR #394, opened 2021-07-27
refactor deploy names, cp2remote & dtb support

### LINKED ISSUE #354: Add support for deploy to Raspberry PI
We can consider this task done when we complete the following tasks:

- [x] Remote/Local deploy to a Rasbian 32bits
- [x] Remote/Local deploy to a Rasbian 64bits
- [x] Create tutorial

alan-barzilay:
This PR adds support for the raspberrypi OS by creating a new plugin, adding support to dtb files and breaking some previous assumptions about kernel names. 

The last commit reverses #339 in order to avoid bug #393, this is not a fix but a mere bandaid to show that this PR is functional while #393 is still being solved.

Closes #354 

PS: the only tests that are failing are the ones that dealt with the vmlinuz assumption, so they are not so much failing as they are just brittle. Before once again engaging in the tedious and time consuming task of fixing our brittle tests I want to have a definitive version of this PR. In short, fixing the tests will be the last thing I will do, after everything is approved and reviewed


src/deploy.sh
```diff
@@ -479,6 +487,31 @@ function modules_install_to()
...
+    1) # VM_TARGET
+      ;;
+    2) # LOCAL_TARGET
+      cmd_manager "sudo cp arch/$architecture/boot/dts/*.dtb /boot/"
+      cmd_manager "sudo cp arch/$architecture/boot/dts/overlays/*.dtb* /boot/overlays/"
+      ;;
+    3) # REMOTE_TARGET
+      cmd_remotely "mkdir -p $REMOTE_KW_DEPLOY/dts/overlays"
+      #copy dtb files to remote
+      cp_host2remote "arch/$architecture/boot/dts/.dtb" "$REMOTE_KW_DEPLOY/dts"
```
melissawen, line 507: Acho que falta um * aqui: ...dts/*.dtb

src/plugins/kernel_install/raspbian.sh
```diff
@@ -0,0 +1,33 @@
...
+
+}
+
+function generate_raspbian_temporary_root_file_system()
+{
+  local name="$1"
+  local target="$2"
+  local flag="$3"
+
+  # there is nothing do be done in th rpi case
```
melissawen, line 31: não entendi

src/deploy.sh
alan-barzilay: nice catch, I will update

src/plugins/kernel_install/raspbian.sh
alan-barzilay: kw is structured in a way that every plugin/ distribution needs to have at least 2 functions, one for the bootloader and one for the temporary root file system like initramfs. But afaik there is nothing equivalent to the initramfs in the raspberry. Since the function needs to exist, I just created a dummy function that returns immediately

melissawen:
Hi Alan,

Nice work!

I did not have time yet to check it carefully. I guess this is a WIP, but I am leaving some comments. 
It is not a full review, sorry.

* It only supports 32bits, right? If so, would be better to explicit it somewhere.
* I also consider it is important to describe how to set kworkflow.config to proper operation (what options should be changed and how). I was also curious to know how it is working for kw build, since I understood you've detached dtb building process.
* After a first version, I think using a tarball to send all files in one shot is a good polishing to avoid crash installation before it completes (leaving the system in an unstable situation).

btw, great work! thanks

alan-barzilay:
> It only supports 32bits, right? If so, would be better to explicit it somewhere.

I haven't tested with 64 bits but I don´t see why it wouldn't work, afaik installing a kernel in a rasp with 32 bits or 64 bits uses the same process. The difference in target architecture is handled in the kworkflow.conflig file by the architecture and cross compiler variables.
> I also consider it is important to describe how to set kworkflow.config to proper operation (what options should be changed and how). 

good call! I will add something in the documentation about this. 

> I was also curious to know how it is working for kw build, since I understood you've detached dtb building process.

the dtb building process happens normally with kw build. Since we don't specify a target, everything marked as `all`in the make file gets build, which includes the dtb files. All i did was include a new step in the deploy stage to also copy the dtb files if needed.


> After a first version, I think using a tarball to send all files in one shot is a good polishing to avoid crash installation before it completes (leaving the system in an unstable situation).

I agree, but I think that's a second step not specific to the rasp deploy, this needs to be a bit more general in kw. There are already a few issues dealing with this (#161 #188) and I believe @joaoseckler may be working in some of them
 
> btw, great work! thanks

thank you :)

alan-barzilay:
This last revision addresses @melissawen feedback. Please let me know what you think about the added tutorial

src/deploy.sh
```diff
@@ -479,6 +487,29 @@ function modules_install_to()
   cmd_manager "$flag" "$cmd"
 }
 
+# This function will copy and install the compiled dtb files
+# @target Where should the dtb files be installed (1-vm, 2-local or 3-remote)
+function dtb_install()
```
rodrigosiqueira, line 492: I'm wondering if this dtb install should be here... Maybe it makes more sense in the plugins folder `plugins/kernel_install/`?

src/kwlib.sh
```diff
@@ -276,6 +276,8 @@ function detect_distro()
   # Debian family
   elif [[ "$distro" =~ "debian" ]] || [[ "$distro" =~ "ubuntu" ]]; then
     echo "debian"
+  elif [[ "$distro" =~ "raspbian" ]]; then
+    echo "raspbian"
```
rodrigosiqueira, line 280: Technically, "raspbian" belongs to the Debian familly...

src/plugins/kernel_install/raspbian.sh
```diff
@@ -0,0 +1,33 @@
...
+# changes here, you must adjust install_modules() and install_kernel()
+# accordingly.
+#
+
+# Update boot loader API
+function update_raspbian_boot_loader()
+{
+  local name="$1"
+  local target="$2"
+  local flag="$3"
```
rodrigosiqueira, line 13: Document these parameters.

src/plugins/kernel_install/raspbian.sh
```diff
@@ -0,0 +1,33 @@
...
+
+  # https://www.raspberrypi.org/documentation/configuration/config-txt/boot.md
+
+  if grep --quiet "kernel" /boot/config.txt; then
```
rodrigosiqueira, line 17: single-quote

src/plugins/kernel_install/raspbian.sh
```diff
@@ -0,0 +1,33 @@
...
+    sudo sed -ri "/kernel/s/=.+/=$name/g" /boot/config.txt
+  else
+    echo "kernel=$name" | sudo tee -a /boot/config.txt
+  fi
+
```
rodrigosiqueira, line 22: drop this extra line

src/plugins/kernel_install/raspbian.sh
```diff
@@ -0,0 +1,33 @@
...
+function update_raspbian_boot_loader()
+{
+  local name="$1"
+  local target="$2"
+  local flag="$3"
+
+  # https://www.raspberrypi.org/documentation/configuration/config-txt/boot.md
+
+  if grep --quiet "kernel" /boot/config.txt; then
+    sudo sed -ri "/kernel/s/=.+/=$name/g" /boot/config.txt
```
rodrigosiqueira, line 18: In this function documentation, explain why do we need to update config.txt with `$name`

src/plugins/kernel_install/raspbian.sh
rodrigosiqueira: Yeah, maybe we can improve this in the future but for now, add a comment in the function doc.

documentation/tutorials/buildlinux.rst
```diff
@@ -82,10 +82,7 @@ And it can be even simpler by just running::
 
   kw b
 
-.. note:
-    kw runs `make olddefconfig` before building.
-
 Well, that's it. kw will automatically infer the number of job slots for
-compiling based on the number of cores of your machine (i.e. when running make
+compiling based on the number of cores of your machine (i.e. when running make
```
rodrigosiqueira, line 86: We don't need this commit. I'm going to revert it in the unstable.

documentation/tutorials/rasp.rst
```diff
@@ -0,0 +1,69 @@
...
+   :depth: 1
+   :local:
+   :backlinks: none
+
+.. highlight:: console
+
+Get the kernel repository
+-------------------------
+After successfully :ref:`installing kw<install-and-remove-kw>`, it's time to
+build your RaspberryPi kernel. Let's start by cloning the repository into your
```
rodrigosiqueira, line 16: "by cloning the **official** repository ..."

documentation/tutorials/rasp.rst
```diff
@@ -0,0 +1,69 @@
...
+current working directory::
+
+  git clone --depth=1 https://github.com/raspberrypi/linux
```
rodrigosiqueira, line 19: I would remove the `--depth` part and add it as a note or tip.

documentation/tutorials/rasp.rst
```diff
@@ -0,0 +1,69 @@
...
+
+
+Build!
+------
+Building and deploying the kernel now is as easy as invoking::
+
+  kw bd --remote
+
+
+
```
rodrigosiqueira, line 67: Drop these extra lines

rodrigosiqueira (changes requested):
Hi @alan-barzilay,

First of all, thanks a lot for your PR. I'm super excited about this new feature; in my view, this work represents another step to make kw more generic :)

I already add some comments, which you can see below. Follows some generic design suggestions:

We can take this opportunity to add a new step in the deploy by adapting and calling the function `learn_device`. First, let's adapt `learn_device` to figure out if the target machine is a Raspberry Pi; I think you already find a way since you suggested this `cat /sys/firmware/devicetree/base/model` in a previous discussion (I'm not sure if this is generic, but it is a good start). Maybe we can get the following from the target machine:
Target architecture
Board name
Did you already test `kw device ` pointing to your Rasp? Could you share the output here?

Anyway, we can invoke `learn_device` at the beginning of the deploy. If we have a raspberry pi we can take specific actions, such as dealing with the dtb. Maybe we can add another parameter in the `plugins/kernel_install/deploy.sh` to specify the target type, and create a new file dedicated to Raspberry Pi specific steps; this might require a refactor in the `debian.sh` and `arch.sh` files.

Looking Forward to Hearing From You

documentation/tutorials/buildlinux.rst
rodrigosiqueira: I just reverted it in the unstable

melissawen:
> > It only supports 32bits, right? If so, would be better to explicit it somewhere.
> 
> I haven't tested with 64 bits but I don´t see why it wouldn't work, afaik installing a kernel in a rasp with 32 bits or 64 bits uses the same process. The difference in target architecture is handled in the kworkflow.conflig file by the architecture and cross compiler variables.

hmm.. not sure. The host directory where aarch64-linux-gnu- installs dtb differs in one more level (/broadcom)
btw, I am asking it just because we should make users clear about that support.

> 
> > I also consider it is important to describe how to set kworkflow.config to proper operation (what options should be changed and how).
> 
> good call! I will add something in the documentation about this.
> 
> > I was also curious to know how it is working for kw build, since I understood you've detached dtb building process.
> 
> the dtb building process happens normally with kw build. Since we don't specify a target, everything marked as `all`in the make file gets build, which includes the dtb files. All i did was include a new step in the deploy stage to also copy the dtb files if needed.

nice. thanks for explaining it, I actually forgot many details in kw build.
> 
> > After a first version, I think using a tarball to send all files in one shot is a good polishing to avoid crash installation before it completes (leaving the system in an unstable situation).
> 
> I agree, but I think that's a second step not specific to the rasp deploy, this needs to be a bit more general in kw. There are already a few issues dealing with this (#161 #188) and I believe @joaoseckler may be working in some of them
> 
good to know.

Thanks,

Melissa
> > btw, great work! thanks
> 
> thank you :)

src/deploy.sh
```diff
@@ -479,6 +487,29 @@ function modules_install_to()
...
+{
+  local target="$1"
+  local architecture="${configurations[arch]:-'arm'}"
+
+  case "$target" in
+    1 | 3) # VM_TARGET or REMOTE_TARGET
+      cmd_remotely "mkdir -p $REMOTE_KW_DEPLOY/dts/overlays"
+      #copy dtb files to remote
+      cp_host2remote "arch/$architecture/boot/dts/*.dtb" "$REMOTE_KW_DEPLOY/dts"
+      cp_host2remote "arch/$architecture/boot/dts/overlays/*.dtb*" "$REMOTE_KW_DEPLOY/dts/overlays"
```
marceloschmitt1, line 502: My old (2019) raspbian installation doesn't have a dts directory. dtb files lay at the file system root dir (a.k.a. /) along with the kernel image and other configuration files. Yet talking about legacy raspberry systems, the overlays lay under "/overlays". I think we shouldn't assume "dts" to be part of either dtb or overlays path. Otherwise, add a note to warn users with older installs.

src/plugins/kernel_install/raspbian.sh
marceloschmitt1, line 18: I wonder what happens without a kernel option at config.txt. I guess we could assume it to be always present, though people may use diverse operating systems with raspberry or no OS at all.

alan-barzilay: @marceloschmitt1 From what I could ascertain, if there is not a kernel parameter in the config.txt the default value is used (which is board dependent)

src/deploy.sh
alan-barzilay: @marceloschmitt1 I believe this may have been a mistake on my part, we create the dts folder on the remote only to better organize and decouple everything that needs to be installed, the idea is to rsync this dts folder to the boot folder. 
Rsyncinc to boot/dts instead of boot/ was a mistake, thank you for catching this!

alan-barzilay:
> hmm.. not sure. The host directory where aarch64-linux-gnu- installs dtb differs in one more level (/broadcom)
> btw, I am asking it just because we should make users clear about that support.

@melissawen  you are right, I missed this detail with the dtbs. I think we should just copy every single dtb file under the dts folder, this way we will always ensure to have everything we need. @marceloschmitt1  I'm not very familiar with dtb files so I'm not sure what is the best approach here, but you seem to be quite knowledgeable about dtb files, do you have any thoughts on this approach?

src/deploy.sh
alan-barzilay: @rodrigosiqueira I don´t think so, the plugins directory is solely composed of scripts that will run in a remote machine while this function is called directly in the host. I think moving it there will just make things more confusing

src/plugins/kernel_install/utils.sh
```diff
@@ -232,7 +232,7 @@ function do_uninstall()
   local target="$1"
   local prefix="$2"
   local flag="$3"
-  local kernelpath="$prefix/boot/vmlinuz-$target"
+  local kernelpath="$prefix/boot/$target"
```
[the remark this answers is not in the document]
alan-barzilay: @rodrigosiqueira sorry, I can't seem to understand your question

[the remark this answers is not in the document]
alan-barzilay: is there something special about the vmlinuz part of a kernel name?

src/deploy.sh
marceloschmitt1: Oh, on may shallow review I missed you are creating the dts dir with `cmd_remotely`. I see your point, though I'm unsure if creating such directory structure worth it. I think it would be enough to `rsync` dtb files to / and dt-overlay files to /overlays directly. Of course, I'm not aware of how is your connection with the remote system so there might be technical issues I'm not seeing here.

marceloschmitt1:
> > hmm.. not sure. The host directory where aarch64-linux-gnu- installs dtb differs in one more level (/broadcom)
> > btw, I am asking it just because we should make users clear about that support.
> 
> @melissawen you are right, I missed this detail with the dtbs. I think we should just copy every single dtb file under the dts folder, this way we will always ensure to have everything we need.
> @marceloschmitt1 I'm not very familiar with dtb files so I'm not sure what is the best approach here but you seem to be quite knowledgeable about them, do you have any thoughts on this approach?

On one scenario we would pass dtb files is when deploying to a raspberry board. We copy the blobs to / of the boot partition. I've done this with 32 bit rpi boards. Need to check out if the same process applies to 64 bit rpi.
There's another scenario in which we deploy to a VM. qemu provides a `-dtb` flag that allows users to pass a dtb file to the system. When using qemu's virt board (`-M virt`), passing a dtb is optional because qemu automatically creates a dtb an passes it to the kernel.
So if deploying to rpi we usually copy all .dtb files from `arch/arm/boot/dts/` or  from `arch/arm64/boot/dts/broadcom/` to somewhere on the system memory. If deploying to a qemu VM, we only need to pass a single dtb file or none if using the virt board.
Hope it didn't get much confusing. Keep asking otherwise.

alan-barzilay:
@marceloschmitt1 I think I understood, thank you!
With your explanation in mind I think we should just copy all dtb files to the /boot, as we have been doing already. Since this PR is aimed at raspberry support I think this should be enough, I think dealing with the dtb in the virtual machine as you described should be left as a future step when we try to add support for embedded systems

melissawen:
Hi @alan-barzilay @rodrigosiqueira

I've tested the deployment option from this patch.

First, you need to rebase on the unstable branch. But I think we can take this opportunity to check arm/cross-compilation support in all features that require those flags. I mean, at least check all "make" commands are working as expected.
Besides that, kw deployment works as expected. Moreover, I enjoy the option of set kernel image by config.txt  - by now, it is possible to change manually, but would be great to have an option to change that "kernel=" variable remotely.
Also, some other options related to "kw d" are not working yet (such as, --uninstall and --list), right? I think the --list options can help to set "kernel=" properly.

Anyway, thanks!

Melissa

alan-barzilay:
@melissawen let me see if I understood your suggestion, you want to be able to remotely update the kernel that the Rpi will boot from? like a command that will remotely manage the rpi bootloader?

melissawen:
> @melissawen let me see if I understood your suggestion, you want to be able to remotely update the kernel that the Rpi will boot from? like a command that will remotely manage the rpi bootloader?

more or less it. I see like a fake management; but it is an option, at least to easily change between custom kernel versions. We still need to boot a kernel, set the name to the other img in config.txt and then reboot the system to switch the kernel versions... but, in my PoV, having this option is fair.

etc/kworkflow_template.config
```diff
@@ -31,6 +31,9 @@ kernel_img_name=bzImage
 # Specify cross-compile name
 #cross_compile=aarch64-linux-gnu-
 
+# Specify if dtb files should be installed, default to False
+#dtb=yes
```
rodrigosiqueira, line 39: Let's use `install_dtb`.

src/deploy.sh
```diff
@@ -120,6 +120,14 @@ function kernel_deploy()
   end=$(date +%s)
   runtime=$((end - start))
 
+  # dtb
+  if [[ ${configurations[dtb]} =~ "yes" ]]; then
```
rodrigosiqueira, line 127: `"` around `${configurations...` and `'yes'` instead of `"yes"`.

src/deploy.sh
```diff
@@ -478,6 +486,23 @@ function modules_install_to()
...
+# This function will copy and install the compiled dtb files
+# @target Where should the dtb files be installed (1-vm, 2-local or 3-remote)
+function dtb_install()
+{
+  local target="$1"
+  local architecture="${configurations[arch]}"
+
+  case "$target" in
+    1 | 3) # VM_TARGET or REMOTE_TARGET
+      cp2remote "arch/$architecture/boot/dts/" "/boot/" "" '--include="*/" --include="*.dtb*" --exclude="*"'
```
rodrigosiqueira, line 392: You need to check if `"arch/$architecture/boot/dts/"` exists before trying to copy it. Also, what is this `'--include="*/" --include="*.dtb*" --exclude="*"'`? Could you rewrite this function in way that become easier to undertand your intention?

src/deploy.sh
```diff
@@ -478,6 +486,23 @@ function modules_install_to()
...
 
+# This function will copy and install the compiled dtb files
+# @target Where should the dtb files be installed (1-vm, 2-local or 3-remote)
+function dtb_install()
+{
+  local target="$1"
+  local architecture="${configurations[arch]}"
+
+  case "$target" in
+    1 | 3) # VM_TARGET or REMOTE_TARGET
```
rodrigosiqueira, line 391: That's not how we handle VM... You cannot suppose the VM is running; actually, it is the opposite of that. See how modules are deploys for reference.

src/deploy.sh
```diff
@@ -478,6 +486,23 @@ function modules_install_to()
...
+      cp2remote "arch/$architecture/boot/dts/" "/boot/" "" '--include="*/" --include="*.dtb*" --exclude="*"'
+      ;;
+    2) # LOCAL_TARGET
+      rsync -LrlptD --include="*/" --include="*.dtb*" --exclude="*" "arch/$architecture/boot/dts/" /boot/ --rsync-path='sudo rsync'
```
rodrigosiqueira, line 395: Use `cmd_manager`, since it will enable you to test this function.

etc/kworkflow_template.config
```diff
@@ -23,6 +23,10 @@ ssh_port=22
 # Specify the default architecture used by KW
 arch=x86_64
 
+# Specify the bootloader being used.
+# Currently supported targets are `grub` (grub2) and `rpi_bootloader` (Raspberrypi native bootloader)
+bootloader=grub
```
rodrigosiqueira, line 28: I'm not sure if I like that... I think kw should be smart enough to detect the bootloader instead of waiting for the user input. Also, enable users to specify the bootloader opens a door for mistakes.

src/plugins/kernel_install/deploy.sh
```diff
@@ -35,6 +35,7 @@ cd "$target_kw_deploy" || exit_msg 'It was not possible to move to deploy dir'
 # Load specific distro script
 . distro_deploy.sh --source-only
 . utils.sh --source-only
+. bootloader.sh --source-only
```
rodrigosiqueira, line 38: Remember that files under `plugins/kernel_install/` are copied to the target machine; in this sense, make sure that we only copy the correct bootloader file.

src/grub.sh
```diff
@@ -0,0 +1,110 @@
+# This function list all the available kernels in a VM, local, and remote
```
rodrigosiqueira, line 1: I like the idea of isolate bootloader-specific code... However, I'm not sure if I like how you isolate the bootloader code; notice that you moved many non-specific codes to `grub.sh`. I expected only grub-related things in this part, not all of the other stuff. Additionally, I would prefer a file in `src/bootloader.sh` that manages generic things and specific bootloader code under `src/plugins/bootloader`.

rodrigosiqueira (changes requested):
Hi @alan-barzilay ,

First of all, thanks a lot for keeping improve your PR. I really appreciate that.

You can see many comments in my review, but I want to highlight two architectural points:

1. Users specify the target bootloader

Tbh, I don't think relying on users to get the bootloader info is a good idea. Mistakes in this area can crash a lot of things, and we don't want that. In my previous review, I suggested how you can handle this different hardware architecture; I'm going to copy-paste my last comment:

> We can take this opportunity to add a new step in the deploy by adapting and calling the function learn_device. First, let's adapt learn_device to figure out if the target machine is a Raspberry Pi; I think you already find a way since you suggested this cat /sys/firmware/devicetree/base/model in a previous discussion (I'm not sure if this is generic, but it is a good start). Maybe we can get the following from the target machine:
Target architecture
Board name
Did you already test kw device pointing to your Rasp? Could you share the output here?
Anyway, we can invoke learn_device at the beginning of the deploy. If we have a raspberry pi we can take specific actions, such as dealing with the dtb. Maybe we can add another parameter in the plugins/kernel_install/deploy.sh to specify the target type, and create a new file dedicated to Raspberry Pi specific steps; this might require a refactor in the debian.sh and arch.sh files.

2. Please, make sure that all previous tests are still passing. In that way, you will avoid regressions; for example, your patch completely breaks the x86 deploy.

src/remote.sh
```diff
@@ -68,27 +68,25 @@ function cmd_remotely()
...
+function cp2remote()
 {
-  local src=${1:-"$KW_CACHE_DIR/$LOCAL_TO_DEPLOY_DIR/*"}
-  local dst=${2:-"$REMOTE_KW_DEPLOY"}
-  local remote=${3:-${configurations[ssh_ip]}}
-  local port=${4:-${configurations[ssh_port]}}
-  local user=${5:-${configurations[ssh_user]}}
-  local flag=${6:-"HIGHLIGHT_CMD"}
+  local src=$1
+  local dst=$2
```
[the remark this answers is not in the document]
alan-barzilay: shellcheck complains only when a variable could be split into multiple ones, which in this case should be impossible, by definition $1 is always a single token. Same reason why it doesnt complain for $#, a single number cant be split into multiple tokens

tests/remote_test.sh
```diff
@@ -141,59 +141,16 @@ function test_cmd_remote()
...
-ssh -p 22 ${user}@localhost sudo \"chown -R root:root $dst\""
-  output=$(cp_host2remote "" "" "" "" "$user" "$flag")
-  assertEquals "Command did not match ($ID)" "$expected_command" "$output"
-
-  ID=6
-  src="$KW_CACHE_DIR/$LOCAL_TO_DEPLOY_DIR/*"
-  dst="$REMOTE_KW_DEPLOY"
-  expected_command="rsync -e 'ssh -p 22' -La $src root@localhost:$dst --rsync-path='sudo rsync'
-ssh -p 22 root@localhost sudo \"chown -R root:root $dst\""
-  output=$(cp_host2remote "" "" "" "" "" "$flag")
```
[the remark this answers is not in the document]
alan-barzilay: tests 3 to 6 dont make sense anymore since we dropped the remote variables, besides that the default values logic is already being tested in the config loader, there is no need to retest it here. Certainly these tests can be expanded, but the ones dropped arent very useful at the moment

src/remote.sh
```diff
@@ -68,27 +68,25 @@ function cmd_remotely()
...
+  local flag=${3:-"HIGHLIGHT_CMD"}
+  local rsync_params=${4:-""}
+
+  local remote=${configurations[ssh_ip]}
+  local port=${configurations[ssh_port]}
+  local user=${configurations[ssh_user]}
 
   if [[ -v configurations['ssh_configfile'] && -v configurations['hostname'] ]]; then
-    cmd_manager "$flag" "rsync -e 'ssh -F ${configurations['ssh_configfile']}' -La $src ${configurations['hostname']}:$dst --rsync-path='sudo rsync'"
+    cmd_manager "$flag" "rsync -e 'ssh -F ${configurations['ssh_configfile']}' -LrlptD $src ${configurations['hostname']}:$dst --rsync-path='sudo rsync' $rsync_params"
```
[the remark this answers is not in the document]
alan-barzilay: how about now?

src/deploy.sh
alan-barzilay: @rodrigosiqueira this function has a single entry point, which happens after line 105 where the target is checked and the vm is mounted. Am I missing something? my understanding is that the vm will necessarily be running at this point

alan-barzilay:
@rodrigosiqueira this last series addresses most of your concerns, please let me know if the new changes are more inline with what you expected.

I'm currently testing if the deploy is broken and if the newly introduced changes broke the rasp support.

alan-barzilay:
@magalilemes could you take a look at this new series? in particular at the `deploy: isolate bootloader logic` commit where I introduce a new function inside device_info.sh

rodrigosiqueira (changes requested):
Hi @alan-barzilay ,

Could you try the remote deploy in your VM using x86? I tried this PR, but it breaks de deploy for x86. See part of the output (I removed part of the log):

```
[..]
-> Execution time: 00:00:04
cat: '/*-release': No such file or directory
ssh -p YY root@XXX.XXX.XXX.XXX  sudo "mkdir -p /root/kw_deploy"
rsync -e 'ssh -p ' /home/siqueira/.local/lib/kw/plugins/kernel_install/debian.sh @:/root/kw_deploy/distro_deploy.sh -LrlptD --rsync-path='sudo rsync' 
Bad port '-l'
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: unexplained error (code 255) at io.c(235) [sender=3.1.3]
rsync -e 'ssh -p ' /home/siqueira/.local/lib/kw/plugins/kernel_install/.sh @:/root/kw_deploy/bootloader.sh -LrlptD --rsync-path='sudo rsync' 
Bad port '-l'
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: unexplained error (code 255) at io.c(235) [sender=3.1.3]
rsync -e 'ssh -p ' /home/siqueira/.local/lib/kw/plugins/kernel_install/remote_deploy.sh @:/root/kw_deploy/ -LrlptD --rsync-path='sudo rsync' 
Bad port '-l'
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: unexplained error (code 255) at io.c(235) [sender=3.1.3]
rsync -e 'ssh -p ' /home/siqueira/.local/lib/kw/plugins/kernel_install/utils.sh @:/root/kw_deploy/ -LrlptD --rsync-path='sudo rsync' 
Bad port '-l'
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: unexplained error (code 255) at io.c(235) [sender=3.1.3]
```

Also, I did not see many of my comments resolved in this version. Did I miss something?

src/device_info.sh
```diff
@@ -218,6 +219,39 @@ function get_os()
...
+# device_info_data variable.
+#
+# @target Target machine
+function get_bootloader()
+{
+  local target="$1"
+  local flag="$2"
+  local ip="$3"
+  local port="$4"
+  local cmd='cat "$etc_path"/*-release | grep -w ID | cut -d = -f 2'
```
magalilemes, line 232: This etc_path variable is not set, so running
`cat "$etc_path"/*-release | grep -w ID | cut -d = -f 2`
leads to
`cat: '/*-release': No such file or directory`

Maybe to get the distribution, you could expand `detect_distro()` in src/kwlib.sh and make it detect raspbian, since it apparently doesn't.

alan-barzilay:
> Also, I did not see many of my comments resolved in this version. Did I miss something?

@rodrigosiqueira what comments specifically? The only thing I think I havent addressed yet is the vmlinuz part which im still working on

src/device_info.sh
```diff
@@ -218,6 +219,39 @@ function get_os()
...
+# This function populates the bootloader variables from the
+# device_info_data variable.
+#
+# @target Target machine
+function get_bootloader()
+{
+  local target="$1"
+  local flag="$2"
+  local ip="$3"
+  local port="$4"
```
rodrigosiqueira, line 231: You can use something like:
```
ip=${3:-${remote_parameters[ssh_ip]}}
[..]
```

src/deploy.sh
```diff
@@ -69,6 +71,9 @@ function kernel_deploy()
   uninstall="${options_values['UNINSTALL']}"
   remote="${remote_parameters['REMOTE']}"
 
+  get_bootloader "$target" 'SILENT'
```
rodrigosiqueira, line 74: Call `learn_device` since we can profit from other data during the deploy; I recommend you to add this `get_bootloader` under `learn_device`.

src/deploy.sh
```diff
@@ -609,7 +616,11 @@ function run_kernel_install()
         cp2remote "$KW_CACHE_DIR/$LOCAL_TO_DEPLOY_DIR/$name.preset" "$REMOTE_KW_DEPLOY" "$flag"
       fi
 
-      cp2remote "arch/$arch_target/boot/$kernel_img_name" "$REMOTE_KW_DEPLOY/vmlinuz-$name" "$flag"
+      if [[ $bootloader == 'grub' ]]; then
+        cp2remote "arch/$arch_target/boot/$kernel_img_name" "$REMOTE_KW_DEPLOY/vmlinuz-$name" "$flag"
+      else
+        cp2remote "arch/$arch_target/boot/$kernel_img_name" "$REMOTE_KW_DEPLOY/$name" "$flag"
+      fi
```
rodrigosiqueira, line 628: I prefer something like this:
```bash
      kernel_name="vmlinuz-$name"
      [[ "$bootloader" != 'grub' ]] && kernel_name="$name"
      cp2remote "arch/$arch_target/boot/$kernel_img_name" "$REMOTE_KW_DEPLOY/$kernel_name" "$flag"
```

src/device_info.sh
```diff
@@ -218,6 +219,39 @@ function get_os()
...
+  local cmd='cat /etc/*-release | grep -w ID | cut -d = -f 2'
+  local os
+
+  case "$target" in
+    1) # VM_TARGET
+      vm_mount
+      os=$(cmd_remotely "$cmd" "$flag")
```
rodrigosiqueira, line 238: Do you mean `cmd_manager`?

src/device_info.sh
```diff
@@ -218,6 +219,39 @@ function get_os()
...
+function get_bootloader()
+{
+  local target="$1"
+  local flag="$2"
+  local ip="$3"
+  local port="$4"
+  local cmd='cat /etc/*-release | grep -w ID | cut -d = -f 2'
+  local os
+
+  case "$target" in
```
rodrigosiqueira, line 235: What is the point of using this? We already have `get_os`.

src/plugins/kernel_install/grub.sh
```diff
@@ -0,0 +1,298 @@
...
+    do_uninstall "$kernel" "" "$flag"
+  done
+
+  # Each distro script should implement update_boot_loader
+  echo "update_boot_loader $kernel $local_deploy $flag"
+  update_boot_loader "$kernel" "$local_deploy" "$flag"
+
+  # Reboot
+  reboot_machine "$reboot" "$local_deploy"
+}
```
rodrigosiqueira, line 83: I don't think that `kernel_uninstall` should be here. We can reuse this for raspberry pi; the only different part is `update_boot_loader`.

alan-barzilay:
this last series addresses all issues raised so far. I have locally tested all deploy features against my raspberry pi and a debian vm as a remote. I have also reworked the unit tests so they should all be passing (travis isnt working anymore)

rodrigosiqueira (changes requested):
Hi @alan-barzilay ,

First of all, thanks for this new version.

I added a couple of comments, but some will probably be outdated because you changed many things from when I started to review. Sorry for that, but this is a tough PR to review.

I feel that your PR is challenging to review and digest because it is enormous and passes through some of the core parts of the kw, which requires a lot of deliberation. In this sense, let's reduce the scope of this PR to the first four commits ("deploy: rename functions and files"). After that, we focus on the bootloader login isolation and finally on the raspberry part.

How about that?

P.s.: Keep in mind that we don't have CI support right now, which means that I need more time to be confident to apply the change.

alan-barzilay:
@magalilemes when we invoke learn_device, sometimes we get some error messages when something it is looking for does not exist. For example, I was working on a vm without a graphical interface and as such got an error:  
`find: ‘/usr/share/xsessions’: No such file or directory`

Something similar happened when working in my rasp with the following errors: 
```
pcilib: Cannot open /proc/bus/pci
lspci: Cannot find any working access method.
cat: /sys/devices/virtual/dmi/id/chassis_type: No such file or directory
/home/barzilay/.local/lib/kw/device_info.sh: line 248: warning: command substitution: ignored null byte in input
```
As a stopgap measure I redirected stderr to /dev/null when I invoked it in deploy.sh

alan-barzilay:
@rodrigosiqueira this last series addresses all issues raised so far. I have locally tested all deploy features against my raspberry pi and a debian vm as a remote. I have also reworked the unit tests so they should all be passing (travis isnt working anymore)


> In this sense, let's reduce the scope of this PR to the first four commits

This is the last week of GSoC, breaking this PR in 2 or more would ultimately make us take more time to merge rasp support since the rasp stuff would stay frozen for a while. Are we sure we want to do this? I have the feeling we will miss our deadline if we do this

alan-barzilay:
The rasp and grub commits have been moved to #418

src/remote.sh
```diff
@@ -68,27 +68,30 @@ function cmd_remotely()
...
-  local dst=${2:-"$REMOTE_KW_DEPLOY"}
-  local remote=${3:-${configurations[ssh_ip]}}
-  local port=${4:-${configurations[ssh_port]}}
-  local user=${5:-${configurations[ssh_user]}}
-  local flag=${6:-"HIGHLIGHT_CMD"}
+  local src="$1"
+  local dst="$2"
+  local flag="${3:-"HIGHLIGHT_CMD"}"
+  local rsync_params=${4:-""}
+  local rsync_target
```
rodrigosiqueira, line 80: Put this variable at the end

tests/deploy_test.sh
```diff
@@ -336,6 +326,7 @@ function test_kernel_install_x86_64()
   configurations=()
   cp "$KW_CONFIG_SAMPLE_X86" "$FAKE_KERNEL/kworkflow.config"
   parse_configuration "$FAKE_KERNEL/kworkflow.config"
+  populate_remote_info
```
rodrigosiqueira, line 329: After a second thought, for making this test self-contained, use:
```
remote_parameters['REMOTE_IP']=${configurations[ssh_ip]}
remote_parameters['REMOTE_PORT']=${configurations[ssh_port]}
remote_parameters['REMOTE_USER']=${configurations[ssh_user]}
```

src/deploy.sh
```diff
@@ -476,6 +485,30 @@ function modules_install_to()
...
+  local architecture="${configurations[arch]}"
+  # the rsync_params variable is passed directly to rsync (also used in cp2remote) and
+  # used to copy only the dtb files and ignore everything else under the dts folder
+  local rsync_params='--include="*/" --include="*.dtb*" --exclude="*"'
+
+  if [[ -d "arch/$architecture/boot/dts/" ]]; then
+    case "$target" in
+      1 | 3) # VM_TARGET or REMOTE_TARGET
+        cp2remote "arch/$architecture/boot/dts/" "/boot/" '' "$rsync_params"
+        ;;
```
rodrigosiqueira, line 502: You need to deal with VM separately. You need to mount the VM, copy the file to there, and umount.

src/deploy.sh
```diff
@@ -476,6 +485,30 @@ function modules_install_to()
...
+      2) # LOCAL_TARGET
+        cmd_manager '' "rsync -LrlptD $rsync_params arch/$architecture/boot/dts/ /boot/ --rsync-path='sudo rsync'"
+        ;;
+    esac
+  else
+    warning "Directory arch/$architecture/boot/dts/ does not exist, no dtb files to be installed."
+  fi
+}
```
rodrigosiqueira, line 510: Where is the unit test for this new function?

src/deploy.sh
```diff
@@ -476,6 +485,30 @@ function modules_install_to()
...
+  # used to copy only the dtb files and ignore everything else under the dts folder
+  local rsync_params='--include="*/" --include="*.dtb*" --exclude="*"'
+
+  if [[ -d "arch/$architecture/boot/dts/" ]]; then
+    case "$target" in
+      1 | 3) # VM_TARGET or REMOTE_TARGET
+        cp2remote "arch/$architecture/boot/dts/" "/boot/" '' "$rsync_params"
+        ;;
+      2) # LOCAL_TARGET
+        cmd_manager '' "rsync -LrlptD $rsync_params arch/$architecture/boot/dts/ /boot/ --rsync-path='sudo rsync'"
```
rodrigosiqueira, line 504: Do we really need this complicated command? I mean, this is a local deploy, a simple copy should be enough.

tests/remote_test.sh
```diff
@@ -141,59 +141,18 @@ function test_cmd_remote()
...
```
rodrigosiqueira, line 196: These tests are stressing this function by removing the optional parameters and check their behave. I still think this has value because you still using the original parameters.

tests/remote_test.sh
```diff
@@ -255,22 +214,18 @@ function test_preapre_host_deploy_dir()
 function test_prepare_remote_dir()
 {
   local cmd="cat /etc/os-release | grep -w ID | cut -d = -f 2"
-  local remote="172.16.224.1"
+  local remote="localhost"
```
rodrigosiqueira, line 217: Since you are here, use ''

tests/remote_test.sh
```diff
@@ -255,22 +214,18 @@ function test_preapre_host_deploy_dir()
...
   local user="root"
-  local port="2222"
+  local port="22"
```
rodrigosiqueira, line 219: No need for "" here.

rodrigosiqueira (changes requested):

src/deploy.sh
alan-barzilay: as I mentioned previously, this function has a single entry point, which happens after line 105 where the target is checked and the vm is mounted. Am I missing something? my understanding is that the vm will necessarily be mounted at this point

src/deploy.sh
alan-barzilay: yes, unless we want to break this into several comands (which personally I find more brittle and error prone)

rodrigosiqueira:
I submitted a new PR with part of these patches.

PR #399, opened 2021-08-02
src: deploy: Fix kw clear-cache

joaoseckler:
Function cleanup inside src/deploy.sh was not deleting anything because
of quoted star globs. This patch unquotes those.

### DIFF
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -389,14 +389,12 @@ function cleanup()
   local flag=${1:-'SILENT'}
   say 'Cleaning up temporary files...'
 
-  if [[ -v options_values['REMOTE'] ]]; then
-    if [[ -d "$KW_CACHE_DIR/$LOCAL_REMOTE_DIR" ]]; then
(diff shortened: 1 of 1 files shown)


src/deploy.sh
```diff
@@ -391,11 +391,11 @@ function cleanup()
 
   if [[ -v options_values['REMOTE'] ]]; then
```
rodrigosiqueira, line 392: Actually, this if is the problem... idk why we added it... anyway, could you update your PR and remove this `if`?

joaoseckler: When deploy and build were together, when interrupting/signal handling, we would check if the current command was a deploy or a build before cleaning. Somehow this test got in the main cleanup function, probably during some of the recent deploy refactorating

joaoseckler:
The last update to this PR removes the unwanted if statement from the cleanup function in `src/deploy`

alan-barzilay (approved):

rodrigosiqueira:
Merged, thanks

Issue #296, opened 2021-06-01
Make kw compliant with the XDG Base Directory Specification


joaoseckler:
"The XDG Base Directory Specification (XDGBDS) defines four categories of so called DotFiles and the corresponding directories in a users home directory that should be used for those" ([source](https://wiki.debian.org/XDGBaseDirectorySpecification)). Kw should conform to this specification. The following resources can be useful:

- [The specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
- [Debian wiki page](https://wiki.debian.org/XDGBaseDirectorySpecification)

Possible breakdown of the work: use `XDG_*` environmental variables, if they exist, in folders definitions at the top of `./kw`

Issue #385, opened 2021-07-14
Respect XDG config standard locations 


aurabindo:
Make kw look for configuration file in ~/XDG_HOME/.config as per the standard https://wiki.archlinux.org/title/XDG_Base_Directory
XDG might only apply for GUI applications, so I'm not sure if a CLI application like kw should bother, but i think a natural second choice for the config file after the system wide location `/etc/kw` could be assumed to be in `~/.config/kw`

**Expected behavior**
Prioritize a config present in `~/.config/kw/` over  `~/.local/etc/kw/`

alan-barzilay:
possible duplicate with #296.

PS: XDG does apply for non-GUI applications :)

aurabindo:
I agree! Other than the XDG dir, my first instinct would be look for a `~/.kworkflow` rather than in `~/.local/etc/`

rodrigosiqueira:
We already have full support for XDG in the unstable and master branch.
Thanks for the help with this issue.

PR #401, opened 2021-08-02
Make kw compliant with XDG Base Directory Specification

joaoseckler:
Make kw compliant with XDG Base Directory Specification

User customizations should now be done with the XDG_*_HOME environmental
variables, instead of PREFIX. This also fixes some problems with
DEV-MODE: corrects the paths to documentation, cache and data folders.

Closes: #296 and #385

This is still a draft: another commit should implement some transitional installation from the previous directory scheme to this one.

### DIFF
--- a/kw
+++ b/kw
@@ -1,11 +1,31 @@
 #!/bin/bash
 
-# Note: the following line should not be modified, and neither the token should
-# be repeated in this file because It will be replaced during installation with
-# the respective installation prefix.
-KW_INSTALL_PREFIX=${KW_INSTALL_PREFIX:-"##KW_INSTALL_PREFIX_TOKEN##"}
-
 KWORKFLOW=${KWORKFLOW:-"kw"}
+
+# Global paths
+
+# Kw shell files
+KW_LIB_DIR="${KW_LIB_DIR:-"$HOME/.local/lib"}/$KWORKFLOW"
+KW_PLUGINS_DIR="${KW_PLUGINS_DIR:-"$KW_LIB_DIR/plugins"}"
+
+# Share files
+KW_SHARE_DIR="${XDG_DATA_HOME:-"$HOME/.local/share"}/$KWORKFLOW"
+KW_DOC_DIR="$KW_SHARE_DIR/doc"
+KW_MAN_DIR="$KW_SHARE_DIR/man"
+KW_SOUND_DIR="$KW_SHARE_DIR/sound"
+
+# User specific data files (currently this collapses with the share dir,
+# but would not for a system-wide installation)
+KW_DATA_DIR="${XDG_DATA_HOME:-"$HOME/.local/share"}/$KWORKFLOW"
+
+# Configuration files
+KW_ETC_DIR="${XDG_CONFIG_HOME:-"$HOME/.config"}/$KWORKFLOW"
+
+# Cache folder
+KW_CACHE_DIR="${XDG_CACHE_HOME:-"$HOME/.cache"}/$KWORKFLOW"
+
+# State files
+KW_STATE_DIR="${XDG_STATE_HOME:-"$HOME/.local/state"}/$KWORKFLOW"
 
 ##BEGIN-DEV-MODE##
 KW_BIN="$(readlink -f "${BASH_SOURCE[0]}")"
@@ -13,19 +33,15 @@ KW_BASE_DIR="$(dirname "${KW_BIN}")"
 if [ -f "${KW_BASE_DIR}/src/kwlib.sh" ]; then
   # running from source directory
   KW_LIB_DIR="${KW_BASE_DIR}/src"
-  KW_INSTALL_PREFIX="${KW_BASE_DIR}"
+  # KW_DATA_DIR # use default data folder
+  KW_DOC_DIR="${KW_BASE_DIR}/documentation"
+  KW_MAN_DIR="${KW_BASE_DIR}/documentation/man"
+  KW_SOUND_DIR="${KW_BASE_DIR}/sound"
+  KW_ETC_DIR="${KW_BASE_DIR}/etc"
+  KW_PLUGINS_DIR="${KW_LIB_DIR}/plugins"
+  # KW_CACHE_DIR # use default cache folder
 fi
 ##END-DEV-MODE##
-
-# Global paths
-KW_LIB_DIR=${KW_LIB_DIR:-"$KW_INSTALL_PREFIX/lib/kw"}
-KW_SHARE_DOC_DIR=${KW_SHARE_DIR:-"$KW_INSTALL_PREFIX/share/doc/kw/"}
-KW_SHARE_MAN_DIR=${KW_SHARE_DIR:-"$KW_INSTALL_PREFIX/share/man"}
-KW_SHARE_SOUND_DIR=${KW_SHARE_DIR:-"$KW_INSTALL_PREFIX/share/sound/kw"}
-KW_ETC_DIR=${KW_ETC_DIR:-"$KW_INSTALL_PREFIX/etc/kw"}
-KW_DATA_DIR=${KW_DATA_DIR:-"$HOME/.kw"}
-KW_PLUGINS_DIR=${KW_PLUGINS_DIR:-"$KW_LIB_DIR/plugins"}
-KW_CACHE_DIR="$HOME/.cache/kw"
 
 # Export external variables required by kworkflow
 export KWORKFLOW
--- a/setup.sh
+++ b/setup.sh
@@ -10,7 +10,6 @@ debian_packages=(qemu git tar python3-docutils pulseaudio-utils dunst sphinx-doc
 SILENT=1
 VERBOSE=0
 FORCE=0
-PREFIX="${PREFIX:-$HOME/.local}"
 
 declare -r app_name="kw"
 
@@ -18,17 +17,16 @@ declare -r app_name="kw"
 ## Following are the install paths
 ##
 # Paths used during the installation process
-declare -r kwbinpath="$PREFIX/bin/kw"
-declare -r binpath="$PREFIX/bin"
-declare -r libdir="$PREFIX/lib/$app_name"
-declare -r sharedir="$PREFIX/share/"
-declare -r sharedocdir="$sharedir/doc"
-declare -r sharemandir="$sharedir/man"
-declare -r sharesounddir="$sharedir/sound/kw"
-declare -r etcdir="$PREFIX/etc/kw"
-# User specific data
-declare -r datadir="$HOME/.$app_name"
-declare -r cachedir="$HOME/.cache/$app_name"
+declare -r kwbinpath="$HOME/.local/bin/$app_name"
+declare -r binpath="$HOME/.local/bin"
+declare -r libdir="$HOME/.local/lib/$app_name"
+declare -r sharedir="${XDG_DATA_HOME:-"$HOME/.local/share"}/$app_name"
+declare -r docdir="$sharedir/doc"
+declare -r mandir="$sharedir/man"
+declare -r sounddir="$sharedir/sound"
+declare -r datadir="${XDG_DATA_HOME:-"$HOME/.local/share"}/$app_name"
+declare -r etcdir="${XDG_CONFIG_HOME:-"$HOME/.config"}/$app_name"
+declare -r cachedir="${XDG_CACHE_HOME:-"$HOME/.cache/$app_name"}"
 
 ##
 ## Source code references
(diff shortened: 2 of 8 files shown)


kw
```diff
@@ -1,32 +1,35 @@
...
 
-# Note: the following line should not be modified, and neither the token should
-# be repeated in this file because It will be replaced during installation with
-# the respective installation prefix.
-KW_INSTALL_PREFIX=${KW_INSTALL_PREFIX:-"##KW_INSTALL_PREFIX_TOKEN##"}
-
 KWORKFLOW=${KWORKFLOW:-"kw"}
 
+# Global paths
+KW_LIB_DIR="${KW_LIB_DIR:-"$HOME/.local/lib"}/$KWORKFLOW"
```
rodrigosiqueira, line 6: Don't we have anything for `.local/lib` in the XDG?

kw
```diff
@@ -1,32 +1,35 @@
...
+KW_SHARE="${XDG_DATA_HOME:-"$HOME/.local/share"}/$KWORKFLOW"
+KW_SHARE_DOC_DIR="$KW_SHARE_DIR/doc"
+KW_SHARE_MAN_DIR="$KW_SHARE_DIR/man"
+KW_SHARE_SOUND_DIR="$KW_SHARE_DIR/sound"
```
rodrigosiqueira, line 10: Just for making this part of the code easier to read, could you add a new line between different sections? Maybe add a comment as well.

setup.sh
```diff
@@ -10,34 +10,32 @@ debian_packages=(qemu git tar python3-docutils pulseaudio-utils dunst sphinx-doc
...
 ##
 # Paths used during the installation process
-declare -r kwbinpath="$PREFIX/bin/kw"
-declare -r binpath="$PREFIX/bin"
-declare -r libdir="$PREFIX/lib/$app_name"
-declare -r sharedir="$PREFIX/share/"
+declare -r kwbinpath="$HOME/.local/bin/$app_name"
+declare -r binpath="$HOME/.local/bin"
+declare -r libdir="$HOME/.local/lib/$app_name"
+declare -r sharedir="${XDG_DATA_HOME:-"$HOME/.local/share"}/$app_name"
```
rodrigosiqueira, line 23: Why did you remove `PREFIX`? I think it looks better with that...

rodrigosiqueira (commented):
Hi @joaoseckler 
In general, it lgtm. Send a new version, I think you can drop the "draft' status.

Also, could test:
- kw b
- kw bd
- kw d --list
- kw d -u KERNEL
- kw p -t 20s
- kw p -l

kw
joaoseckler: Apparently not...

setup.sh
joaoseckler: What I originally thougth is that "$PREFIX" conveys the idea that all of kw is installed under one directory, which is not true anymore. Also, we start working with a lot of variables at the same time (see example below). For me is a little bit less clear what's going on. What do you think?

```
PREFIX="$HOME/.local"
KW_SHARE="${XDG_DATA_HOME:-"$PREFIX/share"}/$KWORKFLOW"
KW_SHARE_DOC_DIR="$KW_SHARE/$KWORKFLOW"
```

joaoseckler:
The last update to this PR:

- Adds comments to the folder definitions
- Adds a legacy_folders function to setup.sh to move from old structure to new one
- Merges share and data folder variables (e.g. KW_SHARE_DIR and KW_DATA_DIR). According to the XDG specification, these two are the same. This means that we'll have something like this in ~/.local/share/kw:

```
ls ~/.local/share/kw
configs  doc  man  pomodoro_current.log  sound  statistics
```

joaoseckler:
The last update to this PR makes src/kw_config_loader.sh compliant with the xdg specification. This means that `.kw/kworkflow.config` is no longer searched for configuration files (unless it is listed in `XDG_CONFIG_DIRS`

kw
```diff
@@ -1,32 +1,48 @@
...
+
+# User specific data files (currently this collapses with the share dir,
+# but would not for a system-wide installation)
+KW_DATA_DIR="${XDG_DATA_HOME:-"$HOME/.local/share"}/$KWORKFLOW"
+
+# Configuration files
+KW_ETC_DIR="${XDG_CONFIG_HOME:-"$HOME/.config"}/$KWORKFLOW"
+
+# Cache folder
+KW_CACHE_DIR="${XDG_CACHE_HOME:-"$HOME/.cache"}/$KWORKFLOW"
```
rodrigosiqueira, line 25: Even if we use `$XDG_{CONFIG,CACHE}_HOME`, I think we should have a `kw` dir (i.e., `$XDG_{CONFIG,CACHE}_HOME/kw`). Fwiu from the spec, that's not a problem.

setup.sh
rodrigosiqueira: Ok, I agree.

setup.sh
```diff
@@ -162,6 +159,27 @@ function confirm_complete_removal()
...
+
+    mv "$HOME/.kw" "$datadir"
+    mv "$prefix/share/doc" "$docdir"
+    mkdir -p "$mandir"
+    mv "$prefix/share/man/kw.rst" "$mandir/"
+    mv "$prefix/share/sound/kw" "$sounddir"
+    mv "$prefix/etc/kw" "$etcdir"
+    rmdir --ignore-fail-on-non-empty "$prefix/share/man" "$prefix/share/sound" \
+      "$prefix/etc"
+  fi
```
rodrigosiqueira, line 180: Technically, if it is a fresh installation, you don't need to move things around; you just need to remove the old code. The only thing you might want to move is the file under `~/.kw/`.

src/kw_config_loader.sh
```diff
@@ -141,9 +141,21 @@ function parse_configuration()
...
+  read -r -a config_dirs <<< "${XDG_CONFIG_DIRS:-"/etc/xdg"}"
+  unset IFS
+
+  for folder in "$PWD" "$config_home" "${config_dirs[@]/%/"/kw"}" \
+    "$KW_ETC_DIR"; do
+    if [[ -f "$folder/$CONFIG_FILENAME" ]]; then
+      parse_configuration "$folder/$CONFIG_FILENAME"
+      return
+    fi
+  done
```
rodrigosiqueira, line 158: Why not use a simple set of `parse_configuration` as we had before. Maybe I did not get the idea, but this looks a little bit overkill to me.

rodrigosiqueira (changes requested):
Hi @joaoseckler ,
In general, lgtm :)
See some of my comments.

kw
joaoseckler: Yes, that's in the code! See the `/$KWORKFLOW` (= `kw`) at the end of the variable

src/kw_config_loader.sh
joaoseckler: You're right, I confused some things. I thought only one file had to be read, but actually all of them must be read. I'll change back to the previous behavior

joaoseckler: You're right, I confused some things. I thought only one file had to be read, but actually all of them must be read. I'll change back to the previous behavior

setup.sh
```diff
@@ -162,6 +159,27 @@ function confirm_complete_removal()
...
+  if [[ -d "$prefix/share/doc" && -f "$prefix/share/man/kw.rst" &&
+    -d "$prefix/share/sound/kw" && -d "$prefix/etc/kw" &&
+    -d "$HOME/.kw" ]]; then
+
+    mv "$HOME/.kw" "$datadir"
+    mv "$prefix/share/doc" "$docdir"
+    mkdir -p "$mandir"
+    mv "$prefix/share/man/kw.rst" "$mandir/"
+    mv "$prefix/share/sound/kw" "$sounddir"
+    mv "$prefix/etc/kw" "$etcdir"
```
rodrigosiqueira, line 177: I don't think we need to move these things around; we just need to remove them.

src/kw_config_loader.sh
```diff
@@ -142,7 +142,14 @@ function parse_configuration()
 function load_configuration()
 {
   parse_configuration "$KW_ETC_DIR/$CONFIG_FILENAME"
-  parse_configuration "$HOME/.kw/$CONFIG_FILENAME"
+
+  # XDG_CONFIG_DIRS is in order of preference, but we must first read
+  # from the folders with less preference
+  while read -d ':' -r dir; do
+    parse_configuration "$(rev <<< "$dir")/$KWORKFLOW/$CONFIG_FILENAME"
+  done <<< "$(rev <<< "${XDG_CONFIG_DIRS:-"/etc/xdg"}"):"
```
rodrigosiqueira, line 150: What is this XDG order preference? Do we really need it? Can we make this easier to read? Also, can you expand the explanation about the order in the comment?

rodrigosiqueira (changes requested):

src/kw_config_loader.sh
joaoseckler: We need it to be compliant with the specification... I made some changes for the ease of reading and added comments.

setup.sh
```diff
@@ -162,6 +159,26 @@ function confirm_complete_removal()
...
+
+  if [[ -d "$prefix/share/doc" && -f "$prefix/share/man/kw.rst" &&
+    -d "$prefix/share/sound/kw" && -d "$prefix/etc/kw" &&
+    -d "$HOME/.kw" ]]; then
+
+    mv "$HOME/.kw/"* "$datadir"
+    rm -rf "$prefix/share/doc"
+    rm -rf "$prefix/share/man/kw.rst"
+    rm -rf "$prefix/share/sound/kw"
+    mv "$prefix/etc/kw" "$etcdir"
```
rodrigosiqueira, line 176: In the `mv` cases, let's add a message that tells the user that we changed those file locations from X to Y.

src/kw_config_loader.sh
```diff
@@ -141,8 +141,22 @@ function parse_configuration()
...
+  unset IFS
+
   parse_configuration "$KW_ETC_DIR/$CONFIG_FILENAME"
-  parse_configuration "$HOME/.kw/$CONFIG_FILENAME"
+
+  # XDG_CONFIG_DIRS is a colon-separated list of directories for config
+  # files to be searched, in order of preference. Since this function
+  # reads config files in a reversed order of preference, we must
+  # traverse it from back to top
+  for ((i = ${#config_dirs[@]} - 1; i >= 0; i--)); do
```
rodrigosiqueira, line 155: I think it will be easier to read this code if you move `${#config_dirs[@]} - 1` outside this `for` loop.
Also, in the comment, could you add one example that justifies this `for`?

src/kw_config_loader.sh
```diff
@@ -141,8 +141,22 @@ function parse_configuration()
...
+    parse_configuration "${config_dirs["$i"]}/$KWORKFLOW/$CONFIG_FILENAME"
+  done
+
+  parse_configuration "${XDG_CONFIG_HOME:-"$HOME/.config"}/$KWORKFLOW/$CONFIG_FILENAME"
```
rodrigosiqueira, line 159: fwiu from the `setup.sh`, we will move `$HOME/.config` to the new folder. If I am correct, do we really need this `${XDG_CONFIG_HOME:-"$HOME/.config"}`?

tests/init_test.sh
```diff
@@ -9,7 +9,7 @@ FAKE_CONFIG_PATH="$FAKE_DIR/.config"
 function setUp()
 {
   export KW_ETC_DIR="tests/samples"
-  export KW_SHARE_SOUND_DIR="tests/samples/share/sound/kw"
+  export KW_SOUND_DIR="tests/samples/share/sound/kw"
```
rodrigosiqueira, line 12: Since you're here, could you replace "" to ''?

rodrigosiqueira (changes requested):
Hi @joaoseckler ,
lgtm, I added some comments.

One question, if we use the XDG variable in an environment without a desktop manager or maybe via ssh, does this XDG variable still available? For example, if I ssh to a machine, install kw, and try to use `kw d` to another machine, the XDG variable will work as expected?

joaoseckler:
> One question, if we use the XDG variable in an environment without a desktop manager or maybe via ssh, does this XDG variable still available? For example, if I ssh to a machine, install kw, and try to use `kw d` to another machine, the XDG variable will work as expected?

The XDG_* variables are only used by users that wish to configure their home page differently from the standard locations. The OS does not initialize those (check you shell: probably only some desktop and session XDG_* variables are defined, not the ones we use here). The typical use case for these is to declare them at a .profile or .bashrc. Complying to the XDG base directory specification means attending to the wishes of these users which declare XDG_* variables for themselves.

If it works in kw: every use of a XDG_* variable should be accompanied by a fallback, e.g.: ${XDG_DATA_DIRS:-/etc/xdg}. If you see some of these variables without the fallback, we have a problem!

src/kw_config_loader.sh
joaoseckler: Actually, $HOME/.config is the new folder!

rodrigosiqueira:
Hi @joaoseckler ,
Thanks a lot for this PR! Merged to the unstable branch.

Issue #147, opened 2020-01-13
Get .config file from a target machine


rodrigosiqueira:
It could be interesting if we can get the `.config` file from a machine in a network. Something like:

```
kw configm --fetch # Get a .config file
kw configm --fetch --optimize # Get a .config file optimized with localmodconfig
kw configm --fetch IP:PORT
```

rodrigosiqueira:
https://github.com/kworkflow/kworkflow/commit/f984b81c456d138bda53e7715e220afea6404b8e

PR #403, opened 2021-08-05
Add option to get .config file to configm

magalilemes:
This PR introduces `--fetch` to `kw configm`, which can be used with `--optimize` and `--remote.` Below are a few usage examples:

`kw configm --fetch` gets the .config file from the local machine.
`kw configm --fetch --remote user@ip:port` gets the .config file from the specified machine.
`kw configm --fetch --optimize --remote user@ip:port` runs make localmodconfig to get an optimized .config from the remote machine.

Relates to #147.

### DIFF
--- a/src/config_manager.sh
+++ b/src/config_manager.sh
@@ -1,4 +1,5 @@
 include "$KW_LIB_DIR/kwio.sh"
+include "$KW_LIB_DIR/remote.sh"
 include "$KW_LIB_DIR/kwlib.sh"
 include "$KW_LIB_DIR/signal_manager.sh"
 
@@ -15,7 +16,7 @@ function config_manager_help()
     return
   fi
   printf '%s\n' 'kw config manager:' \
-    '  configm --fetch [(-o | --output) <filename>] [-f | --force] [--optimize] - Fetch a config' \
+    '  configm --fetch [(-o | --output) <filename>] [-f | --force] [--optimize] [--remote [<user>@<ip>:<port>]] - Fetch a config' \
     '  configm (-s | --save) <name> [(-d | --description) <description>] [-f | --force] - Save a config' \
     '  configm (-l | --list) - List config files under kw management' \
     '  configm --get <name> [-f | --force] - Get a config file based named <name>' \
@@ -114,14 +115,19 @@ function cleanup()
 #        overwritten.
 # @output File name. This option requires an argument, which will be the name of
 #         the .config file to be retrieved.
-# @optimize Optimization flag. If it's set, then an optimized version of the
-#           .config file will be retrieved.
+# @optimize Optimize flag, if set then 'make localmodconfig' is used.
+# @target Target can be 2 (LOCAL_TARGET), and 3 (REMOTE_TARGET).
 function fetch_config()
 {
   local flag="$1"
   local force="$2"
   local output="$3"
   local optimize="$4"
+  local target="$5"
+  local user="${remote_parameters['REMOTE_USER']}"
+  local ip="${remote_parameters['REMOTE_IP']}"
+  local port="${remote_parameters['REMOTE_PORT']}"
+  local kernel_release
   local mods
   local cmd
   local arch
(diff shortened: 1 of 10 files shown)


magalilemes:
I have yet to write tests, but would like to have some early feedback to know if I'm approaching the problem correctly.

rodrigosiqueira (changes requested):
Hi @magalilemes ,

It looks like this feature is not working with the remote option; I recommend you check it. Also, you first need to conduct a code refactor in the configm to separate the parameter parser options from the execution; I recommend that your first commit create a function named `parse_configm_options`, and `option_values` map. Inside `parse_configm_options`, use getopt and `option_values` to handle the user parameters (you probably need to refactor the tests).

Thanks

rodrigosiqueira:
@magalilemes , just for checking, is this still a draft?

magalilemes:
> @magalilemes , just for checking, is this still a draft?

@rodrigosiqueira  Not anymore. You may review it, please.

melissawen:
Hi @magalilemes @rodrigosiqueira,

is there any chance to include get .config for raspbian using --remote?
for this, just need: `sudo modprobe configs` before get from /proc/config.gz

hmm.. or maybe it should be extended when rpi deploy is ready...

src/config_manager.sh
```diff
@@ -327,6 +347,10 @@ function parse_configm_options()
         options_values['FETCH']=1
         shift
         ;;
+      --optimize)
+        options_values['OPTIMIZE']=1
```
rodrigosiqueira, line 379: By default, let's initialize this variable as ''.

src/config_manager.sh
```diff
@@ -80,21 +81,45 @@ function save_config_file()
 # @flag How to display a command.
 # @optimize 1 if .config should be optimized, 0 otherwise.
 # @distro distribution name from target machine
+# @remote a string such as user@ip:port
 function fetch_config()
 {
   local flag="$1"
   local optimize="$2"
+  local remote="$3"
```
rodrigosiqueira, line 89: I think this should be `target`.

rodrigosiqueira (changes requested):
Hi @magalilemes ,

In general, it looks good, but it still requires more effort. Aside from my inline comments, please take a look at:

1. Add tests
2. Ask users before replacing the current `.config`
3. Add an option to save the `.config` with a different name, maybe `-o | --output`
4. The option `--optimize` still not working well, I'm getting this error:

```
using config: '.config'
glue_helper config not found!!
System keyring enabled but keys "debian/canonical-certs.pem" not found. Resetting keys to default value.
module nfs_ssc did not have configs CONFIG_NFS_V4_2_SSC_HELPER
*
* Restart config...
*
*
* BPF subsystem
*
Enable bpf() system call (BPF_SYSCALL) [Y/n/?] y
Enable BPF Just In Time compiler (BPF_JIT) [Y/n/?] y
  Permanently enable BPF JIT and remove BPF interpreter (BPF_JIT_ALWAYS_ON) [Y/n/?] y
Disable unprivileged BPF by default (BPF_UNPRIV_DEFAULT_OFF) [N/y/?] (NEW) ^Cmake[1]: *** [scripts/kconfig/Makefile:53: localmodconfig] Interrupt
make: *** [Makefile:614: localmodconfig] Interrupt
```

src/config_manager.sh
```diff
@@ -75,12 +75,18 @@ function save_config_file()
...
 # @flag How to display a command.
+# @optimize 1 if .config should be optimized, 0 otherwise.
+# @distro distribution name from target machine
 function fetch_config()
 {
   local flag="$1"
+  local optimize="$2"
+  local mods
+
+  mods=$(cmd_manager 'SILENT' 'lsmod')
```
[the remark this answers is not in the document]
magalilemes: In this specific case, it has to be set to SILENT since we don't want the command to appear above the output of the command.

src/config_manager.sh
```diff
@@ -197,53 +198,103 @@ function execute_config_manager()
...
+        config_manager_help
+        exit 0
+        ;;
+      -f)
+        options_values['FORCE']=1
+        shift
+        ;;
+      --save)
+        # Validate string name
+        if [[ "$2" =~ ^- || -z "${2// /}" ]]; then
```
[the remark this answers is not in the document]
magalilemes: I believe this is supposed to avoid things like `kw configm --save --get`. We were using the exact same thing before getopt as well.

rodrigosiqueira:
Hi @magalilemes , did you update this series? I noticed that you did not address some of my comments, such as "add `--description`". Also, could you rebase your change?

magalilemes:
> Hi @magalilemes , did you update this series? I noticed that you did not address some of my comments, such as "add `--description`". Also, could you rebase your change?

@rodrigosiqueira I haven't updated them yet; still working on it

magalilemes:
A few remarks:

1) @rodrigosiqueira  I think the errors you're getting are something you'd get using `make localmodconfig`, I don't know how to avoid that.
2) `make localmodconfig` writes directly to .config, so using `--output` only works if we're not using `--optimize`. Should this behavior be kept, or should a way around it be found?
3) The only way I could think of to write tests was declaring a variable named _root_, assigning it to '/', and when testing assigning it to _SHUNIT_TMPDIR_, but I'm not sure if this is an OK approach.

src/config_manager.sh
```diff
@@ -11,10 +14,11 @@ function config_manager_help()
...
   fi
   printf '%s\n' 'kw config manager:' \
-    '  configm --save <name> [-d <description>] [-f] - save a config' \
+    '  configm --fetch [(-o | --output) <filename>] [-f | --force] [--optimize] [--remote [<remote>:<port>]] - Fetch a config' \
+    '  configm (-s | --save) <name> [(-d | --description) <description>] [-f | --force] - Save a config' \
     '  configm (-l | --list) - List config files under kw management' \
-    '  configm --get <name> [-f] - Get a config file based named <name>' \
-    '  configm (-rm | --remove) <name> [-f] - Remove config labeled with <name>'
+    '  configm --get <name> [-f | --force] - Get a config file based named <name>' \
+    '  configm (-r | --remove) <name> [-f | --force] - Remove config labeled with <name>'
```
joaoseckler, line 21: Nice change (from -rm to -r)

src/config_manager.sh
```diff
@@ -77,6 +81,110 @@ function save_config_file()
...
+  local kernel_release
+  local mods
+  local cmd
+  local ret
+
+  output=${output:-'.config'}
+
+  case "$target" in
+    2) # LOCAL_TARGET
+      if [[ -f "${root}proc/config.gz" ]]; then
```
joaoseckler, line 113: Maybe `if [[ -f "${root}proc/config.gz" ]] && command_exists 'zcat'; then`?

joaoseckler:
>     2. `make localmodconfig` writes directly to .config, so using `--output` only works if we're not using
>     `--optimize`. Should this behavior be kept, or should a way around it be found?


It seems like localmodconfig only calls other kernel scripts... maybe a good but hard way is to go down this path:

```make
PHONY += localmodconfig localyesconfig
localyesconfig localmodconfig: $(obj)/conf
	$(Q)$(PERL) $(srctree)/$(src)/streamline_config.pl --$@ $(srctree) $(Kconfig) > .tmp.config
	$(Q)if [ -f .config ]; then 				\
		cmp -s .tmp.config .config ||			\
		(mv -f .config .config.old.1;			\
		 mv -f .tmp.config .config;			\
		 $< $(silent) --oldconfig $(Kconfig);		\
		 mv -f .config.old.1 .config.old)		\
	else							\
		mv -f .tmp.config .config;			\
		$< $(silent) --oldconfig $(Kconfig);		\
	fi
	$(Q)rm -f .tmp.config
```
If I understood this correctly, we'd only need to call: `perl scripts/kconfig/streamline_config.pl --localmodconfig . Kconfig > tmp.config` and we're done! (did not test this)

>     3. The only way I could think of to write tests was declaring a variable named _root_, assigning it to '/', and when testing assigning it to _SHUNIT_TMPDIR_, but I'm not sure if this is an OK approach.

I think this is ok, similar thing is done in `src/plugins/kernel_install/utils.sh` in the function list_installed_kernels. The approach there was to use a local variable passed as an argument to the function. I think that your approach to use a global variable is cleaner (but I'd declare it with `declare -g` in the top of the file)

magalilemes:
> I think this is ok, similar thing is done in `src/plugins/kernel_install/utils.sh` in the function list_installed_kernels. The approach there was to use a local variable passed as an argument to the function. I think that your approach to use a global variable is cleaner (but I'd declare it with `declare -g` in the top of the file)

@joaoseckler I hadn't seen that. Really cool, I might take some inspiration from it. Thanks

rodrigosiqueira:
@magalilemes could you explore Joao's suggestions? I think it might work for us.

magalilemes:
In a discussion, it was noticed that this condition `[[ -f "/boot/config-$(uname -r)" ]]` is not always true since _kw deploy_ doesn't copy the .config file to the /boot folder of the target machine. So we thought that we might address this by copying .config when deploying, but after trying to do this, I faced a few problems:
1) It might happen that, at the time of the deploy, the .config file in the kernel directory is different from the one that was used during the compilation, so we'd end up copying a somehow inaccurate (?) file.
2) We could copy the .config generated by running `scripts/extract-ikconfig <kernel_image>`, but this might fail if the kernel wasn't compiled with _CONFIG_IKCONFIG_ enabled.

I don't know which of these options should be used, or if there's another one.

rodrigosiqueira:
Hi @magalilemes ,

It is interesting how something that we think is simple turn out to be a very challenging task hahaha
Anyway, thanks a lot for keeping dig in this topic.

I have some thoughts to try to solve this problem:

# Scenario 1: kw bd

If the user uses the command `kw bd`, we can safely copy the current `.config` file.

# Scenario 2: kw deploy

Maybe we can try this heuristic:

1. Get the kernel name that will be deployed (make kernelrelease)
2. Check if the kernel name matches with the current config file version (probably we can check CONFIG_LOCALVERSION=)
2.1. If it matches, deploys this config file
2.2. If not, do not deploy (maybe add a message like "undefined .config file for the target kernel. Consider to use kw bd")

# Back to the configm problem

Getting back to `configm --fetch --optimize` problem. Although the above approach may work for 80-90% of the cases, it does not solve the case where the target kernel does not have a config file associate with it. In that case, I propose the following strategy:

1. Check if there is a config file associate with the kernel running in the target machine.
1.1. If we have, it is the happy case, and you know what to do.
1.2. If we don't have.
1.2.1. Do we have a config-VERSION file in the folder /boot directory? The one distributed by the distribution? If yes, let grab it and use it.
1.2.2. Do we have access to /proc/config? If so, use it.
1.2.3. None of the above cases matches; let's grab the standard config for that architecture and warning the user about that.

magalilemes:
Hey, @rodrigosiqueira.
> ### Back to the configm problem
> 
> [...]
> 
>     1. Check if there is a config file associate with the kernel running in the target machine.
>        1.1. If we have, it is the happy case, and you know what to do.
How do I check that? From what I first thought, we would check if there's a matching config file in the /boot folder, but you presented this option below, so I guess what I thought isn't the correct approach for this case.

>        1.2. If we don't have.
>        1.2.1. Do we have a config-VERSION file in the folder /boot directory? The one distributed by the distribution? If yes, let grab it and use it.
>        1.2.2. Do we have access to /proc/config? If so, use it.
>        1.2.3. None of the above cases matches; let's grab the standard config for that architecture and warning the user about that.
This means running `make ARCH=<arch> defconfig`?

rodrigosiqueira:
> Hey, @rodrigosiqueira.
> 
> > ### Back to the configm problem
> > [...]
> > ```
> > 1. Check if there is a config file associate with the kernel running in the target machine.
> >    1.1. If we have, it is the happy case, and you know what to do.
> > ```
> 
> How do I check that? From what I first thought, we would check if there's a matching config file in the /boot folder, but you presented this option below, so I guess what I thought isn't the correct approach for this case.

Just check for config-KERNEL-VERSION. For example, on Ubuntu generic, we have:

config-5.11.0-25-generic

> 
> > ```
> >    1.2. If we don't have.
> >    1.2.1. Do we have a config-VERSION file in the folder /boot directory? The one distributed by the distribution? If yes, let grab it and use it.
> >    1.2.2. Do we have access to /proc/config? If so, use it.
> >    1.2.3. None of the above cases matches; let's grab the standard config for that architecture and warning the user about that.
> > ```
> 
> This means running `make ARCH=<arch> defconfig`?

yes

rodrigosiqueira:
Hi @magalilemes  just for checking... do you have plans to finish this PR? If you are busy, I can take over and finish it for you (I'll keep your authorship).

magalilemes:
> Hi @magalilemes just for checking... do you have plans to finish this PR? If you are busy, I can take over and finish it for you (I'll keep your authorship).

@rodrigosiqueira I plan on finishing it.

magalilemes:
This last version addresses the following points:
1) When `kw bd` is run, .config is copied.
2) Depending on what we have, when `kw deploy` is run, we may copy .config if it matches the kernel name.
3) Now we use `make defconfig` if we cannot find an appropriate configuration file.
4) I realized that this feature wasn't dealing with interruptions, which could lead to losing configuration files. Now, it uses `signal_manager` to restore files and remove any other temporary file that we might have created along the way.

rodrigosiqueira:
Hi @magalilemes, sorry for my colossal delay in reviewing this PR. Due to my delay, I rebased this change and updated it myself. I hope you don't mind.

Anyway, change applied to the unstable branch.

PR #404, opened 2021-08-05
Make kworkflow.config customizable with kw init 

magalilemes:
This allows customization of a few variables (arch, target and remote ones) from command line at the time a new `kworkflow.config` is created. For instance, one could run:

`kw init --arch arm --target remote` which would make the brand new kworkflow.config file have these variables set as desired.


src/init.sh
```diff
@@ -5,25 +5,60 @@ include "$KW_LIB_DIR/kwio.sh"
...
 #
 # Returns:
-# In case of failure, this function return ENOENT.
+# In case of failure, this function returns ENOENT.
 function init_kw()
 {
   local config_file_template="$KW_ETC_DIR/kworkflow_template.config"
   local name="kworkflow.config"
 
+  init_parser "$@"
```
rodrigosiqueira, line 24: In my view, `kw init` should only be invoked inside the kernel. Let's add `if ! is_kernel_root "$PWD"; then` before `init_parser`.

src/init.sh
```diff
@@ -5,25 +5,60 @@ include "$KW_LIB_DIR/kwio.sh"
...
+
   if [[ -f "$config_file_template" ]]; then
     mkdir -p "$PWD/$KW_DIR"
     cp "$config_file_template" "$PWD/$KW_DIR/$name"
     sed -i -e "s/USERKW/$USER/g" -e "s,SOUNDPATH,$KW_SHARE_SOUND_DIR,g" -e "/^#?.*/d" \
       "$PWD/$KW_DIR/$name"
+
+    if [[ -n "${options_values['arch']}" ]]; then
+      sed -i -r "s/(arch=)x86_64/\1${options_values['arch']}/" "$PWD/$KW_DIR/$name"
```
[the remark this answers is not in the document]
rodrigosiqueira: Additionally, let's validate the arch passed by the user. Basically, we can accept any arch described in the folder `arch/`; if the arch specified by the user does not exist, we can fail and ask the user to use `--force` option if they are sure that they want to use that arch. We can also list the valid archs in the failure.

src/init.sh
```diff
@@ -28,6 +32,17 @@ function init_kw()
     if [[ -n "${options_values['arch']}" ]]; then
       sed -i -r "s/(arch=)x86_64/\1${options_values['arch']}/" "$PWD/$KW_DIR/$name"
     fi
+
+    if [[ -n "${options_values['remote']}" ]]; then
+      user=$(get_based_on_delimiter "${options_values['remote']}" "@" 1)
+      remote=$(get_based_on_delimiter "${options_values['remote']}" "@" 2)
+      ip=$(get_based_on_delimiter "$remote" ":" 1)
+      port=$(get_based_on_delimiter "$remote" ":" 2)
```
rodrigosiqueira, line 40: This is the deprecated way to get these sorts of info, `populate_remote_info` and `remote_parameters[..]`

src/init.sh
```diff
@@ -43,6 +43,10 @@ function init_kw()
       sed -i -r "s/(ssh_ip=)localhost/\1$ip/" "$PWD/$KW_DIR/$name"
       sed -i -r "s/(ssh_port=)22/\1$port/" "$PWD/$KW_DIR/$name"
     fi
+
+    if [[ -n "${options_values['target']}" ]]; then
```
rodrigosiqueira, line 47: Let's validate the target. It can only be `vm`, `local`, or `remote`.

rodrigosiqueira (changes requested):
Hi @magalilemes ,
You are on the right path; I added a couple of comments. Try to address all of them.

src/kw_config_loader.sh
```diff
@@ -127,7 +127,16 @@ function load_configuration()
 {
   parse_configuration "$KW_ETC_DIR/$CONFIG_FILENAME"
   parse_configuration "$HOME/.kw/$CONFIG_FILENAME"
-  parse_configuration "$PWD/$CONFIG_FILENAME"
+
+  # New users may not have kworkflow.config at $PWD
```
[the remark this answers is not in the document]
magalilemes: It's something like: new users may not have the file, so we check before running the command that relies on the file so it doesn't throw an error for not having found that file.
The second comment `# Old users may not have used kw init yet, so they wouldn't have .kw` follows the same logic: since old users may not have .kw, we'll check if the directory exists before running a command that uses it. Does that make sense?

magalilemes:
PR updated

tests/init_test.sh
```diff
@@ -3,49 +3,64 @@
...
+    "Initialized kworkflow directory in $SHUNIT_TMPDIR/$KWORKFLOW/$KW_DIR based on $USER data"
+  )
+
   output=$(init_kw)
+  assertEquals "($LINENO)" 'This command should be run in a kernel tree.' "$output"
 
-  kworkflow_content=$(grep "$USER" -o "$path_config" | head -n 1)
+  mk_fake_kernel_root "$SHUNIT_TMPDIR/$KWORKFLOW/"
+  cd "$SHUNIT_TMPDIR/$KWORKFLOW/" || {
+    fail "($LINENO) It was not possible to move to temporary directory"
```
rodrigosiqueira, line 37: Replace `($LINENO)` for `($LINENO):` in all your tests.

src/init.sh
```diff
@@ -46,6 +47,14 @@ function init_kw()
         find "$PWD/arch" -maxdepth 1 -mindepth 1 -type d -printf '%f\n'
       fi
     fi
+
+    if [[ -n "${options_values['remote']}" ]]; then
+      populate_remote_info "${options_values['remote']}"
```
rodrigosiqueira, line 52: `populate_remote_info` might fail if the user gives a bad parameter. Check this function return and let the user know if they pass something wrong.

tests/init_test.sh
```diff
@@ -57,6 +57,16 @@ function test_init_kw()
   kworkflow_content=$(grep arch= "$path_config")
   assertEquals "($LINENO)" 'arch=baroque' "$kworkflow_content"
 
+  output=$(init_kw --remote juca@123.456.789.123:2222)
```
rodrigosiqueira, line 60: Try to set an invalid remote.

src/init.sh
```diff
@@ -33,6 +54,53 @@ function init_kw()
...
+
+  options="$(getopt \
+    --name 'kw init' \
+    --options "$short_options" \
+    --longoptions "$long_options" \
+    -- "$@")"
+
+  if [[ "$?" != 0 ]]; then
+    return 22 # EINVAL
+  fi
```
rodrigosiqueira, line 102: Could you add a test that reaches this condition?

src/kw_config_loader.sh
```diff
@@ -143,7 +145,22 @@ function load_configuration()
...
+  if [[ -f "$PWD/$CONFIG_FILENAME" ]]; then
+    warning 'We will stop supporting kworkflow.config soon.'
+    if is_kernel_root "$PWD" &&
+      [[ $(ask_yN 'Do you want to update your configurations to .kw?') =~ 1 ]]; then
+      mkdir -p "$PWD/.kw/"
```
rodrigosiqueira, line 154: Could you add a test that validates this part?

src/kw_config_loader.sh
```diff
@@ -143,7 +145,22 @@ function load_configuration()
...
+      mv -i "$PWD/$CONFIG_FILENAME" "$PWD/.kw/$CONFIG_FILENAME"
+    fi
+    parse_configuration "$PWD/$CONFIG_FILENAME"
+  fi
+
+  # Old users may not have used kw init yet, so they wouldn't have .kw
+  if [[ -f "$PWD/.kw/$CONFIG_FILENAME" ]]; then
+    parse_configuration "$PWD/.kw/$CONFIG_FILENAME"
+  fi
```
rodrigosiqueira, line 163: Also, add a test for this part.

rodrigosiqueira (changes requested):
Hi @magalilemes ,
Great PR! Thanks a lot for your new version.

Please, see my comments in the code.

src/init.sh
```diff
@@ -20,11 +23,29 @@ function init_kw()
...
+
+    if [[ -n "${options_values['arch']}" ]]; then
+      if [[ -d "$PWD/arch/${options_values['arch']}" || -n "${options_values['force']}" ]]; then
+        set_config_value 'arch' "${options_values['arch']}"
+      elif [[ -z "${options_values['force']}" ]]; then
+        complain 'This arch was not found in the arch directory'
+        complain 'You can use --force next time if you want to proceed anyway'
+        say 'Available architectures:'
+        find "$PWD/arch" -maxdepth 1 -mindepth 1 -type d -printf '%f\n'
+      fi
```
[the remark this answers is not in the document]
magalilemes: I don't know. Even if we return 22, kw directory will have been initialized, so I thought it was a good idea not to return anything and let the message tell the user that, even though the parameter was wrong, we initialized the directory.

rodrigosiqueira (changes requested):
@magalilemes ,
Everything lgtm. The only problem is the message `We will stop supporting kworkflow.config soon.` showed during the test execution; could you create a new commit where you update all tests that uses `$PWD/kworkflow.config` instead of `$PWD/.kw/kworkflow.config`. If you run all tests, you will see these warnings.

documentation/man/features/init.rst
```diff
@@ -10,6 +10,40 @@ SYNOPSIS
 
 DESCRIPTION
 ===========
-This command creates a **kworkflow.config** file in the current directory. The
-primary reason for running kw init is to pick up a freshly created config
-file.
+This command creates a **.kw** folder containing a kworkflow.config file in the
```
kwy95, line 13: Enclose `kworkflow.config` in double asterisks, do this for files and folder names.

documentation/man/features/init.rst
```diff
@@ -10,6 +10,40 @@ SYNOPSIS
...
+current kernel directory. The primary reason for running kw init is to pick up a
```
kwy95, line 14: `kw init` should be inside double backticks, as it's a command

documentation/man/features/init.rst
```diff
@@ -10,6 +10,40 @@ SYNOPSIS
...
+
+OPTIONS
+=======
+\--arch <arch>:
+  Set the variable arch from the newly created kworkflow.config file. Before
+  actually changing it, this option checks if <arch> is a valid arch found in
+  the arch folder from the kernel directory.
+
+\--remote <user>@<ip>:<port>:
+  Set the variables ssh_user, ssh_ip, and ssh_port to <user>, <ip>, and <port>,
```
kwy95, line 25: Enclose placeholders and variable names in single asterisks.

kwy95 (changes requested):
Small documentation changes. I commented only once per problem, but apply the changes everywhere to all occurrences.

kwy95 (approved):

src/kw_config_loader.sh
```diff
@@ -142,8 +145,25 @@ function parse_configuration()
 function load_configuration()
 {
   parse_configuration "$KW_ETC_DIR/$CONFIG_FILENAME"
-  parse_configuration "$HOME/.kw/$CONFIG_FILENAME"
-  parse_configuration "$PWD/$CONFIG_FILENAME"
+  parse_configuration "$HOME/$KW_DIR/$CONFIG_FILENAME"
+
+  # New users may not have kworkflow.config at $PWD
+  if [[ -f "$PWD/$CONFIG_FILENAME" ]]; then
+    warning 'We will stop supporting kworkflow.config soon.'
```
rodrigosiqueira, line 152: Maybe "We will stop to support kworkflow.config in the kernel root directory in favor of using a .kw/ directory."

src/kw_config_loader.sh
```diff
@@ -142,8 +145,25 @@ function parse_configuration()
...
+    if is_kernel_root "$PWD" &&
+      [[ $(ask_yN 'Do you want to update your configurations to .kw?') =~ 1 ]]; then
```
rodrigosiqueira, line 154: Maybe something like, "Do you want to migrate to the new configuration file approach? (Recommended)"

src/kw_config_loader.sh
```diff
@@ -142,8 +145,25 @@ function parse_configuration()
...
+      mkdir -p "$PWD/$KW_DIR/"
+      mv -i "$PWD/$CONFIG_FILENAME" "$PWD/$KW_DIR/$CONFIG_FILENAME"
```
rodrigosiqueira, line 156: Why do we need `-i`? I don't think we need this since we already asked about the action in the `if` condition.

src/init.sh
```diff
@@ -18,25 +24,136 @@ function init_kw()
...
+  fi
+
+  parse_init_options "$@"
+  if [[ "$?" -gt 0 ]]; then
+    complain "${options_values['ERROR']}"
+    return 22 # EINVAL
+  fi
+
+  if [[ -f "$PWD/$KW_DIR/$name" ]]; then
+    if [[ -n "${options_values['FORCE']}" || $(ask_yN "$KW_DIR/$name already exists, do you wish to overwrite it?") =~ '1' ]]; then
```
rodrigosiqueira, line 39: Maybe something like, "It looks like that you already have a kw config file. Do you want to overwrite it?"

rodrigosiqueira (changes requested):
Hi @magalilemes ,
I added some comments.

PR #405, opened 2021-08-05
Refactor init

kwy95:
While testing kw I ran `kw init` on my kernel folder and lost my configurations, with this change we avoid losing any special configurations.

I also included some small changes to the init files.

### DIFF
--- a/src/init.sh
+++ b/src/init.sh
@@ -18,6 +18,15 @@ function init_kw()
     exit 0
   fi
 
+  if [[ -f "$PWD/$name" ]]; then
+    if [[ "$*" =~ --?f(orce)? || $(ask_yN "$name already exists, do you wish to overwrite it?") =~ '1' ]]; then
+      mv "$PWD/$name" "$PWD/$name.old"
+    else
+      say 'Initialization aborted!'
+      exit 0
+    fi
+  fi
+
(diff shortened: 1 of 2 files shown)


src/init.sh
```diff
@@ -18,12 +18,21 @@ function init_kw()
     exit 0
   fi
 
+  if [[ -f "$PWD/$name" ]]; then
+    if [[ "$*" =~ --?f(orce)? || $(ask_yN "$name already exists, do you wish to overwrite it?") =~ '1' ]]; then
+      mv "$PWD/$name" "$PWD/$name.old"
+    else
+      say 'Initialization aborted!'
+      exit 0
```
rodrigosiqueira, line 26: Add a test that validates this condition.

rodrigosiqueira (changes requested):
In general, lgtm; however, I want to see @magalilemes feedback since she is working on init.

magalilemes (approved):
Looks good to me. Although if this is merged before #404, I'll have to make a few adjustments, but no big deal.

src/init.sh
kwy95: Done!

rodrigosiqueira:
In this case, merged to the unstable branch. Thanks

PR #406, opened 2021-08-05
src: build: Refactor parsing of options

joaoseckler:
Use getopt to parse options. Implement a function for that.


rodrigosiqueira:
Nice patch! Thanks a lot!

P.s: You forgot to add the S-o-b, I added for you.

PR #410, opened 2021-08-06
Ignore comments in kworkflow.config

magalilemes:
We were ignoring only lines that started with '#', so any other line that had a '#' in the middle of it wasn't ignored.


alan-barzilay (approved):
lgtm

rodrigosiqueira:
Nice patch! Thanks a lot for that :)

PR #413, opened 2021-08-08
Compress and extract functions

magalilemes:
This PR adds compression and extraction functions to kwlib.

### DIFF
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -466,3 +466,44 @@ function generate_tarball()
     exit "$ret"
   fi
 }
+
+# This function extracts a .tar.gz file to a given path
+#
+# @file_to_extract The path to the file to extract from
+# @path Where to extract the file
+# @compression_type compression program used
+# @flag How to display (or not) the command used
+function extract_tarball()
+{
+  local file_to_extract="$1"
+  local path="$2"
+  local compression_type="$3"
+  local flag="$4"
+  local cmd
+
+  flag=${flag:-'SILENT'}
+
+  if [[ ! -f "$file_to_extract" ]]; then
+    complain 'We could not find' "$file_to_extract"
+    exit 22 #EINVAL
+  fi
+
+  if [[ ! -d "$path" ]]; then
+    complain "$path" 'does not exist'
+    exit 22 #EINVAL
+  fi
+
+  if [[ -n "$compression_type" ]]; then
+    if [[ "${compression_programs[*]}" =~ $compression_type ]]; then
+      compression_type="--$compression_type"
+    else
+      complain 'Invalid compression type:' "$compression_type"
+      return 22 # EINVAL
+    fi
+  fi
+
+  compression_type=${compression_type:-'--auto-compress'}
+
+  cmd="tar $compression_type -xzf $file_to_extract -C $path"
+  cmd_manager "$flag" "$cmd"
+}
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -1,5 +1,9 @@
 # NOTE: src/kw_config_loader.sh must be included before this file
 
+# Array with compression programs accepted by tar
+declare -ga compression_programs=('gzip' 'bzip2' 'lzip' 'lzma' 'lzop' 'zstd' 'xz'
+  'auto-compress')
+
 # A common task used inside kw is a string separation based on a delimiter, for
 # this reason, this function tries to handle this scenario by getting a
 # delimiter character followed by the position that the users want to retrieve.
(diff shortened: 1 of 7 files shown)


rodrigosiqueira (changes requested):
Hi @magalilemes ,

First of all, thanks a lot for your PR.

I added a couple of comments.

Thanks

src/kwlib.sh
```diff
@@ -387,3 +387,19 @@ function compress()
   cmd="tar -C $path_to_compress -czf $file_path ."
   cmd_manager "$flag" "$cmd"
 }
+
+# This function extracts a .tar.gz file to a given path
+#
+# @file_to_extract The path to the file to extract from
+# @path Where to extract the file
+# @flag How to display (or not) the command used
+function extract()
```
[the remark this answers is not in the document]
magalilemes: Maybe, but I feel like _extract_ is a more common word for this purpose.

tests/kwlib_test.sh
```diff
@@ -392,4 +392,16 @@ function test_exit_msg()
...
+function test_compress()
+{
+  local path_to_compress='tests/samples/statistics'
+  local file_path="$SHUNIT_TMPDIR/compressed.tar.gz"
+  local output
+
+  output=$(compress "$path_to_compress" "$file_path" 'SUCCESS')
+  assertEquals "($LINENO)" "tar -C $path_to_compress -czf $file_path ." "$output"
+
+  assertTrue 'Compressed file was not created' "[[ -f $SHUNIT_TMPDIR/compressed.tar.gz ]]"
```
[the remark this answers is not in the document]
magalilemes: This one test is doing that: it tries to compress the directory `tests/samples/statistics`, sees if the command used is correct and then checks if the compressed file was created

src/kwlib.sh
```diff
@@ -371,3 +371,38 @@ function exit_msg()
...
+#
+# @path_to_compress The directory to compress
+# @file_path Where the compressed directory will be stored
+# @compression_type compression program used
+# @flag How to display (or not) the command used
+function generate_tarball()
+{
+  local path_to_compress="$1"
+  local file_path="$2"
+  local compression_type="$3"
```
rodrigosiqueira, line 385: Can we define a default compression type here?

src/kwlib.sh
```diff
@@ -371,3 +371,38 @@ function exit_msg()
...
+  local flag="$4"
```
rodrigosiqueira, line 386: Let's define SILENT as the default flag.

src/kwlib.sh
```diff
@@ -371,3 +371,38 @@ function exit_msg()
...
+  local ret
+  local cmd
+
+  if [[ ! -d "$path_to_compress" ]]; then
+    complain "$path_to_compress" 'does not exist'
+    exit 22 #EINVAL
+  fi
+
+  if [[ -n "$compression_type" ]]; then
+    compression_type="--$compression_type"
```
rodrigosiqueira, line 396: Can you check if we can validate this compression flag? Maybe the `tar` command provides something; if not, we can try to compress a dummy file just for checking (I would prefer the first option).

src/kwlib.sh
```diff
@@ -371,3 +371,38 @@ function exit_msg()
...
+  fi
+
+  cmd="tar -C $path_to_compress $compression_type -cf $file_path ."
+  cmd_manager "$flag" "$cmd"
+
+  ret=$?
```
rodrigosiqueira, line 402: Double-quote

src/kwlib.sh
```diff
@@ -371,3 +371,38 @@ function exit_msg()
...
+
+  if [[ "$ret" != 0 ]]; then
+    complain "Error archiving modules."
+    exit $ret
```
rodrigosiqueira, line 406: Double-quote

tests/kwlib_test.sh
```diff
@@ -392,4 +392,37 @@ function test_exit_msg()
...
+
+  output=$(tar -taf "$file_path" | sort -d)
+  while read -r f; do
+    if [[ ${expected_files[$count]} != "${f}" ]]; then
+      fail "$ID - Expected file \"${f}\" to be \"${expected_files[$count]}\""
+    fi
+    ((count++))
+  done <<< "$output"
```
rodrigosiqueira, line 422: I think you can replace all of these codes by using `compare_command_sequence`.

src/kwlib.sh
```diff
@@ -406,3 +406,35 @@ function generate_tarball()
...
+#
+# @file_to_extract The path to the file to extract from
+# @path Where to extract the file
+# @compression_type compression program used
+# @flag How to display (or not) the command used
+function extract_tarball()
+{
+  local file_to_extract="$1"
+  local path="$2"
+  local compression_type="$3"
```
rodrigosiqueira, line 420: I'm not sure, but I think the option `-a` from tar command is an automatic mechanism to check the compression type and use it (see `-taf`). Let's use this option as a default.

rodrigosiqueira (changes requested):
In general, lgtm. I added a couple of comments.

rodrigosiqueira:
Hi @magalilemes , any progress here?

src/kwlib.sh
```diff
@@ -1,5 +1,9 @@
 # NOTE: src/kw_config_loader.sh must be included before this file
 
+# Array with compression programs accepted by tar
+declare -ga compression_programs=('gzip' 'bzip2' 'lzip' 'lzma' 'lzop' 'zstd' 'xz'
+  'auto-compress')
```
rodrigosiqueira, line 5: Can you confirm if we need to add a dependency in the `setup.sh` for some of these compression options? For example, I think lzma requires something extra.

src/kwlib.sh
rodrigosiqueira: I think you forgot to set a default compression.

src/kwlib.sh
rodrigosiqueira: Did you check the `-a` option? Can we use it as the default?

src/kwlib.sh
```diff
@@ -418,3 +422,88 @@ function kw_parse_get_errors()
...
+
+  cmd="tar -C $path_to_compress $compression_type -cf $file_path ."
+  cmd_manager "$flag" "$cmd"
+
+  ret="$?"
+
+  if [[ "$ret" != 0 ]]; then
+    complain "Error archiving modules."
+    exit "$ret"
+  fi
```
rodrigosiqueira, line 467: It looks like that we are not testing this path. Could you add a test for that? Also, use single quotes.

rodrigosiqueira (changes requested):
Thanks @magalilemes ,
I just added a few comments.

src/kwlib.sh
magalilemes: Yes, it's the same as `--auto-compress`, and I opted to set it as default a few lines below (line 457)

rodrigosiqueira: Nice, I missed that. Thanks.

src/kwlib.sh
magalilemes: From what I searched, lzip and zstd don't come by default on Ubuntu, so users need to install them before using them with tar

rodrigosiqueira:
Merged!
Thanks

PR #416, opened 2021-08-10
tests: utils: Be clearer when comparing sequences

joaoseckler:
The previous output of function compare_command_sequance wasn't clear
about what was the expected output and what was the real output. This
improves that.


rodrigosiqueira:
Nice improvement!
Merged to the unstable branch

PR #418, opened 2021-08-10
Add Rasp support

alan-barzilay:
based on #394, see there for a full discussion


rodrigosiqueira:
We will delay the Raspberry support; we first need to refactor part of the deploy.

PR #419, opened 2021-08-12
.github: workflows: Update CI

joaoseckler:
There is no need to install the virtualization tools (qemu and
libvirt-guestfs) in the CI machine. We were also checking out branch
unstable, which is not what is supposed to happen. We also split the
installation and documentation in two different steps.


rodrigosiqueira:
Thanks!

PR #420, opened 2021-08-12
Fix warning from configm_test

magalilemes:
This PR changes the order from the arguments in 'find' to get rid of the warning stating that -mindepth must come before -type.


rodrigosiqueira:
Nice PR! Thanks a lot.

PR #421, opened 2021-08-13
explore: fix explore help text

kwy95:
The help text for the explore command was a bit confusing, there was a
weird parameter. I realized its purpose was to show that strings with
spaces need to be quoted. I don't think this is needed in the help text
as it is clearly stated in the man page.


rodrigosiqueira:
Thanks @kwy95 , merged to the unstable branch.

PR #422, opened 2021-08-14
Refactor tests in order to increase compatibility

joaoseckler:
Kw's CI was recently failing in some unexpected ways. This PR tries to fix that. The first commit changes the functioning of compare_command_sequence. Instead of using the hard to read

`expected=("${!1}")`

we now simply use

`local -n expected="$1"`

using bash's -n option to declare variables, which assigns them by reference. The second commit removes the `&` from `command_exists` call, which was causing some race condition problems.

### DIFF
--- a/src/kwio.sh
+++ b/src/kwio.sh
@@ -27,7 +27,7 @@ function alert_completion()
 
   while read -rN 1 option; do
     if [ "$option" == "v" ]; then
-      if command_exists "${configurations[visual_alert_command]} &"; then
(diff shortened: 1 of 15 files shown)


joaoseckler:
I also added a fix to the test of mounted filesystems in vm_test.sh

rodrigosiqueira:
Hi @joaoseckler , I already merged this series under the deploy parser PR. If I missed something, feel free to reopen this PR.
Thanks.

PR #424, opened 2021-08-17
Improve ssh and minor fixes

rodrigosiqueira:
This PR comprises a couple of problems that I identified in the last few weeks.  Keep in mind that the commits are not directly related to each other; see this as a clean-up series.


rodrigosiqueira:
@magalilemes @joaoseckler @kwy95 any review here? If not, I want to apply this PR.

src/kw_ssh.sh
```diff
@@ -0,0 +1,127 @@
...
+    esac
+  done
+}
+
+function ssh_help()
+{
+  echo -e "kw ssh|s options:\n" \
+    "\tssh|s [--script|-s=\"SCRIPT PATH\"]\n" \
+    "\tssh|s [--command|-c=\"COMMAND\"]"
+}
```
kwy95, line 127: Why did you use the old help function? Shouldn't this use the new format with man support?

PR #428, opened 2021-08-24
Fix regression in generate_tarball

magalilemes:
The way generate_tarball was implemented in 30b0459c51f2 led to the deploy malfunctioning in remote machines. This commit fixes this by adding a new parameter to the function (namely _dir_name_).


rodrigosiqueira:
Thanks for the quick action,

PR #431, opened 2021-08-26
setup: Fix legacy_folders condition

rodrigosiqueira:
We recently updated kw to make it compliant with XDG, and we added a
function to migrate the old kw config to the new model automatically.
However, one of the conditions is always false, and for this reason, we
could not update the file location for features like configm and
pomodoro. This commit fixes this issue by handling the migration case by
case.

Fixes: e30c0a86 ("Make kw compliant with XDG Base Directory Specification")

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -165,22 +165,29 @@ function legacy_folders()
 {
   local prefix="$HOME/.local"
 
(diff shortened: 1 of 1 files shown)


magalilemes (approved):

rodrigosiqueira:
Thanks @magalilemes ,

PR #436, opened 2021-08-28
run shellcheck on push

kwy95:
Shellcheck was our only action that was only run upon creating a PR,
this adds the check to pushes as well.


rodrigosiqueira:
Merged

PR #437, opened 2021-08-28
Sendemail setup

kwy95:
Support to setting up `git send-email`


src/mail.sh
```diff
@@ -0,0 +1,170 @@
...
+  return 0
+}
+
+function mail_setup()
+{
+  local global=''
+
+  if [[ ! -d "$PWD/.git" ]] && [[ $(ask_yN 'Not in a git repository, apply changes globally?') =~ 1 ]]; then
+    global='--global'
+  fi
```
rodrigosiqueira, line 36: Maybe we should have a flag named `--global` in the `mail` option. If the user does not provide it, we assume local change.

src/mail.sh
```diff
@@ -0,0 +1,170 @@
...
+
+  # say 'arrived'
+
+  for option in 'SMTP-USER' 'SMTP-ENCRYPTION' 'SMTP-SERVER' 'SMTP-PORT'; do
+    if [[ -n "${options_values[$option]}" ]]; then
+      echo "git config $global ${options_values[$option]}"
+    fi
+  done
```
rodrigosiqueira, line 44: After reading this function, we could retrieve those options from the kw config as well. Now that we have a `.kw` folder, we probably want to do something like `.kw/mail_config` to save local configurations, and we can have something similar in the `$KW_ETC_DIR`. Ofc, we can also update that file via this mail feature.

src/mail.sh
```diff
@@ -46,10 +46,39 @@ function mail_setup()
   return 0
 }
 
+function mail_verify()
```
rodrigosiqueira, line 49: Add function documentation

src/mail.sh
```diff
@@ -46,10 +46,39 @@ function mail_setup()
...
+{
+  local count=0
+  local -a missing_conf
+  local -a min_conf=('user.name' 'user.email' 'sendemail.smtpencryption'
```
rodrigosiqueira, line 53: min stands for minimal, right? Maybe we can call this `essential_config_options_list` or similar.

src/mail.sh
```diff
@@ -46,10 +46,39 @@ function mail_setup()
...
+    'sendemail.smtpserver' 'sendemail.smtpuser' 'sendemail.smtpserverport')
```
rodrigosiqueira, line 54: Keep in mind that git send-email is not part of the git package. You need to update the `setup.sh` to add git send email as an external dependency.

src/mail.sh
```diff
@@ -46,10 +46,39 @@ function mail_setup()
...
+  local set_confs
+
+  # say "${min_conf[*]}"
```
rodrigosiqueira, line 57: You can drop this

src/mail.sh
```diff
@@ -46,10 +46,39 @@ function mail_setup()
...
+  set_confs=$(git config --get-regexp 'user|sendemail\.smtp')
+  # say "*****$set_confs*****"
+  for config in "${min_conf[@]}"; do
+    if ! echo "$set_confs" | grep -cF "$config" - &> /dev/null; then
```
rodrigosiqueira, line 61: Can we move this function outside this if? You can just check `"$?"`.
Also, add a comment explaining this line.

src/mail.sh
```diff
@@ -46,10 +46,39 @@ function mail_setup()
...
+      missing_conf[$count]="$config"
+      count=$((count + 1))
+    fi
+  done
+  # echo "COUNT: $count"
+  if [[ $count -gt 0 ]]; then
+    complain "Missing configurations needed for send-email: ${missing_conf[*]}"
+    return 1
+  fi
+  say "correct config"
```
rodrigosiqueira, line 72: Do you know if git-send-email has any feature that checks config? Maybe we can try to use the `dry` option just to check if everything is fine?

rodrigosiqueira (changes requested):
Hi @kwy95 ,

I think you are on the right path. I already add inline comments, now follows some generic comments:

1. Add tests per commit;
2. Add a commit that introduces the interactive option;

I also had one idea while reviewing your PR, we could have an option named `kw mail --setup --suggestions` where we show common email provider setup. Something like:

```
kw mail --setup --suggestions

Gmail:
 Source: LINK
 SMTP Server: ...
 ...
Outlook:
 Source: 

```

Maybe we can start with a list of the top 10 email services. Also, we should save this info in a text file. We don't want to put this info in the source; we could provide a mechanism for using one of the suggested options, something like this:

```
kw mail --setup --suggestion "Gmail"
```

Finally, I would like to see a backup option because I already have a configuration, and I don't want to mess with that (you probably want to check that before replacing anything... maybe dump the current info to the user and as if they really want to replace that). Maybe you can expand the backup option to provide an option to save git configurations.

src/mail.sh
```diff
@@ -0,0 +1,170 @@
...
+
+  # say "parse2"
+
+  while [[ "$#" -gt 0 ]]; do
+    case "$1" in
+      --help | -h)
+        mail_help "$1"
+        exit
+        ;;
+      --setup | -c)
```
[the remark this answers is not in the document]
kwy95: Futureproofing, I was going to use `-s` as shorthand for  `--send`. I chose `-c` for configure, but am open to better suggestions.

src/mail.sh
```diff
@@ -0,0 +1,170 @@
...
+        options_values['SETUP']=1
+        shift
+        ;;
+      --smtpuser)
+        options_values['SMTP-USER']="sendemail.smtpuser $2"
+        shift 2
+        ;;
+      --smtpencryption)
+        options_values['SMTP-ENCRYPTION']="sendemail.smtpencryption $2"
+        shift 2
```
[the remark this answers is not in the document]
kwy95: Added some validation, let me know what you think

setup.sh
```diff
@@ -4,8 +4,8 @@
 . src/kwlib.sh --source-only
 
 # List of dependences per distro
-arch_packages=(qemu bash git tar python-docutils pulseaudio libpulse dunst python-sphinx imagemagick graphviz python-virtualenv texlive-bin librsvg bzip2 lzip lzop zstd xz)
-debian_packages=(qemu git tar python3-docutils pulseaudio-utils dunst sphinx-doc imagemagick graphviz dvipng python3-venv latexmk librsvg2-bin texlive-xetex python3-sphinx python3-dask-sphinx-theme bzip2 lzip xz-utils lzop zstd)
+arch_packages=(qemu bash git tar python-docutils pulseaudio libpulse dunst python-sphinx imagemagick graphviz python-virtualenv texlive-bin librsvg bzip2 lzip lzop zstd xz perl-authen-sasl perl-io-socket-ssl)
+debian_packages=(qemu git tar python3-docutils pulseaudio-utils dunst sphinx-doc imagemagick graphviz dvipng python3-venv latexmk librsvg2-bin texlive-xetex python3-sphinx python3-dask-sphinx-theme bzip2 lzip xz-utils lzop zstd git-email)
```
rodrigosiqueira, line 8: Maybe it is time to make this single line became a multi-line list.

src/mail.sh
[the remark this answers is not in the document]
rodrigosiqueira: How about `-t`? From se**t**up

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+        mail_help "$1"
+        exit
+        ;;
+      --setup | -c)
+        options_values['SETUP']=1
+        shift
+        ;;
+      --smtp*)
+        local option="${1#--}"
+        options_values["${option^^}"]="sendemail.$option $2"
```
rodrigosiqueira, line 151: If I'm not wrong, `^^` capitalize `$option`, right? How about add `upcase_str` in `kw_string.sh` and use it here?

rodrigosiqueira: Also, don't we need to validate the `$option`? We might try to set something invalid. Maybe you can create an array with all the possible valid values for `sendemail.SOMETHING` and check if `$option` matches one of that options.

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+        return 0
+        ;;
+    esac
+  fi
+}
+
+function parse_mail_options()
+{
+  local short_options='h,c,'
+  local long_options='help,setup,global,'
```
rodrigosiqueira: Add a new line between these two parts.

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+      validate_encryption "$option"
+      # TODO change to eval
+      echo "git config ${options_values['GLOBAL']}${options_values[$option]}"
+    fi
+  done
+
+  return 0
+}
+
+# encryption must be 'ssl' or 'tls' this validates this option
```
rodrigosiqueira, line 57: Use the default comment style (see build or deploy) and improve this comment because it does not provide a good explanation.

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+function validate_encryption()
+{
+  if [[ "$1" =~ .*ENCRYPTION ]]; then
```
rodrigosiqueira, line 60: I would prefer that you create a `local` variable that receives `$1`.

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+    case "${options_values["$1"]: -3}" in
```
rodrigosiqueira, line 61: What are you trying to achieve here? From the code, it is not clear. Also, rely on a position sounds fragile.

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+      ssl | tls)
+        return 0
+        ;;
+      *)
+        say 'Invalid encryption, must be "ssl" or "tls"'
```
rodrigosiqueira, line 66: How about something like: `Invalid encryption, you must choose between "ssl" or "tls".`

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+        say 'Defaulting to plain smtp'
+        options_values["$1"]="sendemail.${1,,} smtp"
+        return 0
```
rodrigosiqueira, line 69: Why return `0`? This does not sound like a standard case.

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+  if [[ "${options_values['SETUP']}" == 1 ]]; then
+    mail_setup
+    exit
+  fi
+
+  return 0
+}
+
+# This function deals with configuring the necessary options to use
+# the git send-email tool to send patches by email
```
rodrigosiqueira, line 37: Improve this description, explain some of the restrictions around this function and other details developers may not figure out from looking at the code for the first time.

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+function mail_setup()
+{
+  if ! git rev-parse --is-inside-work-tree &> /dev/null && [[ -z "${options_values['GLOBAL']}" ]]; then
```
rodrigosiqueira, line 40: Can we move this `git rev-parse ..` outside this `if`? (you can check `$?`).

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+{
+  if [[ "$1" =~ .*ENCRYPTION ]]; then
+    case "${options_values["$1"]: -3}" in
+      ssl | tls)
+        return 0
+        ;;
+      *)
+        say 'Invalid encryption, must be "ssl" or "tls"'
+        say 'Defaulting to plain smtp'
+        options_values["$1"]="sendemail.${1,,} smtp"
```
rodrigosiqueira, line 68: Can we have something more readable than `${1,,}`? If not, add a single-line comment explaining it.

src/mail.sh
```diff
@@ -0,0 +1,125 @@
...
+    complain 'Not in a git repository, aborting setup!'
+    say 'To apply settings globally rerun with "--global" flag.'
+    exit 1 # EPERM
+  fi
+
+  for option in "${!options_values[@]}"; do
+    if [[ "$option" =~ ^SMTP.* ]] && [[ -n "${options_values[$option]}" ]]; then
+      validate_encryption "$option"
+      # TODO change to eval
+      echo "git config ${options_values['GLOBAL']}${options_values[$option]}"
```
rodrigosiqueira, line 50: You probably want to use cmd_manager instead of `eval`.

documentation/man/features/mail.rst
```diff
@@ -0,0 +1,27 @@
...
+OPTIONS
+=======
+-c, \--setup:
+  Initialize and configure **mail** functionality. Each argument specifies a
+  configuration to be set, current accepted options are::
+
+    --smtpuser       <email>
+    --smtpserver     <domain>
+    --smtpserverport <port>
+    --smtpencryption <encryption>
```
rodrigosiqueira, line 29: Add an example section.

src/mail.sh
```diff
@@ -8,6 +8,10 @@ include "$KW_LIB_DIR/kw_config_loader.sh"
 # Hash containing user options
 declare -gA options_values
 
+declare -a essential_config_options=('user.name' 'user.email' 'fail' # for testing only
+  'sendemail.smtpserver' 'sendemail.smtpuser' 'sendemail.smtpserverport')
+declare -a recommended_config_options=('sendemail.smtpencryption' 'sendemail.smtppass')
```
rodrigosiqueira, line 13: I think it is worth adding a more detailed comment about these two arrays on top of the essential_config_options. Additionally, update the name to something that portraits the idea that those are used for tests only.

src/mail.sh
```diff
@@ -63,10 +67,42 @@ function validate_encryption()
   fi
 }
 
+# This function checks that the necessary options to use
```
rodrigosiqueira, line 75: What do you mean by `necessary`? What are the conditions? Please, improve this comment.

src/mail.sh
```diff
@@ -63,10 +67,42 @@ function validate_encryption()
...
+# the git send-email have been set
+function mail_verify()
+{
+  local count=0
+  local -a missing_conf
+  local set_confs
+
+  set_confs=$(git config --get-regexp 'user|sendemail\.smtp')
+  for config in "${essential_config_options[@]}"; do
+    if ! printf '%s\n' "$set_confs" | grep -cF "$config" - &> /dev/null; then
```
rodrigosiqueira, line 85: Can we move this `printf` outside this `if`?

src/mail.sh
```diff
@@ -63,10 +67,42 @@ function validate_encryption()
...
+      missing_conf[$count]="$config"
```
rodrigosiqueira, line 86: Don't you need to add double quotes around `$count`?

src/mail.sh
```diff
@@ -63,10 +67,42 @@ function validate_encryption()
...
+      count=$((count + 1))
+    fi
+  done
+
+  for config in "${recommended_config_options[@]}"; do
+    if ! printf '%s\n' "$set_confs" | grep -cF "$config" - &> /dev/null; then
+      missing_conf[$count]="$config"
+      count=$((count + 1))
+    fi
+  done
```
rodrigosiqueira, line 96: Is this config essential or not? If not, you probably want to separate `missing_conf` for this specific case and just warn the user. Does it make sense, or did I miss something?

src/mail.sh
```diff
@@ -63,10 +67,42 @@ function validate_encryption()
...
+
+  if [[ $count -gt 0 ]]; then
+    complain "Missing configurations needed for send-email: ${missing_conf[*]}"
+    return 1
```
rodrigosiqueira, line 100: I think we have two cases:
1. Essential: We must fail if something is missing.
2. Recommended: It is not ideal, but we can proceed since it will not compromise the` mail` feature. Let's just warn the user.

src/mail.sh
```diff
@@ -63,10 +67,42 @@ function validate_encryption()
...
+  fi
+  say 'correct config'
```
rodrigosiqueira, line 102: Use `success` instead of `say`.

src/mail.sh
```diff
@@ -98,6 +134,10 @@ function parse_mail_options()
         options_values["${option^^}"]="sendemail.$option $2"
         shift 2
         ;;
+      --verify | -v)
+        mail_verify
+        shift
```
rodrigosiqueira, line 156: If you call some function inside the parse option, you must exit after that. If that is not what you want, move this function to another place.

src/mail.sh
```diff
@@ -121,5 +161,6 @@ function mail_help()
     exit
   fi
   printf '%s\n' 'kw mail:' \
-    '  mail (-c | --setup) [--global] - Configure mailing functionality'
+    '  mail (-c | --setup) [--global] - Configure mailing functionality' \
+    '  mail (-v | --verify) - Check if need configuration is set'
```
rodrigosiqueira, line 182: `need` or `needed`? Maybe `required`?

documentation/man/features/mail.rst
```diff
@@ -29,3 +30,7 @@ OPTIONS
 
 -v, \--verify:
   Verify that all the settings needed are set and valid.
+
+-i, \--interactive:
+  This checks your current configuration and prompts about missing options and
+  changes.
```
rodrigosiqueira, line 36: I think we should pass through all configurations, even if they are already set. Maybe we can ask the user if they want to replace or not in each option.

src/mail.sh
```diff
@@ -99,10 +104,17 @@ function mail_verify()
   return 0
 }
 
+# This function makes the setup proccess interactive
+function interactive_setup()
+{
+  :
```
rodrigosiqueira, line 110: Where is the function?

rodrigosiqueira (changes requested):
Hi,

For your next version, I would like you to try to address most of my comments from **this review**, and I also want you to take action based on my **previous review**.

For the interactive part, I suspect you will need to expand kwio.sh to add new functions that will help you in this process.

Ps.: It is time to add tests, do it per commit.

src/mail.sh
kwy95: Option validation happens in the `kw_parse` function, since it will return an error if the options are not listed in either long_options or short_options.

src/mail.sh
kwy95: I agree, I wasn't super happy with this, but essentially it tries to get the type of encryption passed. Since the possible values are 3 characters long I used that, ideally I wanted to split the value at the space character but I wasn't able to. I realize that if someone passes something like `stls` this code would accept it. I explore further options, this is definitely unusable.

src/mail.sh
kwy95: Yeah, I just missed it. Will change it to 22

src/mail.sh
kwy95: Definitely, this is always just `smtpencryption`

src/mail.sh
kwy95: I used to have them inside the verify function, but decided to move them to the top so they could be used for the interactive and setup functions as well. The comment pertains only to the `'fail'` entry, which is there temporarily to force a failure. Will add descriptive comments.

src/mail.sh
kwy95: The `printf` is just sending the value of `set_confs` to the `grep` function. This is testing if they are set, I think I will move them outside and capture the output to check the current value that is set.

src/mail.sh
kwy95: Yes, you are correct.

src/mail.sh
rodrigosiqueira: Yeah, but from what I can see here, there is no validation for `$option`. For example, the user passes `--smpt-lalaxpto`, it looks like that we will accept it as a valid option. I suggest you check `--smtp$option` in the easy stages; you can do it by creating an array with valid options for `smtp`.

src/mail.sh
kwy95: Not as far as I could tell. I tried a few tests hiding some of my configurations, and `git send-email` just tried to assume most of them and get them from the commit messages when possible. While using the `--dry-run` option the only settings it complained about were `user.name` and `user.email`. Without it it obviously failed, but it attempts to send it.

setup.sh
kwy95: Ok, should I use this PR or create a dedicated one?

src/mail.sh
```diff
@@ -0,0 +1,206 @@
...
+    done
+
+    if [[ "$warn" == 1 ]]; then
+      printf '%s\n' '' "You are currently editing at a $curr_scope scope." \
+        'If you continue the value at this scope will be overwritten.' \
+        "The new value will be $value"
+
+      if [[ $(ask_yN 'Do you wish to proceed?') == 0 ]]; then
+        complain "Skipping $setting..."
+        return 1
```
rodrigosiqueira, line 136: Maybe return 125 - ECANCELED (Operation canceled)

tests/mail_test.sh
```diff
@@ -0,0 +1,142 @@
...
+  expected_result="sendemail.smtpserver 'test.email.com'"
+  assert_equals_helper 'Set smtp server' "$LINENO" "${options_values['SMTPSERVER']}" "$expected_result"
+
+  parse_mail_options '--smtpserverport' '123'
+  expected_result="sendemail.smtpserverport '123'"
+  assert_equals_helper 'Set smtp serverport' "$LINENO" "${options_values['SMTPSERVERPORT']}" "$expected_result"
+
+  parse_mail_options '--smtppass' 'verySafePass'
+  expected_result="sendemail.smtppass 'verySafePass'"
+  assert_equals_helper 'Set smtp pass' "$LINENO" "${options_values['SMTPPASS']}" "$expected_result"
```
rodrigosiqueira, line 97: What happens if we try to use `--smtpxpto`? I suppose we will accept it, right?

tests/mail_test.sh
```diff
@@ -0,0 +1,142 @@
...
+    "git config sendemail.smtpuser 'test@email.com'"
+    "git config user.email 'test@email.com'"
+    "git config user.name 'Xpto Lala'"
+  )
+
+  cd "$FAKE_GIT" || {
+    ret="$?"
+    fail "($LINENO): Failed to move to fake git repo"
+    exit "$ret"
+  }
```
rodrigosiqueira, line 119: In the `setup` function, you already `cd` this directory.

tests/mail_test.sh
```diff
@@ -0,0 +1,142 @@
...
+
+function setUp()
+{
+  declare -gA options_values
+
+  cd "$SHUNIT_TMPDIR" || {
+    ret="$?"
+    fail "($LINENO): Failed to move to shunit temp dir"
+    exit "$ret"
+  }
```
rodrigosiqueira, line 27: It is a bit weird to me that you used `cd` in the setup function. Anyway, for symmetric sake, I think we should use cd to the original path in the `tearDown` function or `cd` inside specific tests.

src/mail.sh
```diff
@@ -130,11 +139,66 @@ function add_config()
...
+  done
+
+  if [[ $count_rec -gt 0 ]]; then
+    warning 'We recommend you configure these options before using send-email:'
+    printf '  %s\n' "${missing_rec_conf[@]}"
+  fi
+
+  if [[ $count -gt 0 ]]; then
+    complain 'Missing configurations required for send-email:'
+    printf '  %s\n' "${missing_conf[@]}"
```
rodrigosiqueira, line 193: I think that return 22 makes more sense here.

src/mail.sh
rodrigosiqueira: Did you address this comment?

rodrigosiqueira (changes requested):
Hi @kwy95 

I think you are close to complete this first part. Aside from my inline comments, I suggest that you keep this PR with the first three commits. After we merge that, you can work on the second PR of this basic setup which is composed of the following parts:
Implement the interactive option (you will probably need to create some new functions for helping you with that)
Implement the template option.

tests/utils.sh
```diff
@@ -213,6 +213,40 @@ function mk_fake_boot()
   cp -r "$SAMPLES_DIR/boot" "$FAKE_BOOT_DIR"
 }
 
+function mk_fake_git()
+{
+  local -r path="$1"
```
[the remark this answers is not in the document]
kwy95: Ok, makes sense. I'll remove the `cd`s from inside the function as well.

src/mail.sh
```diff
@@ -0,0 +1,206 @@
...
+
+include "$KW_LIB_DIR/kw_config_loader.sh"
+include "$KW_LIB_DIR/kw_string.sh"
+
+# Hash containing user options
+declare -gA options_values
+
+function kw_mail()
+{
+  local flag="$1"
```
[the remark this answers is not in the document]
kwy95: I'm thinking of using it on tests. I'll reevaluate as I implement the tests.

src/mail.sh
```diff
@@ -0,0 +1,206 @@
...
+  local cmd=''
+  local warn=0
+  local i=0
+
+  flag=${flag:-'SILENT'}
+  curr_scope=${curr_scope:-'local'}
+
+  # Gets set configurations and their scope
+  # The parameter '-d' is used to read all lines at once
+  IFS=$'\t\n' read -r -a set_values -d '' <<< "$(get_all_git_config "$flag" "$setting")"
```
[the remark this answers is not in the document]
kwy95: Not entirely sure I understand why, but using it with read doesn't mess with the systems settings, it is only valid for the line with the read command.

src/mail.sh
```diff
@@ -0,0 +1,206 @@
...
+        ;;
+      --email | --name)
+        local index
+        local option="${1#--}"
+        index="$(str_uppercase "$option")"
+        options_values["$index"]="user.$option '$2'"
+        shift 2
+        ;;
+      --smtpencryption)
+        validate_encryption "$2"
```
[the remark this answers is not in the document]
kwy95: There is no reason that the rest of the configuration can't continue even with a bad encryption, so there is no need to halt execution. But what do you suggest we do in case of failure?

tests/mail_test.sh
kwy95: Parser fails, added a test just in case.

documentation/man/features/mail.rst
rodrigosiqueira: We have a few options for encryption, add them here.

rodrigosiqueira (changes requested):
Hi @kwy95 ,

You can see some of my inline comments, follows some other issues that I found:

1. kw mail -h and --help does not provide the right output. I getting:
```
These are the essential configurations for git send-email:
  user.name
  user.email
  sendemail.smtpuser
  sendemail.smtpserver
  sendemail.smtpserverport
These are the recommended configurations for git send-email:
  sendemail.smtpencryption
  sendemail.smtppass
``
2. When I tried:

a. Clone kw in the /tmp folder
b. Try:
```
kw mail --setup --email juca@lucas.br --name "Juca" --smtpserver "someging@smtp.la" --smtpencryption "tls" --smtpserverport 3333
```
The above command provided the same output like the one in (1) and nothing changed.

Try to test it in your system.

Thanks

File: src/mail.sh [hunk 44]
```diff
@@ -0,0 +1,216 @@
...
+# @curr_scope: The scope being edited
+# @set_values: The values of the option that are already set
+#
+# Return:
+# Returns 0 if successful; non-zero otherwise
+function add_config()
+{
+  local flag="$1"
+  local option="$2"
+  local setting="${options_values["$option"]% \'*}"
```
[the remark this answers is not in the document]
kwy95: Sure, the line stored in options_values contains de both, the setting and the value, e.g. `sendemail.smtpserver 'test.server.com'` this line removes the ending of the line that matches the ` '*` pattern. I created a function called str_remove_suffix that mirror str_remove_prefix, but for some reason they wouldn't match the right parts of the string. I suspect the functions were taking the asterisk literally.

src/mail.sh
```diff
@@ -0,0 +1,216 @@
...
+  local value="${options_values["$option"]#* }"
```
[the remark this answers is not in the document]
kwy95: This is similar to above, but it removes from the start of the string. In this case the example would remove `sendemail.smtpserver `

src/kwlib.sh
```diff
@@ -525,3 +525,59 @@ function get_file_name_from_path()
...
+# @flag: How to display (or not) the command used
+# @regexp: Given regular expression to find associated values
+# @scope: Scope to use
+#
+# Returns:
+# All config values that match the given regular expression
+function get_git_config_regex()
+{
+  local flag="$1"
+  local scope="$2"
```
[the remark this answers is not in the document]
kwy95: I think currently the scope is always passed to this function but I've been using local as the default to avoid messing the system on accident. I'll add these changes.

kwy95:
@rodrigosiqueira I accidentally left some commands on the kw_mail function that I was using for testing, that caused the weird behavior you experienced. I addressed this and all the other comments you left.

kwy95:
@rodrigosiqueira I fixed the errors we encountered on our last meeting and found a work around the usage of `--show-scope`. I feel like some things could still be improved, but this should be working as intended for now.

src/mail.sh
```diff
@@ -0,0 +1,289 @@
...
+  local cmd_scope=${options_values['CMD_SCOPE']}
+  local set_option="${curr_scope}_$option"
+  local cmd
+
+  flag=${flag:-'SILENT'}
+
+  if [[ "${options_values['FORCE']}" == 0 ]]; then
+    if [[ -n "${set_confs[$set_option]}" ]]; then
+      warning "The configuration $option is already set with the following value(s):"
+      for scp in {'global','local'}; do
```
rodrigosiqueira, line 145: scp? What is the meaning of this? Consider replacing it with a better name and declare it as a local variable.

src/mail.sh
```diff
@@ -0,0 +1,289 @@
...
+function get_configs()
+{
+  local cmd_scope=${options_values['CMD_SCOPE']}
+  local -a set_values
+  local option
+  local value
+  local scope
+  local i=0
+
+  IFS=$'\t\n' read -r -a set_values -d '' <<< "$(get_git_config_regex 'user|sendemail\.smtp' "$cmd_scope")"
```
rodrigosiqueira, line 190: iirc, the safe way to deal with `IFS` is declaring it as a local.

rodrigosiqueira (changes requested):

src/mail.sh
```diff
@@ -0,0 +1,396 @@
...
+  done
+
+  for config in "${essential_config_options[@]}"; do
+    if [[ ! "${!set_confs[*]} " =~ (local|global)_"$config"[[:space:]] ]]; then
+      missing_conf[$count]="$config"
+      count=$((count + 1))
+    fi
+  done
+
+  if [[ $count -gt 0 ]]; then
```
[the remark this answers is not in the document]
kwy95: I think it's because the `-gt` operator makes the left side be interpreted as an arithmetic expression, and therefore expects any variable to just be a number, which wouldn't cause any expansion problems. But I will add the quotes anyway.

src/mail.sh
kwy95: When used with the read command like this the `IFS` variable is only changed for the duration of the command. I have tested and can confirm that the value of the `IFS` variable was the same before and after the execution of this line.

rodrigosiqueira:
Hi @kwy95 ,
Thanks a lot for your patch. I already applied it to the unstable branch.

PR #438, opened 2021-08-31
time_and_date: remove cal dependency

kwy95:
A previous commit introduced a dependency on the `ncal` package, in an
effort to keep kw as minimal as possible the functions affected have
been refactored to remove that dependency.

### DIFF
--- a/src/kw_time_and_date.sh
+++ b/src/kw_time_and_date.sh
@@ -86,6 +86,8 @@ function days_in_the_month()
 {
   local month_number="$1"
   local year="$2"
+  local days=31
+  local short=(4 6 9 11) # list of months with 30 days
 
   if [[ -n "$month_number" && "$month_number" -lt 1 || "$month_number" -gt 12 ]]; then
     return 22 # EINVAL
@@ -94,5 +96,21 @@ function days_in_the_month()
   month_number=${month_number:-$(date +%m)}
   year=${year:-$(date +%Y)}
 
-  cal "$month_number" "$year" | awk 'NF {DAYS = $NF}; END {print DAYS}'
+  # check if it's a leap year
+  if [[ "$month_number" == 2 ]]; then
+    if ((year % 4 != 0)); then
+      days=28
+    elif ((year % 100 != 0)); then
+      days=29
+    elif ((year % 400 != 0)); then
+      days=28
+    else
+      days=29
+    fi
+  # check if it's a short month
+  elif [[ "${short[*]}" =~ (^|[[:space:]])$month_number($|[[:space:]]) ]]; then
+    days=30
+  fi
+
+  printf '%s\n' "$days"
 }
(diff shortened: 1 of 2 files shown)


tests/kw_time_and_date_test.sh
```diff
@@ -96,13 +99,6 @@ function test_days_in_the_month()
   total_days=$(days_in_the_month 8)
   assert_equals_helper 'Use this year' "$LINENO" "$total_days" 31
 
-  # Empty year should be converted to the present year
-  total_days=$(days_in_the_month)
-  this_year=$(date +%Y)
-  this_month=$(date +%m)
-  this_month_total_days=$(cal "$this_month" "$this_year" | awk 'NF {DAYS = $NF}; END {print DAYS}')
-  assert_equals_helper 'Use the current month' "$LINENO" "$total_days" "$this_month_total_days"
```
kwy95, line 104: I didn't really know how to test this without essentially using the function again, which is what this did. The only problem is now we don't test the case where a month is not provided.

src/kw_time_and_date.sh
```diff
@@ -94,5 +96,21 @@ function days_in_the_month()
   month_number=${month_number:-$(date +%m)}
   year=${year:-$(date +%Y)}
 
-  cal "$month_number" "$year" | awk 'NF {DAYS = $NF}; END {print DAYS}'
+  if [[ $month_number == 2 ]]; then
```
rodrigosiqueira, line 99: Add quotes around $month_number

src/kw_time_and_date.sh
```diff
@@ -86,6 +86,8 @@ function days_in_the_month()
 {
   local month_number="$1"
   local year="$2"
+  local days=31
+  local short=(4 6 9 11)
```
rodrigosiqueira, line 90: It is not clear to me why you created this array with these numbers.

src/kw_time_and_date.sh
```diff
@@ -94,5 +96,21 @@ function days_in_the_month()
...
+  if [[ $month_number == 2 ]]; then
+    if ((year % 4 != 0)); then
+      days=28
+    elif ((year % 100 != 0)); then
+      days=29
+    elif ((year % 400 != 0)); then
+      days=28
+    else
+      days=29
+    fi
```
rodrigosiqueira, line 108: Can we use `||` in the `if` conditions for the shared cases?

rodrigosiqueira: Also, did you check if the current set of tests covers these new cases?

rodrigosiqueira (changes requested):

src/kw_time_and_date.sh
kwy95: These are the months that have 30 days, the rest have 31. Except for February.

src/kw_time_and_date.sh
kwy95: Not so simple to concatenate the tests as the previous tests must fail in order, but I'll see what can be done.

rodrigosiqueira:
Tbh, I'm not super happy with the new implementation. I think it could be simple... however, since this fixed a real bug that causes many failures in our CI, I'll go ahead and apply it to the unstable branch.
Thanks for your quick action on this issue.

PR #439, opened 2021-09-02
echo -> printf

### LINKED ISSUE #382: Use `printf` instead of `echo` for compatibility
After some research I think we should refactor some of our code to replace our usage of `echo` with `printf`.
The main concern is that `echo`'s behavior is very implementation and system dependent, specifically with regards to its options and text replacement options, where as `printf` seems to be more consistent and reliable.

Further reading:
https://www.in-ulm.de/~mascheck/various/echo+printf/
https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo
https://unix.stackexchange.com/questions/58310/difference-between-printf-and-echo-in-bash

kwy95:
This PR replaces all instances of echo with printf to close #382 
A second commit fixes most misused quotes.


rodrigosiqueira:
Tbh, I did not review this PR in detail because it is huge. However, I tested it, and everything looks fine. I already applied it to the unstable branch; if we see issues, we need to fix them.

Thanks

PR #448, opened 2021-09-08
build_test: add #!/bin/bash

### LINKED ISSUE #447: build_test missing #!
I noticed that `build_test.sh` is the only test file that does not contain the hash bang at the beginning, as far as I can tell this doesn't cause any problems but I think we should be consistent whenever possible. Was this a deliberate decision? Is there a reason for it to not be there?
If not I made a simple PR to add the `#!/bin/bash` line to it.

kwy95:
Adds `#!/bin/bash` to the top of the build_test.sh file as it was the
only one missing it.

Closes #447 

### DIFF
--- a/tests/build_test.sh
+++ b/tests/build_test.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 include './src/build.sh'
 include './tests/utils.sh'
 


rodrigosiqueira:
Merged! Thanks

Issue #275, opened 2021-05-22
Review documentation


rodrigosiqueira:
We need to review our documentation with these two things in mind:
* Update old information;
* Fix mistakes;
In particular, I have this TODO list:

- [ ] Update "How to contribute" page
- [ ] Update front page
- [ ] Codying style

alan-barzilay:
@rodrigosiqueira when you say fix mistakes, do you mean grammatical mistakes or content mistakes? like mentioning deprecated options

rodrigosiqueira:
Both

rodrigosiqueira:
@kwy95 could you review and update the "how to contribute section"? (I sent an email about it last week)

@joaoseckler could you also take a look at the codestyle page? Review it and update it based on our latest discussions.

rodrigosiqueira:
https://github.com/kworkflow/kworkflow/pull/565

PR #450, opened 2021-09-10
doc: Update page layout

rodrigosiqueira:
Currently, we use the standard Sphinx template, which is excellent;
however, we finally have our official website, and it will be nice to
have something more user-friendly. For this reason, this commit
introduces a new Sphinx theme named book; this is part of issue #275.


kwy95 (changes requested):
While testing this patch I noticed that the new theme forces sphinx version <4. This caused problems when generated a warning while generating the docs when I had a version of the docs already done with sphinx 4.0.2, which is what I was using before. Deleting the `build/` folder and running `./setup.sh --docs` again got rid of the issue.
With this in mind I think we should consider deleting a previous documentation build. But I do realize this is probably not a big issue.
I also added a couple of in line comments with more specific changes.

rodrigosiqueira:
Nice catch @kwy95 !
I already update this PR.

setup.sh
```diff
@@ -76,6 +79,13 @@ function check_dependencies()
     fi
     eval "sudo $cmd"
   fi
+
+  for package in "${pip_packages[@]}"; do
+    pip list | grep -F "$package" > /dev/null
+    [[ "$?" != 0 ]] && pip_package_list="$package $pip_package_list"
+  done
+  cmd="pip install $pip_package_list"
+  eval "$cmd"
```
kwy95, line 88: Shouldn't we ask the user before running this?

kwy95 (commented):
Just the one question about user approval.

setup.sh
rodrigosiqueira: Why?

kwy95: I mean, we ask the user before installing the other dependencies, why should this one be different?

rodrigosiqueira: Before installing a system dependency, we ask if the user wants to proceed; we then install everything, including this pip package. Also, pip installation in the userspace sounds like a routine task that nobody cares about. I can make it optional if you want, but I don't see the value in doing that.

kwy95: Ok, how about we move the for loop to line 72-73 and add `pip_package_list`  to the `ask_yN` prompt? Also you need to check if `pip_package_list` is not zero, as running `pip install` without arguments produces an error.
I left a comment suggestion with what I mean.

kwy95 (approved):

rodrigosiqueira:
Merged, thanks

PR #457, opened 2021-09-15
setup: break dependencies into lines

kwy95:
The list of package dependencies in the setup file was too big, they
have now been separated into multiple files.


setup.sh
```diff
@@ -4,8 +4,13 @@
...
 # List of dependences per distro
-arch_packages=(qemu bash git tar python-docutils pulseaudio libpulse dunst python-sphinx imagemagick graphviz python-virtualenv texlive-bin librsvg bzip2 lzip lzop zstd xz python-pip)
-debian_packages=(qemu git tar python3-docutils pulseaudio-utils dunst sphinx-doc imagemagick graphviz dvipng python3-venv latexmk librsvg2-bin texlive-xetex python3-sphinx python3-dask-sphinx-theme python3-pip bzip2 lzip xz-utils lzop zstd)
+arch_packages=(qemu bash git tar python-docutils pulseaudio libpulse dunst
+  python-sphinx imagemagick graphviz python-virtualenv texlive-bin librsvg
+  bzip2 lzip lzop zstd xz python-pip)
+debian_packages=(qemu git tar python3-docutils pulseaudio-utils dunst
+  sphinx-doc imagemagick graphviz dvipng python3-venv latexmk librsvg2-bin
+  texlive-xetex python3-sphinx python3-dask-sphinx-theme python3-pip bzip2
+  lzip xz-utils lzop zstd)
```
rodrigosiqueira, line 13: Can we have something like?
```
arch_packages=(
  qemu
  git
  tar
  ...
```

rodrigosiqueira (changes requested):

setup.sh
rodrigosiqueira: Alternatively, how about refactor this so that we add these package's names in a separate file and load them here?
Maybe we can have a file named `archlinux.dependencies` that we load based on the target distro. Another idea is to put this info in the documentation folder as an rst file and load it in the documentation and this script (reuse the same info twice).

rodrigosiqueira:
Thanks a lot for this PR.
Change merged to the unstable branch.

PR #462, opened 2021-09-23
Create vm option and move the mount, umount and up under it. 

eduardopinheiro:


src/vm.sh
```diff
@@ -109,14 +109,58 @@ function vm_up()
...
+
+  options="$(getopt \
+    --name 'kw vm' \
+    --options "$short_options" \
+    --longoptions "$long_options" \
+    -- "$@")"
+
+  if [[ "$?" != 0 ]]; then
+    return 22 # EINVAL
+  fi
```
rodrigosiqueira, line 125: Use `kw_parse` and `kw_parse_get_errors` here; take a look at `src/deploy.sh`.

src/vm.sh
```diff
@@ -109,14 +109,58 @@ function vm_up()
...
+      --mount | mo)
+        vm_mount "$@"
+        exit
+        ;;
+      --umount | um)
+        vm_umount "$@"
+        exit
+        ;;
+      --up | u)
+        vm_up "$@"
```
rodrigosiqueira, line 144: For this part, I recommend that you create an associative array named `options_values` on top of this file (e.g., `declare -gA options_values`) with these entries:
1. `MOUNT`: Set 0 by default at the beginning of the parse function. Set 1 if we match this case.
2. `UMOUNT`: Set 0 by default and 1 if the user requests it.
3. `UP`: Same idea.

After that, you can create an "entry point" for this feature. For example, let's call it `vm_main`, and this function will call the parse function first, and next it will check the option. Something like this:

```bash
function vm_main()
{
  if [[ ... --help or -h .. ]]; then
   vm_help
   exit
  fi

  parse_vm_options "$@"
  if [[ "$?" != 0 ]]; then
   error...
  fi

  if [[ ${options_values['MOUNT']} == 1 ]]; then
    vm_mount "PARAMETERS"
  fi
  ...
}
```

tests/vm_test.sh
```diff
@@ -152,12 +152,12 @@ function test_vm_umount()
   assertEquals "($ID) - Expected 125" "$expected_ret" "$ret"
 
   ID=2
-  output=$(vm_umount "TEST_MODE" "" "$mount_point")
+  output=$(vm_umount "--umount" "TEST_MODE" "" "$mount_point")
```
rodrigosiqueira, line 155: Since you are working on this part, could you also replace these `ID` variables with the `$LINENO` in the error message? This `ID` variable is a legacy thing that we want to remove.

rodrigosiqueira: Also, use single quotes when you don't have a bash variable to be expanded.

tests/vm_test.sh
```diff
@@ -152,12 +152,12 @@ function test_vm_umount()
...
   ret="$?"
   assertTrue "($ID): We got: $ret" "$ret"
 
   ID=3
-  output=$(vm_umount "TEST_MODE" "" "$mount_point")
+  output=$(vm_umount "--umount" "TEST_MODE" "" "$mount_point")
```
rodrigosiqueira, line 160: Drop `ID` in favor of `$LINENO` and use single quotes.

tests/vm_test.sh
```diff
@@ -166,4 +166,56 @@ function test_vm_umount()
   }
 }
 
+function test_vm_parse_options()
```
rodrigosiqueira, line 169: After you refactor the parser option, review this test.

rodrigosiqueira (changes requested):
Hi @eduardopinheiro ,

First of all, thanks a lot for your PR. In general, it looks good, and you are on the right path. I added a couple of comments and also noticed that you have some shfmt problems. For the new version, try to address the problems that I pointed out and fix the format issue.

tests/vm_test.sh
```diff
@@ -166,4 +166,56 @@ function test_vm_umount()
...
+  option_output="$(vm_help)"
+  unset options_values
+  declare -gA options_values
+  output="$(parse_vm_options -h)"
+  assertEquals "($LINENO)" "$option_output" "$output"
+
+  function vm_mount()
+  {
+    echo 'mount'
+  }
```
kwy95, line 186: If possible avoid creating functions inside functions, if you want to test what command is being executed you can use `cmd_manager` to  print the command instead of executing it.

src/vm.sh
```diff
@@ -55,14 +55,14 @@ function vm_mount()
 
 function vm_umount()
 {
-  local flag="$1"
-  local qemu_img_path="$2"
-  local mount_point_path="$3"
+  local flag="$2"
+  local qemu_img_path="$3"
+  local mount_point_path="$4"
```
kwy95, line 60: If you are not using the first parameter it's a sign you are passing extra stuff to the functions, avoid this. You can use `shift` to remove the first parameter before passing the rest here.

src/vm.sh
```diff
@@ -109,14 +109,58 @@ function vm_up()
   cmd_manager "$flag" "$cmd"
 }
 
+function parse_vm_options()
+{
+  local long_options='help,mount,umount,up'
+  local short_options='h,mo,um,u'
```
kwy95, line 115: You will probably have to refactor these options to deal with the particular options of mount umount and up

kw
```diff
@@ -69,33 +69,17 @@ function kw()
...
-      ;;
-    umount | um)
-      (
-        include "$KW_LIB_DIR/vm.sh"
+        parse_vm_options "$@"
 
-        vm_umount "$@"
         local ret="$?"
-        alert_completion 'kw umount' "$1"
+        alert_completion 'kw vm' "$1"
```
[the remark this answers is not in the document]
kwy95: Try to use the `alert_completion` inside the new entry point function, that way you can use it only on the mount and umount options.

tests/vm_test.sh
```diff
@@ -166,4 +166,56 @@ function test_vm_umount()
...
+
+  option_output="$(vm_mount)"
+  unset options_values
+  declare -gA options_values
+  output="$(parse_vm_options --mount)"
+  assertEquals "($LINENO)" "$option_output" "$output"
+
+  function vm_umount()
+  {
+    echo 'umount'
```
kwy95, line 196: Don't use `echo`, always use `printf` instead. I also noticed there are a few `echo`s that were introduced to this file after #439, please replace them as well.

kwy95 (changes requested):
Good PR, try and address the problems I pointed out, and also take a look at the shellcheck problems. We have a script `scripts/kwreview.sh` that automatically checks your code for any formatting errors.

eduardopinheiro:
Thanks for the reviews, will be working to fix the problems and update the PR as soon as possible.

rodrigosiqueira:
@eduardopinheiro any update here?

eduardopinheiro:
So I finish up the review changes and I think I covered it all, let me know if missed something.

eduardopinheiro:
Hi, thanks again for the review. I've done the changes requested and also added the vm_up test, the branch is now rebased with unstable.

PR #463, opened 2021-09-25
setup: doc: integrate dependencies with docs

kwy95:
Moved dependencies list to the documentation folder, with this change
any documentation can reference the dependencies directly and they will
always be updated.
This also avoids executing unnecessary sudo command depending on system
configuration.


documentation/content/installanduninstall.rst
```diff
@@ -13,34 +13,25 @@
...
 
-* libguestfs
-* Qemu
-* Bash >= 4.4
-* git
-* tar
-* python-docutils
-* rsync
+.. include:: ../dependencies/arch.dependencies
+   :literal:
```
rodrigosiqueira, line 20: Is it possible to make these dependencies be displayed in a single line in the documentation? This will make it easier for someone to just copy and paste in the terminal.

kwy95: Yes, without the `:literal:` that is the behavior, I just thought this was easier to read. But a single line would be easier for copying.

rodrigosiqueira:
Merged, thanks

PR #464, opened 2021-09-26
Configm improvements

rodrigosiqueira:
We recently had a great PR from @magalilemes that reworks part of the config manager and expands its functionalities. This PR is a follow-up of that original patch which introduces code refactoring, bug fixes, and extra tests.


rodrigosiqueira:
Merged

Issue #473, opened 2021-10-02
Includes are failing


isinyaaa:
**Describe the bug**
The includes are failing, so `kw` won't run properly.

**To Reproduce**
1. Install kw
2. Run `kw`

**Expected behavior**
kw should be working as normal.

**Screenshots**
![image](https://user-images.githubusercontent.com/39812919/135734395-9c7eb2b8-be12-4fed-bff2-f997b8c825c3.png)

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Version: happens on latest stable and unstable branches versions, i.e. e18d2af and d4253be.
 - Bash Version: 5.1.008-1

**Additional context**
I'm using fish as my main shell, but I'm not sure how it might be interfering.

PR #474, opened 2021-10-02
Fix include variables naming for some systems

isinyaaa:
In some systems, kw couldn't deal properly with hidden folders (i.e. take a look at #473 ), as the variables created for naming includes had dots on them, which made them improper (i.e. the user would see 'not a valid identifier' errors popping up for every `kw` call).


kwy95 (changes requested):
This looks good. First, can you confirm that this solved the issue #473 for you? Second, add tests for this new line, if possible replicate a situation where this code would fail without this fix.
Thanks for the patch.

isinyaaa:
Hi, @kwy95 , this did manage to fix my issue (#473), so it can be closed :)
To add the test I had to do another small change to the `declare` retval as it seems it always returns 0 for some weird reason (?), I hope that's okay

src/kw_include.sh
```diff
@@ -30,6 +30,9 @@ function include()
     return 0
   fi
 
-  declare -g "${varname}"=1
+  ret=$(declare -g "${varname}"=1 2>&1)
```
rodrigosiqueira, line 34: What kind of error have you seen associated with `declare`?

Notice that this code changed the overall behavior of the include function; after your change, we start to have these warnings in the CI:

```
./src/kwio.sh: line 2: declare: BLUECOLOR: readonly variable
./src/kwio.sh: line 3: declare: REDCOLOR: readonly variable
./src/kwio.sh: line 4: declare: YELLOWCOLOR: readonly variable
./src/kwio.sh: line 5: declare: GREENCOLOR: readonly variable
./src/kwio.sh: line 6: declare: SEPARATOR: readonly variable
```
https://github.com/kworkflow/kworkflow/pull/474/checks?check_run_id=3798086709

Tbh, I prefer the original code in this part.

src/kw_include.sh
```diff
@@ -30,6 +30,9 @@ function include()
...
+  if [[ -n "$ret" ]]; then
+    return 1
```
rodrigosiqueira, line 36: In `kw` we use the errno code for handling failures; in this sense, I think 1 is not the correct code here (EPERM 1 Operation not permitted)

tests/samples/.hidden/include_test.sh
```diff
@@ -0,0 +1,4 @@
+function test_hidden()
+{
+  printf '%s\n' 'output of test hidden'
+}
```
rodrigosiqueira, line 4: Do we need this function?

src/kw_include.sh
```diff
@@ -24,6 +24,7 @@ function include()
   varname="${varname#"$KW_LIB_DIR/"}" # leave path until KW_LIB_DIR
   varname="${varname//\//_}"          # change bars to underlines
   varname="${varname%.*}"             # remove extension
+  varname="${varname/\./}"            # remove "." from hidden folder
```
rodrigosiqueira, line 27: I think you can squash this commit with the previous one.

rodrigosiqueira (changes requested):
Hi @isinyaa,
First of all, thanks a lot for your PR; that is a great contribution.
You can see some of my comments in the PR. Just for curiosity, did you run all tests locally? Is everything ok on your side?

src/kw_include.sh
isinyaaa: I was following the reference at the beginning of the file, but now I think I have a better idea for doing this!

tests/samples/.hidden/include_test.sh
isinyaaa: Not really :)

src/kw_include.sh
isinyaaa: I had separated them in an unusual way, let me see if I can make this any better.

isinyaaa:
> Just for curiosity, did you run all tests locally? Is everything ok on your side?

So, I thought I was alright when I ran the tests partially, but it was a little lax on my part not checking it thoroughly. I apologize!

src/kw_include.sh
```diff
@@ -24,12 +24,19 @@ function include()
...
   if [[ -v "${varname}" ]]; then
     return 0
   fi
 
   declare -g "${varname}"=1
+  retval="$?"
+
+  if [[ "$retval" != 0 ]]; then
+    return "$retval"
+  fi
```
rodrigosiqueira, line 39: `[[ "$?" != 0 ]] return 22 # EINVAL`

isinyaaa:
This code should be good now, but I'd like to point out it'd be interesting to complement the docs for `shellcheck`, as the `pre-commit` hooks for it should also ignore the `SC1090,SC1091` warnings, which are already ignored in the project's CI but not in the reference file in the docs.

With that said, how can I contribute to the docs (I mean, [these docs](https://kworkflow.org/content/developmentworkflow.html)? :)

Also: It'd be interesting to sort out the return value thing for this file before merging this, as I mentioned in the review comment.

kwy95 (changes requested):
I was doing some more extensive testing on this PR and I noticed a few edge cases I think we should consider:

1. Currently if the file is inside a hidden folder and has no extension the include function will remove most of the path. E.g.: `/some/.hidden/path/file` -> `_SOME__IMPORTED`
2. If the file is inside nested hidden folders it will still get problems as we only remove the first occurrence of the dot. E.g.: `/some/.hidden/.twice/path/file.sh` -> `_SOME_HIDDEN_.TWICE_PATH_FILE_IMPORTED`
3. If there are two files with the same name inside different folders that have the same name except one of them is hidden, the resulting variable would be identical and we wouldn't be able to include both files at the same time. E.g.: `/some/path/file.sh` -> `_SOME_PATH_FILE_IMPORTED`; and `/some/.path/file.sh` -> `_SOME_PATH_FILE_IMPORTED`

With this in mind I suggest we stop removing the file extensions and simply replace all dots with underscores. This should avoid all cases pointed out above.
I also think we should consider removing spaces from the path, as some systems allow for files and folder names with spaces.

kwy95:
@isinyaaa Some of the documentation is in fact outdated. The docs are in the `documentation` folder, you can edit them on a separate branch and create a PR for the change.

src/kw_include.sh
```diff
@@ -24,12 +24,17 @@ function include()
...
   varname="${varname%.*}"             # remove extension
+  varname="${varname/\./}"            # remove "." from hidden folder
   varname="${varname^^}_IMPORTED"     # capitalize and append "_IMPORTED"
 
   if [[ -v "${varname}" ]]; then
     return 0
   fi
 
   declare -g "${varname}"=1
+  retval=""
```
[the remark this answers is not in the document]
isinyaaa: oops, my bad!

isinyaaa:
> I was doing some more extensive testing on this PR and I noticed a few edge cases I think we should consider:
> 
> 1. Currently if the file is inside a hidden folder and has no extension the include function will remove most of the path. E.g.: `/some/.hidden/path/file` -> `_SOME__IMPORTED`
> 2. If the file is inside nested hidden folders it will still get problems as we only remove the first occurrence of the dot. E.g.: `/some/.hidden/.twice/path/file.sh` -> `_SOME_HIDDEN_.TWICE_PATH_FILE_IMPORTED`
> 3. If there are two files with the same name inside different folders that have the same name except one of them is hidden, the resulting variable would be identical and we wouldn't be able to include both files at the same time. E.g.: `/some/path/file.sh` -> `_SOME_PATH_FILE_IMPORTED`; and `/some/.path/file.sh` -> `_SOME_PATH_FILE_IMPORTED`
> 
> With this in mind I suggest we stop removing the file extensions and simply replace all dots with underscores. This should avoid all cases pointed out above. I also think we should consider removing spaces from the path, as some systems allow for files and folder names with spaces.

I think this makes a lot of sense! Will also add tests for these cases :)

isinyaaa:
I added those tests but then we've a whole lot of files, which bothered me a little. Those could be empty, though, what do you think?

run_tests.sh
```diff
@@ -175,7 +175,7 @@ check_files="$?"
 #shellcheck disable=SC2086
 if [[ "$#" -eq 0 ]]; then
   check_required_files
-  files_list=$(find ./tests -name '*_test.sh' | grep --invert-match '/shunit2/')
+  files_list=$(find ./tests -name '*_test.sh' | grep --invert-match 'samples/.*' | grep --invert-match '/shunit2/')
```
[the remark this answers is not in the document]
isinyaaa: I just noticed those are actually **not** equivalent:
```bash
$ diff <(find ./tests -name '*_test.sh' | grep --invert-match 'samples/.*' | grep --invert-match '/shunit2/') <(find ./tests -name '*_test.sh' | grep --invert-match 'samples/.*|/shunit2/')
16a17,19
> ./tests/samples/include_test_dir/include_test.sh
> ./tests/samples/.hidden/include_test.sh
> ./tests/samples/.include_test_dir/include_test.sh
```

PR #477, opened 2021-10-04
src: deploy: fix shellcheck 2076

kwy95:
The deploy file had some regular expressions that were within quotes.
That causes the regex to be interpreted literally instead of as
intended. This moved the regex characters outside of the quotes.


rodrigosiqueira:
Nice catch! Merged.

PR #478, opened 2021-10-07
src: kw_time_and_date: fix octal bug

kwy95:
Inside the `days_in_the_month` function the months 08 and 09 were being
interpreted as octal values due to the leading zero, and as they are
greater than 7 it would cause problems related to invalid integer.
This uses bc to correctly convert any numbers to base 10 without any
leading zeros.

The 'short' array has been expanded to accept the months numbers with
leading zeros as this is the usual output by the `date` command.
The leap year check has also been updated for the same reason.

Additional tests have been added to account for these situations.

Unwanted error messages were being displayed whenever statistics was
invoked with wrong date parameters.
This redirects the error messages so they are no longer displayed.

### DIFF
--- a/documentation/dependencies/arch.dependencies
+++ b/documentation/dependencies/arch.dependencies
@@ -18,3 +18,4 @@ lzop
 zstd
 xz
 python-pip
+bc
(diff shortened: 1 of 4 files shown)


rodrigosiqueira:
Very nice catch!
Applied to the unstable branch.

Issue #458, opened 2021-09-19
Add autocomplete per feature


rodrigosiqueira:
Currently, we only have auto-complete for major features. It would be nice to have auto-completed per parameters inside each feature.

lucastheo:
Team UFABC are working this issue

rodrigosiqueira:
PR merged to the unstable branch. Thanks.

rodrigosiqueira:
https://github.com/kworkflow/kworkflow/commit/f984b81c456d138bda53e7715e220afea6404b8e

PR #487, opened 2021-10-15
src: Improve autocompletion on all features

eshiraishi:
The idea to fix Issue #458 is to use two bash features to improve the autocompletion for all features: using arrays that work as a hash map and `mapfile` to read a string and generate an array of possible combinations. This commit is only a proof-of-concept (as it only covers the base `kw` command and the `build` / `b` feature). If this idea is accepted, we can work on replicating it for the rest of the features.

While testing the autocompletion, we realized that adding both the full name options and the short name equivalents (like `kw build` and `kw b`) cluttered the screen without adding any information, so we suggest hiding them from the autocompletion. This does not make the autocompletion stop working if the user types a short name command. Also, the user can still find out about short name equivalents by adding the `--help` flag on each feature, so it's not hidden from them.

Examples:

Without the short name equivalents:

Image

image

image

Can we cut out the short name equivalents for all features or is it preferred to keep both possibilities?

### DIFF
--- a/src/bash_autocomplete.sh
+++ b/src/bash_autocomplete.sh
@@ -1,31 +1,69 @@
 function _kw_autocomplete()
 {
-  local current_command previous_command kw_options
-  COMPREPLY=()
-  current_command="${COMP_WORDS[COMP_CWORD]}"
-  previous_command="${COMP_WORDS[COMP_CWORD - 1]}"
-  kw_options="explore e bd build b init ssh s clear-cache
-                mount mo umount um vars up u codestyle c configm g
-                maintainers m deploy d help h version statistics
-                pomodoro p report r device drm diff df --version -v man
-                backup debug mail"
+  declare -A kw_options
 
-  # By default, autocomplete with kw_options
-  if [[ ${previous_command} == kw ]]; then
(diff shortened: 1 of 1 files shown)


rodrigosiqueira:
Hi @eshiraishi ,

First of all, thanks a lot for this PR. You are on the right path, in my opinion.

> Can we cut out the short name equivalents for all features, or is it preferred to keep both possibilities?

Let's ignore the short version ;)

kwy95 (changes requested):
Ok, the idea is solid, and seems to work on bash. It would be nice if we could have support for zsh as well, but I can try and add that later, as I seem to be the only one using it in this project.
Also, you need to rebase your changes to the latest unstable, and add autocompletion to the mail functionality. And I think this PR can be on a single commit.
**Important:** something is messed up with this PR, you seem to have merged the master branch into it, which has a few merge commits that are not supposed to be on the unstable branch. Please fix this.

rodrigosiqueira:
Hi,
I made some tiny fixes in the commit message, and I removed some extra spaces in the code. Aside from that, everything looks good, and for this reason, I have already applied this change to the unstable branch.

Thanks a lot for this PR. This is a very nice contribution.

Issue #308, opened 2021-06-07
test bug: get_today_info_Test sometimes fail


alan-barzilay:
as noticed in #307, sometimes the get_today_info_Test in kw_time_and_date_Test.sh may fail. 
Looking at the test, this seems to be caused by a race condition. If it so happens that a second is passed between lines 38 and 40, the test will fail. We probably shouldn´t be testing the seconds minutes and hours parts.

kwy95:
Looking at the tests I think this is an inevitability of the way these tests and `get_today_info()` were implemented, currently during the test the `date` command is called 3 times in succession, which will yield 3 slightly different times, when those times happen to cross the seconds mark this test will fail.
I can think of 3 ways to alleviate this problem:

1. We stop testing the seconds.
2. We add an arbitrary margin of error for this test.
3. We redo the test if it fails once.

Not sure which one we should go for if we decide to change anything, but I think the first two would definitely be the easier but the third one is probably more robust.

alan-barzilay:
I agree. Honestly I think we should just pick whichever option is easier to implement, all your suggestions are reasonable

rodrigosiqueira:
https://github.com/kworkflow/kworkflow/commit/f984b81c456d138bda53e7715e220afea6404b8e

PR #493, opened 2021-10-21
tests: kw_time_and_date_test: fix bug

pauloguilhermepp:
The function test_get_today_info sometimes failed when a second passed
between calling date and get_today_info functions. Solves #308

Adds a new variable to check the date return before and after the
calling of the get_today_info function. Checks if they are the same
and, if they are not, repeat the process. Assuring that both calls
were made in the same second.

It was also considered to use a while loop instead of an if, but
to avoid problems with possible endless loops, this solution
was disregarded.

Signed-off-by: Paulo Guilherme <pauloguilherme@usp.br>
Co-authored-by: Lucas Cardoso <lucas.cardoso.santos@usp.br>

### DIFF
--- a/tests/kw_time_and_date_test.sh
+++ b/tests/kw_time_and_date_test.sh
@@ -30,12 +30,18 @@ function test_get_today_info()
   local today
 
   today=$(date +%Y/%m/%d)
-  formated_today=$(get_today_info '+%Y/%m/%d')
-  assert_equals_helper 'Today info did not match' "$LINENO" "$today" "$formated_today"
+  formatted_today=$(get_today_info '+%Y/%m/%d')
+  assert_equals_helper 'Today info did not match' "$LINENO" "$today" "$formatted_today"
 
-  formated_today=$(get_today_info)
   today=$(date)
-  assert_equals_helper 'No parameter' "$LINENO" "$today" "$formated_today"
+  formatted_today=$(get_today_info)
+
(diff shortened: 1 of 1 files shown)


tests/kw_time_and_date_test.sh
```diff
@@ -28,13 +28,22 @@ function test_sec_to_format()
...
 
-  formated_today=$(get_today_info)
   today=$(date)
+  formated_today=$(get_today_info)
+  today_double_check=$(date)
+
+  if [[ "$today" != "$today_double_check" ]]; then
+    today=$(date)
+    formated_today=$(get_today_info)
+    today_double_check=$(date)
```
kwy95, line 44: This second assignment is not needed, as this value will not be used past this point.

kwy95 (changes requested):
I like the idea, but the problem here is that the second may have passed between the `today` and `today_double_check` when the actual test would have passed. I think it's better to test `formated_today` directly. Also since you are already here, change occurences of `formated` in the variable names of this file to the correct spelling of `formatted`.

tests/kw_time_and_date_test.sh
Pauloguilhermepp: Thanks to the comment about line 44, I do not know how we forget this line there. But I am not sure if I understood what you said about the problem with this solution. Would you suggest just comparing the variables `today` and `formatted_today` instead of creating a new variable?

PR #498, opened 2021-10-25
setup: add kw man page to mandb

kwy95:
Adds a compressed copy of the main man page of kw to a location that the
man command can find so that running 'man kw' becomes possible.

Also stops synchronization of raw documentation files as they are not
needed for any features to work, so no need to clutter the users system.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -309,15 +309,22 @@ function synchronize_files()
   # Old kworkflow.config uses complete.wav instead of bell
   ln -s "$sounddir/bell.wav" "$sounddir/complete.wav"
 
-  # Documentation files
-  mkdir -p "$docdir"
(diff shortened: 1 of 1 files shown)


isinyaaa:
> Tbh, I'm not super excited with the idea of installing kw outside `/home` via `setup.sh`; in my opinion, an official package should do that. However, I'm not against the idea of creating a new option in the `setup.sh` named `--system` or similar where we install kw as system software. What do you think?

I recently opened an `orphan request` [on the AUR](https://aur.archlinux.org/packages/kworkflow-git) to try and update kw's PKGBUILD in there, so I think this packaging could be a good primer to test this sort of feature :)

Maybe you can quote that in the new issue, @kwy95 ?

rodrigosiqueira:
@kwy95 any comment to Isabella's question?

kwy95:
Take a look at #508

PR #502, opened 2021-10-28
Add interaction support to init.sh for ssh setup and remote config

melloguel:
In this commit we modified init.sh to look for the --interactive
(or -i) options when parsing arguments. If these options are found,
the script will attempt to interactively detect and setup SSH in
the user's system. Prompts are added to allow the user to interact
with this process and skip steps if desired. We also added an
interactive configuration step for the user's remote connection
settings after SSH is setup. These functionalities are intended to
provide a better user experience for new users who may never have
setup KW before. Functionalities yet to be implemented for this
goal are commented with a #TODO: in the body of the script.

Co-authored-by: Miguel de Mello Carpi <49130937+melloguel@users.noreply.github.com>
Co-authored-by: Lourenço Henrique Moinheiro Martins Sborz Bogo <62181861+louhmmsb@users.noreply.github.com>
Co-authored-by: Eduardo Brancher Urenha <49043437+EduBrancher@users.noreply.github.com>

### DIFF
--- a/documentation/man/features/init.rst
+++ b/documentation/man/features/init.rst
@@ -29,6 +29,12 @@ OPTIONS
   Set the variable `default_deploy_target` from **kworkflow.config** to
   *<target>*, which can be any of vm, local or remote.
 
+\--interactive:
+  Initiates interactive setup mode. In this mode, kw will suggest a series of
+  configurations, explaining their purpose and recommending default options.
+  User confirmation and input is required to configure every feature.
+  Recommended for first time users.
+
 EXAMPLES
 ========
 For these examples, we suppose that the kernel directory is your current
--- a/src/init.sh
+++ b/src/init.sh
@@ -229,5 +229,6 @@ function init_help()
     '  init - Creates a kworkflow.config file in the current directory.' \
     '  init --arch <arch> - Set the arch field in the kworkflow.config file.' \
     '  init --remote <user>@<ip>:<port> - Set remote fields in the kworkflow.config file.' \
-    '  init --target <target> Set the default_deploy_target field in the kworkflow.config file'
+    '  init --target <target> Set the default_deploy_target field in the kworkflow.config file' \
+    '  init --interactive - Perform first time setup in an interactive manner. Recommended for new users.'
 }
--- a/src/init.sh
+++ b/src/init.sh
@@ -39,6 +39,13 @@ function init_kw()
 
   if [[ "${options_values['INTERACTIVE']}" ]]; then
 
+    say 'Welcome to the interactive mode!'     
+    say 'We will ask you a series of questions regarding how do you want to configure kernel workflow.'
+    say 'Before each question there will be a brief explanation of the feature and why it is used.'
+    say 'This is not exhaustive, we recommend you search for more information online.'
+    say 'Please follow the prompts to continue.'
+    echo 
+    
     local cmd=''
     local distro=''
     local has_ssh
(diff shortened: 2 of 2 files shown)


documentation/man/features/init.rst
```diff
@@ -29,6 +29,12 @@ OPTIONS
   Set the variable `default_deploy_target` from **kworkflow.config** to
   *<target>*, which can be any of vm, local or remote.
 
+\--interactive:
```
kwy95, line 32: This is missing the `-i` option. Also, you need to add this option to the synopsis.

kwy95: Look at `build.rst` for reference.

src/init.sh
```diff
@@ -229,5 +229,6 @@ function init_help()
     '  init - Creates a kworkflow.config file in the current directory.' \
     '  init --arch <arch> - Set the arch field in the kworkflow.config file.' \
     '  init --remote <user>@<ip>:<port> - Set remote fields in the kworkflow.config file.' \
-    '  init --target <target> Set the default_deploy_target field in the kworkflow.config file'
+    '  init --target <target> Set the default_deploy_target field in the kworkflow.config file' \
+    '  init --interactive - Perform first time setup in an interactive manner. Recommended for new users.'
```
kwy95, line 255: I just noticed the synopsis on the documentation is missing some of these options, could you create a separate commit to add these options to the man page?
Also, this documentation for the feature can be in the same commit as the feature implementation.

src/init.sh
```diff
@@ -39,17 +39,37 @@ function init_kw()
 
   if [[ "${options_values['INTERACTIVE']}" ]]; then
 
+    say 'Welcome to the interactive mode!'     
+    say 'We will ask you a series of questions regarding how do you want to configure kernel workflow.'
+    say 'Before each question there will be a brief explanation of the feature and why it is used.'
+    say 'This is not exhaustive, we recommend you search for more information online.'
+    say 'Please follow the prompts to continue.'
+    echo 
+
```
kwy95: These should happen after variable declarations.

src/init.sh
```diff
@@ -39,17 +39,37 @@ function init_kw()
 
   if [[ "${options_values['INTERACTIVE']}" ]]; then
 
+    say 'Welcome to the interactive mode!'     
+    say 'We will ask you a series of questions regarding how do you want to configure kernel workflow.'
+    say 'Before each question there will be a brief explanation of the feature and why it is used.'
+    say 'This is not exhaustive, we recommend you search for more information online.'
+    say 'Please follow the prompts to continue.'
+    echo
```
kwy95: As per our codestyle, don't use `echo`, prefer `printf` instead.

src/init.sh
```diff
@@ -39,17 +39,37 @@ function init_kw()
 
   if [[ "${options_values['INTERACTIVE']}" ]]; then
 
+    say 'Welcome to the interactive mode!'     
+    say 'We will ask you a series of questions regarding how do you want to configure kernel workflow.'
```
kwy95, line 43: How about this?
```suggestion
    say 'We will ask a series of questions to help you configure kw.'
```

src/init.sh
```diff
@@ -39,17 +39,37 @@ function init_kw()
...
+    say 'operating network services securely over an unsecured network.'
+    say 'This is useful if you want to deploy a kernel remotely or inside a virtual machine (VM).'
+    say 'Deploying in a remote machine or VM is recommended so you do not break your own kernel.'
+    say 'For a tutorial on how to setup a VM compatible with KW refer to:' 
+    say 'https://flusp.ime.usp.br/others/use-qemu-to-play-with-linux/'
+    say 'KW will now attempt to automatically locate SSH.'
+    warning 'If your Linux distribution is not Debian or Arch this will likely fail.'
+    warning 'We do not support any other distributions (yet).'
+    warning 'We strongly recommend using one of these distributions to work with KW.'
+    echo
```
kwy95, line 67: Again, no `echo`s

src/init.sh
```diff
@@ -39,17 +39,37 @@ function init_kw()
...
+
     case "$distro" in
       none)
         warning "We do not support your distro (yet). We cannot check if SSH is installed."
-	if [[ $(ask_yN "Do you wish to proceed without configuring SSH?") =~ '0' ]]; then
+	      if [[ $(ask_yN "Do you wish to proceed without configuring SSH?") =~ '0' ]]; then
```
kwy95: I think this is a mistake, right?

src/init.sh
```diff
@@ -27,7 +27,7 @@ function init_kw()
   if ! is_kernel_root "$PWD"; then
     complain 'This command should be run in a kernel tree.'
     if [[ $(ask_yN "Do you want to continue?") =~ '0' ]]; then
-    	exit 125 # ECANCELED
+      exit 125 # ECANCELED
```
kwy95, line 30: These problems pointed out by `shfmt` should be fixed before each commit, as to not introduce any commits that are not compliant with our github workflows.

kwy95 (changes requested):
Thanks for this PR, please review my comments and address the problems I pointed out.
Also, don't forget to add tests.

rodrigosiqueira:
Hi @melloguel ,
Any update since our discussion?

melloguel:
Hello, @rodrigosiqueira,
We'll close this PR and do another one with more proper commits.

PR #504, opened 2021-10-29
src: improve usability in ask_yN

LucasCardoso910:
There was no way to select a default answer in ask_yN and to
change what letter would be capitalized. The pattern, in other
Bash applications, is to use the capitalized letter as a default
answer, so the user must only press ENTER to select it.

Add a second parameter to change the default option, capitalizing
the letter corresponding to it. Also, add a first test in the
user's answer to check if the answer is empty. This allows the user
to select the default answer by typing anything.

Signed-off-by: Lucas Cardoso <lucas.cardoso.santos@usp.br>
Co-authored-by: Paulo Guilherme <pauloguilherme@usp.br>

### DIFF
--- a/src/kwio.sh
+++ b/src/kwio.sh
@@ -99,20 +99,35 @@ function success()
   colored_print GREENCOLOR "$@"
 }
 
-# Ask for yes or no
+# Allows the user to select a given question with yes or no answers. Shows a
+# default option (as a capitalized letter). If the user inserts anything
+# different from a "y" or a "yes" (capitalization doesn't matter here) it will
+# be treated as a "no".
 #
 # @message A string with the message to be displayed for the user.
+# @default_option A string (simply y or n) to set the default answer for the
+#                 user. If this parameter is not given, or if it is invalid, 'n'
+#                 will be used.
 #
-# Returns:
+# Return:
 # Return "1" if the user accept the question, otherwise, return "0"
-#
 # Note: ask_yN return the string '1' and '0', you have to handle it by
 # yourself in the code.
 function ask_yN()
 {
-  local message="$*"
+  local message="$1"
+  local default_option="${2:-'n'}"
+  local yes='y'
+  local no='N'
 
-  read -r -p "$message [y/N] " response
+  if [[ "$default_option" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+    yes='Y'
+    no='n'
+  fi
+
+  message="$message [$yes/$no]"
+  response=$(ask_with_default "$message" "$default_option" false)
+
   if [[ "$response" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
     printf '%s\n' '1'
   else
--- a/src/kwio.sh
+++ b/src/kwio.sh
@@ -119,3 +119,41 @@ function ask_yN()
     printf '%s\n' '0'
   fi
 }
+
+# Asks for user input
+#
+# @message A string with the message to be displayed to the user.
+# @default_option String with default answer to be used if no input is given.
+# @show_default A string that defines if the default value will be printed to
+#               the user. Any non-empty value causes this behavior to be false.
+# @flag String to mark special conditions in this function. To activate the test
+#       mode, set this to 'TEST_MODE'.
+#
+# Return:
+# The user answer, guaranteed not to be empty.
+# Note: this function does not verify the given answer. You have to handle this
+# somewhere else.
+function ask_with_default()
+{
+  local message="$1"
+  local default_option="$2"
+  local show_default="$3"
+  local flag="$4"
+  local value
+
+  if [[ -z "$show_default" ]]; then
+    message+=" ($default_option)"
+  fi
+  message+=': '
+
+  [[ "$flag" == 'TEST_MODE' ]] && printf '%s\n' "$message"
+
+  read -r -p "$message" response
+
+  if [[ "$?" -ne 0 || -z "$response" ]]; then
+    printf '%s\n' "$default_option"
+    return
+  fi
+
+  printf '%s\n' "$response"
+}
--- a/tests/kwio_test.sh
+++ b/tests/kwio_test.sh
@@ -124,4 +124,33 @@ function test_alert_completion_sound_alert()
   assertEquals 'Variable s should exist.' "$expected" "$output"
 }
 
+function test_ask_with_default()
+{
+  local output=''
+  local expected_output=''
+  local assert_equals_message=''
+
+  # Default option showing
+  expected_output=$'Insert something here (lala): \nsomething'
+  assert_equals_message='Default answer and user answer are different.'
+  output=$(printf 'something\n' | ask_with_default 'Insert something here' 'lala' '' 'TEST_MODE')
+  assert_equals_helper "$assert_equals_message" "$LINENO" "$expected_output" "$output"
+
+  expected_output=$'Insert something here (lala): \nlala'
+  assert_equals_message='User selected default answer.'
+  output=$(printf '\n' | ask_with_default 'Insert something here' 'lala' '' 'TEST_MODE')
+  assert_equals_helper "$assert_equals_message" "$LINENO" "$expected_output" "$output"
+
+  # Default option not showing (third parameter not empty)
+  expected_output=$'Insert something here: \nsomething'
+  assert_equals_message='Not showing default answer, user answered different.'
+  output=$(printf 'something\n' | ask_with_default 'Insert something here' 'lala' 'false' 'TEST_MODE')
+  assert_equals_helper "$assert_equals_message" "$LINENO" "$expected_output" "$output"
+
+  expected_output=$'Insert something here: \nlala'
+  assert_equals_message='Not showing default answer, user selected it.'
+  output=$(printf '\n' | ask_with_default 'Insert something here' 'lala' 'false' 'TEST_MODE')
+  assert_equals_helper "$assert_equals_message" "$LINENO" "$expected_output" "$output"
+}
+
 invoke_shunit
--- a/tests/kwio_test.sh
+++ b/tests/kwio_test.sh
@@ -153,4 +153,88 @@ function test_ask_with_default()
   assert_equals_helper "$assert_equals_message" "$LINENO" "$expected_output" "$output"
 }
 
+function test_ask_yN()
+{
+  local assert_equals_message=''
+
+  assert_equals_message='Default answer: no, user answer: y'
+  output=$(printf 'y\n' | ask_yN 'Test message')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '1' "$output"
+
+  assert_equals_message='Default answer: no, user answer: Y'
+  output=$(printf 'Y\n' | ask_yN 'Test message')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '1' "$output"
+
+  assert_equals_message='Default answer: no, user answer: Yes'
+  output=$(printf 'Yes\n' | ask_yN 'Test message')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '1' "$output"
+
+  assert_equals_message='Default answer: no, user answer: invalid (sim)'
+  output=$(printf 'Sim\n' | ask_yN 'Test message')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '0' "$output"
+
+  assert_equals_message='Default answer: no, user answer: No'
+  output=$(printf 'No\n' | ask_yN 'Test message')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '0' "$output"
+
+  assert_equals_message='Default answer: no, user answer: N'
+  output=$(printf 'N\n' | ask_yN 'Test message')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '0' "$output"
+
+  # Tests with default option selected
+  assert_equals_message='Default answer: N, user answer: y'
+  output=$(printf 'y\n' | ask_yN 'Test message' 'N')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '1' "$output"
+
+  assert_equals_message='Default answer: y, user answer: Y'
+  output=$(printf 'Y\n' | ask_yN 'Test message' 'y')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '1' "$output"
+
+  assert_equals_message='Default answer: y, user answer: default'
+  output=$(printf '\n' | ask_yN 'Test message' 'y')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '1' "$output"
+
+  assert_equals_message='Default answer: Y, user answer: n'
+  output=$(printf 'n\n' | ask_yN 'Test message' 'Y')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '0' "$output"
+
+  assert_equals_message='Default answer: n, user answer: N'
+  output=$(printf 'N\n' | ask_yN 'Test message' 'n')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '0' "$output"
+
+  assert_equals_message='Default answer: n, user anser: default'
+  output=$(printf '\n' | ask_yN 'Test message' 'n')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '0' "$output"
+
+  # Invalid default
+  assert_equals_message='Default answer: invalid (lala), user answer: default'
+  output=$(printf '\n' | ask_yN 'Test message' 'lala')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '0' "$output"
+
+  assert_equals_message='Default answer: invalid (lala), user answer: n'
+  output=$(printf 'n\n' | ask_yN 'Test message' 'lala')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '0' "$output"
+
+  assert_equals_message='Default answer: invalid (lala), user answer: y'
+  output=$(printf 'y\n' | ask_yN 'Test message' 'lala')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '1' "$output"
+
+  assert_equals_message='Default answer: invalid (lalaYes), user answer: default (no)'
+  output=$(printf '\n' | ask_yN 'Test message' 'lalaYes')
+  assert_equals_helper "$assert_equals_message" "$LINENO" '0' "$output"
+
+  assert_equals_message='Default answer: invalid (lalaNo), user answer: default (no)'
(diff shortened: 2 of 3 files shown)


src/kwio.sh
```diff
@@ -110,10 +111,25 @@ function success()
...
 {
-  local message="$*"
+  local message="$1"
+  local default="${2:-"n"}"
+  local yes
+  local no
 
-  read -r -p "$message [y/N] " response
-  if [[ "$response" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+  if [[ "$default" =~ ^([yY]|[yY][eE][sS])+$ ]]; then
```
rodrigosiqueira, line 119: Let's drop `eE` and `sS` match. I think this might confuse users.

src/kwio.sh
```diff
@@ -110,10 +111,25 @@ function success()
 # yourself in the code.
 function ask_yN()
 {
-  local message="$*"
+  local message="$1"
+  local default="${2:-"n"}"
```
rodrigosiqueira, line 115: Single-quote around `n`. Also, how about using `default_option`?

src/kwio.sh
```diff
@@ -110,10 +111,25 @@ function success()
...
+  local yes
+  local no
 
-  read -r -p "$message [y/N] " response
-  if [[ "$response" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+  if [[ "$default" =~ ^([yY]|[yY][eE][sS])+$ ]]; then
+    default='1'
+    yes="Y"
+    no="n"
```
rodrigosiqueira, line 122: Use single-quote

src/kwio.sh
```diff
@@ -110,10 +111,25 @@ function success()
...
+  else
+    default='0'
+    yes="y"
+    no="N"
+  fi
```
rodrigosiqueira, line 127: You can probably remove this `else` by setting this value as the default when you declare it. For example:
```
local yes='y'
local no='n'
```

tests/plugins/kernel_install/utils_test.sh
```diff
@@ -48,6 +48,35 @@ function test_ask_yN()
...
+
+  output=$(printf '%s\n' 'N' | ask_yN 'Test message' 'n')
+  assert_equals_helper 'TEST_MODE' "$LINENO" '0' "$output"
+
+  output=$(printf '%s\n' '' | ask_yN 'Test message' 'n')
+  assert_equals_helper 'TEST_MODE' "$LINENO" '0' "$output"
+
+  # Invalid default
+  output=$(printf '%s\n' '' | ask_yN 'Test message' 'lala')
+  assert_equals_helper 'TEST_MODE' "$LINENO" '0' "$output"
```
rodrigosiqueira, line 73: After seeing this test, I believe that you need to document this behavior in the function comment. Explain that anything different from `Y/y`, will be treated as `n`.

rodrigosiqueira (changes requested):
Hi @LucasCardoso910 ,
Nice PR, I added a few comments.

src/kwio.sh
LucasCardoso910: I liked the idea, but we found a problem removing this `else`, because the `default_option` integer value is set in this block and before it has the 'y' or 'n' value. We thought that we could create a new variable to assume the integer value, while `default_option` remains only with the char given to the function, but this function would be with five local variables already. What is the best solution?

tests/plugins/kernel_install/utils_test.sh
LucasCardoso910: Well, it depends. In this case, it was given an invalid default option to the function, so it will be used 'n' (just like if it was given nothing). If the user chooses nothing, in this case, will be chosen 'n', but if the function was given 'y' as the default option and the user chooses nothing, it will be chosen 'y'. Is this the behavior you are referring to?

LucasCardoso910:
We changed a little the commit. It was added a new function (ask_with_default) for a more general-purpose task than ask_yN. This function deals with default inputs if the user just presses enter. So, we remodeled ask_yN to call this function and now it just deals with the functions print. We also found some unused variables and removed them (but did in a separate commit because we weren't certain if they were not there for a more obscure reason).

Speaking of prints, it was talked about changing the behavior of ask_yN and instead of printing the result, returning it. It is a good idea, but to do so (in my tests I discovered) would be needed to change all the function calls since they are made in the model of this: `if [[ $(ask_yN 'Do you want to make your host kernel image readable?') =~ 0 ]]; then` (example found in line 63 in the file vm.sh). These calls break when you change it to return instead of print.

src/kwio.sh
```diff
@@ -119,3 +119,36 @@ function ask_yN()
...
+
+# Asks for user input
+#
+# @message A string with the message to be displayed to the user.
+# @default_option A string with the default answer of the user (if the user simply
+#                 inputs nothing, this will be considered the answer)
+# @show_default A boolean value that defines if the default value will be printed
+#               to the user. If this value is empty, this will be considered true.
+#
+# Returns: The user answer, garanted it is not empty.
```
kwy95, line 143: There is a typo here, I'm not sure what you meant, but I think it was either `granted` or `guaranteed`

src/kwio.sh
```diff
@@ -119,3 +119,36 @@ function ask_yN()
...
+#
+# Note: this function does not any other process of answer verification. You have
+# to handle this somewhere else.
```
kwy95, line 146: How about this:
```suggestion
# Note: this function does not verify the given answer. You have to handle this
# somewhere else.
```

tests/plugins/kernel_install/utils_test.sh
kwy95: I think what you should document is the fact that the function only checks if the`response` is a valid yes, otherwise it is considered as a no.

kwy95 (changes requested):
Good PR, just a few problems that need to be addressed.

src/kwio.sh
```diff
@@ -119,3 +119,36 @@ function ask_yN()
...
+# Asks for user input
+#
+# @message A string with the message to be displayed to the user.
+# @default_option A string with the default answer of the user (if the user simply
+#                 inputs nothing, this will be considered the answer)
+# @show_default A boolean value that defines if the default value will be printed
+#               to the user. If this value is empty, this will be considered true.
+#
+# Returns: The user answer, guaranteed it is not empty.
+#
```
rodrigosiqueira, line 148: Use:
```
Return:
The user answer, guaranteed it is not empty
```

src/kwio.sh
```diff
@@ -119,3 +119,36 @@ function ask_yN()
...
+# Note: this function does not verify the given answer. You have to handle this
+# somewhere else.
+
```
rodrigosiqueira, line 151: Drop this extra line

src/kwio.sh
```diff
@@ -119,3 +119,36 @@ function ask_yN()
...
+  else
+    message="$message: "
+  fi
+
+  read -r -p "$message" response
+  if [[ -z "$response" ]]; then
+    printf '%s\n' "$default_option"
+  else
+    printf '%s\n' "$response"
+  fi
```
rodrigosiqueira, line 169: How about?
```bash
if [[ -z "$response" ]]; then
    printf '%s\n' "$default_option"
    return
fi
printf '%s\n' "$response"
```

tests/plugins/kernel_install/utils_test.sh
```diff
@@ -68,6 +68,28 @@ function test_ask_yN()
   assert_equals_helper 'TEST_MODE' "$LINENO" '0' "$output"
 }
 
+function test_ask_with_default()
+{
```
rodrigosiqueira, line 109: Add:
```bash
local output=''
```

tests/plugins/kernel_install/utils_test.sh
```diff
@@ -68,6 +68,28 @@ function test_ask_yN()
   assert_equals_helper 'TEST_MODE' "$LINENO" '0' "$output"
 }
 
+function test_ask_with_default()
```
rodrigosiqueira, line 108: This test is in the wrong place.
It should be inside `tests/kwio_test.sh`

tests/plugins/kernel_install/utils_test.sh
```diff
@@ -68,6 +68,28 @@ function test_ask_yN()
...
+{
+  output=$(printf '%s\n' 'something' | ask_with_default 'Test message' 'lala')
+  assert_equals_helper 'TEST_MODE' "$LINENO" 'something' "$output"
```
rodrigosiqueira, line 111: The first parameter of assert_equal_helper should be a message that a developer can read and understand.

src/kwio.sh
```diff
@@ -119,3 +119,36 @@ function ask_yN()
...
+  local default_option="$2"
+  local show_default="$3"
+
+  if [[ -z "$show_default" ]]; then
+    message="$message ($default_option): "
+  else
+    message="$message: "
+  fi
+
+  read -r -p "$message" response
```
rodrigosiqueira, line 164: Could you try to add a test that captures the default behavior? I mean, you don't have any test that checks the `$message`.

tests/plugins/kernel_install/utils_test.sh
```diff
@@ -66,6 +66,48 @@ function test_ask_yN()
 
   output=$(printf '%s\n' 'N' | ask_yN 'Test message')
   assert_equals_helper 'TEST_MODE' "$LINENO" '0' "$output"
+
+  # Tests with default option selected
+  output=$(printf '%s\n' 'y' | ask_yN 'Test message' 'N')
+  assert_equals_helper 'Test_MODE' "$LINENO" '1' "$output"
```
rodrigosiqueira, line 67: Add a meaningful message and move this test to kwio_test.

rodrigosiqueira (changes requested):
Thanks for this new version.
I added a few sets of comments; please, try to address all of them. Also, rebase your change with the latest code from the unstable branch.

kwy95 (changes requested):
I think @rodrigosiqueira already mentioned most problems, but here are some comments anyway.

tests/plugins/kernel_install/utils_test.sh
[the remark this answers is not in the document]
LucasCardoso910: We used as a pattern to build this function the test_ask_yN function that was already here. It used `TEST_MODE`, so we used it too, but yes, a descriptive message sounds better for me too, we are going to change it.

src/kwio.sh
LucasCardoso910: Do you mean to test if the received `$message` is empty or something like that?

rodrigosiqueira: Try to capture the output of the `read` command and check if the message is the default or not.

LucasCardoso910:
We edited partially the requested changes. There is only one missing that is the one about checking `$message`, we wish to understand a little bit better. But we already added the check if the output of `$answer` is 0, if so, it returns the default answer (is this the better thing to do?). 

We are sending this partially done commit only to allow a better discussion about it.

LucasCardoso910:
It was a little bit more complicated to create a test that would evaluate also the `$message` variable. Unfortunately, capturing the `read` value didn't return the printed text, so we separated in a `printf` and a `read`. But, for more complications, `ask_yN` performed the same operation to get its return value. We tested some options to assure everything was working, but the simpler, in our opinion, was to just get also the message inside `ask_yN` and remove everything behind the last `'\n'` in the `output`, so we used Bash expansion for it.

Pauloguilhermepp:
I and Lucas talked about this PR and we had a better idea to solve it. We will do another commit to this PR soon.

src/kwio.sh
```diff
@@ -119,3 +119,42 @@ function ask_yN()
...
   fi
 }
+
+# Asks for user input
+#
+# @message A string with the message to be displayed to the user.
+# @default_option A string with the default answer to be used if no input is given
+# @show_default A boolean value that defines if the default value will be printed
+#               to the user. If this value is empty, this will be considered true.
+#
```
rodrigosiqueira, line 144: Add @flag.

src/kwio.sh
```diff
@@ -119,3 +119,42 @@ function ask_yN()
...
+
+  if [[ -z "$show_default" ]]; then
+    message+=" ($default_option)"
+  fi
+  message+=': '
+
+  if [[ "$flag" == 'TEST_MODE' ]]; then
+    printf '%s\n' "$message"
+  fi
+
```
rodrigosiqueira, line 165: `[[ "$flag" == 'TEST_MODE' ]] && printf '%s\n' "$message"`

tests/kwio_test.sh
```diff
@@ -124,4 +124,33 @@ function test_alert_completion_sound_alert()
...
 }
 
+function test_ask_with_default()
+{
+  local output=''
+  local desired_output=''
+  local assert_equals_message=''
+
+  # Default option showing
+  desired_output=$'Insert something here (lala): \nsomething'
```
rodrigosiqueira, line 134: let's use `expected_output`.

tests/kwio_test.sh
```diff
@@ -124,4 +124,33 @@ function test_alert_completion_sound_alert()
...
+
+  desired_output=$'Insert something here (lala): \nlala'
+  assert_equals_message='User selected default answer.'
+  output=$(printf '%s\n' '' | ask_with_default 'Insert something here' 'lala' '' 'TEST_MODE')
+  assert_equals_helper "$assert_equals_message" "$LINENO" "$desired_output" "$output"
+
+  # With third parameter
+  desired_output=$'Insert something here: \nsomething'
+  assert_equals_message='Not showing default answer, user answered different.'
+  output=$(printf '%s\n' 'something' | ask_with_default 'Insert something here' 'lala' 'false' 'TEST_MODE')
```
rodrigosiqueira, line 147: How about replace `%s` by the target string.

rodrigosiqueira (changes requested):

tests/kwio_test.sh
```diff
@@ -124,4 +124,33 @@ function test_alert_completion_sound_alert()
...
+  assert_equals_message='Default answer and user answer are different.'
+  output=$(printf '%s\n' 'something' | ask_with_default 'Insert something here' 'lala' '' 'TEST_MODE')
+  assert_equals_helper "$assert_equals_message" "$LINENO" "$desired_output" "$output"
+
+  desired_output=$'Insert something here (lala): \nlala'
+  assert_equals_message='User selected default answer.'
+  output=$(printf '%s\n' '' | ask_with_default 'Insert something here' 'lala' '' 'TEST_MODE')
+  assert_equals_helper "$assert_equals_message" "$LINENO" "$desired_output" "$output"
+
+  # With third parameter
```
rodrigosiqueira, line 144: Update this comment to something that explains what is the third parameter.

src/kwio.sh
```diff
@@ -119,3 +119,42 @@ function ask_yN()
...
+    message+=" ($default_option)"
+  fi
+  message+=': '
+
+  [[ "$flag" == 'TEST_MODE' ]] && printf '%s\n' "$message"
+
+  read -r -p "$message" response
+  value="$?"
+
+  if [[ "$value" -ne 0 || -z "$response" ]]; then
```
kwy95, line 169: You can check the value of `$?` directly.

tests/kwio_test.sh
```diff
@@ -153,4 +153,89 @@ function test_ask_with_default()
   assert_equals_helper "$assert_equals_message" "$LINENO" "$expected_output" "$output"
 }
 
+function test_ask_yN()
+{
+  local count=0
+  local assert_equals_message=''
+
+  assert_equals_message='Default answer: no, user answer: y'
+  output=$(printf '%s\n' 'y' | ask_yN 'Test message')
```
kwy95, line 162: Use the `printf 'y\n'` format for these tests.

kwy95 (changes requested):
Nice update, this is almost done.

rodrigosiqueira:
Hi @LucasCardoso910 , could you address Rubens's comments?

rodrigosiqueira:
Also, do not forget to rebase your PR in the latest code from the unstable branch.

LucasCardoso910:
Done and done!

src/kwio.sh
```diff
@@ -99,23 +99,76 @@ function success()
...
+  local yes='y'
+  local no='N'
+
+  if [[ "$default_option" =~ ^([yY][eE][sS]|[yY])+$ ]]; then
+    yes='Y'
+    no='n'
+  fi
+
+  message="$message [$yes/$no]"
+  response=$(ask_with_default "$message" "$default_option" false)
```
kwy95, line 129: The `false` option is not needed here.

kwy95 (changes requested):
Fix below for approval.

src/kwio.sh
LucasCardoso910: Actually, it is. The default behavior of the ask_with_default is to print a message in this format: `{message} ({defaul_option}): `. The existence of the third string, in this case, false, is what changes this to print just `{message}: `. Since we want in the ask_yN function to print a message like `{message} [y|N]: `, we change this behavior and append to the original message inside ask_yN the possible options.

Just to be sure, I tested it here and go the following results:
When the false parameter is omitted, using this command `ask_yN 'Do you want to do this?' 'y'` prints `Do you want to do this? [Y/n] (y):`, while when there is the false parameter, the exact same command prints this: `Do you want to do this? [Y/n]: ` .

tests/plugins/kernel_install/utils_test.sh
```diff
@@ -41,8 +41,6 @@ function tearDown()
 
 function test_cmd_manager()
 {
-  local count=0
```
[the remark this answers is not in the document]
LucasCardoso910: I believe a recent commit to this file removed the other two. When I did the rebase, it had a conflict on it. Should we change the commit message or just take this commit out since there is only one variable being removed now? I already thought that this was too small to be a commit, but now even more.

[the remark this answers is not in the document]
kwy95: I think you can remove this commit from this PR and create a separate PR with this change but also see if you can find some other simple updates to this file, like replacing double quotes with single quotes where appropriate.

src/kwio.sh
kwy95: I must have missed something then, my bad.

LucasCardoso910: No worries, when you said it I was unsure about it too, I thought we may have taken something out that wasn't supposed to.

tests/plugins/kernel_install/utils_test.sh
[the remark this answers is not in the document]
LucasCardoso910: Ok, I believe is for the best too. We will do this way

kwy95 (approved):
LGTM

rodrigosiqueira:
Thanks a lot for your effort; the change merged to the unstable branch.

PR #505, opened 2021-11-03
src: remote: Introduce ssh connection check

rodrigosiqueira:
ssh connection is a fundamental part of the kw operation; for this
reason, this commit introduces a simple function for checking
connectivity between the dev system and remote.


kwy95 (approved):
LGTM

PR #509, opened 2021-11-11
Mail setup using templates

kwy95:
Make it possible to setup the git send-email environment using
templates.

### DIFF
--- a/src/mail.sh
+++ b/src/mail.sh
@@ -485,7 +485,12 @@ function parse_mail_options()
         options_values['SETUP']=1
         shift
         ;;
-      --email | --name)
+      --email)
+        if [[ -z "${options_values['sendemail.smtpuser']}" ]]; then
+          options_values['sendemail.smtpuser']="$2"
+        fi
+        ;& # this continues executing the code for --name
+      --name)
         setup_token=1
         option="$(str_remove_prefix "$1" '--')"
         index="user.$option"
--- a/src/mail.sh
+++ b/src/mail.sh
@@ -415,12 +415,14 @@ function template_setup()
 
   if [[ -z "$template" ]]; then
     mapfile -t available_templates < <(find "$KW_ETC_DIR/mail_templates" -type f -printf '%f\n' | sort -d)
-    available_templates+=('exit')
+    [[ -n "${options_values['INTERACTIVE']}" ]] && available_templates+=('skip template')
+    available_templates+=('exit kw mail')
 
     say 'You may choose one of the following templates to start your configuration.'
     printf '(enter the corresponding number to choose)\n'
     select user_choice in "${available_templates[@]^}"; do
-      [[ "$user_choice" == 'Exit' ]] && exit
+      [[ "$user_choice" =~ ^Skip ]] && return
+      [[ "$user_choice" =~ ^Exit ]] && exit
 
       template="${user_choice,,}"
       break
@@ -428,6 +430,11 @@ function template_setup()
   fi
 
   load_template "$template"
+
+  if [[ -z "${options_values['NO_INTERACTIVE']}" &&
+    -z "${options_values['INTERACTIVE']}" ]]; then
+    options_values['INTERACTIVE']='template'
+  fi
 }
 
 # Loads the values from the template file to the options_values array
@@ -474,18 +481,27 @@ function interactive_setup()
 
   flag=${flag:-'SILENT'}
 
-  success 'Welcome to the interactive setup of the mail capabilities.'$'\n'
+  if [[ "${options_values['INTERACTIVE']}" == 'parser' ]]; then
+    success 'Welcome to the interactive setup of the mail capabilities.'$'\n'
 
-  [[ "$(ask_yN 'Do you wish to list the currently set values?')" == '1' ]] && mail_list
-  printf '\n'
+    [[ "$(ask_yN 'Do you wish to list the currently set values?')" == '1' ]] && mail_list
+    printf '\n'
 
-  say 'We will start with the essential configuration options!'$'\n'
+    if [[ -z "${options_values['TEMPLATE']}" ]]; then
+      template_setup
+      printf '\n'
+    fi
+
+    say 'We will start with the essential configuration options!'$'\n'
+  fi
 
   interactive_prompt 'essential_config_options'
 
-  say 'These are the optional configuration options.'$'\n'
+  if [[ "${options_values['INTERACTIVE']}" != 'template' ]]; then
+    say 'These are the optional configuration options.'$'\n'
 
-  interactive_prompt 'optional_config_options' false
+    interactive_prompt 'optional_config_options' false
+  fi
 
   for option in "${essential_config_options[@]}" "${optional_config_options[@]}"; do
     if [[ -n "${options_values["$option"]}" ]]; then
@@ -611,8 +627,8 @@ function parse_mail_options()
   local index
   local option
   local setup_token=0
-  local short_options='t,f,v,i,l,'
-  local long_options='setup,local,global,force,verify,template::,interactive,list,'
+  local short_options='t,f,v,i,l,n,'
+  local long_options='setup,local,global,force,verify,template::,interactive,no-interactive,list,'
 
   long_options+='email:,name:,'
   long_options+='smtpuser:,smtpencryption:,smtpserver:,smtpserverport:,smtppass:,'
@@ -630,6 +646,7 @@ function parse_mail_options()
   options_values['VERIFY']=0
   options_values['TEMPLATE']=''
   options_values['INTERACTIVE']=''
+  options_values['NO_INTERACTIVE']=''
   options_values['SCOPE']='local'
   options_values['CMD_SCOPE']=''
 
@@ -681,7 +698,7 @@ function parse_mail_options()
         ;;
       --interactive | -i)
         options_values['SETUP']=1
-        options_values['INTERACTIVE']='parser'
+        [[ -z "${options_values['NO_INTERACTIVE']}" ]] && options_values['INTERACTIVE']='parser'
         shift
         ;;
       --local)
@@ -700,6 +717,10 @@ function parse_mail_options()
         ;;
       --force | -f)
         options_values['FORCE']=1
+        ;&
+      --no-interactive | -n)
+        options_values['INTERACTIVE']=''
+        options_values['NO_INTERACTIVE']=1
         shift
         ;;
       --)
@@ -729,5 +750,5 @@ function mail_help()
     '  mail (-i | --interactive) - Setup interactively' \
     '  mail (-v | --verify) - Check if required configurations are set' \
     '  mail (-l | --list) - List the configurable options' \
-    '  mail --template[=<template>] - Set send-email configs based on <template>'
+    '  mail --template[=<template>] [-n] - Set send-email configs based on <template>'
 }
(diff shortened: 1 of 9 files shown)


rodrigosiqueira (changes requested):
Hi @kwy95 ,

You can see my comments in the code, but I also have these observations:

I tried:

kw mail --template

and I selected Riseup, and I got this message:

Invalid template: Riseup


Additionally, if I try:
 kw mail --template riseup --email "test@lala"
 Invalid email: test@lala
 sendemail.smtpserverport at the [local] scope was successfully set to '465'
 sendemail.smtpencryption at the [local] scope was successfully set to 'tls'
 sendemail.smtpserver at the [local] scope was successfully set to 'mail.riseup.net'

I suppose we should not set up anything if we have a failure.

kwy95:
> I tried:
> 
> kw mail --template
> 
> and I selected Riseup, and I got this message:
> 
> Invalid template: Riseup
> 
> Additionally, if I try: kw mail --template riseup --email "test@lala" Invalid email: test@lala sendemail.smtpserverport at the [local] scope was successfully set to '465' sendemail.smtpencryption at the [local] scope was successfully set to 'tls' sendemail.smtpserver at the [local] scope was successfully set to 'mail.riseup.net'
> 
> I suppose we should not set up anything if we have a failure.

The problem here is that I capitalize the options to display them to the user, but forgot to restore the value to load the templates.

src/mail.sh
```diff
@@ -238,6 +238,29 @@ function get_configs()
...
+# Checks what configs from given list are not set
+#
+# @_config_options: reference to array containing the relevant options
+#
+# Returns:
+# Array with first element as the amount of configs missing, followed by the
+# missing configs
+function missing_opts()
+{
+  local -n _config_options="$1"
```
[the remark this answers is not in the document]
kwy95: I got this pattern from the `compare_command_sequence` function used in our tests. It is used because this is a reference.

rodrigosiqueira (changes requested):
Can we check if the new option is equal to the one that we already have set? For example, I had tls set, but the template option asked me again about it.

src/mail.sh
```diff
@@ -263,23 +285,13 @@ function mail_verify()
...
-  for config in "${essential_config_options[@]}"; do
-    if [[ ! "${!set_confs[*]} " =~ (local|global)_"$config"[[:space:]] ]]; then
-      missing_conf["$count"]="$config"
-      count=$((count + 1))
-    fi
-  done
+  mapfile -t missing_conf < <(missing_options 'essential_config_options')
 
-  if [[ "$count" -gt 0 ]]; then
+  if [[ "${missing_conf[0]}" -gt 0 ]]; then
```
rodrigosiqueira, line 301: fwiu, `missing_options` will return the index and fill out the target array. In this sense, I suppose that `${missing_conf[0]}` should have a value or maybe be empty. I did not get why you are comparing this value to 0? Is it correct?

src/mail.sh
```diff
@@ -130,7 +130,7 @@ function validate_email()
 # @curr_scope: The scope being edited
 # @cmd_scope:  The scope being imposed on the commands
 # @set_option: The relevant index to access set_confs
-# @scp:        Used to go through all scopes
+# @scope:      Used to go through all scopes
```
rodrigosiqueira: Instead of "all scopes" use "local and global scope".

src/mail.sh
```diff
@@ -126,39 +126,39 @@ function validate_email()
 # the configuration
 #
 # @option:     The option being edited
-# @value:      The value to update the option to
 # @curr_scope: The scope being edited
 # @cmd_scope:  The scope being imposed on the commands
-# @set_option: The relevant index to access set_confs
 # @scope:      Used to go through all scopes
+# @values:     Array to store relevant values for each option
```
rodrigosiqueira: I think this documentation does not match the function behavior; I cannot see value getting any parameter.

src/mail.sh
```diff
@@ -167,13 +167,32 @@ function check_add_config()
...
+function config_values()
+{
+  local -n _values="$1"
+  local option="$2"
+  local scope
+
+  for scope in {'global','local'}; do
+    _values["$scope"]="${set_confs["${scope}_$option"]}"
+  done
+  _values['loaded']="${options_values["$option"]}"
```
rodrigosiqueira, line 184: Add a new line after `done`.

src/mail.sh
```diff
@@ -76,6 +76,13 @@ function mail_setup()
     config_values 'values' "$option"
 
     if [[ -n "${values['loaded']}" ]]; then
+      if [[ "$option" =~ 'user.email'|'sendemail.smtpuser' ]]; then
+        validate_email "${values['loaded']}" || continue
+      fi
+      if [[ "$option" == 'sendemail.smtpencryption' ]]; then
```
rodrigosiqueira, line 91: New line between `fi` and `if`

src/mail.sh
```diff
@@ -450,6 +553,7 @@ function parse_mail_options()
   options_values['FORCE']=0
   options_values['VERIFY']=0
   options_values['TEMPLATE']=''
+  options_values['INTERACTIVE']=0
```
rodrigosiqueira, line 565: Can we set `''` instead of `0` and use `-n` or `-z` to check this value in the first part of this commit?

src/mail.sh
```diff
@@ -427,13 +432,111 @@ function load_template()
...
 }
 
+# This function prompts the user for the config values needed to setup the mail
+# capabilities.
+#
+# Returns:
+# 0 if successful; non-zero otherwise
+function interactive_setup()
+{
+  local flag="$1"
```
rodrigosiqueira, line 447: You did not document the `flag` option.

rodrigosiqueira (changes requested):
Hi @kwy95 ,

To me, the below code is the perfect setup:

```
kw mail --template="riseup" --email "another@email.com" --name "Another Email"
```

I love it :)

I think your code is great (I just added some tiny observations), and how you split your commits is excellent! Great job in this PR!

Nevertheless, I have to admit that I'm still not comfortable with the usability. For the below comments, I tried to put myself in a newcomer's foot.

First time user:
----------------

The first time that I used the command:

```
kw mail --template
```

I selected the Gmail option, and I got this question:

```
Do you wish to finish setting up interactively? [y/N]
...
Would you like to preload some value from a template? [y/N]
```

If I were a newcomer, I would not understand what else I needed to do; I would expect that you guide me through this process. In other words, I don't think you should ask here, just go ahead and ask for the missing information (we are guiding the user here). Maybe this path:

```
You may choose one of the following templates to start your configuration.
1) Gmail
2) Office365
3) Outlook
4) Riseup
5) Exit
#? 1
Gmail configuration:
smtpserver: smtp.gmail.com
smtpencryption: tls
smtpserverport: 587

Setup your name:
It looks like that you have:
[global] Juca da Silva

Do you want to set another name for this repository? [y/N]

Setup your email:
It looks like that you have:
[global] JucaDaSilva@xpto.la

Do you want to set another email for this repository? [y/N]
```

In my view, this is more friendly for a newcomer. What do you think?

Ask for replacing something for the same value:
---------------------------------------------

While I was playing with the template option, I noticed you do not recognize equal values. For example:

```
Editing [local]: sendemail.smtpserverport
  'sendemail.smtpserverport' is already set with the following value(s):
    [global]: 587
  It seems you already have a value ready to be applied: 587
  Do you wish to input a new value? [y/N]
```

What is the point to botter the user with this question? Just skip this.

Make template parameter case-insensitive
----------------------------------------

I notice that the template is case-sensitive, which is not good from the user perspective:

kw mail --template="Riseup" --email "another@email.com"
Invalid template: Riseup

Can we make it case-insensitive? Also, handle extra spaces in the beginning and the end.

Ask for preload template
------------------------

I noticed the following question after I was asked about the template that I wanted to use:

```
Would you like to preload some value from a template? [y/N]
```

Imho, it does not make sense to ask it later if I already told you which template I want to use. In other words, if the person is using the template, you can assume that they want to use the template that they choose.

Create a tutorial page
----------------------

Could you start a tutorial page? It would be nice to provide a small tutorial that guides the user through the basic setup. Try to write it in a friendly way for a newcomer, which means that you can decide to provide the best patch in your view.

src/mail.sh
kwy95: Currently `${missing_conf[0]}` stores the number of missing configs, but I suppose I could change it to check for the number of arguments or if it's empty.

src/mail.sh
kwy95: `values` is passed to the `config_options` function that fills it with the values that are set in each scope and the value loaded in `options_values` for the given config

src/mail.sh
kwy95: We can, I used `0` and `1` because this is just a flag, but I think I'll change this a bit and maybe it'll make more sense as an empty string

kwy95:
This PR is getting out of hand in size, I hope this last update is the last one that changes the functioning of this feature.
Something that I think can improve the user friendliness of the interactive setup is the line loading from #519, with it we could add short explanations to each configuration.

src/mail.sh
```diff
@@ -388,25 +662,45 @@ function parse_mail_options()
...
         shift
         ;;
-      --email | --name)
-        validate_setup_opt "$1" || exit 95 # ENOTSUP
+      --email)
+        if [[ -z "${options_values['sendemail.smtpuser']}" ]]; then
+          options_values['sendemail.smtpuser']="$2"
+          smtpuser_autoset=1
+        fi
+        ;&
```
rodrigosiqueira, line 670: iirc case has different types of stops, is this one of them? Or is it a bug?

src/mail.sh
```diff
@@ -315,24 +353,35 @@ function mail_list()
...
+        printf '    [global: %s]' "${values['global']}"
+        la=1
       fi
     fi
-    printf '%s\n' ''
+    if [[ -n "${values['loaded']}" ]]; then
+      if [[ "$la" -gt 0 ]]; then
+        printf ', [loaded: %s]' "${values['loaded']}"
+      else
+        printf '    [loaded: %s]' "${values['loaded']}"
```
rodrigosiqueira, line 381: It looks like that you are not testing this line.

src/mail.sh
```diff
@@ -341,23 +390,245 @@ function print_configs()
...
+        prompt+=" [default: ${values['global']}]"
+      fi
+
+      while true; do
+        if [[ "$option" == 'sendemail.smtpuser' && "$smtpuser_autoset" == 1 ]]; then
+          warning "  kw will set this option to ${values['loaded']}."
+          if [[ "$(ask_yN "  Do you want to change it?")" == 0 ]]; then
+            printf '\n'
+            break
+          fi
```
rodrigosiqueira, line 564: It looks like that you are not testing this part.

rodrigosiqueira (changes requested):

src/mail.sh
kwy95: Yes, this is the "Also execute the next case". This way the code shared by `name` and `email` still runs.

rodrigosiqueira: Since this is not common on kw code base, could you just add one comment next to it?

rodrigosiqueira:
Hi @kwy95 , could you rebase this PR in the latest code from the unstable branch?

rodrigosiqueira:
Merged, thanks a lot for all your effort here.

Issue #510, opened 2021-11-13
`-` in path makes includes fail


isinyaaa:
This is a follow up to #473. It is almost the same problem only with `-` characters instead of dots.

Output:
```sh
src/kw_include.sh: line 34: declare: `_HOME_ISINYAAA_PROJECTS_KWORKFLOW-GIT_SRC_KWORKFLOW_SRC_KWIO_SH_IMPORTED=1': not a valid identifier
src/kw_include.sh: line 34: declare: `_HOME_ISINYAAA_PROJECTS_KWORKFLOW-GIT_SRC_KWORKFLOW_SRC_KWLIB_SH_IMPORTED=1': not a valid identifier
./setup.sh: line 402: say: command not found
./setup.sh: line 49: detect_distro: command not found
./setup.sh: line 64: warning: command not found
./setup.sh: line 65: warning: command not found
./setup.sh: line 66: warning: command not found
./setup.sh: line 409: say: command not found
```

**Note:** This is needed in order to package kw on the AUR, as the package is called "kworkflow-git".

PR #511, opened 2021-11-13
src: kw_include: fix dashed variables naming

isinyaaa:
This fixes `bash`'s declare function errors that occur when the included path has dashes on it (see #510).


kwy95 (approved):
LGTM.

rodrigosiqueira:
Merged, thanks

Issue #500, opened 2021-10-26
Differentiate includes more robustly


isinyaaa:
As per @kwy95's suggestion, multiple characters are being substituted by underscores in `kw_include.sh`'s included paths, which may lead to various paths having the same variable name. So some method for differentiating those paths is necessary, among the suggested ones is some sort of hashing.

> It also dawned on me that since we are substituting all invalid characters with underscores we might eventually have different files that produce the same variable name. A potential workaround for this is to add some sort of hash based on the full path to the end of the variable name. But this might increase the overhead too much, so maybe we should just add this PR once it gets updated and worry about this hashing business later.

_Originally posted by @kwy95 in https://github.com/kworkflow/kworkflow/pull/474#pullrequestreview-787548192_

It might also be interesting to notice the security implication of different versions of `sha`: _https://www.keycdn.com/support/sha1-vs-sha256_

Also, notice that these versions don't have much bearing on runtime:

![image](https://user-images.githubusercontent.com/39812919/138974005-4bcb31c6-874a-42f3-9c1f-d638ce652809.png)

isinyaaa:
It's interesting to notice that the **very weak** `sum` command is actually really fast, so it might be interesting for our particular use case so that we can maintain kw responsive.

![image](https://user-images.githubusercontent.com/39812919/141871818-9079c12e-cf59-4dc4-a3ee-abb977ec01df.png)

@kwy95 @rodrigosiqueira comments?

rodrigosiqueira:
Thanks a lot for this great PR!
Merged to the unstable branch.

PR #514, opened 2021-11-14
Implement include hashing

isinyaaa:
As per #500, implement include hashing so that kw can differentiate its included files more robustly.

Also performs more efficiently :)

### DIFF
--- a/src/kw_include.sh
+++ b/src/kw_include.sh
@@ -8,7 +8,7 @@
 function include()
 {
   local filepath="$1"
-  local varname
+  local fullpath
 
   if [[ ! -r "$filepath" ]]; then
     if [[ ! -e "$filepath" ]]; then
@@ -19,21 +19,16 @@ function include()
     return 1 # EPERM
   fi
 
-  varname="$(realpath "$filepath")"
-  varname="${varname#"$KW_LIB_DIR/"}" # leave path until KW_LIB_DIR
-  varname="${varname//\//_}"          # change bars to underlines
-  varname="${varname//\./_}"          # change dots into underscores
-  varname="${varname// /_}"           # change spaces into underscores
-  varname="${varname//-/_}"           # change dashes into underscores
-  varname="${varname^^}_IMPORTED"     # capitalize and append "_IMPORTED"
+  fullpath="$(realpath "$filepath")"
 
-  if [[ -v "${varname}" ]]; then
-    return 0
+  if [[ -v KW_INCLUDES_SET ]]; then
+    [[ -v KW_INCLUDED_PATHS["$fullpath"] ]] && return 0
+
+    KW_INCLUDED_PATHS["$fullpath"]=1
+  else
+    declare -g KW_INCLUDES_SET=1
+    declare -gA KW_INCLUDED_PATHS=(["$fullpath"]=1)
   fi
-
-  declare -g "${varname}"=1
-
-  [[ "$?" != 0 ]] && return 22 # EINVAL
 
   . "$filepath" --source-only
 }
--- a/src/kw_include.sh
+++ b/src/kw_include.sh
@@ -10,12 +10,11 @@ function include()
   local filepath="$1"
   local varname
 
-  if [[ ! -e "$filepath" ]]; then
-    printf '%s\n' "File $filepath could not be found, check your file path."
-    return 2 # ENOENT
-  fi
-
   if [[ ! -r "$filepath" ]]; then
+    if [[ ! -e "$filepath" ]]; then
+      printf '%s\n' "File $filepath could not be found, check your file path."
+      return 2 # ENOENT
+    fi
     printf '%s\n' "File $filepath could not be read, check your file permissions."
     return 1 # EPERM
   fi
--- a/tests/kw_include_test.sh
+++ b/tests/kw_include_test.sh
@@ -7,37 +7,22 @@ function oneTimeSetUp()
   # specifies where to source the lib
   KW_LIB_DIR="$PWD/src"
 
-  # we'll use these folders to run our tests
-  KW_INC_TEST_PATH="$SHUNIT_TMPDIR/inc_test"
-  KW_INC_TEST_SPACED_PATH="$SHUNIT_TMPDIR/inc test"
-  KW_INC_TEST_HIDDEN_PATH="$SHUNIT_TMPDIR/.inc_test"
-  mkdir "$KW_INC_TEST_PATH" \
-    "$KW_INC_TEST_SPACED_PATH" \
-    "$KW_INC_TEST_HIDDEN_PATH"
-
-  # those test files can be empty, as we're only testing the created varname
+  # this test file can be empty, as we're only testing the created varname
   test_files=(
-    "$KW_INC_TEST_HIDDEN_PATH/include_hidden.sh"
-    "$KW_INC_TEST_HIDDEN_PATH/include_test.sh"
-    "$KW_INC_TEST_SPACED_PATH/include_spaced.sh"
-    "$KW_INC_TEST_PATH/include_test.sh"
-    "$KW_INC_TEST_PATH/include-dashed.sh"
+    "$SHUNIT_TMPDIR/.hidden spaced-dashed include.sh"
+    "$SHUNIT_TMPDIR/include_hash"
   )
 
   touch "${test_files[@]}"
 
   # the next files will be checked for name collisions
   printf "%s\n" "function test1(){ printf '%s\n' 'output of test1';}" > \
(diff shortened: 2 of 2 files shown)


src/kw_include.sh
```diff
@@ -1,4 +1,17 @@
...
+{
+  local varname="$1"
+
+  varname="${varname#"$KW_LIB_DIR/"}" # leave path until KW_LIB_DIR
+  varname="${varname//\//_}"          # change bars to underlines
+  varname="${varname//\./_}"          # change dots into underscores
+  varname="${varname// /_}"           # change spaces into underscores
+  varname="${varname^^}_IMPORTED"     # capitalize and append "_IMPORTED"
+
+  printf "%s" "$varname"
```
kwy95, line 12: Use single quotes around the formatting string fir `printf`

tests/utils.sh
```diff
@@ -306,8 +307,10 @@ function invoke_shunit()
 {
   command -v shunit2 > /dev/null
   if [[ "$?" -eq 0 ]]; then
+    # shellcheck source=/dev/null
```
kwy95, line 310: Were you getting errors here? I think we excluded this error in our CI.

src/kw_include.sh
```diff
@@ -23,6 +23,8 @@ function include()
   local filepath="$1"
   local fullpath
   local varname
+  local hashname
+  local hash
```
kwy95, line 27: `hash` is a shell builtin, I recommend you change the name of this variable. Maybe `hash_str`?

src/kw_include.sh
```diff
@@ -37,14 +39,21 @@ function include()
   fi
 
   varname="$(get_varname "$fullpath")"
+  hashname="${varname%"_IMPORTED"}_HASH"
+  hash="$(echo "$fullpath" | sha512sum)"
```
kwy95, line 43: Don't use `echo`

src/kw_include.sh
```diff
@@ -37,14 +39,21 @@ function include()
   fi
 
   varname="$(get_varname "$fullpath")"
+  hashname="${varname%"_IMPORTED"}_HASH"
```
kwy95, line 42: Won't this give us problems if we have a third+ file with the same name?

src/kw_include.sh
```diff
@@ -37,14 +39,21 @@ function include()
...
+    [[ "${!hashname}" == "$hash" ]] &&
     return 0
-  fi
-
-  declare -g "${varname}"=1
 
+  declare -g "$varname"=1
   [[ "$?" != 0 ]] && return 22 # EINVAL
 
+  declare -g "$hashname"="$hash"
```
kwy95, line 52: Isn't this gonna recreate this variable just with a different value?

kwy95 (changes requested):
Thanks for this change, but I think the current implementation might still be susceptible to duplicates. From what I understood you are creating a secondary variable to store the hash corresponding to each file, but the name of this secondary variable would be the same for conflicting files, so the hash would be overwritten. I think this would allow two files to be included multiple times if you keep alternating them.
The first thought I had about solving #500 was to add the first 6 to 8 characters from the hash to the beginning or end of the variable name, this way we'd have unique variable names for all files.
Let me know if I understood something incorrectly or any other thoughts you might have.

src/kw_include.sh
isinyaaa: you're quite right on that! I've actually spent a little time thinking about that since I committed this, and now I think this might be sorted :)

tests/utils.sh
isinyaaa: Oh, I'm sorry, this was pretty naïve to be honest, I just updated the docs but had the out-of-date version of that .pre-commit-config.yaml locally

rodrigosiqueira (changes requested):
Just a few additional comments.

isinyaaa:
> Thanks for this change, but I think the current implementation might still be susceptible to duplicates. From what I understood you are creating a secondary variable to store the hash corresponding to each file, but the name of this secondary variable would be the same for conflicting files, so the hash would be overwritten. I think this would allow two files to be included multiple times if you keep alternating them.

Since I've committed this I was kinda worried about this sort of behavior and I think you're right, actually.

> The first thought I had about solving #500 was to add the first 6 to 8 characters from the hash to the beginning or end of the variable name, this way we'd have unique variable names for all files.

This sounds like a nice idea, but since Siqueira mentioned your worry with keeping kw responsive I've actually thought of something that utilizes bash 4's implementation of [hash tables](https://www.gnu.org/software/bash/manual/html_node/Arrays.html), which then keeps our implementation fairly similar to what it is and is fast enough so that (I hope) no one notices...

> Let me know if I understood something incorrectly or any other thoughts you might have.

Thanks :)

Also check out this little guide for using those associative arrays:
[1] - _https://www.artificialworlds.net/blog/2012/10/17/bash-associative-array-examples/_

isinyaaa:
Also, as for 823cddc, I added it to this PR only because I'd commented on it with @rodrigosiqueira  earlier today, but I can open another PR if you think this is more appropriate.

src/kw_include.sh
```diff
@@ -1,4 +1,27 @@
...
+function get_varname()
+{
+  local fullpath="$1"
+  local varname
+
+  varname="${fullpath#"$KW_LIB_DIR/"}" # leave path until KW_LIB_DIR
+  varname="${varname//\//_}"           # change bars to underlines
+  varname="${varname//\./_}"           # change dots into underscores
+  varname="${varname// /_}"            # change spaces into underscores
+  varname="${varname//-/_}"            # change dashes into underscores
```
kwy95, line 18: I wonder, can't we do all of this at once? Maybe with `sed`, I think it could look something like this: `varname="$(printf '%s\n' "$varname" | sed -e 's/[/.[:space:]-]/_/g')"`
or even: `varname="$(printf '%s\n' "$varname" | sed -e 's/[^[:alnum:]_]/_/g')"`
I think I prefer my second suggestion, as it will replace any characters that are not allowed as variable names to underscores. The only potential problem would be a `varname` starting with a number, but we don't currently deal with that.

src/kw_include.sh
```diff
@@ -42,14 +43,19 @@ function include()
...
+    [[ -v KW_INCLUDED_PATHS["$fullpath"] ]] && return 0
 
-  declare -g "${varname}"=1
+    KW_INCLUDED_PATHS["$fullpath"]=1
+  else
+    declare -g KW_INCLUDES_SET=1
+    declare -gA KW_INCLUDED_PATHS=(["$fullpath"]=1)
+  fi
 
+  declare -g "$varname"=1
```
kwy95, line 58: With this new implementation, do we even need `varname`?

kwy95 (changes requested):
This is evolving well. As I mention in one of my comments, I think using an associative array may be able to solve all of the problems we've been having and I think we could potentially get rid of `varname` entirely.

src/kw_include.sh
isinyaaa: You have a really good point :) I think we don't!

isinyaaa: @kwy95 what tests should we keep in this new implementation? I thought of only keeping these three:

image

src/kw_include.sh
isinyaaa: This is a really good suggestion, though if we are removing varname I don't see the point in keeping this function anyway.

src/kw_include.sh
kwy95: I don't think we can have too many tests. I especially would like to keep the tests that deal with problems we've had before, this way we can avoid regressions in the future.

src/kw_include.sh
```diff
@@ -8,6 +8,7 @@
 function include()
 {
   local filepath="$1"
+  local fullpath
   local varname
```
kwy95, line 12: You can remove this now.

tests/kw_include_test.sh
```diff
@@ -2,55 +2,20 @@
...
 
-function oneTimeSetUp()
+function test_include_hash()
 {
-  # specifies where to source the lib
-  KW_LIB_DIR="$PWD/src"
+  local fullpath
+  local output
+  local relpath
+  local varname
```
kwy95, line 10: Not used

kwy95 (changes requested):
Minor changes only. I do think we should keep our previous tests as they don't even need much refactoring, and would help avoid regressions.

src/kw_include.sh
isinyaaa: @kwy95 any suggestions on effective name collision testing?

rodrigosiqueira:
Thanks a lot for this great PR!

PR #515, opened 2021-11-14
Fix `test_get_config_from_proc` local test

### LINKED ISSUE #499: `test_get_config_from_proc` failing on some systems
**Describe the bug**
When running tests the `test_get_config_from_proc` fails.

**To Reproduce**
1. clone the repo
2. `cd` into the cloned path
3. run `./run_tests.sh`

**Expected behavior**

**Screenshots**
image

 - OS: Arch Linux
 - Version: Still happens as of latest unstable version [cdf6cd8](cdf6cd8a8c419949cd85cc63b2220f9f1719ad58).
 - Bash Version: 5.1.008-1

**Additional context**
> I suspect that `PROC_CONFIG_PATH='/proc/config.gz'` in the function `get_config_from_proc` is checking your machine filesystem, which means that the test has a bug since it should never rely on anything in the dev system.

See:
_https://github.com/kworkflow/kworkflow/issues/474#issuecomment-951396950_

isinyaaa:
As not all Linux machines are configured the same, set up expectations according to the actual host kernel config.

Fixes #499.

### DIFF
--- a/tests/configm_test.sh
+++ b/tests/configm_test.sh
@@ -474,12 +474,17 @@ function test_get_config_from_proc()
   assert_equals_helper 'proc/config.gz is not supported for VM' "$LINENO" "$?" 95
 
   # 2) Local
-  declare -la expected_cmd=(
-    'sudo modprobe -q configs && [ -s /proc/config.gz ]'
-    'zcat /proc/config.gz > .config'
-  )
-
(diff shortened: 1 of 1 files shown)


isinyaaa:
Will gather test fixes in #516.

Issue #484, opened 2021-10-09
Broken pipe message in the CI


rodrigosiqueira:
**Describe the bug**
In our CI, we have errors like this:

```
./src/kwio.sh: line 73: printf: write error: Broken pipe
./src/kwio.sh: line 73: printf: write error: Broken pipe
./src/kwio.sh: line 73: printf: write error: Broken pipe
./src/kwio.sh: line 73: printf: write error: Broken pipe
```
Ref: https://github.com/kworkflow/kworkflow/pull/483/checks?check_run_id=3848328876#step:5:179

**To Reproduce**
1. Push any change
2. Take a look at the CI log

**Expected behavior**
We do not expect any error messages.

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04.3 LTS

**Additional context**
This only happens in the CI; nobody can reproduce it locally.

PR #516, opened 2021-11-14
fix broken tests

isinyaaa:
This attemps at fixing the broken pipe error that occurs in the project's CI [1], the broken local config manager test and also `tput`'s behavior in a non-configured environment [2].

Attempted fixes:
[1] - _https://superuser.com/questions/554855/how-can-i-fix-a-broken-pipe-error_
[2] - _https://stackoverflow.com/questions/48536791/error-tput-no-value-for-term-and-no-t-specified_

Fixes: #484, #499, #517.

### DIFF
--- a/src/diff.sh
+++ b/src/diff.sh
@@ -106,7 +106,10 @@ function diff_side_by_side()
   local diff_cmd
   local columns
 
-  columns=$(tput cols)
+  # In case TERM is gibberish tput won't work properly
+  # specify dummy terminal option to manage that
+  [[ "$TERM" == '' || "$TERM" == 'dumb' ]] && TPUTTERM=' -T xterm-256color'
+  columns=$(eval tput"${TPUTTERM}" cols)
   diff_cmd="diff -y --color=always --width=$columns"
   flag=${flag:-''}
 
--- a/tests/build_test.sh
+++ b/tests/build_test.sh
@@ -68,7 +68,7 @@ function test_kernel_build()
   ret="$?"
   assertEquals "($LINENO)" "$ret" "22"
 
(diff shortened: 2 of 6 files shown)


tests/build_test.sh
```diff
@@ -68,7 +68,7 @@ function test_kernel_build()
   ret="$?"
   assertEquals "($LINENO)" "$ret" "22"
 
-  output=$(kernel_build 'TEST_MODE' | head -1) # Remove statistics output
+  output=$(kernel_build 'TEST_MODE' | tail -n +1 | head -1) # Remove statistics output
```
kwy95, line 71: Please add a commit message.

src/diff.sh
```diff
@@ -106,7 +106,8 @@ function diff_side_by_side()
   local diff_cmd
   local columns
 
-  columns=$(tput cols)
+  [[ "$TERM" == '' || "$TERM" == 'dumb' ]] && TPUTTERM=' -T xterm-256color'
```
kwy95, line 109: Please add a commit message.

kwy95 (changes requested):
Thanks a lot for this patch, the unit tests output is much cleaner now. The only weird output we still get on our CI is from statistics not having permissions when trying to create files during `build_test`.
Please just add a small descriptive commit message to your commits apart from `Fixes #XYZ`.

rodrigosiqueira (changes requested):
Hi,
I just have a few comments.

kwy95 (approved):
LGTM

rodrigosiqueira:
Thanks a lot!

PR #518, opened 2021-11-16
Debug ftrace

rodrigosiqueira:


src/debug.sh
```diff
@@ -269,6 +278,233 @@ function event_trace()
...
+# A string with the final command is composed.
+function build_ftrace_command_string()
+{
+  local ftrace_syntax="$1"
+  local ftrace_disable="$2"
+  local char_repetition
+  local ftrace_filters
+  local ftrace_type
+  local cmd_disable_ftrace="echo 0 > $TRACING_ON"
+  local cmd_enable_ftrace="echo 1 > $TRACING_ON"
```
kwy95, line 441: Can't we use `printf` here? I don't know much about the kernel's debug feature, but if there are no limitations regarding this it would be better to stick to our codestyle.

kwy95 (changes requested):
Some comments about documentation, and as I mention if possible avoid using `echo`. But maybe there is something I don't understand about this feature that requires `echo` instead of `printf`, but if possible it would be best to change them.

src/debug.sh
```diff
@@ -269,6 +278,233 @@ function event_trace()
...
+}
+
+# This is the function responsible for manager ftrace tasks.
+#
+# @target Target can be 1 (VM_TARGET), 2 (LOCAL_TARGET), and 3 (REMOTE_TARGET)
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/kwlib.sh` function `cmd_manager`
+# @ftrace_syntax: Raw string with event syntax
+# @keep_history: If set to a value different from empty or 0, it will create a
+# directory structure for keeping the trace history. Otherwise, it will create
```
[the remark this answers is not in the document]
isinyaaa: I think for keeping is correct grammar here, actually.

src/debug.sh
rodrigosiqueira: I decided to use `echo` in this case because this will be used in the ssh command in the target machine. This can simplify the debug and also reduce the side of the command a little bit.

PR #519, opened 2021-11-17
Kw first setup

melloguel:
Creates an interactive setup mode for automated SSH configuration accessible by using the new option
```kw init --interactive``` or ```kw init -i``` for short.  This feature includes a quick explanation of the configurations
being made and is intended for newcomers to KW.

For this, we created a new module, interactive.sh, that contains a new function ```interactive_init()```. We added
a call to this function in init.sh that is called when ```-i``` or ```--interactive``` is passed as an option to ```kw init```.

We also added this information in the ```kw init -h``` and ```kw init --help``` prompts.

### CODE REVIEW

**Review by rodrigosiqueira — CHANGES_REQUESTED** [review]
Hi,

I did not review all the changes in the details because I think you need to rework your PR. Try this approach:

* Commit 1: Modify kernel tree check
* Commit 2: Add support to read long strings from a file.

I noticed that you have some big strings in this PR. Instead of having them as part of the code, let's create a mechanism to isolate long strings like this in a single file. This is my suggestion:
* Create a new folder under src. For example, `src/strings/init.txt`
* Define a syntax for this file, let's say:
[REFERENCE_NAME]:
TEXT TEXT TEXT
In other words, `\[\w*\]:` is how you mark a key.
* Create a new function in the kwio or kw_lib that looks like this:

```bash
declare -gA string_file

function load_string()
{
  local path="$1"
  local key
  local capture data

  # First lie
  key=$(head -1 "$path" | grep -o '\w*')

  while read line; do
    if [[ "$line" =~ \[.*\]: ]]; then
      string_file["$key"]="$string"
      string=''
      key=$(echo "$line" | grep -o '\w*')
      continue
    fi
    string+="$line"
  done < "$path"

  # last element
  string_file["$key"]="$string"
}

load_string 'strings/test.txt'
for x in "${!string_file[@]}"; do
  echo ":----:"
  echo "=> key: "$x""
  echo "=> value: ${string_file["$x"]}"
done

```
( This is just a draft to give some ideas, please make something better than that)

* Commit 3: Create a function that checks if ssh is set in the target machine and another function that creates keys.
* Commit 4: Add interactive mode that checks ssh key in the local machine

src/interactive.sh
```diff
@@ -0,0 +1,151 @@
...
+If you press enter without typing anything, we will
+use the default values: root for user and localhost for ip.
+EOF
+
+function interactive_init()
+{
+  distro=$(detect_distro '/')
+
+  say "$text_interactive_start"
+  printf "\n"
```
kwy95, line 54: Single quotes

src/interactive.sh
```diff
@@ -0,0 +1,151 @@
...
+  read -r -p '>>> Press ENTER to continue'
+
+  say "$text_ssh_start"
+  printf "\n"
+  read -r -p '>>> Press ENTER to continue'
+
+  case "$distro" in
+    none)
+      warning "We do not support your distro (yet). We cannot check if SSH is installed."
```
kwy95, line 63: Single quotes

src/interactive.sh
```diff
@@ -0,0 +1,151 @@
...
+
+    ssh_keys=$(find "$HOME/.ssh" -type f -name "*.pub")
+
+    if [[ -n "$ssh_keys" ]]; then
+      say "$text_ssh_keys"
+    fi
+
+    #TODO: Fix cases and make code more efficient
+    select item in "Create new LOCAL RSA key for KW" "Create new GLOBAL RSA key for KW" $ssh_keys; do
+
```
kwy95, line 103: You can remove this extra line.

src/interactive.sh
```diff
@@ -0,0 +1,151 @@
+text_interactive_start=''
```
kwy95, line 1: Add a comment to the top of this file describing it's purpose and use.

src/init.sh
```diff
@@ -45,134 +46,7 @@ function init_kw()
...
-        read -r -p 'IP  : ' ssh_ip
-        options_values['REMOTE']="${ssh_user:-root}@${ssh_ip:-127.0.0.1}:22"
-      fi
-    fi
-    #TODO: Help to setup git;
-
-    #TODO: Help to setup kworkflow.config;
-
-    #TODO: Check for required;
+    interactive_init
```
kwy95, line 49: This commit should be squashed with the previous one.

kwy95 (changes requested):
Left a few comments with extra changes to consider on top of @rodrigosiqueira 's suggestions.

melloguel:
> I noticed that you have some big strings in this PR. Instead of having them as part of the code, let's create a mechanism to isolate long strings like this in a single file. This is my suggestion:
> 
>     * Create a new folder under src. For example, `src/strings/init.txt`
> 
>     * Define a syntax for this file, let's say:
>       [REFERENCE_NAME]:
>       TEXT TEXT TEXT
>       In other words, `\[\w*\]:` is how you mark a key.
> 
>     * Create a new function in the kwio or kw_lib that looks like this:
> 
> 
> ```shell
> declare -gA string_file
> 
> function load_string()
> {
>   local path="$1"
>   local key
>   local capture data
> 
>   # First lie
>   key=$(head -1 "$path" | grep -o '\w*')
> 
>   while read line; do
>     if [[ "$line" =~ \[.*\]: ]]; then
>       string_file["$key"]="$string"
>       string=''
>       key=$(echo "$line" | grep -o '\w*')
>       continue
>     fi
>     string+="$line"
>   done < "$path"
> 
>   # last element
>   string_file["$key"]="$string"
> }
> 
> load_string 'src/strings/init.txt'
> for x in "${!string_file[@]}"; do
>   echo ":----:"
>   echo "=> key: "$x""
>   echo "=> value: ${string_file["$x"]}"
> done
> ```
> 
> ( This is just a draft to give some ideas, please make something better than that)

We came up with this:

```BASH
declare -gA string_file

function load_string()
{
  local path="$1"
  local key=''
  local capture
  local data
  local nl=0
  local error=0

  while read line; do ((nl++))
    if [[ "$line" =~ \[(.*)\]: ]]; then
      key=$(echo "$line" | grep -o -E '\w+')
      if [[ -z "$key" ]]; then
        complain "[ERROR]:$path:$nl: keys should be alphanum chars" && error=1
      else
          string_file["$key"]=''
      fi
    elif [[ -n "$key" ]]; then
      string_file["$key"]+=$'\n'"$line"
    fi
  done < "$path"

  if [[ "${#string_file[@]}" -eq 0 ]]; then
    error=2
    complain "[ERROR]:$path: no key found"
  fi

  if [[ $error -ne 0 ]]; then
    exit $error
  fi
}

load_string "$1" #'src/strings/init.txt'
for x in "${!string_file[@]}"; do
  printf ":----:\n"
  printf "=> key: %s\n" "$x"
  printf "=> value:%s\n" "${string_file["$x"]}"
done
```

src/kwio.sh
```diff
@@ -119,3 +122,52 @@ function ask_yN()
...
+# Return: In case of success, returns zero. If a key is not found, exits 2.
+# If an invalid key is found, exits 1 and prints the line with a bad key.
+
+function load_string()
+{
+  local path="$1"
+  local key=''
+  local capture
+  local data
+  local nl=0
```
rodrigosiqueira, line 148: Let's find a better name for `nl`

src/kwio.sh
```diff
@@ -119,3 +122,52 @@ function ask_yN()
...
+
+# KEY must be non-empty, alphanumeric and between square brackets followed by a
+# colon. The global array string_file can then be queried by key as in
+# string_file[KEY] to obtain the string. load_string must receive the
+# full path of the text file (kept by default in src/strings) as the
+# first argument.
+
+# Return: In case of success, returns zero. If a key is not found, exits 2.
+# If an invalid key is found, exits 1 and prints the line with a bad key.
+
```
rodrigosiqueira, line 141: Drop this extra line.

src/kwio.sh
```diff
@@ -119,3 +122,52 @@ function ask_yN()
...
+# [KEY]:
+# text
+
+# KEY must be non-empty, alphanumeric and between square brackets followed by a
+# colon. The global array string_file can then be queried by key as in
+# string_file[KEY] to obtain the string. load_string must receive the
+# full path of the text file (kept by default in src/strings) as the
+# first argument.
+
+# Return: In case of success, returns zero. If a key is not found, exits 2.
```
rodrigosiqueira, line 139: Add a new line after `:`

src/kwio.sh
```diff
@@ -119,3 +122,52 @@ function ask_yN()
...
   fi
 }
+
+# load strings from file into a dictionary.
+# This function requires a key before a body of text to
+# name that particular body of text, as in this example:
+
+# [KEY]:
+# text
+
```
rodrigosiqueira, line 132: Add documentation for `@path`.

src/kwio.sh
```diff
@@ -119,3 +122,52 @@ function ask_yN()
...
+function load_string()
+{
+  local path="$1"
+  local key=''
+  local capture
+  local data
+  local nl=0
+  local error=0
+
```
rodrigosiqueira, line 150: Let's validate the `$path` here.

src/kwio.sh
```diff
@@ -119,3 +122,52 @@ function ask_yN()
...
+  while read -r line; do
+    ((nl++))
+    if [[ "$line" =~ \[(.*)\]: ]]; then
```
rodrigosiqueira, line 153: How about: `^\[(.*)\]:$`

src/kwio.sh
```diff
@@ -119,3 +122,52 @@ function ask_yN()
...
+      key=$(printf "%s" "$line" | grep -o -E '\w+')
```
rodrigosiqueira, line 154: replace `"` with `'`.

src/kwio.sh
```diff
@@ -119,3 +122,52 @@ function ask_yN()
...
+      if [[ -z "$key" ]]; then
+        complain "[ERROR]:$path:$nl: keys should be alphanum chars" && error=1
+      else
+        string_file["$key"]=''
+      fi
```
rodrigosiqueira, line 159: How about:
```sh
 if [[ -z "$key" ]]; then
        complain "[ERROR]:$path:$nl: keys should be alphanum chars" && error=1
        continue
fi
string_file["$key"]=''
```

src/kwio.sh
```diff
@@ -119,3 +122,52 @@ function ask_yN()
...
+    elif [[ -n "$key" ]]; then
+      string_file["$key"]+=$'\n'"$line"
```
rodrigosiqueira, line 161: Can we find a better name for this?

src/kwio.sh
```diff
@@ -119,3 +122,52 @@ function ask_yN()
...
+# KEY must be non-empty, alphanumeric and between square brackets followed by a
+# colon. The global array string_file can then be queried by key as in
+# string_file[KEY] to obtain the string. load_string must receive the
+# full path of the text file (kept by default in src/strings) as the
+# first argument.
+
+# Return: In case of success, returns zero. If a key is not found, exits 2.
+# If an invalid key is found, exits 1 and prints the line with a bad key.
+
+function load_string()
```
rodrigosiqueira, line 142: Can we find a better name?

rodrigosiqueira (changes requested):

EduBrancher:
@rodrigosiqueira We just finished the changes requested to our function and the commit tree. Please take a look when you are able

src/kwio.sh
```diff
@@ -1,3 +1,5 @@
+# Associative array to read strings from files
+
```
rodrigosiqueira, line 2: Why do we have this comment here?

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
     printf '%s\n' '0'
   fi
 }
+
+# load text used in the module from a file into a dictionary.
```
rodrigosiqueira, line 125: Could you rephrase this? It is a little bit confusing.

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
...
+# This function requires a key before a body of text to
+# name that particular body of text, as in this example:
+
+# [KEY]:
+# text
+
+# @path the full path of the text file to be read
+# as the first argument.
```
rodrigosiqueira, line 133: Add a couple of spaces before `as` until it gets aligned with `the`.

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
...
 }
+
+# load text used in the module from a file into a dictionary.
+# This function requires a key before a body of text to
+# name that particular body of text, as in this example:
+
+# [KEY]:
+# text
+
+# @path the full path of the text file to be read
```
rodrigosiqueira, line 132: maybe replace `read` by `load`

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
...
+# as the first argument.
+# @reset optional parameter. If not null, will reset the dictionary.
```
rodrigosiqueira, line 134: Maybe you can add a phrase that describes a use case for this parameter.

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
...
+
+# KEY must be non-empty, alphanumeric and between square brackets followed by a
+# colon. The global array string_file can then be queried by key as in
+# ${string_file[KEY]} to obtain the string. KEY should be named according
+# to its respective module for compatibility. That is, if we have modules A and B,
+# name A's keys as [KEY_A] and B's keys as [KEY_B]. This makes it so the modules
+# keys will be compatible with each other
+
```
rodrigosiqueira, line 142: Add `#`

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
...
+  if [[ -n "$reset" ]]; then
+    unset module_text_dictionary
+  fi
+
+  declare -gA module_text_dictionary
+
+  if ! [[ -f "$path" ]]; then
+    complain "[ERROR]:$path: Does not exist or is not a text file."
+    return 3
+  fi
```
rodrigosiqueira, line 167: By default, we try to use `errno` codes to manage errors. In this case, maybe you can use:
```
ENOENT 2 No such file or directory
```

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
...
+  local error=0
+  local key_set=0
+
+  if [[ -n "$reset" ]]; then
+    unset module_text_dictionary
+  fi
+
+  declare -gA module_text_dictionary
+
+  if ! [[ -f "$path" ]]; then
```
rodrigosiqueira, line 164: Why not add `!` inside `[[`?

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
...
+    complain "[ERROR]:$path: Does not exist or is not a text file."
+    return 3
+  fi
+
+  if ! [[ -s "$path" ]]; then
+    complain "[ERROR]:$path: File is empty."
+    return 4
+  fi
```
rodrigosiqueira, line 172: Maybe we can handle this error as `22` (EINVAL 22 Invalid argument)

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
...
+      key=$(printf '%s' "$line" | grep -o -E '\w+')
+      if [[ -z "$key" ]]; then
+        error=1
+        complain "[ERROR]:$path:$line_counter: keys should be alphanum chars"
+        continue
+      fi
+
+      if [[ -n "${module_text_dictionary[$key]}" ]]; then
+        warning "[WARNING]:$path:$line_counter: overwriting '$key' key."
+      fi
```
rodrigosiqueira, line 186: I think we don't need this check, or do you see any potential error? Anyway, I'm in favor of dropping this check.

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
...
+
+      key_set=1
+      module_text_dictionary["$key"]=''
+    elif [[ -n "$key" ]]; then
+      module_text_dictionary["$key"]+=$'\n'"$line"
+    fi
+  done < "$path"
+
+  if [[ "$key_set" -eq "0" ]]; then
```
rodrigosiqueira, line 195: Just use `0` (drop quotes)

src/kwio.sh
```diff
@@ -119,3 +121,81 @@ function ask_yN()
...
+    error=2
+    complain "[ERROR]:$path: no key found"
```
rodrigosiqueira, line 197: Maybe `ENOKEY 126 Required key not available`?

tests/kwio_test.sh
```diff
@@ -124,4 +124,26 @@ function test_alert_completion_sound_alert()
...
+
+  load_module_text "$path/file_without_key" "1"
+  assertEquals 'This file has no keys, this should return an error.' "2" "$?"
+
+  load_module_text "$path/file_does_not_exist_(do not create)" "1"
+  assertEquals 'This file does not exist, this should return an error.' "3" "$?"
+
+  load_module_text "$path/file_empty" "1"
+  assertEquals 'This file is empty, this should return an error.' "4" "$?"
+
```
rodrigosiqueira, line 146: Drop this new line.

tests/kwio_test.sh
```diff
@@ -124,4 +124,26 @@ function test_alert_completion_sound_alert()
   assertEquals 'Variable s should exist.' "$expected" "$output"
 }
 
+function test_load_module_text()
+{
+
```
rodrigosiqueira, line 129: Drop this extra line.

tests/kwio_test.sh
```diff
@@ -124,4 +124,26 @@ function test_alert_completion_sound_alert()
...
+  path="$PWD/tests/samples/load_module_text_test_samples/"
+
+  load_module_text "$path/file_correct" "1"
+  assertEquals 'Should work without any errors.' "0" "$?"
```
rodrigosiqueira, line 133: Use the actual value without `"`; this is the same for all of these tests.

src/interactive.sh
```diff
@@ -0,0 +1,40 @@
+function create_ssh_key()
```
rodrigosiqueira, line 1: I think this code should be inside `init.sh`, I don't see why we want a new file for handling this.

Document this function.

src/interactive.sh
```diff
@@ -0,0 +1,40 @@
...
+{
+  local ssh_dir="$1"
+  local key_name="${2:-kw_ssh_key}"
+
+  if [[ -n "$key_name" ]]; then
+    mkdir -p "$ssh_dir"
+    eval "ssh-keygen -t rsa -f $ssh_dir/$key_name"
```
rodrigosiqueira, line 8: Why do we need `eval`? If we really need to use `eval`, replace it with `cmd_manager`.

src/interactive.sh
```diff
@@ -0,0 +1,40 @@
...
+  fi
```
rodrigosiqueira, line 9: Don't we need to return something for validation?

src/interactive.sh
```diff
@@ -0,0 +1,40 @@
...
+}
+
+function check_ssh_installation()
```
rodrigosiqueira, line 12: Documentation...

src/interactive.sh
```diff
@@ -0,0 +1,40 @@
...
+{
+
```
rodrigosiqueira, line 14: drop this extra line.

src/interactive.sh
```diff
@@ -0,0 +1,40 @@
...
+  cmd=''
+  distro=$(detect_distro '/')
```
rodrigosiqueira, line 16: add `local`

src/interactive.sh
```diff
@@ -0,0 +1,40 @@
...
+
+  case "$distro" in
+    none)
+      complain "We do not support your distro (yet). We cannot check if SSH is installed."
```
rodrigosiqueira, line 20: Single-quotes.
Also, I think you should exit or return here.

src/interactive.sh
```diff
@@ -0,0 +1,40 @@
...
+      if [[ "$?" != 0 ]]; then
+        #not found
+        cmd='pacman -S openssh'
+      fi
+      ;;
+    debian)
+      installed=$(dpkg-query -W --showformat='${Status}\n' openssh-client 2> /dev/null | grep -c 'ok installed')
+      if [[ "$installed" -eq 0 ]]; then
+        #not found
+        cmd='apt install openssh-client'
```
rodrigosiqueira, line 33: We don't want to install anything inside `kw`, add these dependencies in the `setup.sh` and `documentation/dependencies/`.

src/init.sh
```diff
@@ -147,6 +159,10 @@ function parse_init_options()
         shift
         options_values['FORCE']=1
         ;;
+      --interactive | -i)
+        options_values['INTERACTIVE']=1
```
rodrigosiqueira, line 163: You need to set a default value before this loop. Set it as `''`.

src/init.sh
```diff
@@ -112,8 +124,8 @@ function set_config_value()
 
 function parse_init_options()
 {
-  local long_options='arch:,remote:,target:,force'
-  local short_options='a:,r:,t:,f'
+  local long_options='arch:,remote:,target:,force,interactive'
```
rodrigosiqueira, line 127: Add `--interactive` in the `src/bash_autocomplete.sh`.

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
   [[ -n "$cmd" ]]
 
 }
+
+function interactive_init()
```
rodrigosiqueira, line 42: Documentation...

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+{
+
```
rodrigosiqueira, line 44: Drop this extra line.

src/kwio.sh
```diff
@@ -1,3 +1,5 @@
+# Associative array to read strings from files
```
rodrigosiqueira, line 1: Commit message does not fully match with this new code.

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+  local cmd=''
+  local distro=''
+  local has_ssh=0
+  local ssh_keys
+  local key_name ssh_dir
+  local ssh_user ssh_ip
+
+  load_string "$KW_LIB_DIR/strings/init.txt"
```
rodrigosiqueira, line 52: You must handle the return of load_string because it can fail. Additionally, you changed this function name in the previous commit, i.e., load_string does not exist anymore :/
Did you test this PR?

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+
+  say "${string_file['text_interactive_start']}"
```
rodrigosiqueira, line 54: I suppose it should be `module_text_dictionary`?

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+
+  printf "\n"
```
rodrigosiqueira, line 56: Single-quotes...

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+  read -r -p '>>> Press ENTER to continue'
+
+  say "${string_file['text_ssh_start']}"
+  printf "\n"
+  read -r -p '>>> Press ENTER to continue'
```
rodrigosiqueira, line 61: hmmm... I would prefer something like `ask_yN`.

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+
+  check_ssh_installation
```
rodrigosiqueira, line 63: At this point, you can expect that the ssh package is available.

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+
+  if [[ -n "$cmd" ]]; then
```
rodrigosiqueira, line 65: You did not set `cmd`...

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+    if [[ $(ask_yN 'SSH was not found in this system, would you like to install it?') =~ '1' ]]; then
+      eval "sudo $cmd"
```
rodrigosiqueira, line 67: Do you really need `eval`? Why not use `cmd_manager`?

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+      has_ssh=$?
```
rodrigosiqueira, line 68: `"$?"`

rodrigosiqueira (changes requested):

src/interactive.sh
EduBrancher: Why is that? Before this point, we have not yet done anything related to ssh. Is this being dealt with in another PR?

src/interactive.sh
EduBrancher: Should we move all the code in interactive.sh to init.sh? We thought of interactive as a sort of auxiliary file to init.

src/interactive.sh
EduBrancher: This was originally intended by us to install ssh at this point. If we are handling it as part of KW's dependencies, then this code is not needed.

src/interactive.sh
rodrigosiqueira: Yes, move all the `interactive.sh` code inside `init.sh`.

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+    fi
+  fi
+
+  if [[ "$has_ssh" =~ "0" ]]; then
```
kwy95, line 72: You should either drop the quotes, or use single quotes

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+    say "${string_file['$text_ssh_found']}"
+    printf "\n"
```
kwy95, line 74: Single quotes

src/interactive.sh
```diff
@@ -38,3 +38,94 @@ function check_ssh_installation()
...
+
+    ssh_keys=$(find "$HOME/.ssh" -type f -name "*.pub")
+
+    if [[ -n "$ssh_keys" ]]; then
+      say "${string_file['text_ssh_keys']}"
+    fi
+
+    #TODO: Fix cases and make code more efficient
+    select item in "Create new LOCAL RSA key for KW" "Create new GLOBAL RSA key for KW" $ssh_keys; do
```
kwy95, line 83: Single quotes where appropriate

kwy95 (changes requested):
Some changes, @rodrigosiqueira already mentioned most of them.

src/kwio.sh
EduBrancher: We thought of using 61: ENODATA (No data available) since we thought an empty text file is technically a valid argument as it is a text file. Do you think that is acceptable, or should we use 22?

kwy95: Makes sense to me

Issue #179, opened 2020-05-15
Add verbose option


rodrigosiqueira:
Currently, kw deploy and other features throw a lot of information on users. Let's clean up the kw output and add a verbose option if the user wants extra information; this can be a huge task; for this reason, let's break it per feature:

- [x] Verbose mode for deploy;
- [ ] Silent mode for build option;

alan-barzilay:
I'm trying to understand this issue, do you mean stuff like this that is printed in every sub command that imports kwio?
```
/home/barzilay/.local//lib/kw/kwio.sh: line 3: declare: BLUECOLOR: readonly variable
/home/barzilay/.local//lib/kw/kwio.sh: line 4: declare: REDCOLOR: readonly variable
/home/barzilay/.local//lib/kw/kwio.sh: line 5: declare: YELLOWCOLOR: readonly variable
/home/barzilay/.local//lib/kw/kwio.sh: line 6: declare: GREENCOLOR: readonly variable
/home/barzilay/.local//lib/kw/kwio.sh: line 7: declare: SEPARATOR: readonly variable
```

rodrigosiqueira:
Ref: https://github.com/kworkflow/kworkflow/issues/857

PR #521, opened 2021-11-25
src: db: introduce db interface

kwy95:
In an effort to standardize our data collection and storage, as well as
improve scalability this introduces a library to create and interact
with `sqlite3` databases. This will eventually replace the files and
folders found in `KW_DATA_DIR` with a unified database.


etc/db_files/tables.sql
```diff
@@ -0,0 +1,19 @@
...
+	dt_end DATETIME NOT NULL,
+	descript VARCHAR(512)
+);
+
+CREATE TABLE IF NOT EXISTS statistic(
+	-- id INTEGER PRIMARY KEY,
+	label VARCHAR(32),
+	dt_start DATETIME NOT NULL,
+	dt_end DATETIME NOT NULL
+);
```
rodrigosiqueira, line 19: I don't think we want to have this file here at this stage. We will need to discuss the data structure later... for now, you can move it to `tests/samples/db_files` and use it for test only.

src/kw_db.sh
```diff
@@ -0,0 +1,83 @@
+include "$KW_LIB_DIR/kwio.sh"
```
rodrigosiqueira, line 1: Remember to add a simple documentation at the top of this file.

src/kw_db.sh
```diff
@@ -0,0 +1,83 @@
...
+include "$KW_LIB_DIR/kw_date_and_time.sh"
+
+DB_NAME='kw.db'
+DB_PATH="$KW_ETC_DIR/db_files/$DB_NAME"
```
rodrigosiqueira, line 5: Let's use `declare -g`.

src/kw_db.sh
```diff
@@ -0,0 +1,83 @@
+include "$KW_LIB_DIR/kwio.sh"
+include "$KW_LIB_DIR/kw_date_and_time.sh"
+
```
rodrigosiqueira, line 3: Let's convert this file in multiple commits with tests and documentation.

src/kw_db.sh
```diff
@@ -0,0 +1,83 @@
...
+  local sql_path="$1"
+  local db="${2:-$DB_PATH}"
+
+  if [[ ! -f "$sql_path" ]]; then
+    printf '%s\n' "Could not find the file: $sql_path"
+    return 22
+  fi
+
+  sqlite3 "$db" < "$sql_path"
+}
```
rodrigosiqueira, line 58: Commit 1: Add support to create a database by using an SQL file.
1. Find a better name for this function
2. Validate each parameter
3. Add documentation
4. Add unit test

src/kw_db.sh
```diff
@@ -0,0 +1,83 @@
...
+
+function apply_command()
+{
+  local cmd="$1"
+  local db="${2:-$DB_PATH}"
+
+  sqlite3 "$db" -batch "$cmd"
+}
```
rodrigosiqueira, line 66: Commit 2: Add support to run SQL commands
See previous comments.

rodrigosiqueira (changes requested):
Nice PR!
Could you break it in small set of commits followed by documentation and tests?
Thanks.

tests/kw_db_test.sh
```diff
@@ -70,11 +70,86 @@ function test_execute_sql_script()
...
+{
+  local output
+  local expected
+  local ret
+
+  cd "$FAKE_DATA" || {
+    ret="$?"
+    fail "($LINENO): Failed to move to fake data dir"
+    exit "$ret"
+  }
```
rodrigosiqueira, line 83: Again, I would prefer that we do not cd this folder.

tests/kw_db_test.sh
```diff
@@ -70,11 +70,86 @@ function test_execute_sql_script()
...
+
+  output=$(format_values_db 2 'first1' 'second1' 'first2' 'second2')
+  ret="$?"
+  expected="('first1','second1'),('first2','second2')"
+  assert_equals_helper 'No error expected' "$LINENO" 0 "$ret"
+  assert_equals_helper 'Wrong output' "$LINENO" "$output" "$expected"
+
+  output=$(format_values_db 1 "some 'quotes'")
+  ret="$?"
+  expected="('some ''quotes''')"
```
rodrigosiqueira, line 136: Don't we have one extra quote here?

rodrigosiqueira (changes requested):

rodrigosiqueira:
Btw, I don't think this is a draft anymore.

src/kw_db.sh
```diff
@@ -0,0 +1,30 @@
...
+{
+  local sql_path="$1"
+  local db="${2:-"$DB_NAME"}"
+  local db_folder="${3:-"$KW_DATA_DIR"}"
+  local db_path
+
+  db_path="$(join_path "$db_folder" "$db")"
+
+  file_exists "$sql_path" || exit_msg "Could not find the file: $sql_path" 2
+  file_exists "$db_path" || warning "Creating database: $db_path"
```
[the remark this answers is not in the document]
kwy95: I'm thinking of removing the message from that function and only check if the file exists and return accordingly. And now I think it might be easier to just use `[[ -f "$path" ]]`. What do you think?

src/kw_db.sh
```diff
@@ -28,3 +28,26 @@ function execute_sql_script()
...
+  local sql_cmd="$1"
+  local db="${2:-"$DB_NAME"}"
+  local db_folder="${3:-"$KW_DATA_DIR"}"
+  local db_path
+
+  db_path="$(join_path "$db_folder" "$db")"
+
+  file_exists "$db_path" || exit_msg 'Database does not exist' 2
+
+  sqlite3 "$db" -batch "$sql_cmd"
```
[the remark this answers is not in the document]
kwy95: I did some tests, and `sqlite3` prints a message to the terminal detailing the problem with the command. I could add the `-bail` flag, which ensures sqlite3 stops as soon as there is an error.
I think we could add a message that displays the command that caused the error, as `sqlite3` only displays the problematic part.

src/kw_db.sh
```diff
@@ -51,3 +51,30 @@ function execute_command_db()
...
+# 0 if succesful; non-zero otherwise
+function insert_into()
+{
+  local table="$1"
+  local entries="$2"
+  local values="$3"
+  local db="${4:-"$DB_NAME"}"
+  local db_folder="${5:-"$KW_DATA_DIR"}"
+  local db_path
+
```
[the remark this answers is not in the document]
kwy95: Because of the way SQL commands work `values` cannot be empty. The way you add an empty value is by either not specifying the specific column, or adding a literal (`''`) empty value. With this in mind I'll validate `table` and `values`. The `entries` variable can be empty if you are adding values to all columns.

src/kw_db.sh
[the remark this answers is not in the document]
rodrigosiqueira: I agree

src/kw_db.sh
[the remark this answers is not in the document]
rodrigosiqueira: We should return something in case of a problem and let the function that invokes it handle the issue. However, I don't think we should add any warning message here because this is not something the user will understand in case of an error. Ideally, we should have the verbose or debug mode for handling these cases. (https://github.com/kworkflow/kworkflow/issues/179)

rodrigosiqueira (changes requested):

tests/kw_db_test.sh
kwy95: Nope. There should be quotes around the full string, and then 1 extra quote for each quote inside the original string.

src/kw_db.sh
```diff
@@ -76,5 +76,62 @@ function insert_into()
...
+# This function takes arguments and assembles them into the correct format to
+# be used as entries in SQL commands
+#
+# @@: Values to be formatted
+#
+# Return:
+# The arguments in formatted string to be used as entries in an INSERT command
+# 22 if no arguments are given
+function format_entries_db()
```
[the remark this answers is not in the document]
kwy95: I'm thinking this could become something like:
```shell
function get_comma_separated()
{
  local IFS=','
  printf '%s\n' "$*"
}
```

rodrigosiqueira:
Thanks a lot for this PR!

PR #529, opened 2021-12-08
Move kernel scripts parameters to kworkflow.config

### LINKED ISSUE #260: Move kernel script parameters to kworkflow.config
kw already have a wrapper to get_maintaners and checkpack scripts. However, the parameters used by these scripts are hardcoded; let's move them to the kworkflow.config and allow users to use the parameter that they want.

kwy95:
This PR moves the parameters used by the `checkpatch_wrapper.sh` and `get_maintainer_wrapper.sh` scripts to the *kworkflow.config* file.

Closes #260

### DIFF
--- a/src/checkpatch_wrapper.sh
+++ b/src/checkpatch_wrapper.sh
@@ -22,7 +22,7 @@ function execute_checkpatch()
   # output. It could be nice if we can add another option just for this sort
   # of check.
 
-  local options='--no-tree --color=always --strict'
+  local options="${configurations[checkpatch_opts]}"
   local -r original_working_dir="$PWD"
   local kernel_root
   local cmd_script
--- a/etc/kworkflow_template.config
+++ b/etc/kworkflow_template.config
@@ -98,6 +98,7 @@ send_opts=--annotate --cover-letter --no-chain-reply-to --thread
 # list of your patches, separate values with commas
 #blocked_emails=
 
-# Options to be used with the checkpatch script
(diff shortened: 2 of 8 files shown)


src/checkpatch_wrapper.sh
```diff
@@ -22,7 +22,7 @@ function execute_checkpatch()
   # output. It could be nice if we can add another option just for this sort
   # of check.
 
-  local options='--no-tree --color=always --strict'
+  local options="${configurations[checkpatch_opts]}"
```
rodrigosiqueira, line 25: What happens if `options` is empty? Do we have a test to check this behavior?

rodrigosiqueira (changes requested):

kwy95:
@rodrigosiqueira
Will update soon.

src/checkpatch_wrapper.sh
kwy95: I think having no options is a valid usage. If the user clears this option we shouldn't assume any options.

rodrigosiqueira:
@kwy95 did you update this PR?

kwy95:
@rodrigosiqueira The only concern you presented was about the script parameters being empty, but I believe that to be a valid usage. What do you think?

rodrigosiqueira:
Hi @kwy95 ,
This PR lgtm, could you just rebase it with the unstable branch?

rodrigosiqueira:
Merged, thanks

PR #533, opened 2021-12-19
src: small optimization when using cat

kwy95:
When using command substitution to access the contents of a file the
`cat` command can be replaced with a `<` which is equivalent but faster.
E.g.: `$(cat "$file")` => `$(< "$file")`


rodrigosiqueira:
Hi @kwy95 ,
This is really interesting. Do you have any reference to the performance improvement? I'm curious to understand why and I also think we should document this optimization in the codestyle.

kwy95:
@rodrigosiqueira I don't have any concrete numbers, I will write a few tests and can add the numbers to the codestyle page. I got this change from the bash reference manual, that explicitly mentions and recommends this usage.

kwy95:
@rodrigosiqueira Apparently this optimization is much more relevant than I anticipated:
cat_redir_time

rodrigosiqueira:
Hi @kwy95 ,
I merged the first commit and dropped the second one. Could you open an issue for the second commit?

PR #534, opened 2021-12-21
src: init:  add git configuration

Pauloguilhermepp:
src: init: create interactive init feature
  
  In an attempt to make KW more user-friendly, this PR adds a new
  feature to help beginners users to configure what is needed to
  work with it.  To do so, it is necessary to set some tools that
  are commonly used by developers (such as git).
  
  The idea is to explain what are those configurations and suggest
  default options helping beginners to set their environment.
  This commit does it with git.
  
  We selected to configure user name, email, editor, and branch name
  since these are the configurations in this tutorial:
  https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup.
  
  Signed-off-by: Lucas Cardoso lucas.cardoso.santos@usp.br
  Signed-off-by: Paulo Guilherme Pinheiro pauloguilherme@usp.br
  Co-authored-by: Eduardo Brancher Urenha eduardo.urenha@usp.br
  Co-authored-by: Miguel de Mello Carpi miguelmello@usp.br
  Co-authored-by: Lourenço Henrique Moinheiro Martins louhmmsb@usp.br

### DIFF
--- a/src/init.sh
+++ b/src/init.sh
@@ -4,6 +4,7 @@
 include "$KW_LIB_DIR/kwio.sh"
 include "$KW_LIB_DIR/kwlib.sh"
 include "$KW_LIB_DIR/remote.sh"
+include "$KW_LIB_DIR/kw_string.sh"
 
 KW_DIR='.kw'
 
@@ -35,6 +36,11 @@ function init_kw()
   if [[ "$?" -gt 0 ]]; then
     complain "${options_values['ERROR']}"
     return 22 # EINVAL
+  fi
+
+  if [[ -n "${options_values['GUIDED-SETUP']}" ]]; then
+    guided_init
+    exit 0
   fi
 
   if [[ -f "$PWD/$KW_DIR/$name" ]]; then
@@ -112,8 +118,8 @@ function set_config_value()
 
 function parse_init_options()
 {
-  local long_options='arch:,remote:,target:,force'
-  local short_options='a:,r:,t:,f'
+  local long_options='arch:,remote:,target:,force,guided-setup'
+  local short_options='a:,r:,t:,f,g'
 
   options="$(kw_parse "$short_options" "$long_options" "$@")"
 
@@ -125,6 +131,7 @@ function parse_init_options()
 
   options_values['ARCH']=''
   options_values['FORCE']=''
+  options_values['GUIDED-SETUP']=''
 
   eval "set -- $options"
 
@@ -146,6 +153,10 @@ function parse_init_options()
       --force | -f)
         shift
         options_values['FORCE']=1
+        ;;
+      --guided-setup | -g)
+        options_values['GUIDED-SETUP']=1
+        shift
         ;;
       --)
         shift
@@ -170,5 +181,179 @@ function init_help()
     '  init - Creates a kworkflow.config file in the current directory.' \
     '  init --arch <arch> - Set the arch field in the kworkflow.config file.' \
     '  init --remote <user>@<ip>:<port> - Set remote fields in the kworkflow.config file.' \
-    '  init --target <target> Set the default_deploy_target field in the kworkflow.config file'
+    '  init --target <target> - Set the default_deploy_target field in the kworkflow.config file' \
+    '  init --guided-setup - Perform first time setup in an interactive manner. Recommended for new users.'
+}
+
+# This function gets four configuration values from git: name, email,
+# editor, and initial branch name and stores them inside an associative array
+# for future access. Checks if these values are empty and store this information
+# in a boolean variable inside the array.
+#
+# @_git_config: associative array received as a nameref attribute to store all
+#              values acquired in this function.
+#
+# Returns: _git_config[configured]: 1, if there is any revelant information in
+#          git missing, and 0, if git is fully configured.
+function get_git_config()
+{
+  local -n _git_config="$1"
+
+  _git_config['name']=$(git config user.name)
+  _git_config['email']=$(git config user.email)
+  _git_config['editor']=$(git config core.editor)
+  _git_config['branch']=$(git config init.defaultBranch)
+
+  _git_config['configured']=0
+
+  if [[ -z "${_git_config['name']}" ||
+    -z "${_git_config['email']}" ||
+    -z "${_git_config['editor']}" ||
+    -z "${_git_config['branch']}" ]]; then
+    _git_config['configured']=1
+  fi
+}
+
+# This function sets if the scope of the git configuration will be
+# local or global.
+#
+# @config_cmd: command in which will be added the configuration scope.
+#
+# Returns: the git config command, adjusted if the scope is local or global
+function set_git_config_scope()
+{
+  local command
+  local scope
+
+  command='git config'
+
+  if ! is_inside_work_tree; then
+    warning 'You are not in a git repository. All modifications will be global' 1>&2
+    if [[ $(ask_yN 'Would you like to proceed?' 'y') =~ '1' ]]; then
+      command+=' --global'
+    else
+      command=''
+    fi
+
+    printf '%s\n' "$command"
+    exit 0
+  fi
+
+  printf '\nSelect the scope of this configuration:\n' 1>&2
+
+  select scope in 'Local' 'Global'; do
+    case "$scope" in
+      'Global')
+        command+=' --global'
+        break
+        ;;
+      'Local')
+        command+=' --local'
+        break
+        ;;
+    esac
+  done
+
+  printf '%s\n' "$command"
+}
+
+# Allows the user to set the empty configurations if any.
+#
+# @_config_git: associative array with data about git configuration.
+# @flag: flag sended to the cmd_manager to change its behaviour.
+#
+# Returns: _config_git: array with the selected information configured on git
+function interactive_set_user_git_info()
+{
+  local -n _config_git="$1"
+  local flag="$2"
+  local git_editor_suggestion
+  local config_cmd
+  local scope
+
+  flag=${flag:-'SILENT'}
+
+  if [[ "${_config_git['configured']}" == '0' ]]; then
+    exit 0
+  fi
+
+  printf '\nNow Git will be configured.\n' 1>&2
+
+  config_cmd=$(set_git_config_scope)
+  if [[ -z "$config_cmd" ]]; then
+    return
+  fi
+
+  if [[ -z "${_config_git['name']}" ]]; then
+    _config_git['name']=$(ask_with_default 'What is your name?' "$USER")
+    cmd_manager "$flag" "$config_cmd user.name \"${_config_git['name']}\""
+  fi
+
+  if [[ -z "${_config_git['email']}" ]]; then
+    while :; do
+      _config_git['email']=$(ask_with_default 'What is your email?' '')
+
+      if validate_email "${_config_git['email']}"; then
+        break
+      fi
+
+      printf 'It was not possible to validate your e-mail, please try again.\n' 1>&2
+    done
+    cmd_manager "$flag" "$config_cmd user.email \"${_config_git['email']}\""
+  fi
+
+  if [[ -z "${_config_git['editor']}" ]]; then
+    if [[ $(ask_yN $'\nWould you like to configure your default editor on Git?' 'y') == '1' ]]; then
+
+      # This follows git precedence to determine used editor
+      git_editor_suggestion='vi'
+      [[ -n "$EDITOR" ]] && git_editor_suggestion="$EDITOR"
+      [[ -n "$VISUAL" ]] && git_editor_suggestion="$VISUAL"
+      _config_git['editor']=$(ask_with_default 'What is your main editor?' "$git_editor_suggestion")
+
+      cmd_manager "$flag" "$config_cmd core.editor \"${_config_git['editor']}\""
+    fi
+  fi
+
+  if [[ -z "${_config_git['branch']}" ]]; then
+    # This is a minor configuration, so default is "no"
+    if [[ $(ask_yN $'\nWould you like to configure your initial default branch on Git?' 'n') == '1' ]]; then
+      _config_git['branch']=$(ask_with_default 'What is your default branch name?' 'main')
+      cmd_manager "$flag" "$config_cmd init.defaultBranch \"${_config_git['branch']}\""
+    fi
+  fi
+}
+
+# This function configures KW in an interactive way
+# by explaining the core configurations required and
+# asking for user input for some configuration decisions.
+# This option is intended for new users.
+#
+# Exit code:
+# ENODATA     : Interaction text could not be loaded.
+# ECANCELED   : User interruption.
+# EINPROGRESS : User selected not implemented feature
+function guided_init()
+{
+  local -A git_configurations
+
+  load_module_text "$KW_LIB_DIR/strings/init.txt"
+
+  if [[ "$?" -ne 0 ]]; then
+    complain '[ERROR]:src/init.sh:interactive_init: Failed to load module text.'
+    exit 61 # ENODATA
+  fi
+
+  say "${module_text_dictionary['text_interactive_start']}"
+  if [[ $(ask_yN '>>> Do you want to continue?' 'y') -ne 1 ]]; then
+    exit 125 # ECANCELED
+  fi
+
+  # Help to setup git
+  get_git_config _config_git
+  interactive_set_user_git_info git_configurations
+
+  #TODO: Help to setup kworkflow.config;
+
+  #TODO: Check for required;
 }
--- a/src/kw_string.sh
+++ b/src/kw_string.sh
@@ -1,3 +1,7 @@
+include "$KW_LIB_DIR/kwio.sh"
+
+declare -gr email_regex='[A-Za-z0-9_\.-]+@[A-Za-z0-9_-]+(\.[A-Za-z0-9]+)+'
+
 # @@ Expect a string
 #
 # Return:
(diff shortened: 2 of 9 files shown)


LucasCardoso910:
We saw that there is an error in the shellcheck test. We read about (https://github.com/koalaman/shellcheck/issues/1225) and it seems that there is a problem in this check in the same context that we were using it. What should we do about it?

rodrigosiqueira:
Hi guys, I was on vacation, and I didn't review this PR. Do you have any updates here?
If you don't want to work on it, just let me know, and I can complete this PR. Otherwise, I'm waiting for a new version.

Pauloguilhermepp:
Hi Siqueira, sorry for the delay. Don't worry, we will finish this PR. We just took a break to be with our families.

LucasCardoso910:
Hi, Siqueira.

As previously said, the error evidenced by shellsheck was a bug. Following the instructions in this issue https://github.com/koalaman/shellcheck/issues/1225 where was found a very similar problem to ours, we renamed one of the variables from `git_config` to `config_git` and the error ceased to exist.

Any other suggestions on how we can improve this code, please say so.

src/init.sh
```diff
@@ -220,6 +218,135 @@ function create_ssh_key()
...
+      if [[ $(ask_yN $'\nWould you like to configure your name on Git?' 'y') =~ '1' ]]; then
+        config_git['name']=$(ask_with_default 'What is your name?' "$USER")
+        cmd_manager "$flag" "$config_cmd user.name \"${config_git['name']}\""
+      fi
+    fi
+
+    if [[ -z "${config_git['email']}" ]]; then
+      if [[ $(ask_yN $'\nWould you like to configure your email on Git?' 'y') =~ '1' ]]; then
+        read -r -p 'What is your email? ' config_git['email']
+        # TODO: Validate if this is a valid email
```
kwy95, line 322: You can use the following regex to validate emails:
```
[A-Za-z0-9_\.-]+@[A-Za-z0-9_-]+(\.[A-Za-z0-9]+)+
```

kwy95 (changes requested):
Please rebase your changes on the unstable branch. I also added some comments to the code for things you can improve.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+      command=''
+    fi
+  fi
+
+  printf '%s\n' "$command"
+}
+
+# Allows the user to set the empty configurations if any.
+#
+# @config_git: associative array with data about git configuration.
```
rodrigosiqueira, line 257: You forgot to document `@flag`

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+#
+# Returns: config_git: array with the selected information configured on git
+function set_git_config()
+{
+  local -n config_git="$1"
```
rodrigosiqueira, line 262: For reference variables, please use `_` in the beginning.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+  local flag="$2"
+  local git_editor_suggestion
+  local config_cmd
+  local scope
+
+  flag=${flag:-'SILENT'}
+
+  if [[ "${config_git['configured']}" =~ '1' ]]; then
```
rodrigosiqueira, line 270: How about something like this:
1. Instead of calling `get_git_config` before this function, why not call it here in the beginning?
2. After you call `get_git_config`, use the check you have in this line but instead of checking for `1`, check for `0` and return if you are not happy or do something else. Also, why are you using `=~`? Why not use a standard comparison?

That way, you can drop this huge `if`.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+    printf '\nNow Git will be configured.\n' 1>&2
```
rodrigosiqueira, line 271: I did not get why we need `1>&2` here.

rodrigosiqueira: Also, in the following lines, you have a check and return, which makes this message outside context.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+
+    config_cmd=$(set_git_config_scope)
+    if [[ -z $config_cmd ]]; then
```
rodrigosiqueira, line 274: Use double-quotes.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+      return
+    fi
+
+    if [[ -z "${config_git['name']}" ]]; then
+      if [[ $(ask_yN $'\nWould you like to configure your name on Git?' 'y') =~ '1' ]]; then
```
rodrigosiqueira, line 279: I think the name in the git is very important; therefore, I would drop this question and enforce users to set their name.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+        config_git['name']=$(ask_with_default 'What is your name?' "$USER")
+        cmd_manager "$flag" "$config_cmd user.name \"${config_git['name']}\""
```
rodrigosiqueira, line 281: @kwy95 do we have helpers for this sort of operation?

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+      fi
+    fi
+
+    if [[ -z "${config_git['email']}" ]]; then
+      if [[ $(ask_yN $'\nWould you like to configure your email on Git?' 'y') =~ '1' ]]; then
```
rodrigosiqueira, line 286: Email is another essential piece of information; for this reason, I would drop this question as well.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+        while :; do
+          printf 'Insert nothing to cancel this.\n' 1>&2
```
rodrigosiqueira, line 288: Should we allow cancellation here? I think we should enforce uses to set an email.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+          config_git['email']=$(ask_with_default 'What is your email?' '')
+          [[ -z "${config_git['email']}" ]] || validate_email "${config_git['email']}" && break
```
rodrigosiqueira, line 290: Use `if` here since this is a long condition.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+      fi
+    fi
+
+    if [[ -z "${config_git['branch']}" ]]; then
+      # This is a minor configuration, so default is "no"
+      if [[ $(ask_yN $'\nWould you like to configure your initial default branch on Git?' 'n') =~ '1' ]]; then
+        config_git['branch']=$(ask_with_default 'What is your default branch name?' 'main')
+        cmd_manager "$flag" "$config_cmd init.defaultBranch \"${config_git['branch']}\""
+      fi
+    fi
```
rodrigosiqueira, line 316: let's drop this.

rodrigosiqueira (changes requested):
Hi @Pauloguilhermepp ,

First of all, I'm sorry for my colossal delay in reviewing this excellent PR. Unfortunately, my time was a bit constrained these days, and I had to prioritize other kw tasks.

If you are still interested in completing this task, I added a detailed review here, and now I have more time to focus on this PR. If you don't want to work on it anymore, just let me know.

src/init.sh
LucasCardoso910: > I did not get why we need `1>&2` here.

We redirect the standard output to the standard error because these messages are for user instruction only. Redirecting to the standard error (which is already done to deal with the select messages), we can easily disregard these messages in the test file.

Pauloguilhermepp: > Also, in the following lines, you have a check and return, which makes this message outside context.

Return only will be executed if the user chooses to cancel the operation in case it is not in a git repository and doesn't want to make global changes in git, so, even if the return statement be executed, there will be some context.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
     '  init - Creates a kworkflow.config file in the current directory.' \
     '  init --arch <arch> - Set the arch field in the kworkflow.config file.' \
     '  init --remote <user>@<ip>:<port> - Set remote fields in the kworkflow.config file.' \
-    '  init --target <target> Set the default_deploy_target field in the kworkflow.config file'
+    '  init --target <target> - Set the default_deploy_target field in the kworkflow.config file' \
+    '  init --interactive - Perform first time setup in an interactive manner. Recommended for new users.'
+}
+
+# This function gets four configuration values from git: user name, user email,
+# editor, and initial branch name and stores them inside an associative array
```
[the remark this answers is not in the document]
Pauloguilhermepp: > Why do we need the branch name?

We selected to configure user name, email, editor, and branch name since these are the configurations in this tutorial: https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup. We think that they put the branch name as an important setup since there is a movement to change the default branch name from master to main, but, for now, master is still used as default. So, this is an important setup to do.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+# Returns: git_config[configured]: 1, if there is any revelant information in
+#          git missing, and 0, if git is fully configured.
+function get_git_config()
+{
+  local -n git_config="$1"
+
+  git_config['name']=$(git config user.name)
+  git_config['email']=$(git config user.email)
+  git_config['editor']=$(git config core.editor)
+  git_config['branch']=$(git config init.defaultBranch)
```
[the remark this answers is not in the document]
LucasCardoso910: The get_all_git_config always returns some text, even if there is no configuration defined (when we tested it, it returned something like: "Local: Global:").
And in our code, we consider that if there is no configuration defined, it will be empty, which would not be true using this function. Besides, it is simpler for our objective here to use the git function.

src/init.sh
```diff
@@ -170,5 +180,173 @@ function init_help()
...
+      case $scope in
+        'Global')
+          command+=' --global'
+          break
+          ;;
+        'Local')
+          command+=' --local'
+          break
+          ;;
+      esac
```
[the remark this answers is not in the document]
Pauloguilhermepp: We tried to use it, but it didn't work. We are not sure why it happened.

src/init.sh
LucasCardoso910: > 1. Instead of calling `get_git_config` before this function, why not call it here in the beginning?

When we tried to move this function call to the beginning of this function, it didn't work. We are not very sure why.

src/kw_string.sh
```diff
@@ -193,3 +197,52 @@ function concatenate_with_commas()
...
 }
+
+# This function validates the encryption. If the passed encryption is not valid
+# this will warn the user and clear the option.
+#
+# @option: The option to determine if it should be an email
+# @value:  The value being passed
+#
+# Return:
+# Returns 0 if valid; 22 if invalid
```
rodrigosiqueira, line 208: This documentation is not correct.

src/kw_string.sh
```diff
@@ -193,3 +197,52 @@ function concatenate_with_commas()
...
+  return 0
+}
+
+# Validates the recipient list given by the user to the options `--to` and
+# `--cc` to make sure the all the recipients are valid.
+#
+# @raw: The list of email recipients to be validated
+#
+# Return:
+# 22 if there are invalid entries; 0 otherwise
```
rodrigosiqueira, line 227: Please, rewrite this documentation. Keep in mind that you are adding this to the kw_string, which is intended to be generic. Also, add an example.

src/init.sh
```diff
@@ -37,6 +38,11 @@ function init_kw()
     return 22 # EINVAL
   fi
 
+  if [[ -n "${options_values['GUIDED-SETUP']}" ]]; then
+    guided_init
+    exit 0
```
rodrigosiqueira, line 43: Use return.

src/init.sh
```diff
@@ -37,6 +38,11 @@ function init_kw()
     return 22 # EINVAL
   fi
 
+  if [[ -n "${options_values['GUIDED-SETUP']}" ]]; then
+    guided_init
```
rodrigosiqueira, line 42: let's call it `guided_setup_init`.

src/init.sh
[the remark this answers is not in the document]
rodrigosiqueira: Ok, but in that case, we should check the current branch and use it as a default. For example, in the case of amd driver the default branch is named `amd-staging-drm-next`, and right after clone, this is the main branch. In that case, we should suppose the current branch is the main branch but give the user the chance to change it.

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+  if ! is_inside_work_tree; then
+    warning 'You are not in a git repository. All modifications will be global' 1>&2
+    if [[ $(ask_yN 'Would you like to proceed?' 'y') =~ '1' ]]; then
+      command+=' --global'
+    else
+      command=''
+    fi
+
+    printf '%s\n' "$command"
+    exit 0
```
rodrigosiqueira, line 239: return

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+  printf '%s\n' "$command"
+}
+
+# Allows the user to set the empty configurations if any.
+#
+# @_config_git: associative array with data about git configuration.
+# @flag: flag sended to the cmd_manager to change its behaviour.
+#
+# Returns: _config_git: array with the selected information configured on git
+function interactive_set_user_git_info()
```
rodrigosiqueira, line 266: Let's call it `interactive_user_git_setup()`

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+  local -n _config_git="$1"
+  local flag="$2"
+  local git_editor_suggestion
+  local config_cmd
+  local scope
+
+  flag=${flag:-'SILENT'}
+
+  if [[ "${_config_git['configured']}" == '0' ]]; then
+    exit 0
```
rodrigosiqueira, line 277: Let use return. Also, I think we should tell the user that git is already configured and maybe show the current config.

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+  fi
+
+  printf '\nNow Git will be configured.\n' 1>&2
```
rodrigosiqueira, line 280: Let's use something like:
`Welcome to kw guided setup`

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+
+  config_cmd=$(set_git_config_scope)
+  if [[ -z "$config_cmd" ]]; then
+    return
+  fi
```
rodrigosiqueira, line 285: Let's use the short version `[[ -z "$config_cmd" ]] && return`

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+
+  if [[ -z "${_config_git['name']}" ]]; then
+    _config_git['name']=$(ask_with_default 'What is your name?' "$USER")
+    cmd_manager "$flag" "$config_cmd user.name \"${_config_git['name']}\""
```
rodrigosiqueira, line 289: We should not change any git value here, let's first collect all data, and in the end, we update everything. I'm suggesting this because users can cancel this operation in the middle, and we will have inconsistent data if we add things before we collect all data.
(same comment for all of these configurations)

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+
+  if [[ -z "${_config_git['email']}" ]]; then
+    while :; do
+      _config_git['email']=$(ask_with_default 'What is your email?' '')
+
+      if validate_email "${_config_git['email']}"; then
+        break
+      fi
+
+      printf 'It was not possible to validate your e-mail, please try again.\n' 1>&2
```
rodrigosiqueira, line 300: Why do we have these `1>&2` spread around this commit?

rodrigosiqueira: Also, use the "complain" function instead of `printf`.

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+      _config_git['editor']=$(ask_with_default 'What is your main editor?' "$git_editor_suggestion")
+
+      cmd_manager "$flag" "$config_cmd core.editor \"${_config_git['editor']}\""
+    fi
+  fi
+
+  if [[ -z "${_config_git['branch']}" ]]; then
+    # This is a minor configuration, so default is "no"
+    if [[ $(ask_yN $'\nWould you like to configure your initial default branch on Git?' 'n') == '1' ]]; then
+      _config_git['branch']=$(ask_with_default 'What is your default branch name?' 'main')
```
rodrigosiqueira, line 321: We should not suppose "main"; we should check the current branch and suggest it as a default.

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+      cmd_manager "$flag" "$config_cmd init.defaultBranch \"${_config_git['branch']}\""
+    fi
+  fi
+}
```
rodrigosiqueira, line 325: How about creating a function per question and calling them here? I think this will be easy to test and debug in the future.

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+  fi
+
+  say "${module_text_dictionary['text_interactive_start']}"
+  if [[ $(ask_yN '>>> Do you want to continue?' 'y') -ne 1 ]]; then
+    exit 125 # ECANCELED
+  fi
+
+  # Help to setup git
+  get_git_config _config_git
+  interactive_set_user_git_info git_configurations
```
rodrigosiqueira, line 354: Why not move `get_git_config` inside `interactive_set_user_git_info`?

src/init.sh
```diff
@@ -170,5 +181,179 @@ function init_help()
...
+    exit 61 # ENODATA
+  fi
+
+  say "${module_text_dictionary['text_interactive_start']}"
+  if [[ $(ask_yN '>>> Do you want to continue?' 'y') -ne 1 ]]; then
+    exit 125 # ECANCELED
+  fi
+
+  # Help to setup git
+  get_git_config _config_git
```
rodrigosiqueira, line 353: Is `_config_git` correct? I expected `git_configurations`.

rodrigosiqueira (changes requested):
Following my first set of reviews, I'll try to test it in more detail and post more comments after that.

rodrigosiqueira:
Hi @Pauloguilhermepp , do you plan to complete this PR? If not, please, let me know.

Pauloguilhermepp:
Hello @rodrigosiqueira, we will finish this PR, do not worry about it. We are currently busy with some activities at our university, but we will finish this PR as soon as possible.

rodrigosiqueira:
I am closing this PR due to the lack of updates.

PR #535, opened 2021-12-23
Mail send

kwy95:
Add send capabilities to the mail feature.

If inside a kernel root will add default options and populate the `to` and `cc` fields.

I believe this to be fully functional now. Some tests could be improved still. I believe it would be beneficial to have a sample `.git` folder so we have control over the resulting patch files.

### DIFF
--- a/documentation/man/features/mail.rst
+++ b/documentation/man/features/mail.rst
@@ -23,8 +23,8 @@ OPTIONS
 =======
 -t, \--setup:
   Initialize and configure **mail** functionality. Each argument specifies a
-  <config> to be set with the corresponding <value>, multiple <config> <value>
-  pairs may be passed at once. Current accepted options are::
+  *<config>* to be set with the corresponding *<value>*, multiple *<config>*
+  *<value>* pairs may be passed at once. Current accepted options are::
 
     --name           <name>
     --email          <email>
@@ -57,7 +57,7 @@ OPTIONS
   Verify that all the settings needed are set and valid.
 
 \--template[=<template>]:
-  This loads the default configuration values based on the given <template>.
+  This loads the default configuration values based on the given *<template>*.
   If no template is given the user will be shown the available templates to
   choose from.
 
--- a/documentation/man/features/mail.rst
+++ b/documentation/man/features/mail.rst
@@ -6,7 +6,7 @@ kw-mail
 
 SYNOPSIS
 ========
-| *kw mail* (-s | \--send) [\--simulate] [\--to='<recipient>,...'] [\--cc='<recipient>,...'] [<rev-range>...] [-v<version>] [\-- <extra-args>...]
+| *kw mail* (-s | \--send) [\--simulate] [\--private] [\--to='<recipient>,...'] [\--cc='<recipient>,...'] [<rev-range>...] [-v<version>] [\-- <extra-args>...]
 | *kw mail* (-t | \--setup) [\--local | \--global] [-f | \--force] (<config> <value>)...
 | *kw mail* (-i | \--interactive) [\--local | \--global]
 | *kw mail* (-l | \--list)
@@ -54,6 +54,9 @@ OPTIONS
 \--simulate:
   Do everything without actually sending the e-mail. This is similar to
   ``git send-email``'s ``--dry-run`` option.
+
+\--private:
+  Supress auto generation of recipients.
 
 <rev-range>...:
   Specify the *<rev-range>* to be sent. The last commit is taken as the
--- a/documentation/man/features/mail.rst
+++ b/documentation/man/features/mail.rst
@@ -6,11 +6,11 @@ kw-mail
 
 SYNOPSIS
 ========
-| *kw mail* (-s | \--send) [\--simulate] [\--to='<recipient>,...'] [\--cc='<recipient>,...']
+| *kw mail* (-s | \--send) [\--simulate] [\--to='<recipient>,...'] [\--cc='<recipient>,...'] [<rev-range>...] [-v<version>] [\-- <extra-args>...]
 | *kw mail* (-t | \--setup) [\--local | \--global] [-f | \--force] (<config> <value>)...
 | *kw mail* (-i | \--interactive) [\--local | \--global]
-| *kw mail* (-v | \--verify) [\--local | \--global]
 | *kw mail* (-l | \--list)
+| *kw mail* \--verify [\--local | \--global]
 | *kw mail* \--template[=<template>] [-n | \--no-interactive] [\--local | \--global] [-f | \--force] [(<config> <value>)...]
 
 DESCRIPTION
@@ -20,11 +20,17 @@ It is common to deal with patch submissions to a mailing list, and
 **mail** functionality aims to wrap this tool to simplify its usage and
 integrate it with other **kw** functionalities.
 
+.. note::
+  Any option recognized by ``git send-email`` can be passed directly to it if
+  placed after the double dash (``--``) argument.
+
 OPTIONS
 =======
 -s, \--send:
-  Send the last commit as a patch using ``git send-email`` to the e-mail
-  adresses specified with ``--to`` and ``--cc``.
+  Send a patch by email using ``git send-email`` to the email adresses
+  specified with ``--to`` and ``--cc``. You can provide *<extra-args>* to be
+  passed directly to ``git send-email``, they should be placed after the double
+  dash (``--``) argument.
 
 \--to='<recipient>,...':
   Specify the recipients that will receive the patch via e-mail. The
@@ -37,6 +43,13 @@ OPTIONS
 \--simulate:
   Do everything without actually sending the e-mail. This is similar to
   ``git send-email``'s ``--dry-run`` option.
+
+<rev-range>...:
+  Specify the *<rev-range>* to be sent. The last commit is taken as the
+  default.
+
+-v<version>:
+  Specify a number *<version>* for your patch.
 
 -t, \--setup:
   Initialize and configure **mail** functionality. Each argument specifies a
@@ -70,7 +83,7 @@ OPTIONS
   Forces the configurations to be added, regardless of conflicts with the
   current values already set in the system. Implies ``--no-interactive``.
 
--v, \--verify:
+\--verify:
   Verify that all the settings needed are set and valid.
 
 \--template[=<template>]:
@@ -97,3 +110,11 @@ Then when you are sure the command executed as expected, drop the
 ``--simulate`` argument to actually send the patch::
 
   kw mail --send --to=some@email.com
+
+To send a range of commits the following can be used::
+
+  kw mail -s <SHA1>..<SHA2>
+
+Extra arguments can be passed directly to ``git send-email`` like this::
+
+  kw mail -s --to='some@email.com' -- --thread
--- a/documentation/man/features/mail.rst
+++ b/documentation/man/features/mail.rst
@@ -25,6 +25,11 @@ mailing lists fetched with the *get_maintainer.pl* script. It will also use
 the union of the recipients of each patch as the recipients of the cover-letter.
 
 .. note::
+  You can block certain e-mail addresses from being automatically added to the
+  recipients list of the patches using the *blocked_emails* option in the
+  *kworkflow.config* file.
+
+.. note::
   Any option recognized by ``git send-email`` can be passed directly to it if
   placed after the double dash (``--``) argument.
 
--- a/documentation/man/features/mail.rst
+++ b/documentation/man/features/mail.rst
@@ -6,6 +6,7 @@ kw-mail
 
 SYNOPSIS
 ========
+| *kw mail* (-s | \--send) [\--simulate] [\--to='<recipient>,...'] [\--cc='<recipient>,...']
 | *kw mail* (-t | \--setup) [\--local | \--global] [-f | \--force] (<config> <value>)...
 | *kw mail* (-i | \--interactive) [\--local | \--global]
 | *kw mail* (-v | \--verify) [\--local | \--global]
@@ -21,6 +22,22 @@ integrate it with other **kw** functionalities.
 
 OPTIONS
 =======
+-s, \--send:
+  Send the last commit as a patch using ``git send-email`` to the e-mail
+  adresses specified with ``--to`` and ``--cc``.
+
+\--to='<recipient>,...':
+  Specify the recipients that will receive the patch via e-mail. The
+  *<recipient>* list can be in any format accepted by ``git send-email``, e.g.:
+  ``some@email.com`` or ``Xpto Lala <lala.xpto@mail.com>``.
+
+\--cc='<recipient>,...':
+  Specify the recipients that will receive a copy of the patch via e-mail.
+
+\--simulate:
+  Do everything without actually sending the e-mail. This is similar to
+  ``git send-email``'s ``--dry-run`` option.
+
 -t, \--setup:
   Initialize and configure **mail** functionality. Each argument specifies a
   *<config>* to be set with the corresponding *<value>*, multiple *<config>*
@@ -71,3 +88,12 @@ variables at once::
 
   kw mail -t --name 'Xpto Lala' --email myemail@gmail.com --smtpencryption tls \
     --smtpserver smtp.gmail.com --smtpserverport 587 --smtpuser myemail@gmail.com
+
+To simulate sending the last commit as a patch just write::
+
+  kw mail --send --simulate --to=some@email.com
+
+Then when you are sure the command executed as expected, drop the
+``--simulate`` argument to actually send the patch::
+
+  kw mail --send --to=some@email.com
--- a/documentation/man/features/mail.rst
+++ b/documentation/man/features/mail.rst
@@ -19,6 +19,10 @@ It is common to deal with patch submissions to a mailing list, and
 ``git send-email`` is one of the most famous tools for handling that. The
 **mail** functionality aims to wrap this tool to simplify its usage and
 integrate it with other **kw** functionalities.
+If used inside of a linux kernel tree the send feature will auto populate the
+*to* and *cc* fields of each patch file with the appropriate maintainers and
+mailing lists fetched with the *get_maintainer.pl* script. It will also use
+the union of the recipients of each patch as the recipients of the cover-letter.
 
 .. note::
   Any option recognized by ``git send-email`` can be passed directly to it if
@@ -125,3 +129,9 @@ To send a range of commits the following can be used::
 Extra arguments can be passed directly to ``git send-email`` like this::
 
   kw mail -s --to='some@email.com' -- --thread
+
+If you are inside of a linux kernel tree, you can send the last three commits
+as a patchset to the maintainers of the subsystems and with copies to the
+appropriate mailing lists using::
+
+  kw mail -s -3
--- a/documentation/man/features/mail.rst
+++ b/documentation/man/features/mail.rst
@@ -30,7 +30,14 @@ OPTIONS
   Send a patch by email using ``git send-email`` to the email adresses
   specified with ``--to`` and ``--cc``. You can provide *<extra-args>* to be
   passed directly to ``git send-email``, they should be placed after the double
-  dash (``--``) argument.
+  dash (``--``) argument. By default this function assumes these arguments to
+  ``git send-email``::
+
+    --annotate --cover-letter --no-chain-reply-to --thread
+
+  .. note::
+    You can change the default arguments used to send emails in the
+    *kworkflow.config* file.
 
 \--to='<recipient>,...':
   Specify the recipients that will receive the patch via e-mail. The
--- a/src/mail.sh
+++ b/src/mail.sh
@@ -4,6 +4,7 @@
 # are set.
 
 include "$KW_LIB_DIR/kw_config_loader.sh"
+include "$KW_LIB_DIR/kwlib.sh"
 include "$KW_LIB_DIR/kw_string.sh"
 
 # Hash containing user options
@@ -84,7 +85,10 @@ function mail_send()
   local to_recipients="${options_values['TO']}"
   local cc_recipients="${options_values['CC']}"
   local dryrun="${options_values['SIMULATE']}"
+  local commit_range="${options_values['COMMIT_RANGE']}"
+  local version="${options_values['PATCH_VERSION']}"
   local extra_opts="${options_values['PASS_OPTION_TO_SEND_EMAIL']}"
+  local kernel_root
   local cmd='git send-email'
 
   flag=${flag:-'SILENT'}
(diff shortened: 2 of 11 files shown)


src/mail.sh
```diff
@@ -77,18 +77,66 @@ function mail_main()
...
+  local out=''
+  local patch_cache="${KW_CACHE_DIR}/patches/"
+  local get_maintainer="perl $kernel_root/scripts/get_maintainer.pl --nogit --nogit-fallback --no-r --no-n --multiline --nokeywords --norolestats"
+
+  mkdir -p "$patch_cache"
+  cmd_manager 'SILENT' "git format-patch --quiet -o $patch_cache $revision"
+
+  for patch in "$patch_cache"*; do
+    ! is_a_patch "$patch" && continue
+    mapfile -t -O "${#to_list}" to_list < <(eval "$get_maintainer --no-l $patch")
```
kwy95, line 121: Think about using associative arrays.

src/mail.sh
```diff
@@ -761,8 +787,23 @@ function parse_mail_options()
         options_values['NO_INTERACTIVE']=1
         shift
         ;;
+      -[[:digit:]])
+        # if this is given without a commit reference we need to avoid adding a
+        # assumed commit
+        commit_count="$1"
+        commit_ref=1
+        shift
+        ;;
```
isinyaaa, line 853: Does this account for double-digits?

kwy95: Hi, no it does not. I think I added a note about it in the documentation, if not I will later. The problem is that the `get_opt` behavior doesn't allow for single dashes for options with more than one character. Also, the accepted options need to be given to `get_opt` and as far I could find there is no support for globs or patterns. Because of this I added a parser for these options when they are passed after the `--`.

tests/mail_test.sh
```diff
@@ -220,6 +228,23 @@ function test_mail_parser()
...
+
+  output=$(mail_send 'TEST_MODE')
+  expected="git send-email @^"
+  assert_equals_helper 'Testing send without options' "$LINENO" "$output" "$expected"
+
+  parse_mail_options '--to=mail@test.com'
+
+  output=$(mail_send 'TEST_MODE')
+  expected="git send-email --to=mail@test.com @^"
+  assert_equals_helper 'Testing send with to option' "$LINENO" "$output" "$expected"
```
rodrigosiqueira, line 266: Can we have more tests but with the `dry` feature?

documentation/man/features/mail.rst
```diff
@@ -20,15 +20,28 @@ It is common to deal with patch submissions to a mailing list, and
 **mail** functionality aims to wrap this tool to simplify its usage and
 integrate it with other **kw** functionalities.
 
+.. warning::
+  If any *<extra-args>* start with a ``-`` they **need** to be placed after the
+  ``--`` argument.
```
rodrigosiqueira, line 28: The first time that I read it, I got a little bit confused because of this part:
`... [\--] [<rev-range>...] [-v<version>] [<extra-args>...]`
In some way, these `rev-range` and `-v<version>` are extra arguments, or did I miss something?

documentation/man/features/mail.rst
```diff
@@ -20,15 +20,28 @@ It is common to deal with patch submissions to a mailing list, and
...
+
 OPTIONS
 =======
 -s, \--send:
   Send an email using ``git send-email`` to the email adresses specified with
-  ``--to``. This sends the last commit.
+  ``--to``. You can provide *<extra-args>* to be passed directly to
+  ``git send-email``, if they start with a ``-`` they **need** to come after
+  the ``--`` argument.
```
rodrigosiqueira, line 36: Try to expand this documentation a little bit more; in particular, try to describe the `--` behavior a little bit better. I recommend you add a paragraph that describes this send mail feature's core idea and add examples.

documentation/man/features/mail.rst
```diff
@@ -20,15 +20,28 @@ It is common to deal with patch submissions to a mailing list, and
...
 
 \--to='<recipient>,...':
   Specify the recipients of the email being sent.
 
+<rev-range>...:
+  Specify the *<rev-range>*s to be sent. The last commit is taken as the
```
rodrigosiqueira: I think this phrase can have two meanings `Specify the *<rev-range>*s to be sent.`:
1. You are sending configurations to git-send-email.
2. You are adding something for sending the patch.
I got what you mean, but let's try to remove ambiguities.

documentation/man/features/mail.rst
rodrigosiqueira: Also, I expected that `-vX` could be placed before `--`.

src/mail.sh
```diff
@@ -78,17 +78,39 @@ function mail_send()
 {
   local flag="$1"
   local recipients="${options_values['TO']}"
+  local extra_opts="${options_values['PASSTHROUGH']}"
```
rodrigosiqueira, line 83: Maybe we can change `PASSTHROUGH` to something that better portrays that these values will be sent to git-send-email. Maybe `PASS_OPTION_TO_SEND_EMAIL`?

src/mail.sh
```diff
@@ -78,17 +78,39 @@ function mail_send()
...
+# @args: arguments to be processed
+#
+# Returns:
+# 0 is any of the arguments is a valid reference to a commit; 1 otherwise
+function check_revision()
+{
+  local args="$*"
+  local arg
+
+  ! is_inside_work_tree && return 0
```
rodrigosiqueira, line 154: Can we use `[[]]` or an `if` condition? I think this line differs slightly from the standard used in the code.

src/mail.sh
```diff
@@ -78,17 +78,39 @@ function mail_send()
...
+
+  #shellcheck disable=SC2086
+  while read -r arg; do
+    [[ $(git rev-parse --verify -q --end-of-options "$arg^{commit}") ]] && return 0
```
rodrigosiqueira: I think this line deserves to be split into multiple lines, and I would use an `if` statement here. Also, replace `-q` with `--quiet`, and consider adding a brief comment.
I believe this change will make this code more readable.

src/mail.sh
```diff
@@ -78,17 +78,39 @@ function mail_send()
...
+# This function checks if any of the arguments in @args is a valid commit
+# reference
+#
+# @args: arguments to be processed
+#
+# Returns:
+# 0 is any of the arguments is a valid reference to a commit; 1 otherwise
+function check_revision()
+{
+  local args="$*"
```
rodrigosiqueira, line 150: Is it possible to have an empty `args`? If so, what is the default behavior? If it is doing nothing, let's return early in this function.

src/mail.sh
```diff
@@ -78,17 +78,39 @@ function mail_send()
...
+  local arg
+
+  ! is_inside_work_tree && return 0
+
+  #shellcheck disable=SC2086
+  while read -r arg; do
+    [[ $(git rev-parse --verify -q --end-of-options "$arg^{commit}") ]] && return 0
+  done <<< "$(git rev-parse $args 2> /dev/null)"
+
+  return 1
```
rodrigosiqueira, line 167: Are you sure we want to return 1? Maybe 0? 1 automatically put my brain in the EPERM message.

src/mail.sh
rodrigosiqueira: Please, correct me if I'm wrong, but we will not support things like `-10` or `-88`?

I wondered if we can handle this parameter outside the get_opt control, maybe check for this value in the "$@" and if we have it, we parse the info by ourselves and then update the "$@" array.

src/mail.sh
```diff
@@ -761,8 +787,23 @@ function parse_mail_options()
...
         ;;
+      -[[:digit:]])
+        # if this is given without a commit reference we need to avoid adding a
+        # assumed commit
+        commit_count="$1"
+        commit_ref=1
+        shift
+        ;;
+      -v)
+        patch_version="$1$2"
```
rodrigosiqueira, line 861: `"$1$2"`? Could you elaborate a little bit about it? Also, add a comment about that in this part of the code.

tests/mail_test.sh
```diff
@@ -140,6 +140,18 @@ function test_mail_parser()
   assert_equals_helper 'Invalid option passed' "$LINENO" "$ret" 22
 
   # valid options
+  parse_mail_options some -- extra -1 args HEAD^
```
rodrigosiqueira, line 144: Why `some` is valid here? I think it should fail since `some` is not a valid option.

tests/mail_test.sh
```diff
@@ -140,6 +140,18 @@ function test_mail_parser()
...
+  expected='some extra -1 args HEAD^'
+  assert_equals_helper 'Set passthrough options' "$LINENO" "${options_values['PASSTHROUGH']}" "$expected"
+
+  parse_mail_options -1
```
rodrigosiqueira, line 148: Test `-0`, `-10`, `-22`, `-100`?

src/mail.sh
```diff
@@ -774,6 +815,13 @@ function parse_mail_options()
   if [[ "$setup_token" == 1 ]]; then
     validate_setup_opt
   fi
+
+  if [[ "$commit_ref" == 0 ]]; then
+    if ! check_revision "${options_values['PASSTHROUGH']}"; then
+      # assume last commit if none given
+      options_values['PASSTHROUGH']="$(str_strip "${options_values['PASSTHROUGH']} @^")"
```
rodrigosiqueira: I forgot the meaning of `@^`; could you add a comment about that in the code?

src/mail.sh
```diff
@@ -78,17 +78,39 @@ function mail_send()
...
 }
 
+# This function checks if any of the arguments in @args is a valid commit
+# reference
+#
+# @args: arguments to be processed
+#
+# Returns:
+# 0 is any of the arguments is a valid reference to a commit; 1 otherwise
+function check_revision()
```
rodrigosiqueira, line 148: Can we have a specific test for this function? I know you indirectly tested it by using the parser option; however, I think a dedicated test to this function provides documentation on how it is expected to work and allows us to test more things.

src/mail.sh
```diff
@@ -77,18 +77,65 @@ function mail_main()
...
+  fi
+
+  [[ -n "$opts" ]] && cmd+=" $opts"
+  [[ -n "$recipients" ]] && cmd+=" $recipients"
   [[ -n "$extra_opts" ]] && cmd+=" $extra_opts"
 
   cmd_manager "$flag" "$cmd"
 }
 
+function generate_kernel_recipients()
```
rodrigosiqueira, line 106: Add documentation.

src/mail.sh
```diff
@@ -77,18 +77,65 @@ function mail_main()
...
+  local kernel_root="$1"
+  local revision="$2"
+  local -a to_list=()
+  local to=''
+  local -a cc_list=()
+  local cc=''
+  local out=''
+  local patch_cache="${KW_CACHE_DIR}/patches/"
+  local get_maintainer="perl $kernel_root/scripts/get_maintainer.pl --nogit --nogit-fallback --no-r --no-n --multiline --nokeywords --norolestats"
+
```
rodrigosiqueira, line 117: Add a note in the documentation that developers should ensure that we are in the kernel tree before calling this function.

src/mail.sh
```diff
@@ -100,14 +147,21 @@ function check_revision()
...
   ! is_inside_work_tree && return 0
 
   #shellcheck disable=SC2086
   while read -r arg; do
-    [[ $(git rev-parse --verify -q --end-of-options "$arg^{commit}") ]] && return 0
+    [[ $(git rev-parse --verify -q --end-of-options "$arg^{commit}") ]] && rev_list+=" $arg"
   done <<< "$(git rev-parse $args 2> /dev/null)"
 
+  if [[ -n "$rev_list" ]]; then
+    # printf '%s\n' "$(str_strip "$rev_list")" 1>&2
```
rodrigosiqueira, line 162: I think you can drop this comment.

tests/mail_test.sh
```diff
@@ -245,6 +246,8 @@ function test_mail_send()
   local expected
   local output
   local ret
+  # local tocmd="--to-cmd='perl scripts/get_maintainer.pl --no-git --no-l --no-n --no-r --multiline --nokeywords --norolestats'"
+  # local cccmd="--cc-cmd='perl scripts/get_maintainer.pl --no-git --no-m --no-r --multiline --nokeywords --norolestats'"
```
rodrigosiqueira, line 254: ?

rodrigosiqueira (changes requested):

src/mail.sh
```diff
@@ -746,6 +784,7 @@ function mail_help()
     exit
   fi
   printf '%s\n' 'kw mail:' \
+    '  mail (-s | --send) [--to=<recipients>] [--rev=<rev-list>] - Send patches via e-mail' \
```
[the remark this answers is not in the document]
kwy95: I think this is a leftover from a previous iteration.

documentation/man/features/mail.rst
kwy95: Yeah, `<rev-range>` and `-vX` work before and after the `--`, only `<extra-args>` that start with dashes **need** to be placed after the `--`. I will try to clarify this.

src/mail.sh
kwy95: After going through `kw_parse` the argument `-vX` will be split in `-v` and `X`, here I am simply joining them again to pass over to `git send-email`.

tests/mail_test.sh
kwy95: `some` does not start with a dash, so `kw_parse` assumes it is a non-option argument, like a file name, and sends it to the back of the arguments list, after the `--`. This is testing this precise behavior, if something like `--some` is passed instead, then we should see an error. I think there are tests for this already.

tests/mail_test.sh
kwy95: Ok. The current implementation only supports multiple digits after the `--`, but I will add more tests.

src/mail.sh
kwy95: Sure, `@` is just shorthand for `HEAD`.

src/mail.sh
```diff
@@ -31,6 +31,11 @@ function mail_main()
     return 22 # EINVAL
   fi
 
+  if [[ "${options_values['SEND']}" == 1 ]]; then
+    mail_send ''
```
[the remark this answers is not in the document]
kwy95: Shellcheck fails without the argument. This is because this function takes the `flag` as an argument and it is only called here. [SC2120](https://github.com/koalaman/shellcheck/wiki/SC2120)

tests/mail_test.sh
kwy95: I didn't implement this feature, but I could add it to the parser.

src/mail.sh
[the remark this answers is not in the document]
rodrigosiqueira: > If the parameters are optional and you currently just don't want to use them, you can ignore this message. In versions strictly greater than v0.6.0, ignoring SC2120 on a function will also disable SC2119 on each of the call sites.

I think we are in the exception case here. Maybe you can just ignore this shellcheck here.

tests/mail_test.sh
rodrigosiqueira: Maybe you can convert the `TEST_MODE` to something that adds the flag `--dry-run`. I think it would be better to run the command instead of checking if it matches the expected string...

tests/mail_test.sh
rodrigosiqueira: Again, I think we should support -100 before `--` for usability sake.

src/plugins/kw_mail/to-cc-cmd.sh
```diff
@@ -0,0 +1,25 @@
+# This file reads the files generated containing the recipients for any given
+# patch, this is supposed to be used by `git send-email` in the `--to-cmd` and
+# `--cc-cmd`. The user should never call this directly.
```
rodrigosiqueira, line 3: Explain this file and how we suppose to use it. If possible, add some examples.

src/mail.sh
```diff
@@ -149,17 +156,43 @@ function validate_email_list()
...
 
+# This function generates the patches beforehand, these are used to count the
+# number of patches and later to generate the appropriate recipients
+#
+# @rev_list: The list of revisions used to generate the patches
+# @version:  The version of the patches
+#
+# Returns:
+# The count of how many patches were created
+function pre_gen_patches()
```
rodrigosiqueira, line 167: use generate instead of gen.

src/mail.sh
```diff
@@ -149,17 +156,43 @@ function validate_email_list()
...
+{
+  local rev_list="$1"
+  local version="$2"
+  local patch_cache="${KW_CACHE_DIR}/patches"
+  local count=0
+
+  rm -rf "$patch_cache"
+  mkdir -p "$patch_cache"
+
+  eval "git format-patch --quiet --output-directory $patch_cache $version $rev_list"
```
rodrigosiqueira, line 177: Why not use `cmd_manager`?

src/mail.sh
```diff
@@ -149,17 +156,43 @@ function validate_email_list()
...
+# Returns:
+# The count of how many patches were created
+function pre_gen_patches()
+{
+  local rev_list="$1"
+  local version="$2"
+  local patch_cache="${KW_CACHE_DIR}/patches"
+  local count=0
+
+  rm -rf "$patch_cache"
```
rodrigosiqueira, line 174: check `$patch_cache` before run `rm -rf`.

rodrigosiqueira (changes requested):
First of all, thanks a lot for this PR. It looks awesome!

I added a couple of inline comments, just minor things. Aside from that, could you create a tutorial page as part of this PR?

Also, I'm creating a kw blog (see https://github.com/kworkflow/blog). Could you write a blog post that explains how you integrated get maintainer with the send email feature? I think your solution was really nice and deserves to be well documented.

tests/mail_test.sh
```diff
@@ -140,6 +140,25 @@ function test_mail_parser()
...
+
+  parse_mail_options '--cc=some@mail.com'
+  expected='some@mail.com'
+  assert_equals_helper 'Set cc flag' "$LINENO" "${options_values['CC']}" "$expected"
+
+  parse_mail_options '--simulate'
+  expected='--dry-run'
+  assert_equals_helper 'Set simulate flag' "$LINENO" "${options_values['SIMULATE']}" "$expected"
+
+  parse_mail_options '--to=name1@lala.com,name2@lala.xpto,name3 second <name3second@lala.com>'
```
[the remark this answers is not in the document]
kwy95: I added validation in the next commit. Do you think I should squash it with this one? Or maybe reverse their order?

documentation/man/features/mail.rst
```diff
@@ -90,3 +103,7 @@ variables at once::
 To send the last commit as a patch just write::
 
   kw mail --send --to=some@email.com
+
+To send a range of commits the following can be used::
+
+  kw mail -s <SHA1>..<SHA2>
```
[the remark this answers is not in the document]
kwy95: This commit adds the ability to specify the range of commits, hence this being here. I could make a dedicated commit before this, but I did it this way because I use the pass through implementation to give these commit references to `git send-email`

src/mail.sh
```diff
@@ -135,6 +147,55 @@ function validate_email_list()
   return 0
 }
 
+# This function generates the appropriate recipients of the patches and saves
+# them in files to be read by the script passed to `git send-email`
+#
+# @kernel_root: The path to the root of the current kernel tree
+# @rev_list:    The list of revisions used to generate the patches
```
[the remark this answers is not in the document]
kwy95: It's essentially Git jargon for references to commits or ranges of commits.
https://git-scm.com/docs/gitrevisions

src/mail.sh
```diff
@@ -135,6 +147,55 @@ function validate_email_list()
...
+  local kernel_root="$1"
+  local rev_list="$2"
+  local version="$3"
+  local to=''
+  local cc=''
+  local to_list=''
+  local cc_list=''
+  local patch_cache="${KW_CACHE_DIR}/patches"
+  local cover_letter_to="${patch_cache}/to/cover-letter"
+  local cover_letter_cc="${patch_cache}/cc/cover-letter"
```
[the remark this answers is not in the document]
kwy95: The curly braces are used to mark the beginning and end of the variable name, this is essential when trying to expand something like: `${my_var}_suffix` because it ensures bash will not try to expand `$my_var_suffix`. In this case it is not in fact needed, as a slash is not a valid character for a variable name, but I chose to use it here to make it clear what is the variable I'm expanding.

src/mail.sh
```diff
@@ -893,14 +962,19 @@ function parse_mail_options()
...
         # if a reference is passed after the -- we need to account for it
-        [[ "$*" =~ -[[:digit:]]+ ]] && commit_ref=1
-        options_values['PASS_OPTION_TO_SEND_EMAIL']="$(str_strip "$* $patch_version")"
+        if [[ "$*" =~ -[[:digit:]]+ ]]; then
+          commit_count="${BASH_REMATCH[0]}"
+          options_values['REV_LIST']+="$commit_count "
+        fi
+        options_values['PASS_OPTION_TO_SEND_EMAIL']="$(str_strip "$* ${options_values['PATCH_VERSION']}")"
+        options_values['REV_LIST']+="$(check_revision "${options_values['PASS_OPTION_TO_SEND_EMAIL']}")"
+        rev_ret="$?"
```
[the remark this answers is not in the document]
kwy95: This is here because the auto generation of recipients depends on the range of commits the user intends to send, therefore a more robust parsing and explicit separation of said arguments was needed.

src/mail.sh
```diff
@@ -85,14 +87,49 @@ function mail_send()
   flag=${flag:-'SILENT'}
 
   [[ -n "$dryrun" ]] && cmd+=" $dryrun"
-  [[ -n "$to_recipients" ]] && cmd+=" --to=\"$to_recipients\""
-  [[ -n "$cc_recipients" ]] && cmd+=" --cc=\"$cc_recipients\""
+  if [[ -n "$to_recipients" ]]; then
+    validate_email_list "$to_recipients" || exit_msg 'Please check your `--to` list.'
```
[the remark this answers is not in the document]
kwy95: The `validate_email_list` function already prints offending entries from this list. Do you think we should also print the full list?

src/plugins/kw_mail/to-cc-cmd.sh
```diff
@@ -0,0 +1,40 @@
+# This file reads the files generated containing the recipients for any given
```
kwy95, line 1: Add tests to this file.

rodrigosiqueira:
Hi,

Thanks a lot for your endurance in this PR, I know it was tough to make it move forward, but you did an excellent job here :)

I decided to merge your change in the unstable branch, and let's start to use it to see if we catch some bugs.

PR #539, opened 2022-01-13
src: kw_time_and_date: fix get_week_beginning_day

kwy95:
The previous implementation used the '%u' parameter to get the day of
the week of a given date, this returns 7 for sundays, this resulted in
the previous sunday being returned by the function, which is incorrect.
This changes the parameter to '%w', which returns 0 for sundays,
yielding the expected behavior.


rodrigosiqueira:
Merged, thanks.

PR #541, opened 2022-01-14
src: kwlib: improve command_exists check

kwy95:
Use only the first "word" from a command line to check if the command
exists or not.


rodrigosiqueira:
Merged!
Thanks

Issue #158, opened 2020-02-07
Speed up kernel build


rodrigosiqueira:
We should explore more techniques for make `kw build` more powerful. I think we could add support for ccache and maybe tmpfs. Follow some interesting links:
1. https://wiki.archlinux.org/index.php/Ccache
2. https://wiki.debian.org/sbuild#Using_.22ccache.22_with_sbuild
3. https://wiki.debian.org/sbuild#sbuild_overlays_in_tmpfs

joaoseckler:
Possible work breakdown for this issue:
- Provide these features as options in the configuration files;
- Implement setup scripts for ccache and tmpfs;
- Implement the logic for using ccache and tmpfs inside the build functionality of kw;

Benchmark the gain of speed considering different combinations would also be helpful.

isinyaaa:
> We should explore more techniques for make `kw build` more powerful. I think we could add support for ccache and maybe tmpfs. Follow some interesting links:
> 
> 1. https://wiki.archlinux.org/index.php/Ccache
> 2. https://wiki.debian.org/sbuild#Using_.22ccache.22_with_sbuild
> 3. https://wiki.debian.org/sbuild#sbuild_overlays_in_tmpfs

Apparently, distros that use `systemd` (most of them) already have `/tmp` as a [tmpfs by default](https://wiki.archlinux.org/title/Tmpfs#Disable_automatic_mount) (can check that using `findmnt /tmp`. What I think we can do, though, is provide some auto-tweaking options like runtime resizing with

```bash
# mount -o remount,size=$(newsize),noatime /tmp
```

We can also help users with a low memory budget by using [zram or zswap](https://wiki.archlinux.org/title/Improving_performance#zram_or_zswap) which is a lot faster than a normal disk, but still not as good as real memory.

isinyaaa:
> Possible work breakdown for this issue:
> 
> * Provide these features as options in the configuration files;

I agree with that, though it would also be interesting to have some sort of `auto` setting, as we can pretty easily check system specs and autoconfigure some stuff from the get-go.

> * Implement setup scripts for ccache and tmpfs;
> * Implement the logic for using ccache and tmpfs inside the build functionality of kw;

Agree.

> Benchmark the gain of speed considering different combinations would also be helpful.

We could maybe even provide statistics for the users at some point. I'm not sure whether the Pomodoro feature already has something like this, but those could be integrated in some fashion.

isinyaaa:
[...]

PR #543, opened 2022-01-16
Speed up kw kernel build

isinyaaa:
This PR addresses the main points of issue #158, but also some other features that were missing, like support for warnings and the LLVM toolchain.

### DIFF
--- a/documentation/dependencies/debian.dependencies
+++ b/documentation/dependencies/debian.dependencies
@@ -24,3 +24,4 @@ bc
 git-email
 sqlite3
 pv
+ccache
--- a/documentation/man/features/build.rst
+++ b/documentation/man/features/build.rst
@@ -11,6 +11,9 @@ SYNOPSIS
 | *kw* (*b* | *build*) [(-d | \--doc)] [\--alert=(s | v | (sv | vs) | n)]
 | *kw* (*b* | *build*) [(-c | \--cpu-scaling)] <percentage> [\--alert=(s | v | (sv | vs) | n)]
 | *kw* (*b* | *build*) [\--ccache] [\--alert=(s | v | (sv | vs) | n)]
+| *kw* (*b* | *build*) [(-w | \--warnings)] [warning-levels] [\--alert=(s | v | (sv | vs) | n)]
+| *kw* (*b* | *build*) [(-s | \--save-log-to)] <path> [\--alert=(s | v | (sv | vs) | n)]
+| *kw* (*b* | *build*) [\--llvm] [\--alert=(s | v | (sv | vs) | n)]
 
 DESCRIPTION
 ===========
@@ -50,6 +53,18 @@ OPTIONS
 \--ccache:
   This option allows the user to enable ``ccache`` usage during compilation
   tasks, which should improve compile times in subsequent compilations. 
+
+-w, \--warnings=(1 | 2 | 3 | 12 | 13 | 23 | 123):
+  This can be used to enable compilation warnings accordingly. Please check the
+  kernel's ``make help`` for more info.
+
+-s, \--save-log-to=path:
+  This option will save the full compilation log with the enabled warnings to
+  the specified path.
+
+\--llvm:
+  This option can be set to enable the usage of the LLVM toolchain during
+  compilation/linking tasks. It can be better for some applications.
 
 \--alert=(s | v | (sv | vs) | n):
   Defines the alert behaviour upon the command completion.
--- a/documentation/man/features/build.rst
+++ b/documentation/man/features/build.rst
@@ -9,6 +9,8 @@ SYNOPSIS
 | *kw* (*b* | *build*) [(-i | \--info)] [\--alert=(s | v | (sv | vs) | n)]
 | *kw* (*b* | *build*) [(-n | \--menu)] [\--alert=(s | v | (sv | vs) | n)]
 | *kw* (*b* | *build*) [(-d | \--doc)] [\--alert=(s | v | (sv | vs) | n)]
+| *kw* (*b* | *build*) [(-c | \--cpu-scaling)] <percentage> [\--alert=(s | v | (sv | vs) | n)]
+| *kw* (*b* | *build*) [\--ccache] [\--alert=(s | v | (sv | vs) | n)]
 
 DESCRIPTION
 ===========
@@ -40,6 +42,14 @@ OPTIONS
   The doc option provides a mechanism for building the kernel-doc; by default,
   it will build htmldocs. Users can change the default documentation output by
   changing the parameter *doc_type* in the **kworkflow.config** file.
+
+-c, \--cpu-scaling:
+  The cpu-scaling option lets the user set whichever usage they want from their
+  CPU, basically setting the ``-j`` flag accordingly.
+
+\--ccache:
+  This option allows the user to enable ``ccache`` usage during compilation
+  tasks, which should improve compile times in subsequent compilations. 
 
 \--alert=(s | v | (sv | vs) | n):
   Defines the alert behaviour upon the command completion.
--- a/etc/build_template.config
+++ b/etc/build_template.config
@@ -1,6 +1,12 @@
 # Specify the default architecture used by kw
 arch=x86_64
 
+# Specify what percentage of the CPU should be utilized when building
+cpu_scaling_factor=100
+
+# Enable compilation cache
+#enable_ccache=true
+
 # Use this flag as a way to indicate to kw the kernel image name. If this
 # option is not specified, kw will try to infer the correct file inside
 # arch/*/boot/.
--- a/etc/build_template.config
+++ b/etc/build_template.config
@@ -7,6 +7,15 @@ cpu_scaling_factor=100
 # Enable compilation cache
 #enable_ccache=true
 
+# Use the LLVM toolchain
+#use_llvm=true
+
+# Set default warning level for compilation tasks
+#warning_level=1
+
+# Path kw should be saving `make` output to
+#log_path=./output.log
+
 # Use this flag as a way to indicate to kw the kernel image name. If this
 # option is not specified, kw will try to infer the correct file inside
 # arch/*/boot/.
--- a/etc/build_template.config
+++ b/etc/build_template.config
@@ -1,0 +1,17 @@
+# Specify the default architecture used by kw
+arch=x86_64
+
+# Use this flag as a way to indicate to kw the kernel image name. If this
+# option is not specified, kw will try to infer the correct file inside
+# arch/*/boot/.
+kernel_img_name=bzImage
+
+# Specify cross-compile toolchain
+cross_compile=aarch64-linux-gnu-
+
+# Default kernel menu config option
+menu_config=nconfig
+
+# Generate kernel-doc, user can use any option currently available in the
+# Kernel (e.g., pdfdocs, latexdocs, cleandocs, etc.)
+doc_type=htmldocs
--- a/src/build.sh
+++ b/src/build.sh
@@ -50,8 +50,10 @@ function kernel_build()
   local doc_type
 
   parse_build_options "$@"
-  if [[ "$?" != 0 ]]; then
-    exit 22 # EINVAL
+  if [[ "$?" -gt 0 ]]; then
+    complain "Invalid option: ${options_values['ERROR']}"
+    build_help
+    return 22 # EINVAL
   fi
 
   cross_compile="${options_values['CROSS_COMPILE']}"
@@ -119,13 +121,11 @@ function parse_build_options()
   local short_options='h,i,n,d'
   local doc_type
 
-  options="$(getopt \
-    --name 'kw build' \
-    --options "$short_options" \
-    --longoptions "$long_options" \
-    -- "$@")"
+  options=$(kw_parse "$short_options" "$long_options" "$@")
 
   if [[ "$?" != 0 ]]; then
+    options_values['ERROR']="$(kw_parse_get_errors 'kw build' "$short_options" \
+      "$long_options" "$@")"
     return 22 # EINVAL
   fi
 
@@ -136,8 +136,6 @@ function parse_build_options()
   options_values['PARALLEL_CORES']=1
   options_values['INFO']=''
   options_values['DOC_TYPE']=''
-
-  eval "set -- $options"
 
   while [[ "$#" -gt 0 ]]; do
     case "$1" in
@@ -161,8 +159,8 @@ function parse_build_options()
         shift
         ;;
       *)
-        complain "Invalid option: $option"
-        exit 22 # EINVAL
+        options_values['ERROR']="$1"
+        return 22
         ;;
     esac
   done
--- a/src/build.sh
+++ b/src/build.sh
@@ -6,7 +6,6 @@ declare -gA options_values
 
 # This function retrieves and prints information related to the kernel that
 # will be compiled.
-# shellcheck disable=2120
 function build_info()
 {
   local flag
@@ -63,7 +62,7 @@ function kernel_build()
   doc_type=${options_values['DOC_TYPE']}
 
   if [[ -n "${options_values['INFO']}" ]]; then
-    build_info
+    build_info ''
     exit
   fi
 
--- a/src/build.sh
+++ b/src/build.sh
@@ -45,8 +45,9 @@ function kernel_build()
   local cross_compile
   local arch
   local menu_config
-  local parallel_cores
   local doc_type
+  local optimizations
+  local cpu_scaling_factor
 
   parse_build_options "$@"
   if [[ "$?" -gt 0 ]]; then
@@ -60,6 +61,7 @@ function kernel_build()
   menu_config=${options_values['MENU_CONFIG']}
   parallel_cores=${options_values['PARALLEL_CORES']}
   doc_type=${options_values['DOC_TYPE']}
+  cpu_scaling_factor=${options_values['CPU_SCALING_FACTOR']}
 
   if [[ -n "${options_values['INFO']}" ]]; then
     build_info ''
@@ -81,22 +83,28 @@ function kernel_build()
     exit 125 # ECANCELED
   fi
 
-  if [ -x "$(command -v nproc)" ]; then
-    parallel_cores=$(nproc --all)
+  if command_exists nproc; then
+    parallel_cores="$(nproc --all)"
   else
-    parallel_cores=$(grep -c ^processor /proc/cpuinfo)
+    parallel_cores="$(grep -c ^processor /proc/cpuinfo)"
+  fi
+
+  optimizations="-j$((parallel_cores * cpu_scaling_factor / 100))"
+
+  if [[ -n "${options_values['CCACHE']}" ]]; then
+    optimizations="CC=\"ccache gcc -fdiagnostics-color\" $optimizations"
   fi
 
   if [[ -n "$doc_type" ]]; then
-    command="make -j$parallel_cores $doc_type"
+    command="make $optimizations $doc_type"
     cmd_manager "$flag" "$command"
     return
   fi
 
+  command="make $optimizations ARCH=$arch $cross_compile"
+
   # Let's avoid menu question by default
   cmd_manager "$flag" "make -j ARCH=$arch $cross_compile olddefconfig --silent"
-
-  command="make -j$parallel_cores ARCH=$arch $cross_compile"
 
   start=$(date +%s)
   cmd_manager "$flag" "$command"
@@ -160,8 +168,8 @@ function load_build_config()
 
 function parse_build_options()
 {
-  local long_options='help,info,menu,doc'
-  local short_options='h,i,n,d'
+  local long_options='help,info,menu,doc,ccache,cpu-scaling:'
+  local short_options='h,i,n,d,c:'
   local doc_type
 
   options=$(kw_parse "$short_options" "$long_options" "$@")
@@ -177,7 +185,8 @@ function parse_build_options()
   options_values['ARCH']="${build_config[arch]:-$arch_fallback}"
   options_values['MENU_CONFIG']=''
   options_values['CROSS_COMPILE']="${build_config[cross_compile]}"
-  options_values['PARALLEL_CORES']=1
+  options_values['CCACHE']="${build_config[ccache]}"
+  options_values['CPU_SCALING_FACTOR']="${build_config[cpu_scaling_factor]:-100}"
   options_values['INFO']=''
   options_values['DOC_TYPE']=''
 
@@ -196,6 +205,21 @@ function parse_build_options()
         options_values['MENU_CONFIG']="${build_config[menu_config]:-$menu_fallback}"
         shift
         ;;
+      --cpu-scaling | -c)
+        if [[ ! "$2" =~ [0-9]+ ]]; then
+          options_values['ERROR']="$2"
+          return 22 # EINVAL
+        fi
+        if [[ "$2" -gt 100 ]]; then
+          complain 'Upscaling CPU performance in compilation tasks may have unintended consequences!'
+        fi
+        options_values['CPU_SCALING_FACTOR']="$2"
+        shift 2
+        ;;
+      --ccache)
+        options_values['CCACHE']=1
+        shift
+        ;;
       --doc | -d)
         doc_type_fallback="${configurations[doc_type]:-htmldocs}"
         options_values['DOC_TYPE']="${build_config[doc_type]:-$doc_type_fallback}"
@@ -206,7 +230,7 @@ function parse_build_options()
         ;;
       *)
         options_values['ERROR']="$1"
-        return 22
+        return 22 # EINVAL
         ;;
     esac
   done
@@ -224,7 +248,9 @@ function build_help()
     '  build - Build kernel' \
     '  build (-n | --menu) - Open kernel menu config' \
     '  build (-i | --info) - Display build information' \
-    '  build (-d | --doc) - Build kernel documentation'
+    '  build (-d | --doc) - Build kernel documentation' \
+    '  build (-c | --cpu-scaling) <percentage> - Scale CPU usage by factor' \
+    '  build (--ccache) - Enable use of ccache'
 }
 
 # This function is used to show the current set up used by kworkflow.
@@ -250,6 +276,8 @@ function show_build_variables()
 
   local -Ar build=(
     [arch]='Target arch'
+    [cpu_scaling_factor]='CPU scaling factor'
+    [enable_ccache]='Enable ccache'
     [kernel_img_name]='Kernel image name'
     [cross_compile]='Cross-compile name'
     [menu_config]='Kernel menu config'
--- a/src/build.sh
+++ b/src/build.sh
@@ -44,6 +44,7 @@ function kernel_build()
   local end
   local cross_compile
   local arch
+  local platform_ops
   local menu_config
   local doc_type
   local optimizations
@@ -69,11 +70,13 @@ function kernel_build()
   fi
 
   if [[ -n "$cross_compile" ]]; then
-    cross_compile="CROSS_COMPILE=$cross_compile"
+    cross_compile=" CROSS_COMPILE=$cross_compile "
   fi
 
+  platform_ops="${arch}${cross_compile}"
+
   if [[ -n "$menu_config" ]]; then
-    command="make -j ARCH=$arch $cross_compile $menu_config"
+    command="make -j ARCH=${platform_ops}${menu_config}"
     cmd_manager "$flag" "$command"
     return
   fi
@@ -101,10 +104,10 @@ function kernel_build()
     return
   fi
 
-  command="make $optimizations ARCH=$arch $cross_compile"
+  command="make $optimizations ARCH=$platform_ops"
 
   # Let's avoid menu question by default
-  cmd_manager "$flag" "make -j ARCH=$arch $cross_compile olddefconfig --silent"
+  cmd_manager "$flag" "make -j olddefconfig --silent ARCH=$platform_ops"
 
   start=$(date +%s)
   cmd_manager "$flag" "$command"
--- a/src/build.sh
+++ b/src/build.sh
@@ -47,8 +47,10 @@ function kernel_build()
   local platform_ops
   local menu_config
   local doc_type
+  local llvm
   local optimizations
   local cpu_scaling_factor
+  local output_path
 
   parse_build_options "$@"
   if [[ "$?" -gt 0 ]]; then
@@ -63,6 +65,9 @@ function kernel_build()
   parallel_cores=${options_values['PARALLEL_CORES']}
   doc_type=${options_values['DOC_TYPE']}
   cpu_scaling_factor=${options_values['CPU_SCALING_FACTOR']}
+  warnings=${options_values['WARNINGS']}
+  llvm=${options_values['USE_LLVM_TOOLCHAIN']}
+  output_path=${options_values['LOG_PATH']}
 
   if [[ -n "${options_values['INFO']}" ]]; then
     build_info ''
@@ -86,6 +91,18 @@ function kernel_build()
     exit 125 # ECANCELED
   fi
 
+  if [[ -n "$warnings" ]]; then
+    warnings=" W=$warnings"
+  fi
+
+  if [[ -n "$output_path" ]]; then
+    output_path=" 2>&1 > $output_path"
+  fi
+
+  if [[ -n "$llvm" ]]; then
+    llvm='LLVM=1 '
+  fi
+
   if command_exists nproc; then
     parallel_cores="$(nproc --all)"
   else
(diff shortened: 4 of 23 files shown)


src/build.sh
```diff
@@ -89,12 +89,18 @@ function kernel_build()
   fi
 
   if [ -x "$(command -v nproc)" ]; then
-    parallel_cores=$(nproc --all)
```
rodrigosiqueira, line 90: Since you are working on this part of the code, we can replace `-x "$(command ...` with `command_exists`.

src/build.sh
```diff
@@ -89,12 +89,18 @@ function kernel_build()
...
+    parallel_cores="-j$(nproc --all)"
   else
-    parallel_cores=$(grep -c ^processor /proc/cpuinfo)
+    parallel_cores="-j$(grep -c ^processor /proc/cpuinfo)"
   fi
 
-  command="make -j$parallel_cores ARCH=$arch $cross_compile"
+  if [[ -n "${options_values['OPTIMIZE']}" ]]; then
+    optimizations="CC=\"ccache gcc -fdiagnostics-color\" $parallel_cores"
```
rodrigosiqueira: I would prefer that we handle `ccache` as a separate option. We can have an entry in the `etc/kworkflow_template.config` to enable it (e.g., `enable_ccache=no`). We can also have a new option under the `kw b` named `--ccache`. Finally, you need to update files in the folder `documentation/dependencies/` to make ccache explicit kw dependencies.

src/build.sh
```diff
@@ -89,12 +89,18 @@ function kernel_build()
...
+  else
+    optimizations="$parallel_cores"
```
rodrigosiqueira, line 116: I think we should also allow users to say how many percent of their CPU they want to utilize, for example, something like `kw build --cpu-utilization "50%"`. I want full-speed compilation most of the time, but sometimes I want to keep some CPU resources for other tasks.

src/build.sh
```diff
@@ -130,7 +136,7 @@ function parse_build_options()
   options_values['ARCH']="${configurations[arch]:-'x86_64'}"
   options_values['MENU_CONFIG']=''
   options_values['CROSS_COMPILE']="${configurations[cross_compile]}"
-  options_values['PARALLEL_CORES']=1
+  options_values['OPTIMIZE']="${configurations[optimize]}"
```
rodrigosiqueira, line 155: The associative array `configurations` is associated with `kworkflow.config` file, in this sense, you need to add `optimize` in the `etc/kworkflow_template.config`. Also, we can use this option to store user-specific flags.

src/build.sh
```diff
@@ -47,6 +47,7 @@ function kernel_build()
   local arch
   local menu_config
   local doc_type
+  local llvm
```
rodrigosiqueira, line 50: This is a very nice commit :)
Let's add these compiler options in the `kworkflow.config` file.

rodrigosiqueira (changes requested):
First of all, thanks a lot for your PR; I really like your PR, and it gave me some ideas.

How about making the build option more flexible? Let's create a new config file dedicated to the compilation setup, something like this:

.kw/build.config

In this config file, we can set things like:

* Optimization flags
* ccache
* Warning level
* Target compiler

This code refactor can make your work with the env option a little bit easier.

src/build.sh
isinyaaa: But do you think a percentage would work alright? How do you propose we make those calculations?

rodrigosiqueira:
Hi @isinyaaa any progress here?

isinyaaa:
Hi, @rodrigosiqueira, I should be getting back to this PR in the next few weeks. I'm sorry for the slow pacing!

etc/build_template.config
```diff
@@ -0,0 +1,29 @@
...
+
+# Set default warning level for compilation tasks
+#warning_level=1
+
+# Use this flag as a way to indicate to kw the kernel image name. If this
+# option is not specified, kw will try to infer the correct file inside
+# arch/*/boot/.
+kernel_img_name=bzImage
+
+# Specify cross-compile name
```
rodrigosiqueira, line 21: Could you replace "name" with "toolchain"?

src/config_manager.sh
```diff
@@ -287,11 +287,11 @@ function get_config_from_defconfig()
...
   fi
 
-  cross_compile=${configurations[cross_compile]}
-  arch=${configurations[arch]}
+  cross_compile=${build_config[cross_compile]}
+  arch=${build_config[arch]}
   # TODO: Right now, the below variable has no meaning and will be empty. Let's
   # expand the config file to enable developers to define ASIC targets.
-  asic=${configurations[asic]}
+  asic=${build_config[asic]}
```
rodrigosiqueira, line 294: tbh, I don't remember why I added it, and this is an excellent example of why we should open issues haha. Let's drop these lines.

src/kw_config_loader.sh
```diff
@@ -159,11 +183,11 @@ function load_configuration()
...
   local -a config_dirs
   local config_dirs_size
-  local IFS=:
-  read -ra config_dirs <<< "${XDG_CONFIG_DIRS:-"/etc/xdg"}"
-  unset IFS
+
+  IFS=: read -ra config_dirs <<< "${XDG_CONFIG_DIRS:-"/etc/xdg"}"
 
   parse_configuration "$KW_ETC_DIR/$CONFIG_FILENAME"
+  parse_configuration "$KW_ETC_DIR/$BUILD_CONFIG_FILENAME" build_config
```
rodrigosiqueira, line 190: What if we move the `build.config` load to `src/build.sh`? In the current implementation, we are loading all config every time that we include this file (`kw_cionfig_loader`). Maybe we can make the load option per feature and save some CPU cycles.

documentation/dependencies/debian.dependencies
```diff
@@ -24,3 +24,4 @@ bc
 git-email
 sqlite3
 pv
+ccache
```
rodrigosiqueira, line 27: We now have `fedora.dependencies` as well.

src/build.sh
```diff
@@ -151,6 +162,20 @@ function parse_build_options()
         options_values['MENU_CONFIG']="${build_config[menu_config]:-nconfig}"
         shift
         ;;
+      --cpu-util | --cpu-scale | -c)
```
rodrigosiqueira, line 183: Why do we want two names here?
Also, you forgot to update the help function and the man page.

src/build.sh
```diff
@@ -151,6 +162,20 @@ function parse_build_options()
...
+        [[ ! "$2" =~ [0-9]+ ]] && {
+          options_values['ERROR']="$2"
+          return 22
+        }
```
rodrigosiqueira, line 187: Let's use the standard if here. Also, add the `# ENVAL` comment after 22.

src/build.sh
```diff
@@ -151,6 +162,20 @@ function parse_build_options()
...
+        [ "$2" -gt 100 ] &&
```
rodrigosiqueira, line 188: Same as the previous comment, and why not use `[[ .. ]]`?

tests/build_test.sh
```diff
@@ -148,8 +148,14 @@ function test_kernel_build_x86()
...
   )
-
   compare_command_sequence '' "$LINENO" 'expected_cmd' "$output"
+
+  output=$(kernel_build 'TEST_MODE' --ccache | tail -n +1 | head -2) # Remove statistics output
+  declare -a expected_cmd=(
+    "make -j ARCH=x86_64  olddefconfig --silent"
+    "make CC=\"ccache gcc -fdiagnostics-color\" -j$PARALLEL_CORES ARCH=x86_64"
+  )
+  compare_command_sequence '' "($LINENO)" "expected_cmd" "$output"
```
rodrigosiqueira, line 158: Create a separate test function for that.\ and add a test for `--cpu-scale` scenario.

src/build.sh
```diff
@@ -70,11 +70,11 @@ function kernel_build()
   fi
 
   if [[ -n "$cross_compile" ]]; then
-    cross_compile="CROSS_COMPILE=$cross_compile"
+    cross_compile=" CROSS_COMPILE=$cross_compile "
   fi
 
   if [[ -n "$menu_config" ]]; then
-    command="make -j ARCH=$arch $cross_compile $menu_config"
+    command="make -j ARCH=$arch$cross_compile$menu_config"
```
rodrigosiqueira, line 80: Since you are concatenating variables, let's use: `${arch}${cross_compile}${menu_config}`.
(same for this commit)

src/build.sh
```diff
@@ -99,12 +110,17 @@ function kernel_build()
   fi
 
   if [[ -n "${options_values['CCACHE']}" ]]; then
-    optimizations="CC=\"ccache gcc -fdiagnostics-color\" $parallel_cores"
+    if [[ -n "$llvm" ]]; then
+      ccache='CC="ccache clang -fdiagnostics-color"'
+    else
+      ccache='CC="ccache gcc -fdiagnostics-color"'
+    fi
```
rodrigosiqueira, line 117: Maybe to avoid coding duplication, we could have a' compiler' variable that uses gcc by default (and you can set `clang` inside the `$llvm` condition). After that, you just need to replace the default compiler and build the `ccache` based on that (e.g., `ccache $compiler -fdiagnostics-color"`).

src/build.sh
```diff
@@ -176,10 +194,28 @@ function parse_build_options()
...
+        ;;
+      --optimize | -o)
+        options_values['OPTIMIZE']=1
+        shift
+        ;;
       --doc | -d)
         options_values['DOC_TYPE']="${build_config[doc_type]:-htmldocs}"
         shift
         ;;
+      --warnings | -w)
```
rodrigosiqueira, line 209: Do not forget to update the help and man.

rodrigosiqueira (changes requested):
Hi @isinyaaa ,
First of all, thanks a lot for this PR! This is a very exciting change, and this change will unblock my work of enabling deploy support to ChromeOS. :)

I added a couple of comments, but could you add one extra commit redirecting all compilation output to a file? Something like `kw b -w 1 --save-log-to <PATH>`.
This could be useful for newcomers looking for new contributions, such as fixing compilation warnings.

src/build.sh
isinyaaa: I feel that using bash syntax for something as simple is kinda unnecessary (i.e. `[[` is a bashism, and not a simple alternative to `[`)

src/build.sh
isinyaaa: really nice suggestion! thanks :)

src/kw_config_loader.sh
isinyaaa: took me a while to do this, and I think some tests didn't end up as they were supposed to. Maybe if we could merge the test fixing PR before this one it'd help in this review.

src/build.sh
isinyaaa: I'm leaving docs for last, sorry about that.

tests/build_test.sh
```diff
@@ -356,7 +385,7 @@ function test_parse_build_options()
   assert_equals_helper 'Default ARCH did not match expectation' "($LINENO)" 'x86_64' "${options_values['ARCH']}"
   assert_equals_helper 'Default MENU_CONFIG did not match expectation' "($LINENO)" '' "${options_values['MENU_CONFIG']}"
   assert_equals_helper 'Default CROSS_COMPILE did not match expectation' "($LINENO)" '' "${options_values['CROSS_COMPILE']}"
-  assert_equals_helper 'Default PARALLEL_CORES did not match expectation' "($LINENO)" '1' "${options_values['PARALLEL_CORES']}"
```
isinyaaa: TODO: change those assertions

src/build.sh
```diff
@@ -86,27 +91,39 @@ function kernel_build()
...
+  fi
+
   if command_exists nproc; then
     parallel_cores="$(nproc --all)"
   else
     parallel_cores="$(grep -c ^processor /proc/cpuinfo)"
   fi
 
-  parallel_cores="-j$((parallel_cores * cpu_scaling_factor / 100))"
+  optimizations="-j$((parallel_cores * cpu_scaling_factor / 100))"
```
isinyaaa, line 112: TODO: put that in the optimizations commit

src/build.sh
```diff
@@ -86,27 +91,39 @@ function kernel_build()
...
+
+  if [[ -n "${options_values['CCACHE']}" ]]; then
+    compiler='gcc'
+    [[ -n "$llvm" ]] && compiler='clang'
+    optimizations="CC=\"ccache $compiler -fdiagnostics-color\" $optimizations"
+  fi
 
   if [[ -n "$doc_type" ]]; then
-    command="make $parallel_cores $doc_type"
+    command="make $optimizations $doc_type"
```
isinyaaa, line 121: also this

src/build.sh
```diff
@@ -147,6 +164,9 @@ function parse_build_options()
   options_values['CROSS_COMPILE']="${build_config[cross_compile]}"
   options_values['CCACHE']="${build_config[ccache]}"
   options_values['CPU_SCALING_FACTOR']="${build_config[cpu_scaling_factor]:-100}"
+  options_values['USE_LLVM_TOOLCHAIN']="${build_config[use_llvm_toolchain]}"
+  options_values['SAVE_LOG_TO']="${build_config[save_log_to]}"
```
isinyaaa, line 168: TODO: add this option to the templates and also test it

src/build.sh
```diff
@@ -147,6 +164,9 @@ function parse_build_options()
   options_values['CROSS_COMPILE']="${build_config[cross_compile]}"
   options_values['CCACHE']="${build_config[ccache]}"
   options_values['CPU_SCALING_FACTOR']="${build_config[cpu_scaling_factor]:-100}"
+  options_values['USE_LLVM_TOOLCHAIN']="${build_config[use_llvm_toolchain]}"
```
isinyaaa, line 167: TODO: test the LLVM, warnings and option composition.

isinyaaa:
I think this PR is pretty much ready, now, as I've resolved all remaining TODOs. Looking forward to your reviews :)

rodrigosiqueira:
Hi @isinyaaa ,
Could you rebase this branch with the latest version of the unstable branch?

isinyaaa:
@rodrigosiqueira Done ;)

tests/build_test.sh
```diff
@@ -53,8 +64,189 @@ tearDown()
...
+  parse_configuration "$TMP_DIR/build.config" build_config
+  assertConfigurations build_config expected_build_configurations "$LINENO"
+
+  true # Reset return value
+}
+
+function assertConfigurations()
+{
+  declare -n configurations_ref=$1
+  declare -n expected_configurations_ref=$2
```
rodrigosiqueira, line 98: hmmm... don't we need `""` around `$1` and `$2`?

tests/build_test.sh
```diff
@@ -53,8 +64,189 @@ tearDown()
...
+  local lineno=${3:-LINENO}
```
rodrigosiqueira, line 99: I think `LINENO` here will not work as expected; basically, it will point to this line instead of where `assertConfigurations` was invoked.

tests/build_test.sh
```diff
@@ -53,8 +64,189 @@ tearDown()
...
+
+  # check if configurations is contained in expected_configurations
+  for k in "${!configurations_ref[@]}"; do
+    if [[ ${expected_configurations_ref[$k]+token} != token ]]; then
```
rodrigosiqueira, line 103: `+token`? That's new for me. Could you elaborate on this?

tests/build_test.sh
```diff
@@ -53,8 +64,189 @@ tearDown()
...
+    fi
+  done
+}
+
+# To test the order of config file loading, we will put a file named
+# kworkflow.config in each place, in order, and remove the previous one.
+# The order is: PWD, XDG_CONFIG_HOME, XDG_CONFIG_DIRS, KW_ETC_DIR
+function test_parse_configuration_files_loading_order()
+{
+  build_config=()
```
rodrigosiqueira, line 123: Move this variable, after the `local` variables.

tests/build_test.sh
```diff
@@ -53,8 +64,189 @@ tearDown()
...
+
+  local expected
+  local original_dir="$PWD"
+
+  KW_ETC_DIR='1'
+  XDG_CONFIG_DIRS='2:3:4'
+  XDG_CONFIG_HOME='5'
```
rodrigosiqueira, line 130: You probably want to restore the original values in the `setUp` function to avoid weird results in other tests.

tests/build_test.sh
```diff
@@ -53,8 +64,189 @@ tearDown()
...
+    "3/$KWORKFLOW/$BUILD_CONFIG_FILENAME"
+    "2/$KWORKFLOW/$BUILD_CONFIG_FILENAME"
+    "5/$KWORKFLOW/$BUILD_CONFIG_FILENAME"
+  )
+
+  output="$(
+    function parse_configuration()
+    {
+      printf '%s\n' "$1"
+    }
```
rodrigosiqueira, line 144: Is this a mock function?

tests/build_test.sh
```diff
@@ -53,8 +64,189 @@ tearDown()
...
+{
+  build_config=()
+
+  local -A shown_options
+  local -A possible_options
+  local output
+
+  # get all assigned options, including commented ones
+  # remove #'s and ='s to get option names
+  output="$(cat "$original_dir/etc/build_template.config")"
```
rodrigosiqueira, line 183: You can use this optimization:
https://kworkflow.org/content/codingstyle.html#command-substitution-and-arithmetic-expression

tests/build_test.sh
```diff
@@ -53,8 +64,189 @@ tearDown()
...
+{
+  build_config=()
+
+  local current_path="$PWD"
+  local -a expected
+
+  function mock_parse_configuration()
+  {
+    :
+  }
```
rodrigosiqueira, line 220: Do we need this?

tests/build_test.sh
```diff
@@ -53,8 +64,189 @@ tearDown()
...
+  for option in "${!shown_options[@]}"; do
+    if [[ ! -v possible_options["$option"] ]]; then
+      fail "($LINENO): show_variable is showing $option not present in kworkflow_template.config"
+    fi
+  done
+}
+
+function test_load_build_config()
+{
+  build_config=()
```
rodrigosiqueira, line 212: move this after the `local` variables

tests/kw_config_loader_test.sh
```diff
@@ -16,12 +16,13 @@ function setUp()
 
 function tearDown()
 {
-  rm -rf "$TMP_DIR"
+  rm -rf "$TMP_DIR" "$SHUNIT_TMPDIR"
```
rodrigosiqueira, line 19: Check `$TMP_DIR` path before removing it, just in case...

documentation/man/features/build.rst
```diff
@@ -9,6 +9,8 @@ SYNOPSIS
 | *kw* (*b* | *build*) [(-i | \--info)] [\--alert=(s | v | (sv | vs) | n)]
 | *kw* (*b* | *build*) [(-n | \--menu)] [\--alert=(s | v | (sv | vs) | n)]
 | *kw* (*b* | *build*) [(-d | \--doc)] [\--alert=(s | v | (sv | vs) | n)]
+| *kw* (*b* | *build*) [(-c | \--cpu-scaling)] <percentage> [\--alert=(s | v | (sv | vs) | n)]
```
rodrigosiqueira, line 12: I think it should be:
`[(-c | \--cpu-scaling) <percentage>]`
@kwy95 ?

src/build.sh
```diff
@@ -81,23 +83,29 @@ function kernel_build()
...
+
+  optimizations="-j$((parallel_cores * cpu_scaling_factor / 100))"
+
+  if [[ -n "${options_values['CCACHE']}" ]]; then
+    optimizations="CC=\"ccache gcc -fdiagnostics-color\" $optimizations"
   fi
 
   if [[ -n "$doc_type" ]]; then
-    command="make -j$parallel_cores $doc_type"
+    command="make $optimizations $doc_type"
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/build.sh
```diff
@@ -81,23 +83,29 @@ function kernel_build()
...
     cmd_manager "$flag" "$command"
     return
   fi
 
+  command="make $optimizations ARCH=$arch $cross_compile"
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#string-concatenation
(same for all commits)

src/build.sh
```diff
@@ -150,6 +159,21 @@ function parse_build_options()
...
         shift
         ;;
+      --cpu-scaling | -c)
+        if [[ ! "$2" =~ [0-9]+ ]]; then
+          options_values['ERROR']="$2"
+          return 22 # EINVAL
+        fi
+        if [[ "$2" -gt 100 ]]; then
+          complain 'Upscaling CPU performance in compilation tasks may have unintended consequences!'
+        fi
```
rodrigosiqueira, line 194: @kwy95 I think we discussed something like this in the `kw mail`. Could you review this part?

src/build.sh
```diff
@@ -209,7 +233,9 @@ function build_help()
     '  build - Build kernel' \
     '  build (-n | --menu) - Open kernel menu config' \
     '  build (-i | --info) - Display build information' \
-    '  build (-d | --doc) - Build kernel documentation'
+    '  build (-d | --doc) - Build kernel documentation' \
+    '  build (-c | --cpu-scaling) <percentage> - Scale CPU usage by factor' \
+    '  build (--ccache) - Enable use of ccache'
```
rodrigosiqueira: Just for curiosity, what will happen in these scenarios:
1. `kw b -n -c`
2. `kw b -n --ccache`
3. `kw b -d --ccache`
4. `kw b -i -c`
5. `kw b -i --ccache`

tests/build_test.sh
```diff
@@ -340,10 +345,42 @@ function test_kernel_build_x86()
...
 }
 
+function test_kernel_isolated_build_options()
+{
+  local expected_result
+  local output
+
+  output=$(kernel_build 'TEST_MODE' --ccache | tail -n +1 | head -2)
+  declare -a expected_cmd=(
+    "make -j ARCH=x86_64  olddefconfig --silent"
```
rodrigosiqueira: Use single quotes.

tests/build_test.sh
```diff
@@ -340,10 +345,42 @@ function test_kernel_build_x86()
...
+
+  output=$(kernel_build 'TEST_MODE' --cpu-scaling 50 | tail -n +1 | head -2)
+  declare -a expected_cmd=(
+    "make -j ARCH=x86_64  olddefconfig --silent"
+    "make -j$SCALING ARCH=x86_64"
+  )
+  compare_command_sequence '' "($LINENO)" "expected_cmd" "$output"
+}
+
+function test_kernel_2composed_build_options()
```
rodrigosiqueira: `2` is a typo?

tests/build_test.sh
```diff
@@ -340,10 +345,42 @@ function test_kernel_build_x86()
...
+{
+  local expected_result
+  local output
+
+  output=$(kernel_build 'TEST_MODE' --ccache --cpu-scaling 50 | tail -n +1 | head -2)
+  declare -a expected_cmd=(
+    "make -j ARCH=x86_64  olddefconfig --silent"
```
rodrigosiqueira: ''

src/build.sh
```diff
@@ -69,11 +70,13 @@ function kernel_build()
   fi
 
   if [[ -n "$cross_compile" ]]; then
-    cross_compile="CROSS_COMPILE=$cross_compile"
+    cross_compile=" CROSS_COMPILE=$cross_compile "
```
rodrigosiqueira, line 78: https://kworkflow.org/content/codingstyle.html#string-concatenation

documentation/man/features/build.rst
```diff
@@ -51,6 +54,18 @@ OPTIONS
   This option allows the user to enable ``ccache`` usage during compilation
   tasks, which should improve compile times in subsequent compilations. 
 
+-w, \--warnings=(1 | 2 | 3 | 12 | 13 | 23 | 123):
+  This can be used to enable compilation warnings accordingly. Please check the
+  kernel's ``make help`` for more info.
```
rodrigosiqueira, line 59: I think it will be more friendly if you explain the warning level 1 and 2... Actually, 1 and 2 are hard to understand; maybe we could have a string alias that is easy to understand.

documentation/man/features/build.rst
```diff
@@ -51,6 +54,18 @@ OPTIONS
...
+
+-s, \--save-log-to=path:
```
rodrigosiqueira, line 61: Do we need `=`? I think it would be better if we didn't need it.

documentation/man/features/build.rst
```diff
@@ -51,6 +54,18 @@ OPTIONS
...
+  This option will save the full compilation log with the enabled warnings to
+  the specified path.
```
rodrigosiqueira, line 63: Can we create a default file if the user does not provide a path? Also, we should document this behavior here.

rodrigosiqueira (changes requested):
Hi @isinyaaa ,
Thanks a lot for your PR; that's an awesome change!

src/build.sh
```diff
@@ -166,6 +166,38 @@ function parse_build_options()
...
 
+# This function loads the kw build configuration files into memory, populating
+# the $build_config hashtable. The files are parsed in a specific order,
+# allowing higher level setting definitions to overwrite lower level ones.
+function load_build_config()
+{
+  local -a config_dirs
+  local config_dirs_size
+
+  IFS=: read -ra config_dirs <<< "${XDG_CONFIG_DIRS:-"/etc/xdg"}"
```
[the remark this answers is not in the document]
isinyaaa: from what I understand, it should split the `XDG_CONFIG_DIRS` string and save each entry separately in the `config_dirs` array. Then, if there's no such variable, use the `/etc/xdg` as default.

As you noted, I feel like this code could improve, but the original one is like this as well, so I just copy pasted it, actually.

tests/build_test.sh
isinyaaa: Actually, this is copy pasted, so I'm not sure about much of its workings, but I can try to improve it.

tests/build_test.sh
isinyaaa: I think it might actually work that way because `lineno=LINENO` but then you'd have to replace it twice when you use it, like `${!lineno}` (like a double pointer).

tests/build_test.sh
isinyaaa: I've done a quick research on this code, and could understand the following:
1. It was introduced by @matheustavares in 8e716c50, and
2. It aims at generalizing the process you use to verify those config arrays, the basic idea being that:
  1. You access array B using a key `$k` that was taken from array A.
  2. If B has such a key, then you end up with the string "token", otherwise it's empty (which is obviously different from "token" on the right hand side).
  3. Now you can check every key in the array easily.

Truly an ingenious solution!

tests/build_test.sh
isinyaaa: seems like it

rodrigosiqueira:
Hi @isinyaaa , could you update this PR?

isinyaaa:
> Hi @isinyaaa , could you update this PR?

I still have many comments to address, but this is a partial update to get these commits up here.

tests/plugins/kernel_install/utils_test.sh
```diff
@@ -72,6 +72,7 @@ function setUp()
 function tearDown()
 {
   rm -rf "$SHUNIT_TMPDIR"
+  mkdir -p "$SHUNIT_TMPDIR"
```
rodrigosiqueira, line 75: I think it makes more sense to move this `mkdir -p` from `terarDown` to `setUp`.

src/kw_config_loader.sh
```diff
@@ -158,11 +158,18 @@ function parse_configuration()
 # higher level setting definitions to overwrite lower level ones.
 function load_configuration()
 {
+  if [[ -v configurations && "$OVERRIDE_CONFIG" != 1 ]]; then
+    return
+  fi
+
```
rodrigosiqueira, line 162: The commit message here does not describe anything... Please, update the commit message.

etc/build_template.config
```diff
@@ -0,0 +1,17 @@
+# Specify the default architecture used by kw
```
rodrigosiqueira, line 1: I recently introduced the `kw init --template` and this commit will be affected by this feature. You will also need to update the template feature to copy the `build.config` file correctly.

1. https://kworkflow.org/man/features/init.html

src/build.sh
```diff
@@ -114,6 +114,50 @@ function kernel_build()
...
+# allowing higher level setting definitions to overwrite lower level ones.
+function load_build_config()
+{
+  if [[ -v build_config && "$OVERRIDE_BUILD_CONFIG" != 1 ]]; then
+    unset OVERRIDE_BUILD_CONFIG
+    return
+  fi
+
+  local -a config_dirs
+  local config_dirs_size
```
rodrigosiqueira, line 158: Let's keep variables at the beginning of the function.

src/build.sh
```diff
@@ -114,6 +114,50 @@ function kernel_build()
...
+
+  if [[ -v XDG_CONFIG_DIRS ]]; then
+    IFS=: read -ra config_dirs <<< "$XDG_CONFIG_DIRS"
+  else
+    [[ -d '/etc/xdg' ]] && config_dirs=('/etc/xdg')
+  fi
+
+  # Old users may not have split their configs yet
```
rodrigosiqueira, line 166: Is this comment related to the below line? It looks unrelated to me.

src/build.sh
```diff
@@ -114,6 +114,50 @@ function kernel_build()
...
+
+  parse_configuration "${XDG_CONFIG_HOME:-"$HOME/.config"}/$KWORKFLOW/$BUILD_CONFIG_FILENAME" build_config
+
+  if [[ -f "$PWD/$KW_DIR/$BUILD_CONFIG_FILENAME" ]]; then
+    parse_configuration "$PWD/$KW_DIR/$BUILD_CONFIG_FILENAME" build_config
+  else
+    # Old users may not have used kw init yet, so they wouldn't have .kw
+    warning "Please use kw init to update your config files"
+  fi
+}
```
rodrigosiqueira, line 189: Can we make this function something generic? I guess other features will benefit from that (e.g., deploy).

src/kw_config_loader.sh
```diff
@@ -1,7 +1,9 @@
 include "$KW_LIB_DIR/kwio.sh"
 include "$KW_LIB_DIR/kwlib.sh"
 
+# these should be the default names for config files
 CONFIG_FILENAME=kworkflow.config
+BUILD_CONFIG_FILENAME=build.config
```
rodrigosiqueira, line 6: Add single quotes around `build.config`

documentation/man/features/build.rst
rodrigosiqueira: @kwy95

rodrigosiqueira (changes requested):
Hi @isinyaaa ,
This PR is a really welcome one! I added some other comments, but let's try to complete it :)

rodrigosiqueira:
@isinyaaa any update here?

isinyaaa:
I just needed to rebase this on my current system, so this is not done yet.

documentation/man/features/build.rst
isinyaaa: indeed you're right

src/build.sh
isinyaaa: @kwy95 ping

src/build.sh
isinyaaa: > 1. `kw b -n -c`
> 2. `kw b -n --ccache`

`-n` will not take any of these flags into consideration.

> 3. `kw b -d --ccache`

This should take `ccache` into consideration just fine :)

> 4. `kw b -i -c`
> 5. `kw b -i --ccache`

I'm really not sure about the `-i` flag, as I've not messed with it at all.

tests/build_test.sh
isinyaaa: nope

documentation/man/features/build.rst
isinyaaa: we actually don't use `=`, this was a typo :)

src/build.sh
```diff
@@ -179,3 +211,45 @@ function build_help()
     '  build (-i | --info) - Display build information' \
     '  build (-d | --doc) - Build kernel documentation'
 }
+
+# This function is used to show the current set up used by kworkflow.
+function show_build_variables()
+{
```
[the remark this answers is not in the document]
isinyaaa: I tried my hand at it but still haven't tested it. I have one question though: when should those variables be printed?

documentation/man/features/build.rst
isinyaaa: Agree!

isinyaaa:
@rodrigosiqueira I'll update this PR with the comments you provided, but I think it'd be best if we somehow split it to make revision easier, as it's already given us so much to talk about.

documentation/man/features/build.rst
kwy95: My bad, I missed this comment somehow.

src/build.sh
kwy95: Hi, the relevant function in the mail.sh file is `reposition_commit_count_arg` and the end of the parser as well, if you want take a look at it. I don't think the use case is the same, but it might help somehow. Anyway, I will review this section further in the new PR.

PR #544, opened 2022-01-17
Fix `debug` parameter handling

isinyaaa:
Previously, calling `kw debug -l -` or even with a whitespace as an argument `' '` would be considered an optional argument for the list subcomand.
    
This change fixes that, removing whitespaces or strings starting in dashes from being considered as possible (optional) arguments, and instead handling those with the parser itself, which doesn't at all affect functionality but provides more coherence in parameter handling throughout the project.
    
This commit also fixes previously incoherent parameter input in long/short options.

### DIFF
--- a/src/debug.sh
+++ b/src/debug.sh
@@ -905,7 +905,7 @@ function parser_debug_options()
   local transition_variables
 
   long_options='remote:,event:,ftrace:,dmesg,cmd:,local,history,disable,list::,follow,reset,help'
-  short_options='f,e,t,g,c,h,d,l,k'
+  short_options='e:,t:,g,f,c:,k,d,l::,h'
 
   options=$(kw_parse "$short_options" "$long_options" "$@")
 
@@ -990,7 +990,7 @@ function parser_debug_options()
         ;;
       --list | -l)
         # Handling optional parameter
-        if [[ -z "$2" ]]; then
+        if [[ "$2" =~ ^- || -z "${2// /}" ]]; then
           options_values['LIST']=1
           shift 2
         else


kwy95 (changes requested):
I noticed there is a problem with how the short and long options of `kw_parse` are being handled in this file. The short options also need the colons to indicate if they receive arguments or not. This makes me believe there are more broken options in this file. Can you update the short option list and see if that solves the problem you had?
The way this is currently implemented the using something like `kw debug -l --local` would not parse the local option. Similarly using `kw debug -c 'some command'` should fail, as the resulting options variable would look like this: `-c -- 'some command'`, and the parser would take the `--` as the command.

isinyaaa:
@kwy95 
> I noticed there is a problem with how the short and long options of `kw_parse` are being handled in this file. The short options also need the colons to indicate if they receive arguments or not. This makes me believe there are more broken options in this file. Can you update the short option list and see if that solves the problem you had?

I actually didn't have any problems and was just reading the code for another PR, so this change I proposed wasn't at all a bug fix, but instead more of a pedantic change to make it more coherent with the rest of the codebase, with the main reasoning being: should something change with the `--list` the parsing has to work properly.

> The way this is currently implemented the using something like `kw debug -l --local` would not parse the local option. Similarly using `kw debug -c 'some command'` should fail, as the resulting options variable would look like this: `-c -- 'some command'`, and the parser would take the `--` as the command.

With that said, I found your observation extremely pertinent and fixed this in the same commit as well :)

rodrigosiqueira:
Merged!
Thanks

PR #561, opened 2022-01-30
Sendemail setup fixes

### LINKED ISSUE #546: User provided values should overwrite template values
**Description:**

If a user tries to configure their mail options using a template and also provides different value to an option set by the template the user given value should have priority.

**How to Reproduce:**
Execute:
```
kw mail --template=riseup --smtpserverport 587
```
The server port option will be set to the template value of 465.

**Expected behavior:**

The user given options should always take precedence over template options.

### LINKED ISSUE #547: smtpuser doesn't have to be a valid email
**Description:**

The mail setup feature validates the smtpuser option as if it had to be a valid email address, but that is not the case.

**How to Reproduce:**

Providing a non email value to the `--smtpuser` option of the mail setup will result in a complaint of invalid email.

**Expected behavior:**

A non email value should be valid.

kwy95:
This patch fixes some issues with the kw mail setup feature. 

Closes #546 
Closes #547

### DIFF
--- a/src/mail.sh
+++ b/src/mail.sh
@@ -336,7 +336,7 @@ function mail_setup()
     config_values 'values' "$option"
 
     if [[ -n "${values['loaded']}" ]]; then
-      if [[ "$option" =~ 'user.email'|'sendemail.smtpuser' ]]; then
+      if [[ "$option" == 'user.email' ]]; then
         validate_email "${values['loaded']}" || continue
       fi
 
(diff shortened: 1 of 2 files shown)


rodrigosiqueira (changes requested):
In general, lgtm.

Could you add a test for the second commit? Since you are fixing a bug, it would be good to have a test that validates that feature with an email or without it. Also, add a comment in this new test that references the issue you open.

rodrigosiqueira:
Merged, thanks

Issue #568, opened 2022-03-11
DB and build tests failing in Arch


isinyaaa:
**Description:**

`kw_db_test.sh` and `build_test.sh` are failing in Arch Linux.

**How to Reproduce:**

1. Run `./run_tests.sh` in Arch.

**Expected behavior:**

Unit tests should be working fine.

**Screenshots:**

[If applicable, add screenshots to help explain your problem.]

**System information:**

 - OS: [e.g. Debian Bullseye]
 - Kw Version: [e.g. use `kw version`]
 - Bash Version [e.g. 5.1.4]

**Additional context:**

<details>
<summary>Tests output</summary>

```raw
Running test [arch_test]
=========================================================
test_generate_arch_temporary_root_file_system

Ran 1 test.

OK
Running test [bootloader_utils_test]
=========================================================
test_discover_device_and_partition
test_discover_all_hard_drive
test_partition_table_type
test_identify_mbr_per_partition
test_identify_bootloader_from_files

Ran 5 tests.

OK
Running test [debian_test]
=========================================================
test_update_debian_boot_loader
test_generate_rootfs_with_libguestfs

Ran 2 tests.

OK
Running test [grub_test]
=========================================================
test_grub
test_run_bootloader_for_vm

Ran 2 tests.

OK
Running test [utils_test]
=========================================================
test_cmd_manager
test_human_list_installed_kernels
test_command_list_installed_kernels
test_list_unmanaged_kernels
test_list_kernels_based_on_grub
test_reboot_machine
test_kernel_uninstall_unmanaged
test_kernel_force_uninstall_unmanaged
test_remove_managed_kernel
test_do_uninstall_cmd_sequence
test_install_modules
test_install_kernel_remote
test_install_kernel_local
test_install_kernel_vm
test_distro_deploy_setup

Ran 15 tests.

OK
Running test [to_cc_cmd_test]
=========================================================
test_to_cc_main

Ran 1 test.

OK
Running test [backup_test]
=========================================================
test_create_backup
test_restore_backup
test_restore_config
test_restore_data_from_dir
test_restore_pomodoro
test_restore_statistics
test_parse_backup_options

Ran 7 tests.

OK
Running test [debug_test]
=========================================================
test_prepare_log_database
test_process_list
test_convert_event_syntax_to_sys_path_hash
test_build_event_command_string
test_parser_debug_options
test_build_ftrace_command_string
test_ftrace_debug
test_ftrace_list
test_dmesg_debug
[...]

PR #569, opened 2022-03-11
Fix broken tests (again)

isinyaaa:
This should be fixing #568.


isinyaaa:
I just had a look at other versions of SQLite, like 3.37.2, and their error messages also differ a little from what we're looking for. I.e.:

* 3.35.5 error messages

  ```raw
  Error: no such table: not_a_table
  ```
  and
  ```raw
  Error: near "SELEC": syntax error
  ```
* 3.37.2 error messages

  ```raw
  Error: in prepare, no such table: not_a_table (1)
  ```
  and
  ```raw
  Error: in prepare, near "SELEC": syntax error (1)
  ```

* 3.38.0 error messages

  ```raw
  Error: in prepare, no such table: not_a_table
  ```
  and
  ```raw
  Error: in prepare, near "SELEC": syntax error
  SELEC * FROM tags;
  ^--- error here
  ```

which leads me to believe we could only look for substrings, i.e. `no such table: not_a_table` and `near "SELEC": syntax error` instead of a perfect match.

kwy95:
Hi @isinyaaa, I just looked over this PR and was thinking the exact same thing. Searching for the substring, either with a regex or other method, would be ideal, as I don't think checking for the different versions is very maintainable.

tests/kw_db_test.sh
```diff
@@ -119,15 +119,21 @@ function test_execute_command_db()
...
   output=$(execute_command_db 'SELECT * FROM not_a_table;' 2>&1)
   ret="$?"
-  expected='Error: no such table: not_a_table'
+  expected='no such table: not_a_table'
   assert_equals_helper 'Invalid table.' "$LINENO" "$ret" 1
-  assert_equals_helper 'Wrong output.' "$LINENO" "$output" "$expected"
+  if ! grep -qi "$expected" <<< "$output"; then
+    fail "line ($LINENO): Wrong output."
+    return
+  fi
```
rodrigosiqueira, line 127: This is something interesting that can be useful in other places; how about creating an assert helper for this? Maybe we can have something like `assert_substring_match` in the `tests/utils.sh`.

tests/utils.sh
```diff
@@ -222,6 +222,13 @@ but got:  \"${f}\"
   done <<< "$result_to_compare"
 }
 
+# This function expects an array of string with the command sequence and a
+# string containing the output.
+#
+# @msg Message to display in case of failure
+# @line $LINENO variable or ID string
```
rodrigosiqueira, line 229: We want to deprecate ID; let's drop the ID part.

rodrigosiqueira (changes requested):
Hi @isinyaaa ,
That's a great PR! Thanks a lot for that. I just added a few tiny comments.

isinyaaa:
This PR just failed for some unknown reason (have a look at _https://github.com/kworkflow/kworkflow/runs/5614006619?check_suite_focus=true_). So I updated the commit dates and pushed again to see if it works.

PR #580, opened 2022-03-22
setup: Change the order of the installation steps

magalilemes:
When installing kw without `pip` installed in the system, the following error occurs:

`./setup.sh: line 78: pip: command not found`

That happens because setup.sh runs a pip command before checking and installing the package dependencies -- pip included.

This commit simply changes the order of the steps, so that pip is not used before it is installed.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -74,11 +74,6 @@ function check_dependencies()
     return 0
   fi
 
-  while IFS='' read -r package; do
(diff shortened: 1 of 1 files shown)


kwy95 (approved):
LGTM

rodrigosiqueira:
Merged! Thanks

PR #589, opened 2022-04-01
tests: plugins: fixup tests

isinyaaa:
This commit fixes both the utils test and the raspberry pi deploy test.

The Raspberry pi test was broken due to the shunit2 temp folder not
being recreated every time it was destroyed, and the utils test was
using a previous format of the `compare_command_sequence` function.

Signed-off-by: Isabella Basso <isabbasso@riseup.net>


rodrigosiqueira:
Merged! Thanks a lot!

PR #593, opened 2022-04-16
Init template

rodrigosiqueira:


kwy95 (changes requested):
I added some corrections, but mostly it seems good. I think the template names could be friendly as I don't think I could remember them to use directly. Also, I guess this would be better as a separate refactor, but the tests having so many separate functions might cause some performance issues, especially in this specific file that removes and recreates the test directory with every test function.

magalilemes (changes requested):
Awesome PR, I'll point just a few things to be fixed.
Also, I followed both tutorials thoroughly, and I was able to -- very easily -- cross compile and deploy the latest kernel on a 32-bit Raspberry Pi OS. The writing is very clear. 😄

Issue #583, opened 2022-03-26
Reenable documentation built with sphinx-book-theme


rodrigosiqueira:
**Description:**

We recently started seeing a random CI failure caused by the sphinx documentation. Unfortunately, I'm not able to reproduce this issue in my local system, which makes the debug a little bit more complicated at the moment; we can only see it in the CI. Anyway, due to my lack of bandwidth, I just disabled sphinx-book-theme.

**How to Reproduce:**

1. Revert the commit SHA - NAME
2. Open a PR, and see the problem

**Expected behavior:**

No failures when generating the documentation with sphinx-book-theme.

**Log:**

Follows the CI error log:

```
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 git-email : Depends: git (< 1:2.25.1-.) but 1:2.35.1-0ppa1~ubuntu20.04.1 is to be installed
             Recommends: libemail-valid-perl but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
Collecting sphinx-book-theme
  Downloading sphinx_book_theme-0.3.0-py3-none-any.whl (334 kB)
Collecting sphinx<5,>=3
  Downloading Sphinx-4.4.0-py3-none-any.whl (3.1 MB)
Requirement already satisfied: pyyaml in /usr/lib/python3/dist-packages (from sphinx-book-theme) (5.3.1)
Collecting pydata-sphinx-theme~=0.8.0
  Downloading pydata_sphinx_theme-0.8.1-py3-none-any.whl (3.3 MB)
Requirement already satisfied: requests>=2.5.0 in /usr/lib/python3/dist-packages (from sphinx<5,>=3->sphinx-book-theme) (2.22.0)
Collecting sphinxcontrib-htmlhelp>=2.0.0
  Downloading sphinxcontrib_htmlhelp-2.0.0-py2.py3-none-any.whl (100 kB)
Requirement already satisfied: imagesize in /usr/lib/python3/dist-packages (from sphinx<5,>=3->sphinx-book-theme) (1.2.0)
[...]

PR #594, opened 2022-04-18
github: workflow: Fix CI failure caused by sphinx-book-theme

hilookas:
A sphinx-book-theme maintainer changed the way of getting the theme path.

Ref: https://github.com/executablebooks/sphinx-book-theme/blame/940d562fad0c8e9d4056b575c68d972ac0f9b2f4/src/sphinx_book_theme/__init__.py#L26

This commit fixes the CI failure by giving sphinx-book-theme a specified old version in pip.dependencies.

Fix: https://github.com/kworkflow/kworkflow/issues/583

### DIFF
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -878,10 +878,11 @@ function pack_kernel_files_and_send()
 {
   local flag="$1"
   local target="$2"
-  local kernel_img_name="$3"
-  local name="$4"
+  local kernel_binary_file_name="$3"
+  local kernel_name="$4"
   local arch=${5-${configurations[arch]}}
   local build_and_deploy="$6"
+  local config_kernel_img_name=${configurations[kernel_img_name]}
   local base_boot_path
   local config_path
   local config_local_version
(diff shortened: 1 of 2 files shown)


hilookas:
A good buid test: https://github.com/hilookas/kworkflow/runs/6059979540

documentation/dependencies/pip.dependencies
```diff
@@ -1 +1 @@
-sphinx-book-theme
+sphinx-book-theme==0.2.0
```
rodrigosiqueira, line 1: I'm ok with this change, but I think this fix is also a workaround. In the commit message, you said that this issue was caused due to a change in how sphinx-book gets the theme path; in this sense, we should fix our doc to work with the latest version of this pip project.

rodrigosiqueira (changes requested):

documentation/dependencies/pip.dependencies
hilookas: Thanks for your review comments. I make a new PR using a better way. https://github.com/kworkflow/kworkflow/pull/603

PR #597, opened 2022-04-24
Add arch linux arm to list of supported systems

isinyaaa:
As an Arch Linux derivative very close to the original, kw supports this
system.

Signed-off-by: Isabella Basso <isabbasso@riseup.net>

### DIFF
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -280,7 +280,7 @@ function detect_distro()
   local distro_id='none'
   local etc_path
   declare -a debian_family=('debian' 'ubuntu' 'raspbian')
-  declare -a arch_family=('arch' 'manjaro')
+  declare -a arch_family=('arch' 'manjaro' 'archarm')
   declare -a fedora_family=('fedora')
 
   etc_path=$(join_path "$root_path" /etc)


rodrigosiqueira:
Hi @isinyaaa , is this archarm the same one used on RPI? Or is it for something else?

rodrigosiqueira:
btw, it looks like you added the execution parameter to the target file.

isinyaaa:
> Hi @isinyaaa , is this archarm the same one used on RPI? Or is it for something else?

Yes, it can be used for that :)

> btw, it looks like you added the execution parameter to the target file.

Sorry, will fix that.

kwy95:
I recently started using a new distro (EndeavourOS) which is essentially arch, and came across the installation problem and circumvented it by adding a new key, similar to what this PR is doing, and I realized that this might not be the best way of doing things. Studying the install steps, I discovered that some distros provide the `ID_LIKE` key to their configurations. This key stores the "parent" distro, so for Ubuntu is Debian, for EndeavourOS is Arch, etc. I am now thinking of reworking the install process to use this key instead, which would allow for instant support to a lot of distros. I expect that not all distros would have the `ID_LIKE` key, so the `ID` would be used as a last resort.

I'm curious to think of your opinions about this. Any potential pitfalls I might not have considered, and if the sentiment is positive I'll open an issue and work on the implementation.

isinyaaa:
> I recently started using a new distro (EndeavourOS) which is essentially arch, and came across the installation problem and circumvented it by adding a new key, similar to what this PR is doing, and I realized that this might not be the best way of doing things. Studying the install steps, I discovered that some distros provide the `ID_LIKE` key to their configurations. This key stores the "parent" distro, so for Ubuntu is Debian, for EndeavourOS is Arch, etc. I am now thinking of reworking the install process to use this key instead, which would allow for instant support to a lot of distros. I expect that not all distros would have the `ID_LIKE` key, so the `ID` would be used as a last resort.
> 
> I'm curious to think of your opinions about this. Any potential pitfalls I might not have considered, and if the sentiment is positive I'll open an issue and work on the implementation.

Sounds good, I can have a look at whether ALarm supports this.

rodrigosiqueira:
Hi @kwy95 @isinyaaa 
The `ID_LIKE` approach sounds like a wonderful idea! It would be nice to investigate it and document it in an issue.

This `ID_LIKE` is something that only derivative distros will have, right? For example, my ArchLinux does not have it, my Manjaro system has `manjaro arch`, and Raspbian does not have it.

Anyway, since we don't have a proper fix now, let's update this PR and merge it. After @kwy95 has a refactor for that, we update it again.

@isinyaaa @kwy95 could some of you open the issue about the `ID_LIKE`?

rodrigosiqueira:
Hi @isinyaaa ,
We now have this ticket:
https://github.com/kworkflow/kworkflow/issues/605

Could you paste your `os-releases` in the above ticket?
Also, I will close this PR since we will work on a definitive solution for that.

PR #602, opened 2022-04-27
src: kwlib: check if the context is inside a git worktree

### LINKED ISSUE #551: src: kwlib: get git config functions should check if inside git repo
**Description:**

The functions `get_git_config_regex` and `get_all_git_config` should check if they are inside a git worktree before executing for the local scope, otherwise git will produce an error.

**How to Reproduce:**

The effects can be seen when attempting to use the list option of the mail feature, when executed outside of a git worktree the first line of output will be a message about the misuse of the `--local` option.

**Expected behavior:**

No git errors expected

mairacanal:
In the functions get_git_config_regex and get_all_git_config, git produce an
error when kw mail -l is running outside a git worktree.

In order to avoid such error, the git config function checks if the context
is inside a git worktree and changes the scope to global.

Closes #551 

Signed-off-by: Maíra Canal <maira.canal@usp.br>

### DIFF
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -583,6 +583,11 @@ function get_all_git_config()
 
   flag=${flag:-'SILENT'}
 
+  # shellcheck disable=2119
+  if ! is_inside_work_tree; then
+    scope='global'
+  fi
+
   for scp in {'global','local'}; do
     if [[ -z "$scope" || "$scope" == "$scp" ]]; then
       output["$scp"]="$(cmd_manager "$flag" "$cmd --$scp $config" | sed -E "s/^/$scp\t/g")"
@@ -612,6 +617,11 @@ function get_git_config_regex()
   local scp
 
   flag=${flag:-'SILENT'}
+
+  # shellcheck disable=2119
+  if ! is_inside_work_tree; then
+    scope='global'
+  fi
 
   for scp in {'global','local'}; do
     if [[ -z "$scope" || "$scope" == "$scp" ]]; then
--- a/tests/kwlib_test.sh
+++ b/tests/kwlib_test.sh
@@ -559,13 +559,11 @@ function test_get_all_git_config()
   }
 
   output=$(get_all_git_config test-config '' 'TEST_MODE' | sort -d)
-  # expected='git config --get-all --show-scope -- test-config'
-  expected=$'global\t'"git config --get-all --global test-config"$'\n'
-  expected+=$'local\t'"git config --get-all --local test-config"
+  expected=$'global\t'"git config --get-all --global test-config"
   assert_equals_helper 'Testing command' "$LINENO" "$expected" "$output"
 
   output=$(get_all_git_config test-config 'local' 'TEST_MODE')
-  expected=$'local\tgit config --get-all --local test-config'
+  expected=$'global\tgit config --get-all --global test-config'
   assert_equals_helper 'Testing command' "$LINENO" "$expected" "$output"
 
   # only possible test at a global scope, as we have limited control over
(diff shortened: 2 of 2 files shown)


mairacanal:
Looking at the unit tests, I believe it is needed to understand the expected behavior. It is wanted to exit the command when it is outside a git worktree or it is wanted to display the global information?

src/kwlib.sh
```diff
@@ -577,12 +577,17 @@ function get_all_git_config()
...
   local scope="$2"
   local flag="$3"
+  local inside_git_repo='git rev-parse --is-inside-git-dir'
   local cmd='git config --get-all'
   local -A output
   local scp
 
   flag=${flag:-'SILENT'}
 
+  if [[ ! $($inside_git_repo 2> /dev/null) ]]; then
```
rodrigosiqueira, line 587: We already have a function for handling this check; look at `is_inside_work_tree`. If you use it, you can delete `inside_git_repo` variable and change this conditional to something like:
`if ! is_inside_work_tree; then`

rodrigosiqueira (changes requested):
Hi @mairacanal ,
Thanks a lot for this PR!

I added just one inline comment. However, notice that the test `test_get_all_git_config` does not pass after your change because you change the behavior of `test_get_all_git_config`. Please update the old test and make sure that we have a test that covers the case where we are in a git tree and outside it.

@kwy95 You probably want to review this as well.
Thanks

kwy95 (changes requested):
My current review I exactly as @rodrigosiqueira 's, try to use the function we already have. Otherwise this looks good.

mairacanal:
This shellcheck failure is kind of problematic for the code's proper functioning because the function's $1 does not mean the script's $1. Have you got any suggestions on how to silence this warning?

rodrigosiqueira:
@mairacanal, why not pass `flag` as a parameter?

mairacanal:
If `flag` is passed as a parameter, I have a couple of problems with the TEST_MODE flag on the unit tests. I am not sure why, but the indentation seems to change on the output.

```test_get_all_git_config
ASSERT:line 565: Testing command expected:<git rev-parse --is-inside-work-tree &> /dev/null
global	git config --get-all --global test-config
local	git config --get-all --local test-config> but was:<global	git config --get-all --global test-config
local	git config --get-all --local test-config>
ASSERT:line 569: Testing command expected:<git rev-parse --is-inside-work-tree &> /dev/null
local	git config --get-all --local test-config> but was:<local	git config --get-all --local test-config>
ASSERT:line 575: Testing command expected:<git rev-parse --is-inside-work-tree &> /dev/null
global	git config --get-all --global test-config> but was:<global	git config --get-all --global test-config>
test_get_git_config_regex
ASSERT:line 619: Testing command expected:<git rev-parse --is-inside-work-tree &> /dev/null
global	git config --get-regexp --global 'test-config'
local	git config --get-regexp --local 'test-config'> but was:<global	git config --get-regexp --global 'test-config'
local	git config --get-regexp --local 'test-config'>
ASSERT:line 623: Testing command expected:<git rev-parse --is-inside-work-tree &> /dev/null
local	git config --get-regexp --local 'test-config'> but was:<local	git config --get-regexp --local 'test-config'>
ASSERT:line 629: Testing command expected:<git rev-parse --is-inside-work-tree &> /dev/null
global	git config --get-regexp --global 'test-config'> but was:<global	git config --get-regexp --global 'test-config'>
```

As far as user experience, the use of the `flag` parameter seems to keep everything running as expected. 

Do you have any idea on this flag issue?

kwy95:
@mairacanal Just add this line to the top of both ifs: `# shellcheck disable=2119` :)

rodrigosiqueira:
Merged! Thanks

PR #603, opened 2022-04-28
github: workflow: Fix CI failure caused by sphinx-book-theme

hilookas:
A sphinx-book-theme maintainer changed the way of getting the theme path.

Ref: https://github.com/executablebooks/sphinx-book-theme/blame/940d562fad0c8e9d4056b575c68d972ac0f9b2f4/src/sphinx_book_theme/__init__.py#L26

For jinja2>=3, it changes the way the variable is wrapped.

Ref: https://github.com/pallets/jinja/blob/5962edeb271d93687eb93f32d53ffe53f86871e0/src/jinja2/loaders.py#L184

This commit fixes the CI failure by updating jinja2 to the version that can work with sphinx-book-theme in `pip.dependencies`.

A pair of double quotes encloses the package name to prevent the shell from misrecognizing angle brackets in `setup.sh`.

Fix: https://github.com/kworkflow/kworkflow/issues/583

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -86,8 +86,8 @@ function check_dependencies()
   fi
 
   while IFS='' read -r package; do
-    pip list | grep -F "$package" > /dev/null
-    [[ "$?" != 0 ]] && pip_package_list="$package $pip_package_list"
+    python3 -c "import pkg_resources; pkg_resources.require('$package')" &> /dev/null
+    [[ "$?" != 0 ]] && pip_package_list="\"$package\" $pip_package_list"
   done < "$DOCUMENTATION/dependencies/pip.dependencies"
 
   if [[ -n "$pip_package_list" ]]; then
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -37,8 +37,6 @@ jobs:
 
       - name: Check documentation
         run: |
-          # FIXME: Workaround, see https://github.com/kworkflow/kworkflow/issues/583
-          echo '' > 'documentation/dependencies/pip.dependencies'
           ./setup.sh --docs --force
 
       - name: Check installation
--- a/documentation/dependencies/pip.dependencies
+++ b/documentation/dependencies/pip.dependencies
@@ -1,1 +1,2 @@
+jinja2>=3
 sphinx-book-theme


hilookas:
A well-run build test: https://github.com/hilookas/kworkflow/runs/6212236464

Compared to https://github.com/kworkflow/kworkflow/pull/594 , this PR fixes the bug by updating jinja2.

setup.sh
```diff
@@ -87,7 +87,7 @@ function check_dependencies()
 
   while IFS='' read -r package; do
     pip list | grep -F "$package" > /dev/null
-    [[ "$?" != 0 ]] && pip_package_list="$package $pip_package_list"
+    [[ "$?" != 0 ]] && pip_package_list="\"$package\" $pip_package_list"
```
rodrigosiqueira, line 90: Why do we need `\"`? Is it because of `jinja2>=3`?

rodrigosiqueira (changes requested):
Hi @hilookas ,
Thanks a lot for this new version; this looks like a proper fix for the issue :)

Some comments:
1. Wrap your commit messages under 75 characters (if you use vim, you can try `gd`). Links are an exception.
2. Use `Closes #ISSUE_NUMBER`. This syntax is parsed by Github, which will automatically close the issue.
3. Is this your real email? `i@18kas.com`. I think your git config is not configured correctly; could you update it? You can use something like `git config user.name "YOUR NAME"` and `git config user.email "YOUR EMAIL"`
4. Rebase your branch with the latest version of the unstable branch

Finally, don't close this PR after making the above change; you just need to use `git push -f origin hilookas:workflow-bugfix-2`. That way, Github will automatically update this PR.

setup.sh
hilookas: Thats right. `pip install jinja2>=3` like command will make the shell exec `pip install jinja2` and put its input into a `=3` file.

hilookas:
Thanks for the suggestion! I prefer to use `i@18kas.com` as my public email. I'll be more careful next time. 

The commit has been updated. :)

rodrigosiqueira:
Something is wrong with your change; we need to say `y` during the `./setup.sh -i` even if we already have the dependencies installed in the system. We just want to ask users when the dependency is not installed; we should not ask anything after that.

hilookas:
A new way to detect if a package needs to be installed or updated has been added. It should be quieter now. :>

hilookas:
I started a VM on Vultr to get a clean OS with the settings of:

````
3.5usb/mo VM
Ubuntu 20.04 LTS x64
````

And executed these commands with the default `root` user:

````
git clone https://github.com/hilookas/kworkflow.git
cd kworkflow
git checkout workflow-bugfix-2
./setup.py -i
sphinx-build -nW -b man documentation /root/.local/share/kw/man
````

The installation run well and documentation was generated.

To position the problem, could you please exec `pip list` and paste the output? Thanks a lot!

Execute `pip uninstall jinja2` and then execute `./setup.sh -i` may solve the problem. :)

File: setup.sh (hunk 1 above)
Comment (tales-aparecida) [reply setup.sh:90]: Consider using `pip freeze -r <dependencies file>` instead of the whole `while` loop :)

```
# Get the list of uninstalled dependencies.
# `pip freeze -r <dependencies file>` compares the installed modules to the dependencies file and outputs to stderr lines like:
# "WARNING: Requirement file [pip.dependencies] contains sphinx-book-theme, but package 'sphinx-book-theme' is not installed"
# `sed` gets the module name from those lines and `paste` joins them in a single line.
pip_package_list=$(pip freeze -r $DOCUMENTATION/dependencies/pip.dependencies 2>&1 1> /dev/null | sed -r 's/^.*contains (.*),.*not installed.*$/\1/' | paste --ser
ial -d' ')
```

hilookas: Thanks for your advice. But for this case, we need to upgrade jinja2, because the lower version of jinja2 can't get along well with sphinx-book-theme, resulting in CI errors. The current solution checks whether the package is installed and whether its version meets the requirements. :>

rodrigosiqueira:
Hi @hilookas ,
After I removed the old Sphinx dependencies, everything was alright. For this reason, I decided to merge this change.

Thanks a lot, and sorry for my delay

PR #606, opened 2022-05-05
Fix large initramfs issue

rodrigosiqueira:
Usually, I work with the same set of test systems and kernel branches.
However, I had to switch hardware and branch recently, but I kept the
same config file since it was the same hardware family. Usually, I have
the CONFIG_DEBUG_INFO option set to yes, and I don't have any issue with
that; but for some reason, it started to generate a huge initramfs
(>256Mb) which caused boot failure. After a lot of research, I found
this page:

https://01.org/linuxgraphics/gfx-docs/drm/kbuild/kbuild.html

I also realize that the Debian package uses this option by default. For
this reason, I decided to add this as a default in the kw, and I also
added a warning in case the user has this option and explicitly disabled
the strip from the kw config file.


src/deploy.sh
```diff
@@ -804,10 +805,19 @@ function modules_install_to()
...
+      load_module_text "${KW_ETC_DIR}/strings/deploy.txt"
+      warning "${module_text_dictionary[large_initramfs_warning]}"
+    fi
+  fi
+
   if [[ "$local_deploy" == 'local' ]]; then
     cmd='sudo -E make modules_install'
   else
-    cmd="make INSTALL_MOD_PATH=$install_to modules_install"
+    cmd="make ${strip_modules_debug}INSTALL_MOD_PATH=$install_to modules_install"
```
tales-aparecida, line 820: Hi, everything looks okay, I just didn't understand why this was not added to the "local" installation.
Could you explain it for me? Is it because of the fore-mentioned "Debian package uses this option by default."?

rodrigosiqueira: I missed it! Really nice catch! Thanks a lot.

rodrigosiqueira:
Thanks Tales for your review, PR merged into the unstable branch.

PR #608, opened 2022-05-10
src: deploy: Fix sudo issue in local deploy

rodrigosiqueira:
Some users reported that `kw bd --local` and `kw d --local` had some
weird behavior during the deploy. Something that looks like this:

    * Preparing modules
    [sudo] password for siqueira: [>

    [>

As a result, sometimes, users type the wrong password and have to type
it multiple times. This problem happens due to the pipe between the
modules_install command and the pv command. This commit fixes this issue
by running a simple sudo command before the task that requires the pipe
utilization.

### DIFF
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -805,7 +805,7 @@ function modules_install_to()
   fi
 
   if [[ "$local_deploy" == 'local' ]]; then
-    cmd='sudo -E make modules_install'
+    cmd='sudo true && sudo -E make modules_install'
   else
     cmd="make INSTALL_MOD_PATH=$install_to modules_install"
   fi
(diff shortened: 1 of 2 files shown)


melissawen:
@rodrigosiqueira, IIRC, I didn't face this issue before, and I keep not seeing it. That means, things keep working fine on this topic.

From my side, I see these warnings:
```
/home/melissawen/.local/lib/kw/plugins/kernel_install/utils.sh: line 38: /opt/kw/status: No such file or directory
* Preparing modules
[sudo] password for melissawen: 
[===========================================================================================================================================================================================================>] 100%
```
then
```
touch: cannot touch '/opt/kw/INSTALLED_KERNELS': No such file or directory
grep: /opt/kw/INSTALLED_KERNELS: No such file or directory
tee: /opt/kw/INSTALLED_KERNELS: No such file or directory
-> Execution time: 00:00:50
Cleaning up temporary files...
Exiting...
```

and it is solved by create kw directory and touch the relevant files

rodrigosiqueira:
I merged this patch because it fixes the issue for other users. However, I created a ticket for the issue that you reported:

https://github.com/kworkflow/kworkflow/issues/609

@melissawen please, add more details to the above issue.

Thanks

PR #612, opened 2022-05-23
Change detect distro approach

rodrigosiqueira:


src/kwlib.sh
```diff
@@ -265,41 +265,49 @@ function join_path()
...
-  # Debian family
-  if [[ "${debian_family[*]}" =~ ${distro_id} ]]; then
-    printf '%s\n' 'debian'
-  # ArchLinux family
-  elif [[ "${arch_family[*]}" =~ ${distro_id} ]]; then
-    printf '%s\n' 'arch'
-  # Fedora family
-  elif [[ "${fedora_family[*]}" =~ ${distro_id} ]]; then
-    printf '%s\n' 'fedora'
+  if [[ ${os_family[*]} =~ ${distro_id} ]]; then
```
magalilemes, line 309: It may be worth taking into consideration that `ID_LIKE` might have more than one operating system listed. According to the manpage:

```
ID_LIKE=
A space-separated list of operating system identifiers in the same syntax as the ID= setting.
```
For instance, Pop!_OS has `ID_LIKE="ubuntu debian"`, and then this comparison is evaluated as false.

magalilemes, line 309: It may be worth taking into consideration that `ID_LIKE` might have more than one operating system listed. According to the manpage:

```
ID_LIKE=
A space-separated list of operating system identifiers in the same syntax as the ID= setting.
```
For instance, Pop!_OS has `ID_LIKE="ubuntu debian"`, and then this comparison is evaluated as false.

kwy95: Based on this comment I think that a for loop can be constructed by creating a `$distro_ids` like so:
```bash
distro_ids=$(printf '%s' "$os_release_processed" | grep -w 'ID\(_LIKE\)\?' | cut -d = -f 2)
```
This gathers all entries, from both `ID` and `ID_LIKE`, then it's possible to loop over the entries with:
```bash
for distro_id in $distro_ids; do
  if [[ ${os_family[*]} =~ ${distro_id} ]]; then
    printf '%s\n' "$distro_id"
    return
  fi
done

printf '%s\n' 'none'
```

tests/samples/os/steamos/etc/os-release
```diff
@@ -0,0 +1,10 @@
+NAME="SteamOS"
```
kwy95, line 1: Try to add one from Pop_OS! that have multiple entries for `ID_LIKE`

kwy95 (changes requested):
I think the comment @magalilemes left is the most important thing really, otherwise this looks great.

rodrigosiqueira:
Hi @kwy95 @magalilemes ,
Thanks a lot for this great review! I applied all your suggestions in this new version.

PR #613, opened 2022-05-24
Add support to Fedora-based systems

### LINKED ISSUE #567: Add support to Fedora-based systems
**Description:**

Currently, kw deploy only works for Debian and ArchLinux families. Let's expand kw distro support and add a Fedora-based system as a target.

**Why:**

Many kernel developers use Fedora as their base system.

**Target user:**

* Kernel developers that use Fedora.
* Developer that needs to support stable distros such as RedHat.

**How to implement it:**

First step:
* Learn how to install a custom kernel in a Fedora distro manually.
* Learn about:
   * Pre-setup (required packages, config file, etc)
   * How to generate the initramfs
   * Default bootloader and how to update it

In a few words, this work can be divided into the following commits:
* Commit 1: Add setup step for fedora system. Here we want to install the required package, and we might need to configure something else. You will need to create the `fedora.sh` file at `src/plugins/kernel_install/`
* Commit 2: Introduce Fedora deploy. In theory, at this stage, you just need to enable the initramfs update for the fedora system and other minor details. You will probably spend most of your time implementing tests and validating the correct behavior.

mairacanal:
Currently, kw deploy only works for Debian and ArchLinux families. This
commit expands kw distro support and adds Fedora-based systems as a target.

Closes #567

Signed-off-by: Maíra Canal <maira.canal@usp.br>

---

The remote and local deploy were tested on Fedora 35. VM deployment was not tested as I could not set up a VM with kw yet (even if other distros).

Unit tests are still missing and I'm currently working on them, but I decided to post the PR while developing them in order to get feedback from the kw community. Any feedback is welcomed!

### DIFF
--- a/src/plugins/kernel_install/grub.sh
+++ b/src/plugins/kernel_install/grub.sh
@@ -4,7 +4,29 @@
 # 1. run_bootloader_update: Update GRUB in a local and remote machine.
 # 2. run_bootloader_for_vm: Update GRUB in a virtual machine.
 
-declare -rg DEFAULT_GRUB_CMD_UPDATE='grub-mkconfig -o /boot/grub/grub.cfg'
+declare -g DEFAULT_GRUB_CMD_UPDATE='grub-mkconfig -o /boot/grub/grub.cfg'
+
+# Some distributions, such as Fedora, use GRUB2 as the default bootloader. On
+# those systems, grub-mkconfig command is replaced by grub2-mkconfig. This function
+# checks if the grub-mkconfig command exists and if doesn't, the default grub
+# update command is set to grub2-mkconfig.
+#
+# Returns:
+# 0 if a grub update command exists and 2 otherwise.
+function define_grub_cmd_update()
+{
+  local grub_cmd='grub-mkconfig'
+  local grub2_cmd='grub2-mkconfig'
+
+  if ! command_exists "$grub_cmd"; then
+    if ! command_exists "$grub2_cmd"; then
+      return 2 # ENOENT
+    fi
+    DEFAULT_GRUB_CMD_UPDATE='grub2-mkconfig -o /boot/grub2/grub.cfg'
+  fi
+
+  return 0
+}
 
 # Update grub bootloader in a target machine.
 function run_bootloader_update()
@@ -21,6 +43,12 @@ function run_bootloader_update()
   elif [[ "$target" == 'vm' ]]; then
     run_bootloader_for_vm "$flag"
     return "$?"
+  fi
+
+  define_grub_cmd_update
+  if [[ "$?" -gt 0 ]]; then
+    complain "There is no grub-mkconfig command in the system."
+    return 125 # ECANCELED
   fi
 
   cmd_grub+="$DEFAULT_GRUB_CMD_UPDATE"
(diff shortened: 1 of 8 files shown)


mairacanal:
> In general, everything looks good to me; I just have two questions:
> 
> 1. Did you try kw deploy remote?
> 2. Did you try kw deploy local?

I tested kw deploy remote (F35 and F36) and local (F35). Probably, tomorrow, I'll finish up the unit tests and finalize the PR.

mairacanal:
I guess the PR is finally ready. The `kw d --local` was tested on Fedora 35 with a Dell Inspiron 15. The `kw d --remote` was tested on Fedora 35 with a Dell Inspiron 15 and Fedora 36 with a Desktop. 

The unit tests were added to Fedora and some small modifications to grub_test.

The branch is already rebased also.

tests/plugins/kernel_install/grub_test.sh
```diff
@@ -17,11 +17,22 @@ function test_grub()
   local expected_cmd
 
   output=$(run_bootloader_update 'TEST_MODE' 'local')
-  expected_cmd='sudo -E grub-mkconfig -o /boot/grub/grub.cfg'
+
+  if [[ ! -x "$(command -v "grub-mkconfig")" ]]; then
+    expected_cmd='sudo -E grub2-mkconfig -o /boot/grub2/grub.cfg'
+  else
+    expected_cmd='sudo -E grub-mkconfig -o /boot/grub/grub.cfg'
+  fi
```
rodrigosiqueira, line 25: Correct me if I'm wrong, but you assume that the developer system has grub, right? I guess this test will fail for the simple fact that the developer is using another bootloader such as u-boot or syslinux, right? I think you can mock some of the grub behavior with an alias in the test; that way, you can also test the `grub` and `grub2`. IIRC `device_test` and `utils_test.sh` might have some good examples.

src/plugins/kernel_install/grub.sh
```diff
@@ -4,7 +4,29 @@
...
+# Returns:
+# 0 if a grub update command exists and 2 otherwise.
+function define_grub_cmd_update()
+{
+  local grub_cmd='grub-mkconfig'
+  local grub2_cmd='grub2-mkconfig'
+
+  if [[ ! -x "$(command -v $grub_cmd)" ]]; then
+    if [[ ! -x "$(command -v $grub2_cmd)" ]]; then
+      return 2 # ENOENT
```
rodrigosiqueira, line 23: Could you add a test that hit this case?

rodrigosiqueira (changes requested):
Very nice PR! I just had two minor comments.
Thanks.

tests/plugins/kernel_install/grub_test.sh
mairacanal: Currently, the test fails if the developer is using another bootloader. For example, all tests run fine on the CI/CD, but, locally with Fedora, tests are failing. I'll try to mock the grub behavior so that the tests can run system wide.

mairacanal: Checking the `device_test` and `utils_test.sh`, I saw how mock was done on unit tests. But, I'm still running into a problem: even if I alias or export grub-mkconfig, `$(command -v 'grub-mkconfig') cannot find the mock command and I end up falling into the case where there is no grub installed.

Have you got any suggestions on what to do?

mairacanal: End up getting a solution for the `$(command -v 'grub-mkconfig')` problem with `kwlib`.

src/plugins/kernel_install/grub.sh
mairacanal: Added!

rodrigosiqueira:
Thanks a lot for this PR! Merged to the unstable branch.

PR #614, opened 2022-05-26
Small code cleanup

isinyaaa:
This is a small cleanup I made during PR's #543 development.

### DIFF
--- a/src/config_manager.sh
+++ b/src/config_manager.sh
@@ -289,14 +289,10 @@ function get_config_from_defconfig()
 
   cross_compile=${configurations[cross_compile]}
   arch=${configurations[arch]}
-  # TODO: Right now, the below variable has no meaning and will be empty. Let's
-  # expand the config file to enable developers to define ASIC targets.
-  asic=${configurations[asic]}
 
   # Build command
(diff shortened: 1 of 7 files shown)


src/config_manager.sh
```diff
@@ -296,7 +296,6 @@ function get_config_from_defconfig()
   # Build command
   [[ -n "$arch" ]] && cmd+=" ARCH=$arch"
   [[ -n "$cross_compile" ]] && cmd+=" CROSS_COMPILE=$cross_compile"
-  [[ -n "$asic" ]] && cmd+=" $asic\_defconfig"
```
kwy95, line 299: Shouldn't this PR also remove line 294 from this file?

kwy95 (commented):
I think there is an extra line that should be removed based on the comments from the previous PR, apart from that this LGTM.

isinyaaa:
@kwy95 thanks for the quick review :) I just think I might need to update some other stuff as I'm rebasing this on some months of other contributors' work, and there might be other stuff that won't work out of the box.

rodrigosiqueira:
Hi @isinyaaa , thanks a lot for this PR. I made some minor changes in the commit number 3 and merged it into the unstable branch.

PR #618, opened 2022-05-27
Enable kernel deploy to SteamOS

rodrigosiqueira:
Hi,

I'm trying to enable kw deploy for working with SteamOS. This PR is the first functional version that I want to share in order to get some ideas. You can safely ignore the first 3 commits since they are part of another PR.

Things that I'm not happy with this PR:
1. Dracut has a lot of error message, even if it works as expected
2. I had to change many things in the Grub (I did it manually)
3. The setup part causes me problems due to the requirement to update pacman keys and others

Maybe @melissawen and @andrealmeid  might be interested in this PR.

### DIFF
--- a/src/plugins/kernel_install/arch.sh
+++ b/src/plugins/kernel_install/arch.sh
@@ -15,6 +15,7 @@ declare -ga required_packages=(
   'lzop'
   'zstd'
   'xz'
+  'os-prober'
 )
 
 # ArchLinux package manager
--- a/src/plugins/kernel_install/arch.sh
+++ b/src/plugins/kernel_install/arch.sh
@@ -29,18 +29,41 @@ function generate_arch_temporary_root_file_system()
   local target="$3"
   local bootloader_type="$4"
   local path_prefix="$5"
+  local prefered_root_file_system="$6"
   local cmd=''
   local sudo_cmd
   local template_path
   local mkinitcpio_destination_path
-  local LOCAL_KW_ETC="$KW_ETC_DIR/template_mkinitcpio.preset"
+  local LOCAL_KW_ETC="${KW_ETC_DIR}/template_mkinitcpio.preset"
+  # mkinitcpio still the default on ArchLinux
+  local root_file_system_tool='mkinitcpio'
+
+  # If the user specify which rootfs they want to use, let's use it then...
+  if [[ -n "$prefered_root_file_system" ]]; then
+    if command_exists "$prefered_root_file_system"; then
+      root_file_system_tool="$prefered_root_file_system"
+    else
+      printf 'It looks like that "%s" does not exists\n' "$prefered_root_file_system"
+      prefered_root_file_system=''
+    fi
+  fi
+
+  if [[ -z "$prefered_root_file_system" ]]; then
+    if ! command_exists 'mkinitcpio'; then
+      if ! command_exists 'dracut'; then
+        return 22 # EINVAL
+      else
+        root_file_system_tool='dracut'
+      fi
+    fi
+  fi
 
   # We do not support initramfs outside grub scope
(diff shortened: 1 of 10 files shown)


src/plugins/kernel_install/arch.sh
```diff
@@ -34,11 +34,21 @@ function generate_arch_temporary_root_file_system()
   local template_path
   local mkinitcpio_destination_path
   local LOCAL_KW_ETC="$KW_ETC_DIR/template_mkinitcpio.preset"
+  # mkinitcpio still the default
+  local root_file_system_tool='mkinitcpio'
+
+  if [[ ! -x "$(command -v 'mkinitcpio')" ]]; then
```
melissawen, line 40: I don't think it is a good way to define the tool to generate initramfs because I've installed mkinitcpio in my system (for example) and running mkinitcpio didn't work. maybe selecting by the distro name is more effective...

src/plugins/kernel_install/arch.sh
```diff
@@ -52,22 +62,30 @@ function generate_arch_temporary_root_file_system()
...
+        cmd+="bash -c \""
+        cmd+="sed 's/NAME/$name/g' '$template_path' > $mkinitcpio_destination_path\""
+
+        cmd="$sudo_cmd mkinitcpio --preset $name"
+        cmd_manager "$flag" "$cmd"
+      ;;
+      'dracut')
+        cmd='dracut --force --persistent-policy by-partuuid '
+        cmd+="--hostonly /boot/initramfs-${name}.img ${name}"
+        cmd_manager "$flag" "$cmd"
```
melissawen, line 86: is it enough to update grub? don't you need to run update-grub?

melissawen:
@rodrigosiqueira I tested and worked fine (even without update-grub). However, I think checking by mkinitcpio command is too fragile. At least, checking by distro name would be better (if we don't have another idea)

melissawen:
Hi,

I got the following error running `kw deploy` after an update of the steam deck:
```
* Sending kernel modules (6.0.0-rc1-mwen-mainline+) to the remote
      8,808,329 100%  155.46MB/s    0:00:00 (xfr#1, to-chk=0/1)
* Sending kernel boot files
     10,015,485 100%  144.63MB/s    0:00:00 (xfr#1, to-chk=0/1)
utils.sh: line 546: complain: command not found
-> Execution time: 00:00:18
Cleaning up temporary files...
Exiting...
```

Using the verbose option, we can see the bootloader is not identified: `[bootloader]= [distro]='steamos arch'`

I solved it by adding `EFI/steamos` to the lits of GRUB files, as below:

```
diff --git a/src/plugins/kernel_install/bootloader_utils.sh b/src/plugins/kernel_install/bootloader_utils.sh
index 807f47e..8e22b1e 100644
--- a/src/plugins/kernel_install/bootloader_utils.sh
+++ b/src/plugins/kernel_install/bootloader_utils.sh
@@ -11,6 +11,7 @@ declare -gar GRUB=(
   'boot/grub2/grub.conf'
   'boot/grub2/grub.cfg'
   'boot/efi/EFI/ubuntu/'
+  'boot/efi/EFI/steamos'
   'NST/menu.lst'
   'boot/grub/menu.lst'
   'ubuntu/disks/boot/grub/menu.lst'
```

melissawen:
As we stopped to be able to use the buttons to manage grub menu options, I am using `grub-reboot` to define the kernel to be load after a reboot (another option is plug a external keyboard). Then I found an interesting script for grub-menu selection here: https://askubuntu.com/questions/1019213/display-grub-menu-and-options-without-rebooting

What you think?

kwy95 (commented):
I realize this PR was already merged, but I had these pending comments and didn't want to simply ignore them.

Issue #626, opened 2022-05-30
Separate report output per days


kwy95:
**Description:**

Within a tag, `kw report` outputs all entries regardless of date, it would be nice to have a date based separation in the output.

**Type:**

Improvement

**Why:**

This makes output clearer as you know when each timebox was activated.

PR #627, opened 2022-05-30
Separate report output per date

kwy95:

Part of #626


src/report.sh
```diff
@@ -173,6 +173,7 @@ function grouping_day_data()
   local timebox_sec
   local total_time_box_sec=0
   local total_repetition=0
+  local -A used
```
rodrigosiqueira, line 176: Can you find a better name here? Fwiu, this variable works like a check if something is already in use or not, right?

Also, in the commit message, add a sample of how the output looks before and after your commit.

kwy95: The variable is used to check which tags have been accounted for in that day, to avoid printing the same date twice on each tag entry. I was heavily inspired by the *kw_include.sh* file.
What do you think of `date_printed`?

rodrigosiqueira:
@kwy95 lgtm, could you just rebase it on the latest code from the unstable branch?

rodrigosiqueira:
Merged! Thanks!

Issue #588, opened 2022-03-31
Add --rfc as an option for kw mail


melissawen:
**Description:**

Add --rfc as an option that automated the use of `-- --subject-prefix=` for RFC patches sent by `kw mail -s`

**Why:**

It aims to simplify daily tasks since developers commonly use the RFC prefix

**Target user:**

any

melissawen:
I've tested and it's working fine. thanks @kwy95

PR #630, opened 2022-06-03
src: mail: add --rfc option

kwy95:
This allows users to set this option that adds a request for comment tag
to the prefix of the email subject.

Part of #588

### DIFF
--- a/src/mail.sh
+++ b/src/mail.sh
@@ -89,6 +89,7 @@ function mail_send()
   local version="${options_values['PATCH_VERSION']}"
   local extra_opts="${options_values['PASS_OPTION_TO_SEND_EMAIL']}"
   local private="${options_values['PRIVATE']}"
+  local rfc="${options_values['RFC']}"
   local kernel_root
   local patch_count=0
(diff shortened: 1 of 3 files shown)


src/mail.sh
```diff
@@ -123,6 +124,7 @@ function mail_send()
 
   [[ -n "$opts" ]] && cmd+=" $opts"
   [[ -n "$private" ]] && cmd+=" $private"
+  [[ -n "$rfc" ]] && cmd+=" $rfc"
```
rodrigosiqueira, line 127: Is `--rfc` part of the git-send-email options? I did not find anything like that. Tbh, I expected something like `--subject-prefix="RFC"`

kwy95: git-send-email uses git-format-patch under the hood to create the email files, and that contains the ``--rfc`` option.

https://git-scm.com/docs/git-format-patch#Documentation/git-format-patch.txt---rfc

rodrigosiqueira:
Nice PR! Merged to the unstable branch.

PR #631, opened 2022-06-05
Update KW setup

isinyaaa:
I know those commit names are a little unhelpful, so I'm just sending this for you to look at the code, then I'll update them.

I tried to separate the commits as best as I could, though there's a lot going on on 3/4 because of all the refactorings, sorry about that.


setup.sh
```diff
@@ -1,106 +1,209 @@
...
-    cmd="dnf install -y $package_list"
+    installed_cmd="dpkg-query -W --showformat='${Status}\n"
+    cmd="apt install -y ${package_list[*]}"
   else
-    warning 'Unfortunately, we do not have official support for your distro (yet)'
-    warning 'Please, try to find the following packages:'
-    warning "$(cat "$DOCUMENTATION/dependencies/arch.dependencies")"
-    return 0
+    installed_cmd="dnf list installed"
+    cmd="dnf install -y ${package_list[*]}"
```
kwy95, line 174: Let's keep the fedora check, I know it's redundant but it adds clarity to the code and wouldn't really impact performance.

kwy95 (changes requested):
Hi, thanks for this PR. But as it stands it's a bit too confusing, please rebase it and add commit messages explaining why you are making the proposed changes. Also, remember to add documentation if you change the way the code works.
I can make a more thorough review once this is updated. Also, do you think this could be related to #635 ? I will make a PR addressing it soon, but your input would help adjust what approach to take.

rodrigosiqueira:
I created the following issue to keep track of this effort:

https://github.com/kworkflow/kworkflow/issues/640

At some point, I'll address it.

PR #634, opened 2022-06-10
Expand build features

rodrigosiqueira:


isinyaaa:
@rodrigosiqueira I just realized we have another type of warning I didn't actually contemplate in this patch series. From the kernel docs we can see we also have a [documentation build warning](https://docs.kernel.org/doc-guide/kernel-doc.html#how-to-format-kernel-doc-comments). Do you want me to open an issue about that?

PR #639, opened 2022-06-18
setup: add global config file

### LINKED ISSUE #635: Lack of global config file makes features misbehave
**Description:**

Since 19440bf834ff241a786e901013b7ce96816cd637 the setup script doesn't create a global config file, this causes features like `kw pomodoro` to misbehave, as the feature relies on the config file to determine the commands to use to trigger the visual and sound alerts.

**How to Reproduce:**

1. Install kw with `./setup.sh -i`
2. Go to a directory where you haven't run `kw init`
3. Execute `kw p -t 1s`
4. You should see the warnings about not being able to find the necessary commands, and the supposed commands are empty strings

**Expected behavior:**

The visual and sound alerts should work correctly.

**System information:**

 - OS: EndeavourOS && Ubuntu 22.04
 - kw Version: unstable (7b90953)
 - Bash Version: 5.1.16(1)-release

**Additional context:**

I believe there are other features that might suffer from this as well, the Pomodoro feature is just the one I noticed first because I frequently use it in whatever directory I'm in.

kwy95:
Some of kw's functions don't need to be used inside of a kernel tree,
for example, the `kw pomodoro` command that can be run on any folder,
but depend on values present in the config files. This means that we
need a global config file to ensure these functions work correctly.

Closes #635

### DIFF
--- a/etc/kworkflow_template.config
+++ b/etc/kworkflow_template.config
@@ -1,0 +1,72 @@
+# Default configuration file
+# Note: You can customize the settings in this file but no setting should be
+# removed. Otherwise, KW may not work properly.
+
+#? Attention, when using "./setup.sh -i":
+#? - USERKW will be replaced by the current user
+#? - SOUNDPATH will be replaced by kw's install path
+
+# Default ssh user
+ssh_user=root
+
+# Default ssh ip
+ssh_ip=localhost
+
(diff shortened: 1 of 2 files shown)


rodrigosiqueira:
Merged

PR #644, opened 2022-06-25
src: remote: Introduce remote interface

rodrigosiqueira:
When we want to manage multiple machines under kw project, we need to
open the config file and add the remote information. This is a far way
from convenient and user-friendly. This commit introduces a new option
named `remote`, which relies heavily on the git remote.


rodrigosiqueira:
Merged

PR #649, opened 2022-07-09
src: help: Fix renaming of configm to kernel-config-manager

mairacanal:
The commit 305d9e1 changed the command from configm to
kernel-config-manager, but didn't change the short command from g to k.
This commit fixes the new short command for kernel-config-manager
command and also reorganize the commands in alphabetical order.

Fixes: 305d9e1 ('src: configm: Rename configm to kernel-config-manager')
Signed-off-by: Maíra Canal <mairacanal@riseup.net>


rodrigosiqueira:
Merged, thanks

Issue #223, opened 2021-04-04
Modernize explore


rodrigosiqueira:
The `explore` feature is old and outdated, and we need to modernize it. I recommend the following steps:

- Create the main entry point to this feature: explore_main.
- Use the new parser system. See `src/build.sh` and `src/deploy` for more references.
- Update and expand unit tests.

See:
https://kworkflow.org/content/project_structure.html#components

alan-barzilay:
I was going to do issue #184, can I tackle both issues in the same PR or should them be separate things?

alan-barzilay:
I think I can work on this one after we merge #180.

Maybe we should create some sort of template for kw to avoid this sort of thing? Something to make it more uniform how we handle different flags, inputs and help

harnaman-hk:
Hi. I can work on this issue.
Currently we use case blocks in both explore_parser and explore. I think we can merge the two.
Are there any other ideas for this?

alan-barzilay:
@harnaman-hk we are also adding a help option for the explore command, to avoid duplicating work I would suggest you make you branch from #180 since it hasn't been merged yet.

rodrigosiqueira:
@hilookas since you already made a PR for the explore feature, could you work on this issue?

hilookas:
@rodrigosiqueira I will investigate it.

PR #651, opened 2022-07-24
src: explore: Add parameter amount check

### LINKED ISSUE #409: Parameter does not handle errors correctly
**Describe the bug**
When using `kw e` with extra parameters, we can see the error:
```
$HOME/.local/lib/kw/kwlib.sh: line 93: ./: Is a directory
```

**To Reproduce**
1. Go to any repository with git
2. Try `kw e VALID_NAME VALID_NAME /path`

**Expected behavior**
Fail without showing any kw specific message.

 - OS: Ubuntu 20.01
 - Version: kw from unstable branch
 - Bash Version: 5.0.17

hilookas:
Without checking the number of arguments can cause unexpected arguments
to be passed into the function.

The `explore_git_log` function in the `explore.sh` file will treat the
third parameter `$3` (extra parameters given by the user) as a `flag`
parameter and pass it to `cmd_manager`.

Before the commit f16c80, this would cause an error as said
in issue #409.

After the commit f16c80, the error will not occur, but the flag was
still passed with a wrong value.

This commit will output an error when the user gives the wrong amount of
parameters.

Closes #409

### DIFF
--- a/src/explore.sh
+++ b/src/explore.sh
@@ -17,17 +17,40 @@ function explore()
   case "$ret" in
     1)        # LOG
       shift 1 # Remove 'log' string
+
+      if [[ "$#" -gt 2 && "$3" != "TEST_MODE" ]]; then
+        complain 'Too many parameters'
+        exit 22 # EINVAL
+      fi
+
       explore_git_log "$@"
       ;;
     2)        # Use GNU GREP
       shift 1 # Remove 'grep' string
+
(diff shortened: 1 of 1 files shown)


rodrigosiqueira:
Hi @hilookas , first of all, thanks a lot for this PR.

I'll merge it, but I realize we have other significant problems. This explore feature is old and does not follow kw code pattern. I created the issue below to modernize this feature; it should be easy since it is a very tiny feature. See the below link:

https://github.com/kworkflow/kworkflow/issues/223

PR #652, opened 2022-07-29
src: report: fix per tag hours of focus

### LINKED ISSUE #598: kw report per tag outputs wrong focus time
**Description:**

I have been using `kw pomodoro` for the past month and whenever I use the `kw report` feature to check how many hours I've done the total focus time of the tag I use most is incorrect.
It seems to be that the counter resets at 24h.

**How to Reproduce:**

1. create a tag
2. have more than 24h logged with said tag
3. generate a report

**Expected behavior:**

The total time spent per tag should be cumulative past 24h.

**System information:**

 - OS: EndeavourOS
 - Kw Version: unstable (commit: 9302673)
 - Bash Version 5.1.16(1)-release

kwy95:
The hours of focus per tag were previously displayed the output of using
the `date` command to convert the value from seconds, this had the side
effect of being capped at 24h. This is because the `date` command
converts seconds as the amount of seconds since the epoch, and so
returns the time of the corresponding date.

Closes #598

### DIFF
--- a/src/report.sh
+++ b/src/report.sh
@@ -279,7 +279,7 @@ function grouping_year_data()
   done
 }
 
-function show_total_work_hours()
+function calculate_total_work_hours()
 {
   local work_hours_sec="$1"
   local hours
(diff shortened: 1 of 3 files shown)


rodrigosiqueira:
Nice change!
Applied to the unstable branch. Thanks

PR #655, opened 2022-08-02
src: build: hide error generated by make

### LINKED ISSUE #398: kw b -i shows some unexpected messages
**Describe the bug**
After building an x86 kernel, I tried `kw b -i` and got some makefile messages.

**To Reproduce**
1. Compile an x86 kernel
2. Use kw b -i

**Expected behavior**
Just info messages handled by kw.

**Message**
`arch/x86/Makefile:148: CONFIG_X86_X32 enabled but no binutils support`

 - OS: Ubuntu  20.01
 - Version: kw from unstable branch
 - Bash Version: 5.0.7

hilookas:
When CONFIG_X86_X32 is enabled but the corresponding tool is not
installed, an error `CONFIG_X86_X32 enabled but no binutils support`
will be reported when executing `make kernelrelease`, although no error
was reported when compiling the kernel.

Since this build_info function only gets the current kernel version by
executing the `make kernelrelease` command, and this error does not
affect the getting process of the version, this commit simply hides the
error message.

Closes #398

### DIFF
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -178,7 +178,7 @@ function find_kernel_root()
 function get_kernel_release()
 {
   local flag="$1"
-  local cmd='make kernelrelease'
+  local cmd='make kernelrelease 2> /dev/null'
 
   [[ "$flag" != 'TEST_MODE' ]] && flag='SILENT'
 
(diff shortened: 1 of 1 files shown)


hilookas:
However, during testing, it was found that when the makefile is corrupted (add random characters to it), the make command will not run normally, but the command will still run normally and output an empty version.

A more comprehensive check needs to be added as future work.

hilookas:
Here's a stray command call `make kernelrelease` at deploy.sh:191
It can be changed to a get_kernel_release call and errors are handled uniformly in get_kernel_release function.

rodrigosiqueira:
Merged, thanks

PR #668, opened 2022-08-10
Remove alert notification from some subcommands

magalilemes:
Some subcommands from both `kw build` and `kw deploy` don't really need to have an alert for completion (such as `kw b -n`, or even an invalid one like `kw d -`), so remove the notification by exiting instead of returning when these subcommands are called.

Also a minor change to one of the arguments of alert_completion, instead of `kw db` say `kw bd`.

### DIFF
--- a/kw
+++ b/kw
@@ -105,7 +105,7 @@ function kw()
 
         kernel_build && deploy_main 1 "$@"
         local ret="$?"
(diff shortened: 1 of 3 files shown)


kwy95 (approved):
LGTM

rodrigosiqueira:
Very nice PR!

Issue #611, opened 2022-05-23
Show distro info via kw device


rodrigosiqueira:
**Description:**

`kw device` shows a lot of data from the target device. We should expand it to show distro info such as:

* Distro name
* Base distro
* Version

We can easily collect that information from os-release.

**Type:**

Other

**Why:**

* Improve report details

PR #669, opened 2022-08-10
Show infos as suggested by issue #611

Kamorst:


src/device_info.sh
```diff
@@ -201,27 +203,41 @@ function get_disk()
...
 {
   local target="$1"
-  local ip="$2"
-  local port="$3"
-  local os
+  local flag="$2"
+  local ip
+  local port
+  local raw_os_release
+  local cmd='cat ROOTPATH/etc/os-release'
```
kwy95, line 210: Use separate path argument and concatenate the ROOTPATH as needed.

tests/device_test.sh
```diff
@@ -149,21 +153,22 @@ function which_distro_mock()
...
-  alias detect_distro='detect_distro_mock'
-  get_os "$LOCAL_TARGET"
-  assert_equals_helper 'Failed to gather local target OS data' "($LINENO)" 'lala' "${device_info_data['os']}"
+  local output
 
-  # VM
-  get_os "$VM_TARGET"
-  assert_equals_helper 'Failed to gather VM target OS data' "($LINENO)" 'lala' "${device_info_data['os']}"
+  # Check local deploy calls the expected commands
+  declare -a expected_cmd=(
```
kwy95, line 159: Make this a string.

src/device_info.sh
```diff
@@ -201,27 +203,41 @@ function get_disk()
...
       ;;
   esac
 
-  device_info_data['os']="$os"
+  if [[ "$flag" == 'TEST_MODE' ]]; then
+    printf '%s\n' "$cmd"
+    return 0
+  fi
+
+  source <(echo "$raw_os_release")
```
kwy95, line 237: Don't use `echo`, use `printf` instead.

PR #680, opened 2022-09-10
documentation: Change en-dash to double dashes

hilookas:
`--force` will be rendered as `-force` (en-dash), which was confusing.

This commit checked all documentation and escape the `--` with `\--`


rodrigosiqueira:
Nice change!! Thanks a lot!

PR #682, opened 2022-09-11
src: deploy: Add `--no-reboot` option

### LINKED ISSUE #542: Add deploy parameter to control the reboot action better
**Description:**

Right now, we check if we need to reboot the system by reading the `kworkflow.config` file. This is a little bit annoying in some scenarios; in this sense, I'm suggesting to new parameter for the deploy option:

```
--reboot
--no-reboot
```

**Type:**

[Enhance]

hilookas:
This commit add `--no-reboot` option and improved the way kw combine cmd option and config file for reboot action. Cmd option will override config file for reboot action.

Docs and tests are also updated.

Closes #542

### DIFF
--- a/documentation/man/features/deploy.rst
+++ b/documentation/man/features/deploy.rst
@@ -8,7 +8,8 @@ SYNOPSIS
 ========
 *kw* (*d* | *deploy*) [\--remote <remote>:<port> | \--local | \--vm]
                       [\--setup]
-                      [-r | \--reboot] [-m | \--modules] [-s | \--ls-line]
+                      [-r | \--reboot] [\--no-reboot]
+                      [-m | \--modules] [-s | \--ls-line]
                       [-l | \--list] [-a | \--list-all]
                       [(-u | \--uninstall) <kernel-name>[,...]] [-f \--force]
                       [\--alert=(s | v | (sv | vs) | n)]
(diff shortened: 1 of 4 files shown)


documentation/man/features/deploy.rst
```diff
@@ -65,6 +66,9 @@ OPTIONS
 -r, \--reboot:
   Reboot machine after deploy.
 
+\--no-reboot:
+  Do not reboot machine after deploy.
```
rodrigosiqueira, line 70: Maybe:
Do not reboot the machine after deploy even if the reboot option is set in the config file.

src/deploy.sh
```diff
@@ -1006,13 +1006,6 @@ function run_kernel_install()
   name=${name:-'kw'}
   flag=${flag:-'SILENT'}
 
-  if [[ "$reboot" == 0 ]]; then
-    reboot_default="${deploy_config[reboot_after_deploy]}"
-    if [[ "$reboot_default" =~ 'yes' ]]; then
-      reboot=1
-    fi
-  fi
```
rodrigosiqueira, line 1014: Maybe I miss something, but I still want to read the config file to get the default behavior for the reboot option. What I expect from the `--reboot` or `--no-reboot` is that those options take high priority over the config file. For example:
* deploy.config set reboot by default
* kw deploy -> it will reboot after deploy
* kw deploy --no-reboot -> it will not reboot, even if the config file set reboot as the default behavior.

rodrigosiqueira (changes requested):

rodrigosiqueira:
I tested it, and everything looks fine.
Thanks! Merged into the unstable branch

PR #683, opened 2022-09-12
src: remote: Accept new key from remote

### LINKED ISSUE #610: kw ssh must handle the first ssh connection
**Description:**

The IP might change when we have a test machine connected to a very simple local network. However, even if you update `kworkflow.config` file to reflect the new change, you will see this message:

```
 IP: IP
 User: root
 Port: 22
```

Next, if you try to manually ssh to the machine, you will notice that kw failed because of this:

```
The authenticity of host 'IP (IP)' can't be established.
ED25519 key fingerprint is SHA256:DDDD.
This host key is known by the following other names/addresses:
    ~/.ssh/known_hosts:5: IP
Are you sure you want to continue connecting (yes/no/[fingerprint])?
```

You just have to say yes, and everything will work fine. This is super annoying and can mislead users; we need to improve how we handle ssh connections inside kw.

We probably need something simple like:

`ssh -o "StrictHostKeyChecking no" user@host`

**Type:**

Others

**Why:**

* We don't want to mislead users

hilookas:
After this commit, ssh will accept new key from remote and save the key to `known_hosts`.

Ref: https://unix.stackexchange.com/questions/33271/how-to-avoid-ssh-asking-permission

Closes #610

### DIFF
--- a/src/remote.sh
+++ b/src/remote.sh
@@ -23,11 +23,11 @@ function is_ssh_connection_configured()
   local user=${4:-${remote_parameters['REMOTE_USER']}}
   local remote_file=${5:-${remote_parameters[REMOTE_FILE]}}
   local remote_file_host=${5:-${remote_parameters[REMOTE_FILE_HOST]}}
-  local ssh_cmd="ssh -q -o BatchMode=yes -o ConnectTimeout=5 -p $port $user@$remote exit"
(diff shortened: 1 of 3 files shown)


rodrigosiqueira:
Nice change!

PR #685, opened 2022-09-12
src: mail: Fix passthrought options with spaces to send-mail

Kamorst:
I Needed to close PR https://github.com/kworkflow/kworkflow/pull/672 to create a new branch and rebase to unstable.


rodrigosiqueira:
@kwy95

src/mail.sh
```diff
@@ -1056,7 +1057,14 @@ function parse_mail_options()
...
           options_values['COMMIT_RANGE']+="$commit_count "
         fi
-        options_values['PASS_OPTION_TO_SEND_EMAIL']="$(str_strip "$* ${options_values['PATCH_VERSION']}")"
+        for i in "$@"; do
+          if [[ "${i}" =~ " " ]]; then
+            pass_option_to_send_email+=" ${i@Q}"
+          else
+            pass_option_to_send_email+=" ${i}"
+          fi
+        done
```
kwy95, line 1066: It should be possible to change all of this to a single use of the `Q` operator, like so:
```
pass_option_to_send_email="${*@Q}"
```

kwy95 (changes requested):
Thansk for this PR.

While testing this fix I came across some interesting behavior that should to be considered.
Some functions that manipulate the arguments could cause the separation of the values due to repeated quote removal, i.e. the `reposition_commit_count_arg` function. This means that by the time the arguments get to your implementation the relevant values could have been already separated.
Another consideration is that most parameters that will be passed through to `git send-email` do not get parsed by kw, which means they retain the `=`, it might be worth considering some sort of preliminary parse to organize the values in a similar way to what is done through `kw_parse`, but I'm also weary of adding unnecessary functionality to what is already one of the largest files we maintain.

src/mail.sh
Kamorst: This change made a lot of test fail. I could not make reasonable changes to the tests to make it work. 
I'll try to look it later, but could you look the failed tests to make sure that the changes is not creating more bugs?

rodrigosiqueira: @kwy95

rodrigosiqueira:
Hi @Kamorst & @davidbtadokoro 
Can we revive this change?

davidbtadokoro:
Of course. Going to work on that!

Kamorst:
Reverted to the old fix method.

src/mail.sh
```diff
@@ -1056,7 +1057,14 @@ function parse_mail_options()
           commit_count="${BASH_REMATCH[0]}"
           options_values['COMMIT_RANGE']+="$commit_count "
         fi
-        options_values['PASS_OPTION_TO_SEND_EMAIL']="$(str_strip "$* ${options_values['PATCH_VERSION']}")"
+        for i in "$@"; do
```
rodrigosiqueira, line 1060: Please, add a TODO comment to find a better way to handle this.

rodrigosiqueira:
Merged! Thanks.

PR #679, opened 2022-09-09
src: explore: Modernize explore

hilookas:
This commit replaces old parser with new `kw_parse`, updates explore entrypoint to `explore_main` and updates relevant tests.

Closes #223

Signed-off-by: Haiqin Cui <i@18kas.com>

PR #686, opened 2022-09-13
src: explore: Add only-source only-header function

### LINKED ISSUE #184: Add rules for searching headers and c files
During the work with Kernel, it is common to search for functions and structs, as a result, we can find the occurrence of a string match in multiple files. However, there are situations that the developer only wants to take a look at the herder file or a C file. For this reason, it could be nice to have a specialized search for the header and C file. For example: 

```bash
kw e vkms_set_crc_source
drivers/gpu/drm/vkms/vkms_composer.c:236:int vkms_set_crc_source(struct drm_crtc *crtc, const char *src_name)
drivers/gpu/drm/vkms/vkms_crtc.c:161:   .set_crc_source         = vkms_set_crc_source,
drivers/gpu/drm/vkms/vkms_drv.h:145:int vkms_set_crc_source(struct drm_crtc *crtc, const char *src_name);

kw e -h vkms_set_crc_source
drivers/gpu/drm/vkms/vkms_drv.h:145:int vkms_set_crc_source(struct drm_crtc *crtc, const char *src_name);

kw e -c vkms_set_crc_source
drivers/gpu/drm/vkms/vkms_composer.c:236:int vkms_set_crc_source(struct drm_crtc *crtc, const char *src_name)
drivers/gpu/drm/vkms/vkms_crtc.c:161:   .set_crc_source         = vkms_set_crc_source,
```

hilookas:
This commit introduces two options `--only-source` and `--only-header`.

This commit also adds a new sample header file to test this function.

Relevant documentations are also updated.

Closes #184

### DIFF
--- a/documentation/man/features/explore.rst
+++ b/documentation/man/features/explore.rst
@@ -6,7 +6,8 @@ kw-explore
 
 SYNOPSIS
 ========
-*kw* (*e* | *explore*) [(-l | \--log) | (-g | \--grep) | (-a | \--all)] <expr>
+*kw* (*e* | *explore*) [(-l | \--log) | (-g | \--grep) | (-a | \--all)]
+                       [(-c | \--only-source) | (-H | \--only-header)] <expr>
                        [-p] [<dir> | <file>]
 
 DESCRIPTION
@@ -37,3 +38,9 @@ OPTIONS
   ignores files inside .git, except if it is called inside .git directory. In
   other words, if you use this option you will notice that ``git grep`` is
   used first, and then GNU grep.
+
+-c | \--only-source:
+  With this option, it is possible to show only the results from the source.
+
+-H | \--only-header:
+  With this option, it is possible to show only the results from the header.
--- a/src/explore.sh
+++ b/src/explore.sh
@@ -5,59 +5,53 @@
 include "$KW_LIB_DIR/kwio.sh"
 include "$KW_LIB_DIR/kwlib.sh"
 
+# Hash containing user options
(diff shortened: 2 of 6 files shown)


hilookas:
See also: #679

documentation/man/features/explore.rst
```diff
@@ -37,3 +38,9 @@ OPTIONS
   ignores files inside .git, except if it is called inside .git directory. In
   other words, if you use this option you will notice that ``git grep`` is
   used first, and then GNU grep.
+
+-c | \--only-source:
+  With this option, it is possible to show only the results from the source.
+
+-h | \--only-header:
```
rodrigosiqueira, line 45: let's use -h for the help function to keep all parameters consistent.

src/explore.sh
```diff
@@ -14,7 +14,7 @@ function explore_main()
 {
   local ret
 
-  if [[ "$1" =~ -h|--help ]]; then
+  if [[ "$1" =~ --help ]]; then
```
rodrigosiqueira, line 17: -h is for a short help, and --help is for open the man page.

rodrigosiqueira (changes requested):

src/explore.sh
hilookas: What about "-H" for header "-h" for help? :D

rodrigosiqueira:
Nice PR!
Let's use just the PR https://github.com/kworkflow/kworkflow/pull/679 , you just need to add the extra commit in that PR.

PR #688, opened 2022-09-14
src: setup: Add root install support

### LINKED ISSUE #687: `sudo: command not found` in setup.sh with root user
**Description:**

I am newbie to arch linux. To test the kworkflow in arch linux, I installed arch linux with only root user.

When I run setup.sh, it gives an error:

`sudo: command not found`

**How to Reproduce:**

1. Fresh install of arch linux
2. Using root as login user
3. `./setup.sh --install`

**Expected behavior:**

Install package without sudo.

**Screenshots:**

```
[root@archlinux kworkflow]# ./setup.sh --install
Checking dependencies ...
Can we install the following dependencies ccache rsync perl-io-socket-ssl perl-authen-sasl python-pip lzop lzip librsvg texlive-bin python-virtualenv graphviz imagemagick python-sphinx dunst libpulse pulseaudio python-docutils qemu ? [y/N]: yes
./setup.sh: line 85: sudo: command not found
Can we install the following pip dependencies "sphinx-book-theme" "jinja2>=3" ? [y/N]: ^C
```

**System information:**

 - OS: Arch Linux
 - kw Version: unstable
 - Bash Version: 5.1.16(1)

hilookas:
After this commit, setup.sh will check if it is running as root and install packages without sudo if so.

Closes #687

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -82,7 +82,11 @@ function check_dependencies()
     fi
 
     # Install system packages
-    eval "sudo $cmd"
+    if [ "$EUID" -eq 0 ]; then
+      eval "$cmd"
+    else
+      eval "sudo $cmd"
+    fi
   fi
 
   while IFS='' read -r package; do


rodrigosiqueira:
Merged! Thanks

PR #689, opened 2022-09-16
src: lib: Add function to get config option and improve fix for #398

hilookas:
This commit will detect kernel version and ignore error messages only in v5.13~v5.17 with `CONFIG_X86_X32` enabled.

Take v5.13 kernel source for example, `Makefile:274` says kernelrelease will not using compiler helper where `try-run` function locate (Also see `Makefile:274` and `Makefile:601`). But in`arch/x86/Makefile`, it uses `try-run` which is not defined causing `x32_ld_ok` is blank failing next `if`.

To fix this problem, we need patch the kernel. Rather than it, hiding the error will less invasive to kernel code.

In kernel v5.18 and later, `CONFIG_X86_X32` was removed so warning was also removed.

In addition, helper function `get_kernel_config` will use `./scripts/config` in kernel source, so it is hard to add tests for it.

See: #398

### DIFF
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -168,6 +168,25 @@ function find_kernel_root()
   printf '%s\n' "$kernel_root"
 }
 
+# Get value for specific config option
+#
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/kwlib.sh` function `cmd_manager`
+# @name Name of option in config
+#
+# Note: Make sure that you called is_kernel_root before trying to execute this
+# function.
+function get_kernel_config()
+{
+  local flag="$1"
+  local name="$2"
+  local cmd="./scripts/config --state $name"
+
+  [[ "$flag" != 'TEST_MODE' ]] && flag='SILENT'
+
+  cmd_manager "$flag" "$cmd"
+}
+
 # Get the kernel release based on the command kernelrelease.
 #
 # @flag How to display a command, the default value is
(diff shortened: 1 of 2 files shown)


src/kwlib.sh
```diff
@@ -178,8 +197,14 @@ function find_kernel_root()
 function get_kernel_release()
 {
   local flag="$1"
-  # TODO: Maybe we need to remove this error redirection
-  local cmd='make kernelrelease 2> /dev/null'
+  local cmd='make kernelrelease'
```
rodrigosiqueira, line 200: I don't have a way to reproduce this bug in my current setup. Could you check if the flag `--silent` removes the requirement of this /dev/null? I mean:

```
make kernelrelease --silent
```

hilookas: It still there.
Here's .config file that I am using from ubuntu server 22.04.1
https://pastebin.com/NGNVzgE8
using kernel source v5.17 (any source from v5.13-v5.17 can reproduce this bug)

src/kwlib.sh
```diff
@@ -178,8 +197,14 @@ function find_kernel_root()
...
+
+  local patch_makefile
+  if [[ "$(get_kernel_version "$flag")" =~ ^5.13|^5.14|^5.15|^5.16|^5.17 ]]; then
```
kwy95, line 203: Use this regexp instead: `^5\.1[3-7]`

kwy95 (changes requested):

rodrigosiqueira:
Hi @hilookas , I'll close this PR due to the lack of updates. Feel free to re-open it with your updates.

PR #690, opened 2022-09-16
documentation: tutorial: Add note on the deploy.config for the RPI

mairacanal:
The RPI needs to keep the modules after they are installed, but the default kworkflow template states ``strip_modules_debug_option=yes``. So, explicitly tell the user that it is necessary to set ``strip_modules_debug_option=no`` on the config files.

Signed-off-by: Maíra Canal <mairacanal@riseup.net>


melissawen:
@mairacanal, thanks for clarifying the documentation for RPi deployment. In this sense, can you also include a fix in the init_template for rpi 32 and 64 bit?
* 32-bit: https://github.com/kworkflow/kworkflow/blob/unstable/etc/init_templates/rpi4-raspbian-32-cross-x86-arm/deploy.config#L20
* 64-bit: https://github.com/kworkflow/kworkflow/blob/unstable/etc/init_templates/rpi4-raspbian-64-cross-x86-arm/deploy.config#L20

mairacanal:
@melissawen I just added the fix to the templates. Thanks for noticing it!

Issue #696, opened 2022-09-23
`kw k --save` uses invalid function when `git commit` fails


davidbtadokoro:
**Description:**

`kw k --save` triggers the creation of a local repository in `kw/configs` (if not already present) to manage the configs files and commits the addition (or update) of the `.config` in question. If, by any reason, the commit fails, instead of calling the function `complain()` (that prints an error in red) it calls `fail()` which is not defined. Although the commit fails and there is an error of `fail: command not found`, the `.config` file is added to `kw` management.
 
**How to Reproduce:**

1. Unset your `git` account default identity with:
- `git config --global user.name ""`
- `git config --global user.email ""` (this second one may be optional)
2. Try to add a `.config` file with `kw k --save | -s "<name>"` 

**Expected behavior:**

The `fail()` call (line 134 of `kernel_config_manager.sh`) should be replaced by a `complain()` call and print the correct error message. Also, it should not add the `.config` file to `kw` management.

**Screenshots:**

<img width="993" alt="Captura de Tela 2022-09-23 às 01 43 05" src="https://user-images.githubusercontent.com/95055784/191893377-83080fd5-2c9f-4d10-89e8-85bc7ea0240d.png">

**System information:**

 - OS: Virtual Machine (QEMU) running Arch Linux
 - kw Version: unstable branch commit [52907dc](https://github.com/kworkflow/kworkflow/commit/52907dccb9956a2e17bc4539af70769b461ca9f4)
 - Bash Version: 5.1.16(1)-release (x86_64-pc-linux-gnu)

rodrigosiqueira:
Interesting...
Tbh, I'm now wondering if we can remove the git dependency from kernel-config-manager. In these 3 years of kw, I never checked the git history under kernel-config-manager, and I never heard of anyone that did that. Additionally, @mairacanal  reported issues related to that recently.
Maybe it is time for us to make things simpler and remove the git part from this feature...

davidbtadokoro:
It makes sense! I'm new to the project, but I could not figure out much of the benefit of having and keeping this local repository. As `kw k` offers good features to manage these configs, maybe the git part is indeed redundant.

Also, from your comment in me PR related to this issue, should I create another issue proposing this change? I'm kind of new to contributing in projects as a whole too ; - D

PR #697, opened 2022-09-23
src: kernel_config_manager: fix `kw --save | -s` behavior when `git commit` fails

davidbtadokoro:
After this patch, failing `kw --save | -s` because of error in commiting save operation of .config file no longer throws wrong error (undefined function) neither consolidates the operation.

Closes: #696

### DIFF
--- a/src/kernel_config_manager.sh
+++ b/src/kernel_config_manager.sh
@@ -131,6 +131,7 @@ function save_config_file()
   if [[ "$ret" == 1 ]]; then
     warning "Warning: $name: there's nothing new in this file"
   elif [[ "$ret" -gt 0 ]]; then
+    git reset --hard # Aborts save operation
     complain "Could not save user config files"
   else
     success "Saved $name"
--- a/src/kernel_config_manager.sh
+++ b/src/kernel_config_manager.sh
@@ -131,7 +131,7 @@ function save_config_file()
   if [[ "$ret" == 1 ]]; then
     warning "Warning: $name: there's nothing new in this file"
   elif [[ "$ret" -gt 0 ]]; then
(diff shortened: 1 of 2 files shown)


src/kernel_config_manager.sh
```diff
@@ -131,7 +131,7 @@ function save_config_file()
   if [[ "$ret" == 1 ]]; then
     warning "Warning: $name: there's nothing new in this file"
   elif [[ "$ret" -gt 0 ]]; then
-    fail "Could not save user config files"
+    complain "Could not save user config files"
```
rodrigosiqueira, line 135: Nice catch! Since we are here, could you replace those double quotes with single quotes?

src/kernel_config_manager.sh
```diff
@@ -131,6 +131,7 @@ function save_config_file()
   if [[ "$ret" == 1 ]]; then
     warning "Warning: $name: there's nothing new in this file"
   elif [[ "$ret" -gt 0 ]]; then
+    git reset --hard # Aborts save operation
```
rodrigosiqueira, line 134: If you are fixing a bug, we need a test that reproduces the issue. However, see my comment on the issue that you created; maybe we can take a different path here.

rodrigosiqueira (changes requested):

rodrigosiqueira:
Hi @davidbtadokoro , could you reduce the total of commits to just 3?
* Commit to replacing fail to complain
* One commit for the cosmetic changes
* One commit for the fix

davidbtadokoro:
For sure @rodrigosiqueira 
Thought of squashing some of those commits, but for the sake of organization I'm going to close this PR and create a clone of it.

PR #708, opened 2022-09-26
documentation: dependencies: fedora.dependencies: add missing deps

Kamorst:
Add "patch" and "sqlite" to the Fedora dependencies file which provides respectively "patch" and "sqlite3". Those dependencies are not installed by default on Fedora and should be installed with kw.

### DIFF
--- a/documentation/dependencies/fedora.dependencies
+++ b/documentation/dependencies/fedora.dependencies
@@ -22,3 +22,5 @@ git-email
 pv
 rsync
 ccache
+patch
+sqlite


documentation/dependencies/fedora.dependencies
```diff
@@ -22,3 +22,5 @@ git-email
 pv
 rsync
 ccache
+patch
```
rodrigosiqueira, line 25: Why do we need `patch`?

Kamorst: Lines 171 and 210 of ./src/backup.sh.

rodrigosiqueira:
Thanks! Merged.

PR #709, opened 2022-09-27
src: diff: implement feature to diff folders in kw

### LINKED ISSUE #661: Diff folders to show files in one folder but not the others
**Description/Why:**

Sometimes a developer might want to find out files that are in one kernel tree but not in another one. It might be useful to support this type of diff under the `kw diff`.

**Targeted users:**

* Developers that work with multiple trees

**API draft:**

```bash
kw diff folder-one folder-two
- file-y
- file-h
```
Maybe a good start point, could be this command:

```
diff <(cd dir1; ls) <(cd dir2; ls)
```
or
```
diff -rq tests/amdgpu/ ../another-tree/tests/amdgpu/
```

emanuellima1:
This commit implements a diff on folders that shows the files on the first target that are not on the other.

kw diff folder-one folder-two

Closes #661

Signed-off-by: Emanuel Lima <emanuel-lima@outlook.com>

### DIFF
--- a/src/diff.sh
+++ b/src/diff.sh
@@ -18,7 +18,7 @@ function diff_manager()
       return 0
     fi
 
-    if [[ ! -f "$file" ]]; then
+    if [[ ! -e "$file" ]]; then
       complain "Invalid path: $file"
       diff_help
       return 2 # ENOENT
@@ -48,6 +48,11 @@ function diff_manager()
   if [[ "$test_mode" == 'TEST_MODE' ]]; then
     printf '%s\n' "$target_1 $target_2 $interactive"
     return 0
+  fi
+
+  if [[ -d "$target_1" && -d "$target_2" ]]; then
+    diff_folders "$target_1" "$target_2"
+    return "$?"
   fi
 
   diff_side_by_side "$target_1" "$target_2" "$interactive"
@@ -88,6 +93,35 @@ function diff_parser_options()
     esac
   done
 
+}
+
+# Show the diff result between folders in two columns equally divided.
+#
+# @folder_1 Path to the first folder
+# @folder_2 Path to the second folder
+# @flag How to display a command, see `src/kwlib.sh` function `cmd_manager`
+#
+# Return:
+# In case of success, return 0, otherwise, return 22.
+function diff_folders()
+{
+  local folder_1="$1"
+  local folder_2="$2"
+  local flag="$3"
+  local diff_cmd
+  local columns
+
+  diff_cmd='diff --color=always --recursive --brief'
+  flag=${flag:-'SILENT'}
+
+  diff_cmd="${diff_cmd} ${folder_1} ${folder_2} | grep ${folder_1}"
+
+  if [[ ! -d "$folder_1" || ! -d "$folder_2" ]]; then
+    complain "Make sure that ${folder_1} and ${folder_2} are valid folders"
+    return 22 # EINVAL
+  fi
+
+  cmd_manager "$flag" "$diff_cmd"
 }
 
 # Show the diff result in two columns equally divided.
--- a/tests/diff_test.sh
+++ b/tests/diff_test.sh
@@ -39,6 +39,36 @@ function test_diff_side_by_side()
 
 }
 
+function test_diff_folders()
+{
+  local folder_1="${SAMPLES_DIR}/db_files"
+  local folder_2="${SAMPLES_DIR}/first_set_of_bytes_from_disk"
+
+  output=$(LANG=en_US.UTF-8 diff_folders "$folder_1" "$folder_2")
+  assertEquals "$output" "Only in $folder_1: init.sql"$'\n'"Only in $folder_1: insert.sql"
+}
+
+function test_diff_folders_no_difference()
+{
+  local folder_1="${SAMPLES_DIR}/db_files"
+  local folder_2="${SAMPLES_DIR}/db_files"
+
+  output=$(diff_folders "$folder_1" "$folder_2")
+  assertEquals "$output" ""
+}
+
+function test_diff_folders_invalid_path()
+{
+  local folder_1="${SAMPLES_DIR}/db_files"
+  local folder_2="${SAMPLES_DIR}/first_set_of_bytes_from_disk"
+
+  output=$(diff_folders 'an_invalid_file' "$folder_2")
+  assertEquals "($LINENO) Expected 22" 22 "$?"
+
+  output=$(diff_folders 'an_invalid_file' 'another_invalid_file')
+  assertEquals "($LINENO) Expected 22" 22 "$?"
+}
+
 function test_diff_manager()
 {
   local file_1="$SAMPLES_DIR/MAINTAINERS"
--- a/documentation/man/features/diff.rst
+++ b/documentation/man/features/diff.rst
@@ -11,7 +11,8 @@ SYNOPSIS
 DESCRIPTION
 ===========
 This is a wrapper to some useful usage of the diff command. By default, it shows
-diff files side-by-side in an interactive way.
+diff files side-by-side in an interactive way. If instead of files, two directories
+are passed to the command, it shows which files are on <folder1> but not on <folder2>.
 
 OPTIONS
 =======


src/diff.sh
```diff
@@ -50,6 +50,11 @@ function diff_manager()
     return 0
   fi
 
+  if [[ -d $target_1 && -d $target_2 ]]; then
```
rodrigosiqueira, line 53: Add double quotes around the bash variable.

src/diff.sh
```diff
@@ -90,6 +95,32 @@ function diff_parser_options()
...
+# @flag How to display a command, see `src/kwlib.sh` function `cmd_manager`
+function diff_folders()
+{
+  local folder_1="$1"
+  local folder_2="$2"
+  local flag="$3"
+  local diff_cmd
+  local columns
+
+  diff_cmd="diff --color=always -rq"
```
rodrigosiqueira, line 111: Replace `-r`with `--recursive` and `-q` with `--brief`.

src/diff.sh
```diff
@@ -90,6 +95,32 @@ function diff_parser_options()
...
+  flag=${flag:-''}
```
rodrigosiqueira, line 112: Let's add SILENCE by default.

src/diff.sh
```diff
@@ -90,6 +95,32 @@ function diff_parser_options()
...
+
+  diff_cmd="$diff_cmd $folder_1 $folder_2 | grep $folder_1"
+
+  if [[ ! -d "$folder_1" || ! -d "$folder_2" ]]; then
+    complain "Make sure that $folder_1 and $folder_2 are valid folders"
```
rodrigosiqueira, line 117: For string concatenation:

tests/diff_test.sh
```diff
@@ -39,6 +39,21 @@ function test_diff_side_by_side()
...
 
+function test_diff_folders()
+{
+  local folder_1="$SAMPLES_DIR/db_files"
+  local folder_2="$SAMPLES_DIR/first_set_of_bytes_from_disk"
+  local diff_cmd="diff --color=always -rq $folder_1 $folder_2 | grep $folder_1"
+
+  output=$(diff_folders 'an_invalid_file' "$folder_2")
+  ret="$?"
+  assertEquals "($LINENO) Expected 22" "22" "$ret"
```
rodrigosiqueira, line 50: You don't need `$ret`, use `$?` directly.

rodrigosiqueira, line 50: You also don't need double quotes around 22.

tests/diff_test.sh
```diff
@@ -39,6 +39,21 @@ function test_diff_side_by_side()
...
+
+  output=$(diff_folders 'an_invalid_file' 'another_invalid_file')
+  ret="$?"
+  assertEquals "($LINENO) Expected 22" "22" "$ret"
+}
```
rodrigosiqueira, line 55: Could you add some tests for valid cases? You can create some fake folders and test the function by calling it and comparing the output.

rodrigosiqueira (changes requested):

src/diff.sh
emanuellima1: I'm not sure what you mean by SILENCE

rodrigosiqueira:
Hi @emanuellima1 , any update here?

emanuellima1:
> Hi @emanuellima1 , any update here?

Writing the tests now

emanuellima1:
Is there anything else to be done on this PR?

src/diff.sh
```diff
@@ -50,7 +50,7 @@ function diff_manager()
     return 0
   fi
 
-  if [[ -d $target_1 && -d $target_2 ]]; then
+  if [[ -d "$target_1" && -d "$target_2" ]]; then
```
kwy95, line 53: Don't add commits fixing problems that were introduced in this PR, you should squash this commit with the previous one. You should only ever create extra commits to fix things that have already been merged to the main branch as you don't want to alter the history, but on PRs you should definetly adapt the commits as to make them "perfect".

kwy95 (changes requested):

src/diff.sh
```diff
@@ -50,6 +50,11 @@ function diff_manager()
     return 0
   fi
 
+  if [[ -d "$target_1" && -d "$target_2" ]]; then
+    diff_folders "$target_1" "$target_2"
+    return 0
```
rodrigosiqueira, line 55: `diff_folders` might return a failure status code, also return it instead of return 0; i.e., `return "$?"`.

src/diff.sh
```diff
@@ -90,6 +95,32 @@ function diff_parser_options()
 
 }
 
+# Show the diff result between folders in two columns equally divided.
+#
+# @folder_1 Path to the first folder
+# @folder_2 Path to the second folder
+# @flag How to display a command, see `src/kwlib.sh` function `cmd_manager`
```
rodrigosiqueira, line 102: Document the return values.

src/diff.sh
```diff
@@ -90,6 +95,32 @@ function diff_parser_options()
...
+function diff_folders()
+{
+  local folder_1="$1"
+  local folder_2="$2"
+  local flag="$3"
+  local diff_cmd
+  local columns
+
+  diff_cmd="diff --color=always --recursive --brief"
```
rodrigosiqueira, line 111: Use single-quotes here.

tests/diff_test.sh
```diff
@@ -39,6 +39,21 @@ function test_diff_side_by_side()
 
 }
 
+function test_diff_folders()
+{
+  local folder_1="$SAMPLES_DIR/db_files"
+  local folder_2="$SAMPLES_DIR/first_set_of_bytes_from_disk"
```
rodrigosiqueira, line 45: https://kworkflow.org/content/codingstyle.html#string-concatenation

tests/diff_test.sh
```diff
@@ -39,6 +39,21 @@ function test_diff_side_by_side()
...
+
+  output=$(diff_folders 'an_invalid_file' "$folder_2")
+  assertEquals "($LINENO) Expected 22" 22 "$?"
+
+  output=$(diff_folders 'an_invalid_file' 'another_invalid_file')
+  assertEquals "($LINENO) Expected 22" 22 "$?"
```
rodrigosiqueira, line 51: Split these feature tests into two parts, let's say:
```bash
test_diff_folder_invalid_path()
...
test_diff_folder()
```
Additionally, add a test of two files that have no difference.

rodrigosiqueira (changes requested):

emanuellima1:
I think this does it. I had to force push to squash the commits.

tests/diff_test.sh
```diff
@@ -39,6 +39,36 @@ function test_diff_side_by_side()
 
 }
 
+function test_diff_folders()
+{
+  local folder_1="${SAMPLES_DIR}/db_files"
+  local folder_2="${SAMPLES_DIR}/first_set_of_bytes_from_disk"
+
+  output=$(LANG=en_US.UTF-8 diff_folders "$folder_1" "$folder_2")
```
rodrigosiqueira, line 47: I did not get yet why we need `LANG=en_US.UTF-8`.

rodrigosiqueira:
Hi @emanuellima1 , this change lgtm but could you address my question about the LANG?

emanuellima1:
> Hi @emanuellima1 , this change lgtm but could you address my question about the LANG?

I thought I had addressed it. Without it the test will potentially give a false negative.

rodrigosiqueira:
> > Hi @emanuellima1 , this change lgtm but could you address my question about the LANG?
> 
> I thought I had addressed it. Without it the test will potentially give a false negative.

But why? If this is something important, we might consider adding it inside `diff_folders` function instead of adding it in the test. The idea of changing a function behavior in the test does not sound good to me since it looks like the change only focuses on making the test pass.

emanuellima1:
> But why? If this is something important, we might consider adding it inside `diff_folders` function instead of adding it in the test. The idea of changing a function behavior in the test does not sound good to me since it looks like the change only focuses on making the test pass.

I wrote the why above:

> As I said above, I am using LANG to force the output of diff to be in english so that the test can pass no matter the language of the tester's system. Because the test is a string comparison, if the tester's machine is in Japanese, the test would fail, even though the function is working correctly.

That is why I don't think it should be inside diff_folders. We can't alter the language of a regular use of the software.

rodrigosiqueira:
Merged.
Thanks

PR #711, opened 2022-09-28
src: remote: fix error message thrown when `kw ssh` fails using a remote config file

### LINKED ISSUE #695: `kw ssh` shows a blank error message
**Description:**

`kw ssh` shows a blank error message

**How to Reproduce:**

1. Disconnect your remote device from the network
2. kw ssh

The output is:
```
 IP:
 User:
 Port:
```

**Expected behavior:**

The message would contain the IP, the user, and the port from the remote.

**System information:**

 - OS: Fedora 36
 - kw Version: kw @ d6e3da105786e274eeef78b43a616d9800a45a78
 - Bash Version [use `bash --version`]: 5.1.16(1)-release (x86_64-redhat-linux-gnu)

davidbtadokoro:
This patch corrects the error message when a failed `kw ssh` occurs using a remote config file instead of explicitly typing the IP, user and port of the host (using `kw ssh --remote | -r`). It also adds a test to verify if the correct error message is being thrown to the user in three distinct cases:
- Explicitly typing the IP, user and port;
- Using a remote config file;
- When no remote config file is found (without explicitly declaring the host infos).

Closes #695

### DIFF
--- a/src/remote.sh
+++ b/src/remote.sh
@@ -21,8 +21,8 @@ function is_ssh_connection_configured()
   local remote=${2:-${remote_parameters['REMOTE_IP']}}
   local port=${3:-${remote_parameters['REMOTE_PORT']}}
   local user=${4:-${remote_parameters['REMOTE_USER']}}
-  local remote_file=${5:-${remote_parameters[REMOTE_FILE]}}
-  local remote_file_host=${5:-${remote_parameters[REMOTE_FILE_HOST]}}
+  local remote_file=${5:-${remote_parameters['REMOTE_FILE']}}
+  local remote_file_host=${5:-${remote_parameters['REMOTE_FILE_HOST']}}
   local ssh_cmd="ssh -q -o StrictHostKeyChecking=accept-new -o BatchMode=yes -o ConnectTimeout=5 -p $port $user@$remote exit"
 
   if [[ -z "$remote" && -z "$port" && -z "$user" ]]; then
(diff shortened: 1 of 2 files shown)


rodrigosiqueira:
Hi @davidbtadokoro ,
 Could you rebase your change in the latest code from the unstable branch?

davidbtadokoro:
Thanks for noticing @rodrigosiqueira !
Doing it now!

davidbtadokoro:
Noticed that there are no tests for the `is_ssh_connection_configured()` function.

Also, when no remote config file is found (without explicitly defining the host with `kw ssh -r`) `kw ssh` tries anyway to establish a connection, resulting in a `Bad port '@'` thrown to the user.

Me and @pomba123 are creating tests for this function and fixing this issue of trying to connect anyway when we already know that the connection is going to fail.

src/remote.sh
```diff
@@ -28,6 +28,8 @@ function is_ssh_connection_configured()
   if [[ -z "$remote" && -z "$port" && -z "$user" ]]; then
     if [[ -n "${remote_file}" ]]; then
       ssh_cmd="ssh -q -o StrictHostKeyChecking=accept-new -o BatchMode=yes -o ConnectTimeout=5 -F ${remote_file} ${remote_file_host} exit"
+    else
+      return 2 # ENOENT
```
kwy95, line 32: Again, you implemented a test that expects this behavior before this behavior was implemented. As I said, you shouldn't introduce broken tests on a commit.

kwy95 (changes requested):
The biggest takeaway is that you should keep your commits as atomic as possible, firstly by ensuring that all the changes in the commit are necessary to achieve the commits purpose, and nothig else. But also, ensure that all tests and functionality works at every commit, so that the tool could be used at any point of it's history.

davidbtadokoro:
Many thanks for the feedback @kwy95 !!

It makes sense, the tool should work at any point. I actually made the fixes before the tests, but idk why I thought it was best to commit the tests before 😭

Also, thanks for pointing the other improvements.
I think I am going to close this PR and make 2 others (of one commit each to correct the failure message bug and the error handling) to clean the messy git history.

PR #716, opened 2022-09-30
setup.sh: fix installed package detection

Kamorst:
The `setup.sh` uses `pacman -Qs "$package"` to query if a package is installed but the `s` flag searches with regex. This is not reliable because creates false positives specially with packages with short names (like `bc`). With `pacman -Ql "$package"` program searches with the exact same name if a `$package` is installed.


Kamorst:
> Hi @Kamorst , which false positive did you see? It is a good idea to describe it in the commit message.

Done.

rodrigosiqueira:
Merged! Thanks

PR #718, opened 2022-10-01
src: remote: handle error of remote config file not found

davidbtadokoro:
In the `is_ssh_connection_configured` function there was no error handling when kw did not find a remote config file. With this commit, rather than ignoring this case and trying to establish a connection anyway we return an error code of no entry (`ENOENT`) and avoid extra computation when we know that it will fail.

Co-authored-by: Everaldo Gomes Junior <everaldogjr@gmail.com>

### DIFF
--- a/tests/remote_test.sh
+++ b/tests/remote_test.sh
@@ -127,6 +127,85 @@ function test_populate_remote_info()
   }
 }
 
+function test_is_ssh_connection_configured()
+{
+  local remote='test_remote'
+  local user='test_user'
+  local port='22'
+  local flag='TEST_MODE'
+  local current_path="$PWD"
+
+  cd "$TEST_PATH" || {
+    fail "($LINENO) It was not possible to move to temporary directory"
+    return
+  }
+
+  remote_parameters['REMOTE_IP']="$remote"
+  remote_parameters['REMOTE_USER']="$user"
+  remote_parameters['REMOTE_PORT']="$port"
+
+  is_ssh_connection_configured "$flag" > /dev/null
+
+  assertEquals "($LINENO):" 0 "$?"
+
+  cd "$current_path" || {
+    fail "($LINENO) It was not possible to move to the original path"
+    return
+  }
+}
+
+function test_is_ssh_connection_configured_with_remote_config_file()
+{
+  local flag='TEST_MODE'
+  local current_path="$PWD"
+
+  cd "$TEST_PATH" || {
+    fail "($LINENO) It was not possible to move to temporary directory"
+    return
+  }
+
+  remote_parameters['REMOTE_IP']=''
+  remote_parameters['REMOTE_USER']=''
+  remote_parameters['REMOTE_PORT']=''
+  remote_parameters['REMOTE_FILE']="${TEST_PATH}/.kw/remote.config"
+  remote_parameters['REMOTE_FILE_HOST']='origin'
+
+  is_ssh_connection_configured "$flag" > /dev/null
+
+  assertEquals "($LINENO):" 0 "$?"
+
+  cd "$current_path" || {
+    fail "($LINENO) It was not possible to move to the original path"
+    return
+  }
+}
+
+function test_is_ssh_connection_configured_no_remote_config_file()
+{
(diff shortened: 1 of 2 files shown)


rodrigosiqueira:
Hi @davidbtadokoro could you rebase your change against the latest code in the unstable branch?

davidbtadokoro:
@rodrigosiqueira done 👍

tests/remote_test.sh
```diff
@@ -127,6 +127,47 @@ function test_populate_remote_info()
...
+{
+  local remote='test_remote'
+  local user='test_user'
+  local port='22'
+  local flag='TEST_MODE'
+
+  cd "$TEST_PATH" || {
+    fail "($LINENO) It was not possible to move to temporary directory"
+    return
+  }
```
rodrigosiqueira, line 140: Before that, add something like:
```
local current_path="$PWD"
... to the end of this test ...
  cd "$current_path" || {
    fail "($LINENO) It was not possible to move to the original path"
    return
  }
```

In some point, we will need to refactor this test to use this cd in and out approach.

tests/remote_test.sh
```diff
@@ -127,6 +127,47 @@ function test_populate_remote_info()
...
+
+  # Case 1: IP, user and port passed as command line arguments (kw ssh -r)
+  remote_parameters['REMOTE_IP']="$remote"
+  remote_parameters['REMOTE_USER']="$user"
+  remote_parameters['REMOTE_PORT']="$port"
+
+  is_ssh_connection_configured "$flag" > /dev/null
+
+  assertEquals "($LINENO):" '0' "$?"
```
rodrigosiqueira, line 149: You don't need quotes around 0.

tests/remote_test.sh
```diff
@@ -127,6 +127,47 @@ function test_populate_remote_info()
...
+  # Case 2: Using a remote config file
+  remote_parameters['REMOTE_IP']=''
+  remote_parameters['REMOTE_USER']=''
+  remote_parameters['REMOTE_PORT']=''
+  remote_parameters['REMOTE_FILE']="${TEST_PATH}/.kw/remote.config"
+  remote_parameters['REMOTE_FILE_HOST']='origin'
+
+  is_ssh_connection_configured "$flag" > /dev/null
+
+  assertEquals "($LINENO):" '0' "$?"
```
rodrigosiqueira, line 160: Could you create one specific function for checking this case?

tests/remote_test.sh
```diff
@@ -127,6 +127,47 @@ function test_populate_remote_info()
...
+
+  # Case 3: No remote config file found
```
rodrigosiqueira, line 162: Same as my previous comment.

tests/remote_test.sh
```diff
@@ -127,6 +127,47 @@ function test_populate_remote_info()
...
+  remote_parameters['REMOTE_FILE']=''
+  remote_parameters['REMOTE_FILE_HOST']=''
+
+  is_ssh_connection_configured "$flag"
+
+  assertEquals "($LINENO):" '2' "$?"
```
rodrigosiqueira, line 168: You can drop the quotes around 2

rodrigosiqueira (changes requested):

tests/remote_test.sh
davidbtadokoro: Do you mean creating another test function for this case? As this case pertains to the `is_ssh_configured()` function, isn't it ok to leave it here? Sorry, I am a little confused 😕

rodrigosiqueira: Yes, every case here can be a single test. I know the other tests are not like that, but this is something that we will need to change soon on this test file. Ideally, we want one function test per test case to make it easier to debug issues in case of failures. You can have a function test named:

```
function test_is_ssh_connection_configured_with_remote_config_file()
{
...
}

function test_is_ssh_connection_no_remote_config_file()
...
```

davidbtadokoro: Oh, I see. Thanks for clarifying this for me. Going to work on that.

PR #720, opened 2022-10-04
src: kernel_config_manager: fix `kw (--save | -s)` behavior when git commit fails (CLONE)

### LINKED ISSUE #696: `kw k --save` uses invalid function when `git commit` fails

davidbtadokoro:
After this patch, failing `kw (--save | -s)` because of error in commiting save operation of .config file no longer throws wrong error (undefined function) neither consolidates the operation.

Closes #696 

Obs.: This PR is a clone of PR #697. Check this thread for more info.

### DIFF
--- a/src/kernel_config_manager.sh
+++ b/src/kernel_config_manager.sh
@@ -138,7 +138,7 @@ function save_config_file()
   ret="$?"
 
   if [[ "$ret" -gt 0 ]]; then
-    fail "Could not save user config files"
+    complain "Could not save user config files"
   else
     success "Saved $name"
   fi


src/kernel_config_manager.sh
```diff
@@ -140,6 +140,7 @@ function save_config_file()
   if [[ "$ret" == 1 ]]; then
     warning "Warning: $name: there's nothing new in this file"
   elif [[ "$ret" -gt 0 ]]; then
+    git reset --hard > /dev/null 2>&1 # Aborts save operation
```
rodrigosiqueira, line 143: Do we need `/dev/null`? Can we just add `--quiet`?

tests/kernel_config_manager_test.sh
```diff
@@ -84,6 +84,8 @@ function test_save_config_file_check_save_failures()
 {
   local current_path="$PWD"
   local ret=0
+  local git_name=""
+  local git_email=""
```
rodrigosiqueira, line 88: Use single-quotes.

tests/kernel_config_manager_test.sh
```diff
@@ -100,6 +102,19 @@ function test_save_config_file_check_save_failures()
   assert_equals_helper "Should return ENOENT, because '.config' != '.configuration'" "$LINENO" "$?" "2"
   rm .configuration
 
+  # Test git commit fail scenario
+  touch .config
+  git_name=$(git config user.name)
+  git_email=$(git config user.email)
+  git config --global user.name ""
+  git config --global user.email ""
```
rodrigosiqueira, line 110: single quotes.

tests/kernel_config_manager_test.sh
```diff
@@ -100,6 +102,19 @@ function test_save_config_file_check_save_failures()
...
+  ret=$(save_config_file "$NO_FORCE $NAME_1" "$DESCRIPTION_1")
```
rodrigosiqueira, line 111: Add newline after ret

rodrigosiqueira (changes requested):

src/kernel_config_manager.sh
davidbtadokoro: You are right 👍 
Doing it!

src/kernel_config_manager.sh
```diff
@@ -140,6 +140,7 @@ function save_config_file()
   if [[ "$ret" == 1 ]]; then
     warning "Warning: $name: there's nothing new in this file"
   elif [[ "$ret" -gt 0 ]]; then
+    git reset --hard --quiet # Aborts save operation
```
kwy95, line 143: Given #723 I don't think that this commit is necessary.

kwy95 (changes requested):
Please resist the urge to close and reopen PRs with the same content, the discussions and reviews present in a PR are part of the history of the development process and you should always try to keep it coherent. To do this you should learn to manipulate the branch and commit you are working on to keep a clean set of updated commits based on the reviews and the upstream state.

src/kernel_config_manager.sh
davidbtadokoro: Thanks for the clarification @kwy95, lesson learned 👍
Also think the commit is outdated with #723

rodrigosiqueira:
Hi @davidbtadokoro ,
Could you drop the second commit?
Also, could you include a test that covers the change in commit 1?

tests/kernel_config_manager_test.sh
kwy95: Some of these tests should be on the previous commit, and as previously stated this commit seems unnecessary considering #723

kwy95 (changes requested):

rodrigosiqueira:
Hey @davidbtadokoro , could you update this PR?

rodrigosiqueira:
Hey @davidbtadokoro , for this PR you just need to drop the second commit and add test for the first one.
Thanks

davidbtadokoro:
Hi @rodrigosiqueira , because of #723 isn't this PR obsolete at this point?

rodrigosiqueira:
> Hi @rodrigosiqueira , because of #723 isn't this PR obsolete at this point?

The first commit still valid.

davidbtadokoro:
Hi @rodrigosiqueira
 
As you requested, I dropped the second commit and actually redone the first because the commit message was outdated. However, I could not make the bug reproducible for testing, because #723 made the check dependent of a `cp` command. I.e. the only way to reproduce it was to find a deterministic way for the `cp` command to fail outside the function and I couldn't find a way to do it.

rodrigosiqueira:
Excellent, now we just need a test to cover that line. Could you add it?

davidbtadokoro:
Hi @rodrigosiqueira 

I am sorry, but I could not figure out a way to reproduce the bug for testing, i.e. make the `cp` command to fail. Please, see last comment I made.

kwy95 (approved):

rodrigosiqueira:
Merged, thanks

Issue #710, opened 2022-09-28
Missing pip dependency on archlinux


Kamorst:
**Description:**

On archlinux if **python-setuptools** is not installed the installer (`./setup.sh -i`) always asks to install pip dependencies even when they were already installed. 

This happens because at line 93: `python3 -c "import pkg_resources; pkg_resources.require('$package')" #&> /dev/null` pkg_resources is imported but it fails because **python-setuptools** provides it. The solution would be to add **python-setuptools** to the arch depencencies or change the way which the `setup.sh` get the information if a certain package is installed. It is advised by the setuptools [site](https://setuptools.pypa.io/en/latest/pkg_resources.html) to do not use pkg_resources.

**How to Reproduce:**

1. Clean install archlinux
2. run `./setup.sh -i`
3. run `./setup.sh -i` again but it will again to install pip packages which were installed at step 2.

**Expected behavior:**

Do not prompt to reinstall pip packages.

**Screenshots:**

![image](https://user-images.githubusercontent.com/31375521/192675603-75382af7-394b-4d5b-9ef0-b517daf1848e.png)

**System information:**

 - OS: archlinux 
 - kw Version: ```alpha
Branch: unstable
Commit: 52907dc```
 - Bash Version: `GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)`

SharmaDhruv2511:
I want to work on this. Is this still open???

PR #721, opened 2022-10-06
dependencies: arch.dependencies: add arch dependency

Kamorst:
Add `python-setuptools` to the arch dependencies list.

The `setup.sh -i` commands uses a python-setuptools package to detect if all pip dependencies packages are installed but it is not at the depencendies list.

Closes issue #710.

Signed-off-by: Pedro Rabello Sato <kamorst@usp.br>


rodrigosiqueira:
Thanks,
Merged

Issue #595, opened 2022-04-23
.config file is not deleted with `kw -r [name]` command


mairacanal:
**Description:**

The .config file is not deleted with the use of the command `kw -r [name]`. This can be checked by the fact that `kw g -l` still shows the config file and also the file is still on the folder `~/.local/share/kw/configs/configs`.

**How to Reproduce:**

```
➜ kw g -l
Name                           | Description

AMDGPU                         |
NO_DCN_CONFIG                  |

➜ kw g -r AMDGPU
This operation will remove AMDGPU from kw management
Are you sure that you want to proceed? [y/N]: y
The AMDGPU config file was removed from kw management

➜ kw g -l
Name                           | Description

AMDGPU                         |
NO_DCN_CONFIG                  |
```

**Expected behavior:**

I would expect to see the AMDGPU .config file deleted in this specific case.

**Screenshots:**

![image](https://user-images.githubusercontent.com/61601791/164943889-16a9d375-ca03-433a-832b-7887167f8eff.png)

**System information:**

 - OS: Fedora 35
 - Kw Version: alpha / Branch: unstable / Commit: 9302673
 - Bash Version: zsh 5.8.1

pomba123:
Can't reproduce this error on  commit [414fe8f](https://github.com/kworkflow/kworkflow/commit/414fe8fbad7118226b6053f632641380e3f0d581).

config,g - Set kw config options - The g option now is used to manage the config files of kw

kernel-config-manager,k - The k option now is used to manage config files of the target kernel

The commands have the expected behavior.

mairacanal:
I just reproduced on 52907dccb9956a2e17bc4539af70769b461ca9f4 on Fedora 36.
![image](https://user-images.githubusercontent.com/61601791/191862561-a1cbecfc-c501-4353-bb42-32bfcebee5a3.png)

mairacanal:
I did:
```
cd ~/.local/share/kw/configs
git config --local user.email <my email>
```

The output was: `fatal: --local can only be used inside a git repository`

Kamorst:
Is there any tracked .config file? If not, first create the git repo with `kw k  -s TMP` inside a linux git repo.
The only way I could replicate this bug is by creating the git which tracks the config files before setting user.email and user.name globally but the above [commands](https://github.com/kworkflow/kworkflow/issues/595#issuecomment-1263995398) fixed the issue.

mairacanal:
[...]

PR #723, opened 2022-10-15
src: kernel_config_manager.sh: Remove git depndency

Kamorst:
The KW subcommand kernel_config_manager used a rather complex scheme to tranck .config files. Now, the management is done using the file system with common commnads (like "rm" and "cp"). This will simplify and remove some complexity from the code.

It is added a function in the setup.sh script to deal with old kw installations that used git to manager .config files.

This could fix/avoid problems like the issue #595.


Kamorst:
This is a implementation of [this](https://github.com/kworkflow/kworkflow/issues/696#issuecomment-1256977525) suggestion.

kwy95 (changes requested):
Please review my comments.

setup.sh
```diff
@@ -165,6 +166,29 @@ function cmd_output_manager()
   return "$?"
 }
 
+# KW used git to track saved configs from kernel-config-manager.
+# It changed so this function removes the unused .git folder from
+# the dot_configs_dir
+function remove_k_legacy_git()
```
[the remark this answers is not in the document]
Kamorst: Replaced `remove_k_legacy_git` with `remove_kernel_config_manager_legacy_git`. I hope it makes things clearer.

setup.sh
```diff
@@ -25,6 +25,7 @@ declare -r sounddir="$sharedir/sound"
 declare -r datadir="${XDG_DATA_HOME:-"$HOME/.local/share"}/$app_name"
 declare -r etcdir="${XDG_CONFIG_HOME:-"$HOME/.config"}/$app_name"
 declare -r cachedir="${XDG_CACHE_HOME:-"$HOME/.cache/$app_name"}"
+declare -r dot_configs_dir="$datadir/configs"
```
rodrigosiqueira, line 28: `/$datadir/${datadir}/`

setup.sh
```diff
@@ -165,6 +166,27 @@ function cmd_output_manager()
   return "$?"
 }
 
+# KW used git to track saved configs from kernel-config-manager.
+# It changed so this function removes the unused .git folder from
+# the dot_configs_dir
```
rodrigosiqueira, line 171: What happens if we do not remove the unused .git?
Also, add a TODO as a reminder to drop this one day.

setup.sh
```diff
@@ -165,6 +166,27 @@ function cmd_output_manager()
...
+function remove_kernel_config_manager_legacy_git()
```
rodrigosiqueira, line 172: maybe:
remove_legacy_git_from_kernel_config_manager

setup.sh
```diff
@@ -165,6 +166,27 @@ function cmd_output_manager()
...
+{
+  local -r original_path="$PWD"
+
+  [[ ! -d "$dot_configs_dir"/.git ]] && return
```
rodrigosiqueira, line 176: `/$dot_configs_dir/.git/"${dot_configs_dir}/.git"/`

setup.sh
```diff
@@ -165,6 +166,27 @@ function cmd_output_manager()
...
+
+  if pushd "$dot_configs_dir"; then
+    rm -rf .git/
+    popd || {
+      complain "Could not return to original path from dot_configs_dir"
+      exit 1
+    }
+  else
+    complain 'Could not cd to dot_configs_dir'
```
rodrigosiqueira, line 185: Add the `dot_configs_dir` value as part of this complain; this can be useful for debugging.

src/kernel_config_manager.sh
```diff
@@ -131,15 +130,14 @@ function save_config_file()
     printf '%s\n' "$description" > "$metadata_dir/$name"
   fi
 
+  if cmp -s "$original_path/.config" "$dot_configs_dir/$configs_dir/$name"; then
```
rodrigosiqueira, line 133: For string concatenation, we use:

src/kernel_config_manager.sh
```diff
@@ -562,14 +560,13 @@ function remove_config()
   basic_config_validations "$target" "$force" 'Remove' "$msg"
 
   cd "$dot_configs_dir" || exit_msg 'It was not possible to move to configs dir'
-  git rm "$configs_dir/$target" "$dot_configs_dir/$metadata_dir/$target" > /dev/null 2>&1
-  git commit -m "Removed $target config: $USER - $(date)" > /dev/null 2>&1
+  rm "$configs_dir/$target" "$dot_configs_dir/$metadata_dir/$target" || exit_msg 'Could not remove config file'
```
rodrigosiqueira, line 563: Use an if condition check here to make it easier to see the error message.

setup.sh
```diff
@@ -165,6 +166,27 @@ function cmd_output_manager()
...
+# KW used git to track saved configs from kernel-config-manager.
+# It changed so this function removes the unused .git folder from
+# the dot_configs_dir
+function remove_kernel_config_manager_legacy_git()
+{
+  local -r original_path="$PWD"
+
+  [[ ! -d "$dot_configs_dir"/.git ]] && return
+
+  if pushd "$dot_configs_dir"; then
```
rodrigosiqueira, line 178: The problem with `pushd` is the fact that this command displays the folder path, which I think is unnecessary. Is it possible to make it quiet? If not, let's use the `cd` command.

rodrigosiqueira (changes requested):

rodrigosiqueira:
Also, update the commit message. It has a typo.

setup.sh
Kamorst: `kw k` will function as intended. The .git folder will be superfluous and it will not be updated but the functionality of `kw k` will stay the same.

setup.sh
Kamorst: It's possible to redirect stdout to null. I changed the code to append `&>/dev/null` but if you think that using `cd` is more elegant i can change the code.

src/kernel_config_manager.sh
```diff
@@ -562,14 +560,15 @@ function remove_config()
   basic_config_validations "$target" "$force" 'Remove' "$msg"
 
   cd "$dot_configs_dir" || exit_msg 'It was not possible to move to configs dir'
-  git rm "$configs_dir/$target" "$dot_configs_dir/$metadata_dir/$target" > /dev/null 2>&1
-  git commit -m "Removed $target config: $USER - $(date)" > /dev/null 2>&1
+  if ! rm "$configs_dir/$target" "$dot_configs_dir/$metadata_dir/$target"; then
```
rodrigosiqueira, line 563: Could you move the `rm` command outside this `if`? Usually, `rm` can be dangerous, and it is better to make it easier to see in the code. Also, make sure that both directories exist.
Finally, for string concatenation, check:

src/kernel_config_manager.sh
```diff
@@ -562,14 +560,15 @@ function remove_config()
...
+    exit_msg 'Could not remove config file'
+  fi
   cd "$original_path" || exit_msg 'It was not possible to move back from configs dir'
 
   say "The $target config file was removed from kw management"
 
   # Without config file, there's no reason to keep config directory
-  if [ ! "$(ls "$dot_configs_dir")" ]; then
+  if [ ! "$(ls "$dot_configs_dir"/"$configs_dir")" ]; then
```
rodrigosiqueira, line 571: https://kworkflow.org/content/codingstyle.html#string-concatenation

Also, can we move this `ls` outside the `if` condition?

rodrigosiqueira (changes requested):

kwy95 (approved):
LGTM

rodrigosiqueira:
Merged! Thanks a lot

PR #725, opened 2022-10-22
src: config: add configurations listing

### LINKED ISSUE #701: Make kw config show all variables
**Description:**

It would be useful if kw config shows all configurations. Something like this:

```
$ kw config
   deploy.reboot yes
   ...
   build.use_llvm yes
   ...
```

**Type:**

Improvement

**Why:**
This will make kw config more friendly

davidbtadokoro:
As requested in #701, this PR implements a feature in `kw config` for listing the configurations in the `kw` environment.

Considering the requests of the issue, there are two differences:
1. Instead of using the suggested option `--list` I went for `--show`, because the clear short option for `--list` (`-l`) was already in use in contrast with `-s` that was available.
2. The issue requested that the `kw vars` functionality was incorporated in the feature. However, I think some of the output of `kw vars` is not consistent with the options configurable through `kw config` and this output seems kind of innacurate.

Given those, how should I proceed? This implementation of the feature is satisfying or should I address these two differences?

**Use:**
`kw (config | config --show | config -s)`: list all options configurable through `kw config` including those there are not set.
`kw config (--show | -s) <config_target>...`: list all options of the targets configurable through `kw config` including those there are not set.

Closes #701

### DIFF
--- a/src/config.sh
+++ b/src/config.sh
@@ -27,6 +27,7 @@ function config_main()
   local option_and_value
   local config_options
   local base_path="${PWD}/.kw"
+  local is_show_configurations=false
   local option
   local value
 
@@ -41,8 +42,15 @@ function config_main()
     return 22 # EINVAL
   fi
 
-  # Validate and decompose options
   parameters="${options_values['PARAMETERS']}"
+  is_show_configurations="${options_values['IS_SHOW_CONFIGURATIONS']}"
+
+  if [[ -n "${is_show_configurations}" ]]; then
+    show_configurations "$parameters"
+    return "$?"
+  fi
+
+  # Validate and decompose options
   validate_option_parameter "$parameters"
   if [[ "$?" != 0 ]]; then
     complain "Invalid options: ${parameters}"
@@ -192,8 +200,8 @@ function set_config_value()
 
 function parse_config_options()
 {
-  local long_options='help,global,local'
-  local short_options='h,g,l'
+  local long_options='help,global,local,show'
+  local short_options='h,g,l,s'
 
   options="$(kw_parse "$short_options" "$long_options" "$@")"
 
@@ -206,6 +214,12 @@ function parse_config_options()
   # Default values
   options_values['SCOPE']='local'
   options_values['PARAMETERS']=''
+  options_values['IS_SHOW_CONFIGURATIONS']=''
+
+  # 'kw config' should list all configurations
+  if [[ "$#" == 0 ]]; then
+    options_values['IS_SHOW_CONFIGURATIONS']=1
+  fi
 
   eval "set -- $options"
 
@@ -223,6 +237,10 @@ function parse_config_options()
         options_values['SCOPE']='local'
         shift
         ;;
+      --show | -s)
+        options_values['IS_SHOW_CONFIGURATIONS']=1
+        shift
+        ;;
       --)
         shift
         ;;
@@ -232,6 +250,64 @@ function parse_config_options()
         ;;
     esac
   done
+}
+
+# This function lists all the configurations if no argument is passed or lists
+# the configurations of one or more specific config files passed as an argument.
+#
+# @target_config_files: specific configs to be listed
+#
+# Return:
+# In case of success return 0, otherwise, return 22.
+function show_configurations()
+{
+  local -a target_config_files="$1"
+  local -a configs
+  local -a options
+  local options_buffer
+  local value
+
+  include "${KW_LIB_DIR}/kw_config_loader.sh"
+
+  # Check which configs we need to show
+  if [[ -n "${target_config_files}" ]]; then
+    read -ra configs <<< "${target_config_files}"
+  else
+    read -ra configs <<< "${!config_file_list[@]}"
+  fi
+
+  # For each config file
+  for config in "${configs[@]}"; do
+    # Check if it is a valid config
+    if ! is_config_file_valid "$config"; then
+      complain "Invalid config target: ${config}"
+      return 22 # EINVAL
+    fi
+
+    # Load corresponding config file
+    eval "load_${config}_config"
+
+    # This part is heavily depedent of the names defined in kw_config_loader
+    if [[ "$config" == 'kworkflow' ]]; then
+      declare -n config_array='configurations'
+    else
+      declare -n config_array="${config}_config"
+    fi
+
+    # For each possible option in a config file
+    options_buffer=$(printf '%s' "${config_file_list[$config]}" | sed --null-data 's/\n/ /g')
+    read -ra options <<< "${options_buffer}"
+    for option in "${options[@]}"; do
+      value="${config_array[$option]}"
+      if [[ -z "$value" ]]; then
+        warning "${config}.${option}=N/A"
+      else
+        printf '%s\n' "${config}.${option}=${value}"
+      fi
+    done
+  done
+
+  return 0
 }
 
 function config_help()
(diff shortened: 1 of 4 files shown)


src/config.sh
```diff
@@ -35,6 +35,11 @@ function config_main()
     exit 0
   fi
 
+  if [[ "$#" == 0 ]]; then
+    show_configurations
+    exit 0
```
rodrigosiqueira, line 40: `show_configurations` may return a failure code. Use `$?` instead of 0, and replace `exit` with `return`.

src/config.sh
```diff
@@ -223,6 +228,11 @@ function parse_config_options()
         options_values['SCOPE']='local'
         shift
         ;;
+      --show | -s)
+        shift 2
+        show_configurations "$@"
```
rodrigosiqueira, line 233: I think it would be better if you use something like `options_values['SHOW_CONFIG']+="$2"`. I guess you'll need to check if `$2` is empty; if so, just add 1 to the variable value.

src/config.sh
```diff
@@ -35,6 +35,11 @@ function config_main()
     exit 0
   fi
 
+  if [[ "$#" == 0 ]]; then
```
rodrigosiqueira, line 38: It would be better if you have a check like this:
```
if [[ -n options_values['SHOW_CONFIG'] ]]
...
```

src/config.sh
```diff
@@ -234,6 +244,64 @@ function parse_config_options()
...
+# In case of success return 0, otherwise, return 22.
+function show_configurations()
+{
+  local -a specific_configs="$*"
+  local -a configs
+  local -a options
+  local options_buffer
+  local value
+
+  include "$KW_LIB_DIR/kw_config_loader.sh"
```
rodrigosiqueira, line 262: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/config.sh
```diff
@@ -234,6 +244,64 @@ function parse_config_options()
...
+    read -ra configs <<< "${!config_file_list[@]}"
+  else
+    read -ra configs <<< "${specific_configs}"
+  fi
+
+  # For each config file
+  for config in "${configs[@]}"; do
+    # Check if it is a valid config
+    if ! is_config_file_valid "$config"; then
+      complain "Invalid config target: $config"
```
rodrigosiqueira, line 275: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/config.sh
```diff
@@ -234,6 +244,64 @@ function parse_config_options()
...
+
+    # This part is heavily depedent of the names defined in kw_config_loader
+    if [[ "$config" == 'kworkflow' ]]; then
+      declare -n config_array='configurations'
+    else
+      declare -n config_array="${config}_config"
+    fi
+
+    # For each possible option in a config file
+    options_buffer=$(printf '%s' "${config_file_list[$config]}" | sed -z 's/\n/ /g')
```
rodrigosiqueira, line 290: use the full option name, i.e., replace `-z` with `--null-data`.

src/config.sh
```diff
@@ -244,5 +312,6 @@ function config_help()
   printf '%s\n' 'kw config <config.option value>:' \
     '  config - Show config values' \
     '  config (-g | --global) <config.option value> - Change global config' \
-    '  config (-l | --local) <config.option value> - Change local config'
+    '  config (-l | --local) <config.option value> - Change local config' \
+    '  config (-s | --show) [config]... - Show all or specific current configurations'
```
rodrigosiqueira, line 316: You also need to update the man documentation and maybe add this to some of the tutorials.

tests/config_test.sh
```diff
@@ -147,4 +151,134 @@ function test_parse_config_options()
...
+  }
+
+  output=$(show_configurations)
+
+  printf '%s' "$output" | grep -qFx 'vm.virtualizer=libvirt'
+  assertTrue "($LINENO) - Wrong configuration output" "$?"
+  printf '%s' "$output" | grep -qFx 'vm.mount_point=/home/lala'
+  assertTrue "($LINENO) - Wrong configuration output" "$?"
+  printf '%s' "$output" | grep -qFx 'vm.qemu_path_image=/home/xpto/p/virty.qcow2'
+  assertTrue "($LINENO) - Wrong configuration output" "$?"
```
rodrigosiqueira, line 172: Could you convert these checks to a function?

rodrigosiqueira (changes requested):
Very nice change!

davidbtadokoro:
Made the changes. Here are some notes:

1. Did a refactoring on the flow of execution. Rather than directly calling `show_configurations()` from `parse_config_options()`, the switch-case sets a boolean to signalize that we want to show the configs and parses the arguments (config targets) like any other argument. Then the function is called from `config_main()`;
2. Don't know if you saw, but options that are not set (but are configurable through `kw config`) are also displayed but using `warning()` with the value as N/A. Do you think this is a valid approach or should we only display the current setted options (this could be easily done)?

src/config.sh
davidbtadokoro: Noticed now that you referred to the check right at the beginning of `config_main()`. I thought you where referring to the check in the beginning of `show_configurations()` 😭 
Fixing it ASAP.

davidbtadokoro: [UPDATE]:
I actually just moved this check from `config_main()` to `parse_config_options()` and unified the entrypoint to `show_configurations()` through a check inside `config_main()` with a check similar to the one suggested.

src/config.sh
```diff
@@ -206,6 +214,12 @@ function parse_config_options()
   # Default values
   options_values['SCOPE']='local'
   options_values['PARAMETERS']=''
+  options_values['IS_SHOW_CONFIGURATIONS']=false
+
+  # 'kw config' should list all configurations
+  if [[ "$#" == 0 ]]; then
```
rodrigosiqueira, line 220: I think you want to move this condition to the `case` below.

src/config.sh
```diff
@@ -223,6 +237,10 @@ function parse_config_options()
         options_values['SCOPE']='local'
         shift
         ;;
+      --show | -s)
+        options_values['IS_SHOW_CONFIGURATIONS']=true
+        shift
+        ;;
```
rodrigosiqueira, line 243: You can probably use something like this:
```bash
options_values['IS_SHOW_CONFIGURATIONS']=1
shift

if [[ -n "$2" ]]; then
  options_values['IS_SHOW_CONFIGURATIONS']="$2" # You can store the new parameter here
  shift
fi
```

P.s.: This is just a draft, I did not test it.

rodrigosiqueira (changes requested):

rodrigosiqueira:
>     2. Don't know if you saw, but options that are not set (but are configurable through `kw config`) are also displayed but using `warning()` with the value as N/A. Do you think this is a valid approach or should we only display the current setted options (this could be easily done)?

I like your approach :)

src/config.sh
davidbtadokoro: I see. However, my implementation accepts more than one config target (`kw config -s vm notification` shows the options for both targets). By following your suggestion I don't see how to mantain this behaviour. Should I remove it and just deal with all or one target?

src/config.sh
davidbtadokoro: That was my intention, but I couldn't find a way to make it work. The problem is that in the `case` below, I don't think there is a way to recuperate the info of "no arguments passed". I based my approach in [this part](https://github.com/kworkflow/kworkflow/blob/unstable/src/kernel_config_manager.sh#L31) of the `kw k` implementation.

rodrigosiqueira:
Very nice work! I had some minor comments, but I decided to merge it anyway. I can deal with the small changes later.

Thanks a lot for this work! I really like it :)

davidbtadokoro:
Thank you! It's being a joy contributing to the project and I'm glad to see that you liked it!

PR #731, opened 2022-11-03
src: update: add self-update mechanism to kw

### LINKED ISSUE #303: Add self-update mechanism to kw
Usually, when a user that installed kw via `setup.sh` file want to update kw, they have to follow these steps:

1. Go to the kw repository
2. Update their branches
3. ./setup -i

We can make users' life much easier by adding an installation management system inside kw. For example:

```bash
kw update
```

This command could identify which branch the user used to install kw and update the current version of kw based on that. Let's say that the user used the `master` branch; when they run the update command, it will take the latest version from the upstream.

Since installing things from the unstable branch is quite common, we could provide something like this:

```bash
kw update --unstable
```

Finally, we could have an installation mechanism that does not require users to clone kw repository directly with this sort of change. We could have something like this:

```
curl -sSL https://an/url/setup.sh | bash -s setup.sh -i
```

Note: This feature should not be available for kw version installed via the distro package, i.e., this feature must only be available for versions installed via `./setup -i`/.

@joaoseckler any thoughts?

davidbtadokoro:
Now the user can update kw by just running `kw update` instead of having to go to a kw repo, checking out to the given branch and running the `setup.sh` script. By default, `kw update` clones the official kw repo in the `/tmp` and installs it from the given branch (just works with the master and unstable branch). If the user wants to update using a local repo, passing the path to this repo with the flag `--local` does the job. Also, if the user wants to install the current unstable version of kw, the flag `--unstable` can be used.

Closes #303

Co-authored-by: Everaldo Gomes Junior <everaldogjr@gmail.com>

### DIFF
--- a/documentation/man/features/self-update.rst
+++ b/documentation/man/features/self-update.rst
@@ -1,0 +1,40 @@
+==============
+kw-self-update
+==============
+
+.. _self-update-doc:
+
+SYNOPSIS
+========
+| *kw* (*u* | *self-update*)
+| *kw* (*u* | *self-update*) [(-u | --unstable)]
+| *kw* (*u* | *self-update*) [(-h | --help)]
+
+DESCRIPTION
+===========
+The `kw self-update` feature facilitates the process of updating kw.
+
+By just running `kw self-update`, the user can update `kw` based on the
+master or unstable branch in a simple way.
+
+By default, `kw self-update` updates `kw` based on the master branch. If
+the user wants to update based on the unstable branch, the `--unstable`
+option should be used.
+
+OPTIONS
+=======
+-u, \--unstable:
+  Update kw based on the unstable branch
+
+\--help:
+  Show this man page
+
+EXAMPLES
+========
+Updating based on the master branch::
+
+  kw self-update
+
+Updating based on the unstable branch::
+
+  kw self-update --unstable
--- a/kw
+++ b/kw
@@ -127,6 +127,13 @@ function kw()
         execute_checkpatch "$@"
       )
       ;;
+    self-update | u)
+      (
+        include "${KW_LIB_DIR}/self_update.sh"
+
+        self_update_main "$@"
+      )
+      ;;
     maintainers | m)
       (
         include "${KW_LIB_DIR}/maintainers.sh"
--- a/src/self_update.sh
+++ b/src/self_update.sh
@@ -1,0 +1,169 @@
+include "${KW_LIB_DIR}/kwlib.sh"
+include "${KW_LIB_DIR}/kwio.sh"
+include "${KW_LIB_DIR}/help.sh"
+
+declare -gA options_values
+
+function self_update_main()
+{
+  local target_branch='master'
+  local path_to_tmp_dir
+  local ret
+
+  parse_self_update_options "$@"
+  if [[ "$?" != 0 ]]; then
+    complain "Invalid option: ${options_values['ERROR']}"
+    self_update_help
+    return 22 # EINVAL
+  fi
+
+  if [[ -n "${options_values['UNSTABLE']}" ]]; then
+    target_branch='unstable'
+  fi
+
+  printf '%s\n' "Updating kw based on the branch ${target_branch}"
+  if [[ $(ask_yN 'Do you want to continue?') =~ '0' ]]; then
+    complain 'Update aborted.'
+    return 125 # ECANCELED
+  fi
+
+  path_to_tmp_dir=$(mktemp --directory)
+  update_from_official_repo "${target_branch}" "${path_to_tmp_dir}"
+
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain "Update from branch ${target_branch} failed"
+    return "$ret"
+  fi
+
+  success "Update based on the branch ${target_branch} successful!"
+}
+
+# Updates kw using the official kw repository.
+#
+# @target_branch Branch used to update
+# @path_to_tmp_dir Path to temp dir used for cloning repo
+# @flag Variable for testing
+#
+# Return:
+# In case of success return 0, otherwise, return the error
+# code.
+function update_from_official_repo()
+{
+  local target_branch="$1"
+  local path_to_tmp_dir="$2"
+  local flag="$3"
+  local original_pwd="$PWD"
+  local cmd
+  local ret
+
+  flag=${flag:-'SILENT'}
+
+  # Just to be sure temp dir is not root dir
+  if [ "${path_to_tmp_dir}" == '/' ]; then
+    complain "Aborting update! Invalid path to temporary directory: '${path_to_tmp_dir}'"
+    return 1 # EPERM
+  fi
+
+  cmd="git -C ${path_to_tmp_dir} clone --quiet 'https://github.com/kworkflow/kworkflow.git'"
+  cmd_manager "$flag" "$cmd"
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain 'Could not clone kworkflow repository'
+    return "$ret"
+  fi
+
+  cmd="git -C ${path_to_tmp_dir}/kworkflow checkout --quiet ${target_branch}"
+  cmd_manager "$flag" "$cmd"
+  if [[ "$?" != 0 ]]; then
+    complain "Could not checkout to branch ${target_branch}"
+    return 95 # EOPNOTSUPP
+  fi
+
+  cmd="git -C ${path_to_tmp_dir}/kworkflow pull --quiet 'origin' ${target_branch}"
+  cmd_manager "$flag" "$cmd"
+  if [[ "$?" != 0 ]]; then
+    complain "Could not pull from branch ${target_branch}"
+    return 95 # EOPNOTSUPP
+  fi
+
+  cmd="cd ${path_to_tmp_dir}/kworkflow"
+  cmd_manager "$flag" "$cmd"
+  if [[ "$?" != 0 ]]; then
+    complain 'Could not change to kworkflow directory'
+    return 2 # ENOENT
+  fi
+
+  cmd='bash setup.sh --install'
+  cmd_manager "$flag" "$cmd"
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain 'Could not update kw'
+    cd "${original_dir}" || complain 'Could not change back to original directory'
+    return "$ret"
+  fi
+
+  cmd="cd ${original_dir}"
+  cmd_manager "$flag" "$cmd"
+  if [[ "$?" != 0 ]]; then
+    complain 'Could not change back to original directory'
+    return 2 # ENOENT
+  fi
+
+  cmd="rm -rf ${path_to_tmp_dir}"
+  cmd_manager "$flag" "$cmd"
+  if [[ "$?" != 0 ]]; then
+    complain 'Could not remove temporary directory'
+    return 1 # EPERM
+  fi
+
+  return 0
+}
+
+function parse_self_update_options()
+{
+  local long_options='unstable,help'
+  local short_options='u,h'
+
+  options="$(kw_parse "$short_options" "$long_options" "$@")"
+
+  if [[ "$?" != 0 ]]; then
+    options_values['ERROR']="$(kw_parse_get_errors 'kw update' "$short_options" \
+      "$long_options" "$@")"
+    return 22 # EINVAL
+  fi
+
+  eval "set -- $options"
+
+  while [[ "$#" -gt 0 ]]; do
+    case "$1" in
+      --unstable | -u)
+        options_values['UNSTABLE']=1
+        shift
+        ;;
+      --help | -h)
+        self_update_help "$1"
+        exit
+        ;;
+      --)
+        shift
+        ;;
+      *)
+        shift
+        ;;
+    esac
+  done
+}
+
+function self_update_help()
+{
+  if [[ "$1" == --help ]]; then
+    include "${KW_LIB_DIR}/help.sh"
+    kworkflow_man 'self-update'
+    return
+  fi
+  printf '%s\n' 'kw self-update:' \
+    '  kw self-update - Update kw based on the master branch' \
+    '  kw self-update (-u | --unstable) - Update kw based on the unstable branch' \
+    '  kw self-update --help - Show the man page for kw self-update'
+}
--- a/documentation/conf.py
+++ b/documentation/conf.py
@@ -176,6 +176,7 @@ man_pages = [
     ('man/features/report', 'report', 'user data report support', [author], 1),
     ('man/features/ssh', 'ssh', 'ssh access', [author], 1),
     ('man/features/vars', 'vars', 'view kw config values', [author], 1),
+    ('man/features/self-update', 'self-update', 'kw self-update mechanism', ['David Tadokoro, Everaldo Junior'], 1),
 ]
 
 
--- a/documentation/man/kw.rst
+++ b/documentation/man/kw.rst
@@ -92,6 +92,7 @@ the previous sections.
   | :ref:`kw-report<report-doc>`
   | :ref:`kw-pomodoro<pomodoro-doc>`
   | :ref:`kw-mail<mail-doc>`
+  | :ref:`kw-self-update<self-update-doc>`
 
 clear-cache
 ~~~~~~~~~~~
--- a/src/bash_autocomplete.sh
+++ b/src/bash_autocomplete.sh
@@ -72,6 +72,9 @@ function _kw_autocomplete()
 
   kw_options['env']='--create --list --use'
 
+  kw_options['self-update']='--unstable --help'
+  kw_options['u']="${kw_options['self-update']}"
+
   mapfile -t COMPREPLY < <(compgen -W "${kw_options[${previous_command}]} " -- "${current_command}")
 
   # TODO:
--- a/src/help.sh
+++ b/src/help.sh
@@ -33,6 +33,7 @@ function kworkflow_help()
     '  man - Show manual pages' \
     '  pomodoro,p - kw pomodoro support' \
     '  report,r - Show kw pomodoro reports and kw usage statistics' \
+    '  self-update,u - kw self-update mechanism' \
     '  ssh,s - SSH support' \
     '  vars - Show variables' \
     '  version,--version,-v - Show kw version'
--- a/tests/self_update_test.sh
+++ b/tests/self_update_test.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+
+include './src/self_update.sh'
+include './tests/utils.sh'
+
+function oneTimeSetUp()
+{
+  export TEST_PATH="${SHUNIT_TMPDIR}/test_path"
+  export original_dir="$PWD"
+  export fake_dir_path="${SHUNIT_TMPDIR}/fake_dir"
+}
+
+function setUp()
+{
+  mkdir -p "${fake_dir_path}" || {
+    fail "($LINENO) It was not possible to create the fake directory"
+    return
+  }
+
+  cd "${SHUNIT_TMPDIR}" || {
+    fail "($LINENO) It was not possible to move to temporary directory"
+    return
+  }
+}
+
+function tearDown()
+{
+  rm -rf "${fake_dir_path}" || {
+    fail "($LINENO) It was not possible to remove the fake directory"
+    return
+  }
+
+  cd "${original_dir}" || {
+    fail "($LINENO) It was not possible to back to the kw folder"
+    return
+  }
+}
+
+function test_update_from_official_repo_based_on_master()
+{
+  update_from_official_repo 'master' "${fake_dir_path}" 'TEST_MODE' > /dev/null 2>&1
+  assertEquals "($LINENO) It should be able to update based on the master." 0 "$?"
+}
+
+function test_update_from_official_repo_based_on_unstable()
+{
+  update_from_official_repo 'unstable' "${fake_dir_path}" 'TEST_MODE' > /dev/null 2>&1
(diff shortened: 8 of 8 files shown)


davidbtadokoro:
Observation: We opted to clone the official repo and install `kw` from it because we can't be sure that the user has a `kw`repo in his fs.

rodrigosiqueira (changes requested):
Hi @davidbtadokoro ,
That's a very nice change! I added some comments in the PR.

documentation/man/features/self-update.rst
```diff
@@ -0,0 +1,47 @@
...
+  Update kw from the unstable branch
+
+-h, \--help:
+  Show this man page
+
+EXAMPLES
+========
+Updating from the master branch::
+
+  kw u
```
rodrigosiqueira, line 42: Use the full argument names here.

src/self_update.sh
```diff
@@ -155,6 +159,11 @@ function parse_self_update_options()
 
 function self_update_help()
 {
+  if [[ "$1" == --help || "$1" == -h ]]; then
+    include "$KW_LIB_DIR/help.sh"
```
rodrigosiqueira, line 163: Use `{}` in the string concatenation.

src/self_update.sh
```diff
@@ -155,6 +159,11 @@ function parse_self_update_options()
 
 function self_update_help()
 {
+  if [[ "$1" == --help || "$1" == -h ]]; then
```
rodrigosiqueira, line 162: Note that the file `help.sh` has a quick summary of all features; add `kw self-update` there.

documentation/man/features/self-update.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Updating from the unstable branch::
+
+  kw u -u.
+
```
rodrigosiqueira, line 47: Remove the extra line at the end of the man.

src/self_update.sh
rodrigosiqueira, line 162: You don't want `-h` in this condition, otherwise, we will open the man page for the `-h` option. The `-h` option should be the short summary of the feature.

src/self_update.sh
```diff
@@ -0,0 +1,161 @@
...
+      *)
+        shift
+        ;;
+    esac
+  done
+}
+
+function self_update_help()
+{
+  printf '%s\n' 'Usage:' \
```
rodrigosiqueira, line 167: Instead of `Usage`, use `kw self-update`.

kw
```diff
@@ -136,6 +136,13 @@ function kw()
         execute_checkpatch "$@"
       )
       ;;
+    self-update | u)
```
rodrigosiqueira, line 139: You forgot to update the `bash_autocomplete.sh` file with this new option.

rodrigosiqueira (changes requested):
That is a very nice feature! I'm excited to merge it since it will make the kw users' life way easier :)

Please, try to address my comments.

davidbtadokoro:
Hi @rodrigosiqueira 

Firstly sorry for the delay.
Secondly I added two commits that refer to the updates in `src/help.sh` and `src/bash_autocomplete.sh`. As they are changes on different files from which we were working on I thought it made sense to separete them in other commits instead of squashing them with the documentation commit. However, I am in doubt if this doesn't pertain to the "documentation" aspect of the feature. In any way, it is simple to squash them, so, if you find necessary, just make the call.

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+
+  printf '%s\n' "Updating kw based on the branch ${target_branch}"
+  if [[ $(ask_yN 'Do you want to continue?') =~ '0' ]]; then
+    complain 'Update aborted.'
+    return 125 # ECANCELED
+  fi
+
+  flag="${options_values['TEST']}"
+  path_to_tmp_dir=$(mktemp --directory)
+  cmd="update_from_official_repo ${target_branch} ${path_to_tmp_dir}"
```
rodrigosiqueira, line 34: `update_from_official_repo` is your function, you don't need to use `cmd_manager` for that. Usually, we use `cmd_manager` for a global system command like `cd`, `rm`, `find`, etc.

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+  fi
+
+  success "Update based on the branch ${target_branch} successful!"
+}
+
+# Updates kw using the official kw repository.
+#
+# @target_branch Branch used to update
+# @path_to_tmp_dir Path to temp dir used for cloning repo
+# @test Indicates if it is a test or not
```
rodrigosiqueira, line 50: Do you mean `@flag`?

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+#
+# Return:
+# In case of success return 0, otherwise, return the error
+# code.
+function update_from_official_repo()
+{
+  local target_branch="$1"
+  local path_to_tmp_dir="$2"
+  local flag
```
rodrigosiqueira, line 59: I guess you forgot `flag="$3"`

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+  local original_dir="$PWD"
+  local cmd
+  local ret
+
+  cmd="cd ${path_to_tmp_dir}"
+  cmd_manager "$flag" "$cmd" > /dev/null
```
rodrigosiqueira, line 65: Why do we need `/dev/null`?

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+  if [[ "$?" != 0 ]]; then
+    complain "Could not change to ${path_to_tmp_dir} dir"
+    return 2 # ENOENT
+  fi
+
+  cmd="git clone --quiet 'https://github.com/kworkflow/kworkflow.git'"
+  cmd_manager "$flag" "$cmd" > /dev/null
```
rodrigosiqueira, line 72: Why `/dev/null`?

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+
+  cmd="cd kworkflow"
+  cmd_manager "$flag" "$cmd" > /dev/null
+  if [[ "$?" != 0 ]]; then
+    handle_error 'Could not change to kworkflow repository' \
+      "${original_dir}" 1 'remove_kw_dir'
+    return "$?"
+  fi
+
+  cmd="git checkout --quiet ${target_branch}"
```
rodrigosiqueira, line 87: btw, instead of `cd` you can use the `-C` option from git. This argument will simplify your code since you will not need to deal with `cd` in and out.

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+  if [[ "$?" != 0 ]]; then
+    cd '..'
+    handle_error "Could not pull from branch ${target_branch}" \
+      "${original_dir}" 95 'remove_kw_dir'
+    return "$?"
+  fi
+
+  if [[ -n "$3" ]]; then
+    flag='TEST_MODE'
+  fi
```
rodrigosiqueira, line 107: Just get this value at the beginning of this function, and you don't need this `if`. Check `build.sh` or `deply.sh` to see how we usually use this option.

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+  cmd="bash setup.sh --install"
```
rodrigosiqueira, line 108: if you stop to `cd`, you will need to use `$PATH_TO/setup.sh...`

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+# @original_dir Path to the original dir to be returned
+# @errno Errno code to be returned
+# @remove If not null string tries to remove kw cloned repository
+#
+# Return:
+# The errno code number passed as argument.
+function handle_error()
+{
+  error_message="$1"
+  original_dir="$2"
```
rodrigosiqueira, line 153: See my previous comment; you don't need to handle folders.

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+
+  complain "$error_message"
+  [[ -n "$remove" ]] && rm -rf 'kworkflow'
+  cd "${original_dir}" || exit 1
+  return "$errno"
+}
+
+function parse_self_update_options()
+{
+  local long_options='unstable,test'
```
rodrigosiqueira: We don't need to expose `test` to final users.

documentation/man/features/self-update.rst
```diff
@@ -0,0 +1,40 @@
+==============
```
rodrigosiqueira, line 1: This commit should be squashed with the first one.

rodrigosiqueira (changes requested):
Hi,
Thanks again for updating this change! See my comments, and remember to update bash_autocomplete.sh file.

src/self_update.sh
davidbtadokoro: The reason I used `cmd_manager` was to remove the logic (the `update_from_official_repo` call) from inside the `if` statement as you requested in the comment referenced below. Could not think of another way to check if it is in test mode (i.e. don't actually execute the function) without using an `if`.

> Fwiu, this `TEST` is only for unit test reasons; if so, it will be better if you invert the logic in a way that the code under this condition is moved outside this `if` statement. The reason for that is because `TEST` is not part of the normal code flow; it is more like an exception made only for test reasons.

src/self_update.sh
davidbtadokoro: As these commands must be executed even in testing (e.g. to check if the update fails using an invalid branch) we cant use the flag `TEST_MODE` or `SILENT`. This results in the command being outputted (with the `say` function) which causes problems for testing and also prints info to the user that should not be printed.

src/self_update.sh
davidbtadokoro: As I commented above, the reason I did this was because all the commands (except this) must be executed even in the testing scenario. Am I doing this wrong? Should I drop these test cases and execute all `cmd_manager` calls with `TEST_MODE` or `SILENT`?

davidbtadokoro:
Hi @rodrigosiqueira 

One last doubt: in the last changes I did update `bash_autocomplete.sh`. Do you mean update it again after these new changes?

kw
rodrigosiqueira: You added one commit for update `bash_autocomplete.sh`; that change should be squashed in this commit since this is the commit that introduces this new operation.

src/self_update.sh
```diff
@@ -0,0 +1,161 @@
...
+  local ret
+
+  parse_self_update_options "$@"
+  if [[ "$?" != 0 ]]; then
+    complain "Invalid option: ${options_values['ERROR']}"
+    self_update_help
+    return 22 # EINVAL
+  fi
+
+  if [[ -n "${options_values['UNSTABLE']}" ]]; then
```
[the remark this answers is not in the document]
rodrigosiqueira: Please, don't forget about this comment.

src/self_update.sh
rodrigosiqueira: I did not get it... what is the problem with simply calling:

```
update_from_official_repo ${target_branch} ${path_to_tmp_dir}
if [[ "$?" != 0 ]]; then ...
```

Do you want to use `cmd_manager` for trying to test this main function? If you test `update_from_official_repo`, you don't need to test it via `self_update_main`.

src/self_update.sh
rodrigosiqueira: Please, address my previous comment.

src/self_update.sh
rodrigosiqueira: In the test, you can create some fake directories to avoid the requirement of using `cmd_manager` for cding a folder. In the test setup phase, you can create a fake directory (see `$SHUNIT_TMPDIR` for inspiration) and use it as a parameter for `update_from_official_repo`; that way, you can test the cd in and out without the need to check the command itself which is more valuable.

src/self_update.sh
rodrigosiqueira: Can we remove this `/dev/null`?

src/self_update.sh
```diff
@@ -0,0 +1,209 @@
...
+  cmd="git clone --quiet 'https://github.com/kworkflow/kworkflow.git'"
+  cmd_manager "$flag" "$cmd" > /dev/null
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain 'Could not clone kworkflow repository'
+    return "$ret"
+  fi
+
+  cmd="cd kworkflow"
+  cmd_manager "$flag" "$cmd" > /dev/null
```
rodrigosiqueira, line 80: See my previous comment about using `cmd_manager` for `cd`.

rodrigosiqueira (changes requested):

davidbtadokoro:
Hi @rodrigosiqueira 
Thank you for addressing all my questions! Due to some other affairs I'm probably not going to be able to work on it in this week or the next 😢 (but until the holidays for sure).

davidbtadokoro:
Hey @rodrigosiqueira 

You were right, especifically about using `-C` for the git commands. This seemed to make cding in and out of directories unnecessary and REALLY simplified the code. However, using `bash path/to/setup.sh --install`  doesn't work as `setup.sh` assumes you are running it from inside a kworkflow repo. Tried searching for ways to work around it, but at the end of the day I had to cd in and out of the kworkflow repo, just to run `setup.sh` (if you know a way to do it without cding I would be glad to change it!). As for the rest of your requests, I think all are solved 😄

davidbtadokoro:
Also, the "Test setup and doc build / setup-and-doc-build" failed, but I think it is unrelated to my changes.

Screen Shot 2022-12-08 at 23 58 08

It seems that actions can't resolve the `python-docutils` dependency.

rodrigosiqueira:
> Hey @rodrigosiqueira
> 
> You were right, especifically about using `-C` for the git commands. This seemed to make cding in and out of directories unnecessary and REALLY simplified the code. However, using `bash path/to/setup.sh --install` doesn't work as `setup.sh` assumes you are running it from inside a kworkflow repo. Tried searching for ways to work around it, but at the end of the day I had to cd in and out of the kworkflow repo, just to run `setup.sh` (if you know a way to do it without cding I would be glad to change it!). As for the rest of your requests, I think all are solved smile

Ideally, we may want to refactor `setup.sh` in the future to support something like `-C`; however, for now, you can deal with the `cd` only for the setup part. For the git part, use `-C`.

rodrigosiqueira:
> Also, the "Test setup and doc build / setup-and-doc-build" failed, but I think it is unrelated to my changes.
> 
> Screen Shot 2022-12-08 at 23 58 08
> 
> It seems that actions can't resolve the `python-docutils` dependency.

I already fixed it, rebase your change against the latest unstable branch.

davidbtadokoro:
Thanks @rodrigosiqueira done the rebase. Think everything is OK now.

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+  cmd_manager 'SILENT' "$cmd"
+  if [[ "$?" != 0 ]]; then
+    complain "Could not pull from branch ${target_branch}"
+    return 95 # EOPNOTSUPP
+  fi
+
+  cmd="cd ${path_to_tmp_dir}/kworkflow"
+  cmd_manager 'SILENT' "$cmd"
+  if [[ "$?" != 0 ]]; then
+    complain "Could not change to kworkflow directory"
```
rodrigosiqueira, line 85: If you do not have any string concatenation, use single quotes.

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+    return 2 # ENOENT
+  fi
+
+  cmd="bash setup.sh --install"
```
rodrigosiqueira, line 89: Use single-quotes, see previous comment.

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+  cmd_manager "$flag" "$cmd"
+  ret="$?"
+  if [[ "$?" != 0 ]]; then
+    complain "Could not update kw"
```
rodrigosiqueira, line 93: Use single quotes; see the previous comment.

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+    return "$ret"
+  fi
+
+  cmd="cd ${original_dir}"
+  cmd_manager 'SILENT' "$cmd"
+  if [[ "$?" != 0 ]]; then
+    complain "Could not change back to original directory"
```
rodrigosiqueira, line 100: Use single quotes; see the previous comment.

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+{
+  local target_branch="$1"
+  local path_to_tmp_dir="$2"
+  local flag="$3"
+  local original_pwd="$PWD"
+  local cmd
+  local ret
+
+  cmd="git -C ${path_to_tmp_dir} clone --quiet 'https://github.com/kworkflow/kworkflow.git'"
+  cmd_manager 'SILENT' "$cmd"
```
rodrigosiqueira, line 61: `cmd_manager` give us good debug flexibility via its flag. If you hardcode SILENT here, you will not take full advantage of it. For this reason, I recommend you to do something like this:
```bash
[..]
  local cmd
  local ret

  flag=${flag:-'SILENT'}
[..]
  cmd_manager "$flag" "$cmd"
```

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain 'Could not clone kworkflow repository'
+    return "$ret"
+  fi
+
+  cmd="git -C ${path_to_tmp_dir}/kworkflow checkout --quiet ${target_branch}"
+  cmd_manager 'SILENT' "$cmd"
```
rodrigosiqueira, line 69: Use: `cmd_manager "$flag" "$cmd"`

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+  if [[ "$?" != 0 ]]; then
+    complain "Could not checkout to branch ${target_branch}"
+    return 95 # EOPNOTSUPP
+  fi
+
+  cmd="git -C ${path_to_tmp_dir}/kworkflow pull --quiet 'origin' ${target_branch}"
+  cmd_manager 'SILENT' "$cmd"
```
rodrigosiqueira, line 76: Use: `cmd_manager "$flag" "$cmd"`

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+  if [[ "$?" != 0 ]]; then
+    complain "Could not change to kworkflow directory"
+    return 2 # ENOENT
+  fi
+
+  cmd="bash setup.sh --install"
+  cmd_manager "$flag" "$cmd"
+  ret="$?"
+  if [[ "$?" != 0 ]]; then
```
rodrigosiqueira, line 92: Two problems here:
1. If `setup.sh` fail, you will leave the user in the `{path_to_tmp_dir}/kworkflow` which is not good. Return after you `cd` back to the original folder.
2. When you move this check, you will need to replace `?` by `ret` in the condition check.

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+    complain "Could not update kw"
+    return "$ret"
+  fi
+
+  cmd="cd ${original_dir}"
+  cmd_manager 'SILENT' "$cmd"
```
rodrigosiqueira, line 98: Use: cmd_manager "$flag" "$cmd"

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+  if [[ "$?" != 0 ]]; then
+    complain "Could not change back to original directory"
+    return 2 # ENOENT
+  fi
+
+  cmd="rm -rf ${path_to_tmp_dir}"
+  cmd_manager 'SILENT' "$cmd"
```
rodrigosiqueira, line 105: Use: cmd_manager "$flag" "$cmd"

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+  fi
+
+  cmd="cd ${original_dir}"
+  cmd_manager 'SILENT' "$cmd"
+  if [[ "$?" != 0 ]]; then
+    complain "Could not change back to original directory"
+    return 2 # ENOENT
+  fi
+
+  cmd="rm -rf ${path_to_tmp_dir}"
```
rodrigosiqueira, line 104: Just to be super safe with `-rf`, I would check `${path_to_tmp_dir}` to ensure it is not `/` or any other important folder.

src/self_update.sh
```diff
@@ -0,0 +1,160 @@
...
+  cmd_manager 'SILENT' "$cmd"
+  if [[ "$?" != 0 ]]; then
+    complain 'Could not remove temporary directory'
+    return 1 # EPERM
+  fi
+
+  return 0
+}
+
+function parse_self_update_options()
```
rodrigosiqueira, line 114: Add a test for this function as well.

rodrigosiqueira (changes requested):
Hi @davidbtadokoro ,
Thanks a lot for keeping improving this PR. We are almost in an excellent state. Please, see my new comments.

rodrigosiqueira:
Hey @davidbtadokoro , any update here?

davidbtadokoro:
Hi @rodrigosiqueira
Thanks for the feedback and I'm to know that we are almost at the point! Sorry for the delay. Got tied up in some affairs, but probably by monday night I'll address your comments.

src/self_update.sh
davidbtadokoro: Done it

src/self_update.sh
davidbtadokoro: Done it

davidbtadokoro:
Hi @rodrigosiqueira addressed your comments. Hope we are almost at that good point 😄!!

rodrigosiqueira:
Merged! Thanks a lot for the great work here!

PR #732, opened 2022-11-05
src: kw_env.sh: Change env dir

Kamorst:
Each env dir was being create inside .kw/. As an attempt to create a more organized dir structure, each env folder is created below .kw/envs/.

The .config is saved and restored with env. Don't know if this is a desirable feature.


rodrigosiqueira (changes requested):
Hi @Kamorst  ,

Thanks a lot for this change. This is the right direction for the env feature! Also, since this feature is not merged in the master branch, we have the freedom to make huge changes here. In other words, now is the right time for changes like that.

In general, your change looks good. However, when I tried it, I noticed some problems that I would like to ask you to address, even though it not directly related to your change.

Inside the kernel tree, if we have an old .kw config folder without the remote.config and we try to create a new env we have this error:

```
kw env --create XPTO
cp: cannot stat '/home/siqueira/Documents/Repositories/kernels/rpi/.kw/remote.config': No such file or directory
```

kw should create that config file if it does not have one yet.

Also, If we already have an env, we will hit this error:
```
kw env --create SOMETHING
cp: cannot stat '/home/siqueira/Documents/Repositories/kernels/amd-staging-drm-next/.kw/envs/X86-AMD-STAGING-DRM-NEXT/.config': No such file or directory
```
.config must be inside the env; otherwise, we will have compilation issues (from the make -h:   make O=dir [targets] Locate all output files in "dir", including .config)

src/kw_env.sh
```diff
@@ -66,9 +67,14 @@ function use_target_env()
...
 
     # Create symbolic link
-    ln --symbolic --force "${local_kw_configs}/${target_env}/${config}.config" "${local_kw_configs}/${config}.config"
+    ln --symbolic --force "${local_kw_configs}/${ENV_DIR}/${target_env}/${config}.config" "${local_kw_configs}/${config}.config"
   done
 
+  # Restore kernel .config from env
+  if [[ -e "${local_kw_configs}/${ENV_DIR}/${target_env}/.config" ]]; then
+    cp "${local_kw_configs}/${ENV_DIR}/${target_env}/.config" .config
+  fi
```
rodrigosiqueira, line 76: Actually, you don't want to copy `.config` inside `.kw` folder; you want it on `${KW_CACHE_DIR}/${target_env}`. Kw env will create a folder in the cache path for the new env, and all products of the kernel compilation will be isolated in the cache path. Also, since `.config` file is specific to each env, we must keep it inside the env.

src/kw_env.sh
```diff
@@ -16,6 +16,7 @@ declare -ga config_file_list=(
 )
 
 declare -gr ENV_CURRENT_FILE='env.current'
+declare -gr ENV_DIR='envs'
```
rodrigosiqueira, line 19: We will need to use this inside build.sh and deploy.sh, let's move it to the `kw` file.

src/kw_env.sh
```diff
@@ -97,31 +98,45 @@ function create_new_env()
...
   if [[ -f "${PWD}/.config" ]]; then
-    mv "${PWD}/.config" "${cache_build_path}/${env_name}"
+    mv "${PWD}/.config" "${cache_build_path}/${ENV_DIR}/${env_name}"
+  elif [[ -e /proc/config.gz ]]; then
+    zcat /proc/config.gz > "${cache_build_path}/${ENV_DIR}/${env_name}/.config"
+  elif [[ -e "/boot/config-$(uname -r)" ]]; then
+    cp "/boot/config-$(uname -r)" "${cache_build_path}/${ENV_DIR}/${env_name}/.config"
   elif [[ "$ret" == 0 ]]; then
-    cp "${cache_build_path}/${current_env_name}/.config" "${cache_build_path}/${env_name}"
+    cp "${cache_build_path}/${ENV_DIR}/${current_env_name}/.config" "${cache_build_path}/${ENV_DIR}/${env_name}/.config"
```
rodrigosiqueira, line 139: Try something like this:
```diff
diff --git a/src/kw_env.sh b/src/kw_env.sh
index 7fc4c3c..39c2303 100644
--- a/src/kw_env.sh
+++ b/src/kw_env.sh
@@ -127,16 +127,23 @@ function create_new_env()
 
   current_env_name=$(get_current_env_name)
   ret="$?"
+  # If we already have an env, we should copy the config file from it.
+  if [[ "$ret" == 0 ]]; then
+    cp "${cache_build_path}/${ENV_DIR}/${current_env_name}/.config" "${cache_build_path}/${ENV_DIR}/${env_name}/.config"
+    return
+  elif [[ -f "${PWD}/.config" ]]; then
+    cp "${PWD}/.config" "${cache_build_path}/${ENV_DIR}/${env_name}"
+    return
+  fi
 
-  # Create a default .config for env
-  if [[ -f "${PWD}/.config" ]]; then
-    mv "${PWD}/.config" "${cache_build_path}/${ENV_DIR}/${env_name}"
-  elif [[ -e /proc/config.gz ]]; then
+  warning "You don't have a config file, git it from default paths"
+  if [[ -e /proc/config.gz ]]; then
     zcat /proc/config.gz > "${cache_build_path}/${ENV_DIR}/${env_name}/.config"
   elif [[ -e "/boot/config-$(uname -r)" ]]; then
     cp "/boot/config-$(uname -r)" "${cache_build_path}/${ENV_DIR}/${env_name}/.config"
-  elif [[ "$ret" == 0 ]]; then
-    cp "${cache_build_path}/${ENV_DIR}/${current_env_name}/.config" "${cache_build_path}/${ENV_DIR}/${env_name}/.config"
+  else
+    warning 'kw was not able to find any valid config file for the new env'
+    return 22 # EINVAL
   fi
 }
```

rodrigosiqueira (changes requested):
Hi @Kamorst ,
Thanks a lot for updating this PR; I added some comments. Also, you will need to change other files that check the current env. For example:

```diff
diff --git a/src/build.sh b/src/build.sh
index 52d19c5..4f74704 100644
--- a/src/build.sh
+++ b/src/build.sh
@@ -67,7 +67,7 @@ function kernel_build()
 
   env_name=$(get_current_env_name)
   if [[ "$?" == 0 ]]; then
-    options_values['ENV_PATH_KBUILD_OUTPUT_FLAG']="${KW_CACHE_DIR}/${env_name}"
+    options_values['ENV_PATH_KBUILD_OUTPUT_FLAG']="${KW_CACHE_DIR}/${ENV_DIR}/${env_name}"
     output_kbuild_flag=" O=${options_values['ENV_PATH_KBUILD_OUTPUT_FLAG']}"
   fi
```

Please, check all files.

src/kw_env.sh
Kamorst: But if it's inside the `kw` how could the kw_env_test `include` it?

rodrigosiqueira:
@Kamorst could you check why this change failed?

rodrigosiqueira:
@Kamorst Could you rebase your branch against the latest unstable branch?

src/kw_env.sh
```diff
@@ -4,6 +4,9 @@ include "${KW_LIB_DIR}/kw_string.sh"
 
 declare -gA options_values
 
+# TODO: REMOVE THIS
+declare -gr ENV_DIR='envs'
```
rodrigosiqueira, line 8: Why not remove it now?

src/kw_env.sh
```diff
@@ -135,15 +159,17 @@ function list_env_available_envs()
...
 
   if [[ ! -d "$local_kw_configs" ]]; then
     complain 'It looks like that you did not setup kw in this repository.'
     complain 'For the first setup, take a look at: kw init --help'
     exit 22 # EINVAL
   fi
 
-  output=$(find "$local_kw_configs" -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | sort -d)
-  if [[ -z "$output" ]]; then
+  output=$(ls "${local_kw_configs}/${ENV_DIR}" 2> /dev/null)
```
rodrigosiqueira, line 170: Just to keep the consistency around the code, could you replace this `ls` with `find`?

src/kw_env.sh
```diff
@@ -135,15 +159,17 @@ function list_env_available_envs()
...
+  ret=$?
```
rodrigosiqueira, line 171: Use double-quotes.

src/kw_env.sh
```diff
@@ -135,15 +159,17 @@ function list_env_available_envs()
...
+  if [[ ret -ne 0 || -z $output ]]; then
```
rodrigosiqueira, line 172: I think you missed `$` before `ret`. Also, add double-quotes around `$output`.

rodrigosiqueira (changes requested):

rodrigosiqueira:
@Kamorst any update?

report_output
```diff
@@ -0,0 +1,93 @@
+# Report:
```
rodrigosiqueira, line 1: I think this file is not part of this PR.

src/kw_env.sh
```diff
@@ -52,7 +52,7 @@ function use_target_env()
   local local_kw_configs="${PWD}/.kw"
   local tmp_trash
 
-  if [[ ! -d "${local_kw_configs}/${target_env}" ]]; then
+  if [[ ! -d "${local_kw_configs}/${ENV_DIR}/${target_env}" ]]; then
```
rodrigosiqueira, line 55: Please, review the `kw_env.sh` file since Aquila added new functions. For example, the destroy and exit env.

rodrigosiqueira (changes requested):

rodrigosiqueira:
Hi @Kamorst , any update here?

Kamorst:
> Hi @Kamorst , any update here?

not yet, I'll take a look at it as soon as possible

rodrigosiqueira (changes requested):
Few notes after my tests:
* It looks like the exit-env does not work as expected, because the current env config file is not copied back to the .kw/ folder.
* When --destroy option is used, the current env file is not copied back to .kw/ folder.

Please, try multiple manual tests. E.g.:

1. Create multiple new envs: check .kw folder and others
2. Use new envs: check folder and config files
3. Exit envs: check files
4. Destroy envs: check files

Keep in mind that kw should be able to work as expected after exit and destroy.

rodrigosiqueira:
LGTM, could you just rebase this PR and update it?

Kamorst:
PR rebased.

rodrigosiqueira:
Merged! Thanks

PR #742, opened 2022-11-21
etc/init_templates: change default deploy target to remote for x86

melissawen:
vm deployment support was removed. change default deploy target on x86 template to match deployment options.

Signed-off-by: Melissa Wen <mwen@igalia.com>


rodrigosiqueira:
Merged! Thanks.

PR #743, opened 2022-11-21
plugins/utils: copy dtb files if in the remote package

melissawen:
Check if there are dtb files in the remote package and copy them to boot, accordingly.

Suggested-by: Rodrigo Siqueira <siqueirajordao@riseup.net>
Signed-off-by: Melissa Wen <mwen@igalia.com>

### DIFF
--- a/src/plugins/kernel_install/utils.sh
+++ b/src/plugins/kernel_install/utils.sh
@@ -641,6 +641,13 @@ function install_kernel()
   cmd="${sudo_cmd}cp ${KW_DEPLOY_TMP_FILE}/kw_pkg/${kernel_image_name} /boot/"
   cmd_manager "$flag" "$cmd"
 
+  # Copy dtbs to /boot if exist
+  dtb_files=$(find "${KW_DEPLOY_TMP_FILE}"/kw_pkg -type f -name "*.dtb")
+  if [[ -n "$dtb_files" ]]; then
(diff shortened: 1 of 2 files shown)


src/plugins/kernel_install/utils.sh
```diff
@@ -630,6 +630,12 @@ function install_kernel()
 
   install_modules "$target" "$flag"
 
+  # Copy dtbs if exist
+  if ls "${KW_DEPLOY_TMP_FILE}"/kw_pkg/*.dtb > /dev/null 2>&1; then
```
rodrigosiqueira, line 634: Avoid `ls` since this command does not have a standard behavior across distros. I recommend you to use `find`. Also, avoid bash commands inside the if condition.

rodrigosiqueira (changes requested):
Hi @melissawen ,
Thanks for this PR. Could you also highlight in the commit message that this commit fixes the deploy for the RPI? Also, add a test that covers these lines.

rodrigosiqueira:
@melissawen , any update?

rodrigosiqueira:
Hi @melissawen , any update?

src/plugins/kernel_install/utils.sh
melissawen: I just updated the commit message and the code to use `find`. I need more time to examine and create a test.

rodrigosiqueira:
Thanks a lot for this fix! Merged in the unstable branch.

PR #745, opened 2022-11-23
draft: add support to kw build --clean

aquilamacedo:
Adding kw build --clean option in such a way that it is integrated into the env.


src/build.sh
```diff
@@ -97,6 +100,10 @@ function kernel_build()
     llvm='LLVM=1 '
   fi
 
+  if [[ -n "$clean" ]]; then
+    build_clean "$output_kbuild_flag"
```
rodrigosiqueira, line 104: You can return at after call `build_clean`

src/build.sh
```diff
@@ -154,6 +161,21 @@ function kernel_build()
   return "$ret"
 }
 
+function build_clean()
```
rodrigosiqueira, line 164: Add code documentation.

src/build.sh
```diff
@@ -154,6 +161,21 @@ function kernel_build()
...
+{
+  local raw_options="$*"
```
rodrigosiqueira, line 166: I don't think you need this.

src/build.sh
```diff
@@ -154,6 +161,21 @@ function kernel_build()
...
+  local output_kbuild_flag=''
+
+  env_name=$(get_current_env_name)
+
```
rodrigosiqueira, line 170: Drop this extra line.

src/build.sh
```diff
@@ -154,6 +161,21 @@ function kernel_build()
...
+  if [[ "$?" == 0 ]]; then
+    options_values['ENV_PATH_KBUILD_OUTPUT_FLAG']="${KW_CACHE_DIR}/${env_name}"
```
rodrigosiqueira, line 172: I guess you don't need this line, you can set the path directly on `output_kbuild_flag`

src/build.sh
```diff
@@ -200,8 +222,8 @@ function load_build_config()
 
 function parse_build_options()
 {
-  local long_options='help,info,menu,doc,ccache,cpu-scaling:,warnings::,save-log-to:,llvm'
-  local short_options='h,i,n,d,c:,w::,s:'
+  local long_options='help,info,menu,doc,ccache,cpu-scaling:,warnings::,save-log-to:,llvm,clean'
+  local short_options='h,i,n,d,c:,w::,s:,cl'
```
rodrigosiqueira, line 226: I think `c` makes more sense here. For the `--cpu-scaling`, we can use `S` instead of `c`.

src/build.sh
```diff
@@ -226,6 +248,8 @@ function parse_build_options()
   options_values['WARNINGS']="${build_config[warning_level]}"
   options_values['LOG_PATH']="${build_config[log_path]:-${configurations[log_path]}}"
   options_values['USE_LLVM_TOOLCHAIN']="${build_config[use_llvm]:-${configurations[use_llvm]}}"
+  options_values['CLEAN']=''
+  options_values['CLEAN_PATH']='.'
```
rodrigosiqueira, line 252: I did not get why do we need this.

rodrigosiqueira (changes requested):
Hi Aquila,
Thanks for your PR; you can see some of my comments. Please, expand your commit message, add man doc, and unit test.

src/build.sh
aquilamacedo: You are right, I was thinking of implementing something like `build --clean folder=driver/gpu/drm/amd/display` but I'll leave it for a future PR.

src/build.sh
rodrigosiqueira, line 104: You probably want:
1. Check if `build_clean` worked as expected or not; in case of error show an error message;
2. Return after this function.

src/build.sh
```diff
@@ -154,6 +161,16 @@ function kernel_build()
   return "$ret"
 }
 
+# This function adds the build clean option
+# integrated to env in order to clean the kernel.
+function build_clean()
+{
+  local flag="$1"
```
rodrigosiqueira, line 168: Search for the use of `flag` in the code; we usually use it for testing and displaying command messages. You probably want this parameter here, but for another reason. Maybe a better name is `output_kbuild_flag` or `clean_flags`.

src/build.sh
```diff
@@ -154,6 +161,16 @@ function kernel_build()
...
+
+  command="make clean${flag}"
+  cmd_manager "$command"
```
rodrigosiqueira, line 171: Search for the use of `cmd_manager` in the code, and you will notice that this function expects a `flag` as its first parameter.

src/build.sh
```diff
@@ -200,8 +217,8 @@ function load_build_config()
 
 function parse_build_options()
 {
-  local long_options='help,info,menu,doc,ccache,cpu-scaling:,warnings::,save-log-to:,llvm'
-  local short_options='h,i,n,d,c:,w::,s:'
+  local long_options='help,info,menu,doc,ccache,cpu-scaling:,warnings::,save-log-to:,llvm,clean'
+  local short_options='h,i,n,d,S:,w::,s:,c'
```
rodrigosiqueira, line 221: Change `c` to `S` should be in a separate commit before this one. You also need to update the man page.

src/build.sh
```diff
@@ -317,11 +339,12 @@ function build_help()
...
     '  build (-i | --info) - Display build information' \
     '  build (-d | --doc) - Build kernel documentation' \
-    '  build (-c | --cpu-scaling) <percentage> - Scale CPU usage by factor' \
+    '  build (-S | --cpu-scaling) <percentage> - Scale CPU usage by factor' \
     '  build (--ccache) - Enable use of ccache' \
     '  build (-w | --warnings) [warning_levels] - Enable warnings' \
     '  build (-s | --save-log-to) <path> - Save compilation log to path' \
-    '  build (--llvm) - Enable use of the LLVM toolchain'
+    '  build (--llvm) - Enable use of the LLVM toolchain' \
+    '  build (-c | --clean) - Clean option integrated into env'
```
rodrigosiqueira, line 347: You also have to update the man page.

rodrigosiqueira (changes requested):
Hi @aquilamacedo ,
See my comments, and please add unit tests.

rodrigosiqueira:
Hi @aquilamacedo , please, address my previous comments and also fix the check failures (rebase your change against the latest unstable branch will fix the doc build error).

aquilamacedo:
Hi @rodrigosiqueira. I just finished making the requested changes. I didn't understand why it gave an error

rodrigosiqueira:
> Hi @rodrigosiqueira. I just finished making the requested changes. I didn't understand why it gave an error
Hi @aquilamacedo ,
If you click on the 'Details' link in the failure, you can see why it failed.

documentation/man/features/build.rst
```diff
@@ -70,6 +71,10 @@ OPTIONS
   compilation/linking tasks. You can enable it by default via `use_llvm` option
   in the `build.config` file.
 
+-c, \--clean:
+  The clean option removes most files generated by the kernel build system but
+  keeps the kernel configuration.
```
rodrigosiqueira, line 76: Maybe this:

The clean option removes files generated by the kernel build system but keeps the kernel configuration. This option considers whether the user is using or not kw env.

src/build.sh
```diff
@@ -97,6 +100,13 @@ function kernel_build()
     llvm='LLVM=1 '
   fi
 
+  if [[ -n "$clean" ]]; then
+    build_clean "$output_kbuild_flag"
+
+    ret="$?"
+    return "$ret"
```
rodrigosiqueira, line 107: You don't need to save `$?` in the `ret` variable, just return `$?` directly.

src/build.sh
```diff
@@ -154,6 +164,17 @@ function kernel_build()
   return "$ret"
 }
 
+# This function adds the build clean option
+# integrated to env in order to clean the kernel.
```
rodrigosiqueira, line 168: How about something like this?

This function runs the 'make clean' command under the hood, with the advantage of checking if the user is using an env or not. In other words, it integrates env with the clean option.

Also, document the output_kbuild_flag parameter and flag.

src/build.sh
```diff
@@ -154,6 +164,17 @@ function kernel_build()
...
+function build_clean()
+{
+  local output_kbuild_flag="$1"
+  local flag
```
rodrigosiqueira, line 172: You should receive flag as a parameter; search for this flag utilization in the code. After you understand that, you will notice how easy it is to test this function.

src/build.sh
rodrigosiqueira: Did you rebase this PR against the latest unstable branch?

src/build.sh
```diff
@@ -268,6 +290,10 @@ function parse_build_options()
         options_values['USE_LLVM_TOOLCHAIN']=1
         shift
         ;;
+      --clean | -c)
+	      options_values['CLEAN']=1
```
rodrigosiqueira, line 294: See shfmt failure.

tests/build_test.sh
```diff
@@ -238,6 +238,18 @@ function test_kernel_build_menu_cross_compilation_flags()
   assertEquals "($LINENO)" "$expected_result" "$output"
 }
 
+function test_kernel_build_clean() {
+  local expected_result
+  local output
+
+  build_config=()
+  parse_configuration "$SAMPLES_DIR/build_no_log.config" build_config
```
rodrigosiqueira, line 246: Please, refer to: https://kworkflow.org/content/codingstyle.html#string-concatenation

tests/build_test.sh
```diff
@@ -238,6 +238,18 @@ function test_kernel_build_menu_cross_compilation_flags()
   assertEquals "($LINENO)" "$expected_result" "$output"
 }
 
+function test_kernel_build_clean() {
```
rodrigosiqueira, line 241: See shfmt failure, and read: https://kworkflow.org/content/codingstyle.html#functions

tests/build_test.sh
```diff
@@ -238,6 +238,18 @@ function test_kernel_build_menu_cross_compilation_flags()
...
+  local expected_result
+  local output
+
+  build_config=()
+  parse_configuration "$SAMPLES_DIR/build_no_log.config" build_config
+
+  output=$(kernel_build 'TEST_MODE' --clean 2> /dev/null)
+  expected_result="make clean"
```
rodrigosiqueira, line 249: Use single quotes.

tests/build_test.sh
```diff
@@ -238,6 +238,18 @@ function test_kernel_build_menu_cross_compilation_flags()
...
 }
 
+function test_kernel_build_clean() {
+  local expected_result
+  local output
+
+  build_config=()
+  parse_configuration "$SAMPLES_DIR/build_no_log.config" build_config
+
+  output=$(kernel_build 'TEST_MODE' --clean 2> /dev/null)
```
rodrigosiqueira, line 248: Why do we need `2> /dev/null`?

tests/build_test.sh
```diff
@@ -877,4 +898,20 @@ function test_kernel_build_menu_inside_env()
...
+function test_kernel_build_clean_inside_env()
+{
+  local expected_result
+  local output
+  local env_output="${KW_CACHE_DIR}/fake_env"
+  build_config=()
+
+  mk_fake_kw_env
+
+  parse_configuration "$SAMPLES_DIR/build_no_log.config" build_config
```
rodrigosiqueira, line 910: Please, refer to: https://kworkflow.org/content/codingstyle.html#string-concatenation

rodrigosiqueira (changes requested):

tests/build_test.sh
aquilamacedo: in the `expected_result` variable or `parse_configuration`?

tests/build_test.sh
aquilamacedo: Because without the `2> /dev/null` it keeps printing this error when I run `./run_tests.sh`

src/build.sh
aquilamacedo: I think it's a good idea.

As for the documentation part of the `output_kbuild_flag` parameter, I think it will not be necessary, as I will remove this variable from the function, since it is global. only the flag variable will remain for the `cmd_manager` function

src/build.sh
aquilamacedo: You are right, I already made the required changes. :)

rodrigosiqueira:
Merged into the unstable branch. Thanks

PR #750, opened 2022-12-11
src: build: changing the short option of the --cpu-scaling parameter.

aquilamacedo:
This commit changes the short option of the `build --cpu-scaling` parameter 
from `c` to `S`, as `c` will be used in the future for the `build --clean` parameter.


rodrigosiqueira:
Nice change! Merged into the unstable branch.
Thanks

Issue #728, opened 2022-11-01
Remote does not fail when missing parameters


rodrigosiqueira:
**Description:**

kw remote is lacking a proper validation of its parameter.

**How to Reproduce:**

The idea is to add a new remote.

1. Make sure that you have a .kw folder (use [kw init](https://kworkflow.org/man/features/init.html))
2. Try: `kw remote xpto-system root@la-lnx-16`
3. It does not fail

**Expected behavior:**

Since the above command does not have the `add|remove|rename`, it should fail.

**System information:**

- OS: Debian Sid
- kw Version: unstable

karinesobrinho:
UFABC group will provide a solution.

Issue #727, opened 2022-11-01
Remote silent fails to add some specific strings


rodrigosiqueira:
**Description:**

kw has the [remote option](https://kworkflow.org/man/features/remote.html), which is used to set up any target machine connected via the network. Under the hood, it setup a feature that will create a [ssh_config](https://man7.org/linux/man-pages/man5/ssh_config.5.html) file, which kw can reference by using the `-F` option.

This is a new feature, and as a result, it has some bugs. One of them is related to how it parses some of the elements.

**How to Reproduce:**

1. Make sure that you have a `.kw` folder (use [kw init](https://kworkflow.org/man/features/init.html))
2. Try:
```bash
kw remote add something99 root@igt-abc99-0.lala.xpto
```
3. Check the `.kw/remote.config` and notice that `something99` is not inserted.

**Expected behavior:**

The above example shout add:

```
[..]
Host something99
  Hostname igt-abc99-0.lala.xpto
  Port 22
  User root
```

**System information:**

 - OS: Debian Sid
 - kw Version: unstable

**Extra notes**

If you work on this ticket, it is worth checking many other types of strings and adding a proper test for them.

WLAraujo:
Nós da UFABC vamos trabalhar nessa

davidbtadokoro:
@rodrigosiqueira could not reproduce it.
Branch unstable, commit 41341d9

aquilamacedo:
@rodrigosiqueira I tried to reproduce this bug but wasn't able to replicate it.
Branch unstable, commit [cca88c2](https://github.com/kworkflow/kworkflow/commit/cca88c224b59cc8bbca5bc85c8e9dc07c43ca8df)

PR #753, opened 2022-12-14
Issues 727 and 728 - Fix of some remote functionalities 

WLAraujo:
Pull request with the implementation of the following issues:
- Issue 728 - https://github.com/kworkflow/kworkflow/issues/728 - Remote does not fail when missing parameters
- Issue 727 - https://github.com/kworkflow/kworkflow/issues/727 - Remote silent fails to add some specific strings

### DIFF
--- a/src/kw_remote.sh
+++ b/src/kw_remote.sh
@@ -11,6 +11,16 @@ function remote_main()
     exit 0
   fi
 
+  if [[ "$1" =~ -l|--list ]]; then
+    if [[ -f "local_remote_config_file" ]]; then
+      cat "$local_remote_config_file"
+      exit 0
+    else
+      echo "There are no remotes"
+      exit 0
+    fi
+  fi
+
   parse_remote_options "$@"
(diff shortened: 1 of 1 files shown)


rodrigosiqueira:
@WLAraujo @GabrielPiVaz 
Thanks for your PR.

Please, spend some time writing a good commit message; this is the third time you have submitted a series without a good commit description. I know it may sound like an unimportant thing to do; however, keep in mind that I have no context of what you guys discussed or want to achieve. We are working remotely and in no-related teams. If you write a good commit message, it will be an invitation to make your reviewer more comfortable to see your change. Additionally, if someone else in the future tries to understand why you make this change, they can use git blame and see your commit message after identifying your change. I guess I already recommend the below link as a good reference, but I want to recommend it again:

https://cbea.ms/git-commit/

Thanks

rodrigosiqueira:
I closed this PR due to the lack of reply. If you have a better version of this change, feel free to re-open this PR.

Issue #746, opened 2022-12-03
Add an option not to use kw env if a user is in an env


rodrigosiqueira:
**Description:**

kw env feature is used for isolating different kernel builds inside envs; however, kw also supports in-tree kernel build. We need a mechanism to allow users to "exit" the env feature. We can have something like this:

```
$kw env --no-env
```

This command should:
1. Copy the current `env` files to `.kw/` folder. Remember that envs are a simple symbolic link to another path; if we don't have any env, we just need to drop those symbolic links in favor of the actual file.
2. Notify users about the consequence and ask them if they really want to proceed.

aquilamacedo:
i'm working on it.

aquilamacedo:
Hi @rodrigosiqueira. I have a doubt about this issue.

![image](https://user-images.githubusercontent.com/43859386/206881750-354e1742-7858-49aa-8457-0e0ae073761f.png)

There is a code snippet in the use_target_env function that moves the default .config files to a temporary path using `mktemp` (line 72 of the code).

I didn't understand the reason to send the .config files from the default environment to a temporary path. I believe that a temporary path is not the best option to store these files, since `/tmp/` writes in volatile memory.

So I thought of creating a directory inside the .kw, something like (_.kw/default/_) to move these .config files from the default environment to this directory instead of a temporary path, it would be easier to retrieve these files to use in `- -no-env` parameter.

What do you think about this?

rodrigosiqueira:
Hi @aquilamacedo ,

If we are not using any env, the `.kw` folder will look like this:
```
build.config  deploy.config  kworkflow.config  mail.config  notification.config  remote.config
```
However, when we create an env, we will replace all of those 'real' config files with a symbolic link pointing to `./kw/ENV_NAME/`. Notice from the comments in lines 68-69 that we don't want those 'real' files there if the user is using envs; for this reason, I moved them to a tmp folder. It was intentional to move to a temporary folder since this is now garbage.

For this specific issue, you can check:
1. If the user is using some specific env and doesn't want to use it anymore, just eliminate all the symbolic links and copy the current env config files into `.kw/` folder. You may also want to ask the user if they want to use the current env as the default for the case without env.
2. If the user is not using any env, you don't need to do anything.

rodrigosiqueira:
@aquilamacedo any update here? I want to complete the two env issues open to update the master branch.

PR #758, opened 2022-12-30
src: kw_env: Adding support to `kw env --no-env`

aquilamacedo:
This commit creates a mechanism allowing the user to "exit" the env resource using the `--no-env` parameter.

Closes: #746

### DIFF
--- a/documentation/man/features/env.rst
+++ b/documentation/man/features/env.rst
@@ -9,7 +9,7 @@ SYNOPSIS
 | *kw* (*env*) [(-c | \--create) <NAME>]
 | *kw* (*env*) [(-u | \--use) <NAME>]
 | *kw* (*env*) [(-l | \--list)]
-
+| *kw* (*env*) [(-e | \--exit-env)]
 
 DESCRIPTION
 ===========
@@ -48,6 +48,11 @@ OPTIONS
 -l, \--list:
   It shows all envs created via `\--create` option.
 
+-n, \--exit-env:
+  Allow users to "exit" the env feature. If the user is using a specific env
+  and doesn't want to use it anymore, the `--exit-env` option will remove all
+  symbolic links and copy the current env's configuration files to the .kw.
+
 EXAMPLES
 ========
 If you want to create a new env, you can use::
@@ -63,3 +68,7 @@ If you want to create a new env, you can use::
 If you want to list the available envs::
 
   kw env --list
+
+If you want to exit the env feature::
+
+  kw env --exit-env
--- a/src/kw_env.sh
+++ b/src/kw_env.sh
@@ -39,6 +39,11 @@ function env_main()
     list_env_available_envs
     return "$?"
   fi
+
+  if [[ -n "${options_values['EXIT_ENV']}" ]]; then
+    exit_env
+    return "$?"
+  fi
 }
 
 # When we switch between different kw envs we just change the symbolic links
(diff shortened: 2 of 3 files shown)


aquilamacedo:
Hi @rodrigosiqueira. I just created this draft issue #746. I would like to know if the implementation is according to your specifications.

documentation/man/features/env.rst
```diff
@@ -48,6 +48,9 @@ OPTIONS
 -l, \--list:
   It shows all envs created via `\--create` option.
 
+-n, \--no-env:
+  Allow users to "exit" the env feature.
```
rodrigosiqueira, line 52: After reading this, I think that `--exit-env` is a better name.

rodrigosiqueira: Also, explain what exit an env means. Explain the behavior as we discussed in the https://github.com/kworkflow/kworkflow/issues/746 .

documentation/man/features/env.rst
```diff
@@ -63,3 +66,7 @@ If you want to create a new env, you can use::
 If you want to list the available envs::
 
   kw env --list
+
+If you want to exit the env feature:::
```
rodrigosiqueira, line 70: Delete one of the `:`

src/kw_env.sh
```diff
@@ -72,6 +77,25 @@ function use_target_env()
   printf '%s\n' "$target_env" > "${local_kw_configs}/${ENV_CURRENT_FILE}"
 }
 
+function use_no_env()
```
rodrigosiqueira, line 80: add function documentation, and let's call this function `exit_env`.

src/kw_env.sh
```diff
@@ -72,6 +77,25 @@ function use_target_env()
...
+{
+  local current_env
+  local local_kw_configs="${PWD}/.kw"
+
+  current_env=$(< "${local_kw_configs}/${ENV_CURRENT_FILE}")
+
+  if [[ $(ask_yN 'Do you want to make this env the default?') =~ '1' ]]; then
```
rodrigosiqueira, line 87: I think you can add a warning followed by the question, something like:
```
warning 'You are about to leave the env setup, and ENV_NAME config files will be used as a default.'
  if [[ $(ask_yN 'Do you really want to proceed?') =~ '1' ]]; then
```

src/kw_env.sh
```diff
@@ -72,6 +77,25 @@ function use_target_env()
...
+    for config in "${config_file_list[@]}"; do
+      if [[ -L "${local_kw_configs}/${config}.config" ]]; then
```
rodrigosiqueira, line 89: Add a comment before this `if` that explains that you are removing the link but not the original file.

src/kw_env.sh
```diff
@@ -72,6 +77,25 @@ function use_target_env()
...
+        rm "${local_kw_configs}/${config}.config"
+      fi
+      cp "${local_kw_configs}/${current_env}/${config}.config" "${local_kw_configs}"
+    done
+    printf '%s\n' 'NO ENV' > "${local_kw_configs}/${ENV_CURRENT_FILE}"
```
rodrigosiqueira, line 94: I think you can remove the current env file.

src/kw_env.sh
```diff
@@ -72,6 +77,25 @@ function use_target_env()
...
+    success "You left the environment feature."
```
rodrigosiqueira, line 95: single quotes.

rodrigosiqueira (changes requested):

aquilamacedo:
Hi @rodrigosiqueira. I just finished making the requested changes :)

rodrigosiqueira:
Thanks a lot for this PR. I already merged it with the following changes:

```bash
diff --git a/src/bash_autocomplete.sh b/src/bash_autocomplete.sh
index 64941a3..e5a01da 100644
--- a/src/bash_autocomplete.sh
+++ b/src/bash_autocomplete.sh
@@ -70,7 +70,7 @@ function _kw_autocomplete()
   kw_options['drm']='--remote --local --gui-on --gui-off --load-module
                      --unload-module --conn-available --modes --help'
 
-  kw_options['env']='--create --list --use'
+  kw_options['env']='--create --list --use --exit-env'
 
   kw_options['self-update']='--unstable --help'
   kw_options['u']="${kw_options['self-update']}"
diff --git a/src/kw_env.sh b/src/kw_env.sh
index ef42135..4163aa6 100644
--- a/src/kw_env.sh
+++ b/src/kw_env.sh
@@ -86,7 +86,7 @@ function exit_env()
 
   current_env=$(< "${local_kw_configs}/${ENV_CURRENT_FILE}")
 
-  warning 'You are about to leave the env setup, and ENV_NAME config files will be used as a default.'
+  warning "You are about to leave the env setup, and ${current_env} config files will be used as a default."
   if [[ $(ask_yN 'Do you really want to proceed?') =~ '1' ]]; then
     for config in "${config_file_list[@]}"; do
       # All symbolic links will be removed, and the current env configuration files
```

Issue #759, opened 2022-12-30
Add support to kw build --distclean


aquilamacedo:
**Description:**

Now that we have support for envs within kw and we have the build clean option built into the env, we need to add the distclean option built into the env, something like:
```
kw build --distclean
```

PR #761, opened 2023-01-06
src: build: add support to `kw build --distclean`

aquilamacedo:
This commit adds the `build --distclean` option.

The distclean option will delete everything that make clean does and remove the configuration.

Closes: #759


aquilamacedo:
Hi @rodrigosiqueira. I've just created this draft of issue #759. I'd like to know what else needs to be added to this issue and if it lives up to your specs.

src/build.sh
```diff
@@ -227,8 +245,8 @@ function load_build_config()
 
 function parse_build_options()
 {
-  local long_options='help,info,menu,doc,ccache,cpu-scaling:,warnings::,save-log-to:,llvm,clean'
-  local short_options='h,i,n,d,S:,w::,s:,c'
+  local long_options='help,info,menu,doc,ccache,cpu-scaling:,warnings::,save-log-to:,llvm,clean,distclean'
+  local short_options='h,i,n,d,S:,w::,s:,c,dc'
```
rodrigosiqueira, line 249: Let's use `C` instead of `dc`.

src/build.sh
```diff
@@ -181,6 +187,18 @@ function build_clean()
...
 
+# This function runs the 'make distclean' command under the hood, with
+# the advantage of checking if the user is using an env or not.
+# In other words, it integrates env with the distclean option.
+function build_distclean()
+{
+  local flag="$1"
+  local cmd
+
+  cmd="make distclean${output_kbuild_flag}"
```
rodrigosiqueira, line 198: Notice that `output_kbuild_flag` is not global; in this sense, you must receive this as a parameter or call `get_current_env_name`.
Btw... this code should not work inside env, did you check it?

src/build.sh
```diff
@@ -354,7 +377,8 @@ function build_help()
     '  build (-w | --warnings) [warning_levels] - Enable warnings' \
     '  build (-s | --save-log-to) <path> - Save compilation log to path' \
     '  build (--llvm) - Enable use of the LLVM toolchain' \
-    '  build (-c | --clean) - Clean option integrated into env'
+    '  build (-c | --clean) - Clean option integrated into env' \
+    '  build (-C | --distclean) - Distclean option integrated into env'
```
rodrigosiqueira, line 381: New options should be added to `src/bash_autocomplete.sh`.

rodrigosiqueira (changes requested):

src/build.sh
aquilamacedo: I don't understand very well, isn't this implementation supposed to be integrated into the env?  

Actually by my tests it seems that it is working inside the env.

rodrigosiqueira: Notice that we never declared `output_kbuild_flag` as global (see `declare -g` at the beginning of this file). If your test passes, it may be because of another test setup. Something like this:

1. Sub-test X: Setup env
2. Sub-test X+n: This is your test, and it is passing because Sub-test X created an env, and we never cleaned it.
(this is a bug in the test since each unit test should be independent)

Ensure that the `tearDown` function is in the unit test and that it cleans everything created by the previous test, and try again.

rodrigosiqueira:
Hi @aquilamacedo , any update here?

aquilamacedo:
> Hi @aquilamacedo , any update here?

Hi @rodrigosiqueira, not yet but i'm almost done

documentation/man/features/build.rst
```diff
@@ -76,6 +77,11 @@ OPTIONS
   keeps the kernel configuration. This option considers whether the user
   is using or not kw env.
 
+-C, \--distclean:
+  The distclean option will delete everything that make clean does and
+  remove the configuration. Like `--clean` This option also considers
+  whether or not the user is using kw env.
```
rodrigosiqueira, line 83: To make this description more valuable, make it clear that under the wood, we are calling make distclean. Also, explain that this command will reset the kernel tree to its default; in other words, it will remove any configuration file and script output.

src/build.sh
rodrigosiqueira: Please, address my previous comment.

The solution should be to use the below code inside this function to get the `output_kbuild_flag`.

```bash
  env_name=$(get_current_env_name)
  if [[ "$?" == 0 ]]; then
    options_values['ENV_PATH_KBUILD_OUTPUT_FLAG']="${KW_CACHE_DIR}/${ENV_DIR}/${env_name}"
    output_kbuild_flag=" O=${options_values['ENV_PATH_KBUILD_OUTPUT_FLAG']}"
  fi
```

rodrigosiqueira (changes requested):
Hi @aquilamacedo ,

I added some other comments; please take some time to read and improve your PR.

rodrigosiqueira:
Very nice change!

PR #764, opened 2023-01-12
src: plugins: kernel_install: utils: Copy config file

rodrigosiqueira:
Recently we updated kw to use a package system, but we forgot to copy the .config file to the /boot folder. This was a perfect storm to hit the below error:

root@debian:~# update-initramfs -c -k [KERNEL-NAME] update-initramfs: Generating /boot/initrd.img-[KERNEL-NAME] grep: /boot/config-vmlinuz-[KERNEL-NAME]: No such file or directory W: zstd compression (CONFIG_RD_ZSTD) not supported by kernel, using gzip grep: /boot/config-vmlinuz-[KERNEL-NAME]: No such file or directory E: gzip compression (CONFIG_RD_GZIP) not supported by kernel update-initramfs: failed for /boot/initrd.img-[KERNEL-NAME] with 1.

Notice that we can only see it in a Debian-based distro that uses the latest initramfs-tool. This is a recent change, as you can see here:

https://salsa.debian.org/kernel-team/initramfs-tools/-/merge_requests/37/diffs#e3b45015ecbf200e358bd3915bb51b4b9d044d5e_188_190

This commit address this issue by ensuring the copy of the .config file.

### DIFF
--- a/src/plugins/kernel_install/utils.sh
+++ b/src/plugins/kernel_install/utils.sh
@@ -637,6 +637,10 @@ function install_kernel()
     cmd_manager "$flag" "$cmd"
   fi
 
+  # Copy kernel config
+  cmd="${sudo_cmd}cp ${KW_DEPLOY_TMP_FILE}/kw_pkg/config-${name} /boot/"
+  cmd_manager "$flag" "$cmd"
+
   # Update kernel image in the /boot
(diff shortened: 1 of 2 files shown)


melissawen:
LGTM. I've already tested this change locally. KW tests seem good too.

Reviewed-by: Melissa Wen <melissa.srw@gmail.com>

PR #767, opened 2023-02-02
tests: report_test: Fix terminal and file outputs from test_save_data_to()

davidbtadokoro:
The test_save_data_to() test function generated a terminal output during its execution and a file outside the Shunit temporary dir, which resulted in a hanging file named 'report_output' inside the repository. The former issue was due to a call to the save_data_to() function without handling its output and the latter was because the same function was being called passing the current directory as an argument (as this test suite does not have a set up to change to the Shunit tmp dir this resulted in the file being generated in the repository root).

The solution for the first was to redirect all function output to /dev/null and for the second was to pass the Shunit tmp dir as argument (although, as mentioned above, the best approach would've been to cd in and out of this dir with set up and teardown functions).

Considered making set up and teardown functions, but, as this commit aims only to fix the improper terminal and file outputs, those changes are beyond the scope.

### DIFF
--- a/tests/report_test.sh
+++ b/tests/report_test.sh
@@ -370,7 +370,7 @@ function test_save_data_to()
   local expected
 
   grouping_day_data '2020/04/04'
-  save_data_to "$SHUNIT_TMPDIR/test"
+  save_data_to "$SHUNIT_TMPDIR/test" > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


davidbtadokoro:
As context, these are the prints of the terminal and file outputs, respectively:

Screen Shot 2023-02-02 at 19 37 48

Screen Shot 2023-02-02 at 19 38 00

rodrigosiqueira:
Very nice PR!

Issue #734, opened 2022-11-13
Make kw deploy work outside the kernel tree


rodrigosiqueira:
Some deploy commands should work outside the kernel tree, and we need to enhance this behavior. The following commands should work in any folder:
* kw deploy --list
* kw deploy --uninstall

PR #768, opened 2023-02-03
Allow some `kw deploy` commands to be run outside kernel tree

davidbtadokoro:
As pointed in #734, some `kw deploy` commands like `--list` and `--uninstall` can't be run outside a kernel tree when they should. This PR implements this.

Also, this PR contains commits fixing issues related with ShellCheck failures that I came across when implementing the above. The issues where:
- An unreachable command in  `deploy.sh`  ([SC2317)](https://www.shellcheck.net/wiki/SC2317))
- Multiple individual redirects in `deploy.sh`  ([SC2129](https://www.shellcheck.net/wiki/SC2129))
- An unreachable command incorrectly pointed out by ShellCheck in `deploy_test.sh`  ([SC2129](https://www.shellcheck.net/wiki/SC2129))

### DIFF
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -1437,8 +1437,8 @@ function parse_deploy_options()
         ;;
       *)
         options_values['ERROR']="Unrecognized argument: $1"
-        return 22 # EINVAL
         shift
+        return 22 # EINVAL
         ;;
     esac
   done
--- a/src/deploy.sh
+++ b/src/deploy.sh
@@ -96,7 +96,7 @@ function deploy_main()
     exit 22 # EINVAL
   fi
 
-  if [[ -z "${options_values['FROM_PACKAGE']}" ]]; then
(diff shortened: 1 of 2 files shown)


src/deploy.sh
```diff
@@ -96,7 +96,8 @@ function deploy_main()
     exit 22 # EINVAL
   fi
 
-  if [[ -z "${options_values['FROM_PACKAGE']}" ]]; then
+  if [[ -z "${options_values['FROM_PACKAGE']}" &&
```
rodrigosiqueira, line 99: We can use `CAN_RUN_OUTSIDE_KERNEL_TREE` to replace `FROM_PACKAGE` as well.

davidbtadokoro: Tested and done 👍

rodrigosiqueira:
Very nice change! Tested, and merged in the unstable branch.

Thanks a lot!

PR #770, opened 2023-02-05
src: kw_remote: Fix not failing when missing valid options

### LINKED ISSUE #728: Remote does not fail when missing parameters

davidbtadokoro:
Before this commit, kw remote did not validate if the command had valid option, i.e., calling kw remote (without option/parameters) or calling kw remote <name> <remote> (without valid option) did not fail, although it did not seems to have any side effect.

Now we validate if any of the options add, remove, rename, list or set default remote were passed at the end of parse_remote_options(). If not, the command fails.

Chose to validate at the end of parse_remote_options() as 1) it seems the responsability of this function and 2) it enables to make a simple unit test for this case.

Closes #728

### DIFF
--- a/src/kw_remote.sh
+++ b/src/kw_remote.sh
@@ -332,6 +332,12 @@ function parse_remote_options()
   elif [[ "${options_values['DEFAULT_REMOTE']}" == 1 ]]; then
     options_values['ERROR']='Expected a string values after --set-default='
     return 22
+  elif [[ -z "${options_values['ADD']}" && -z "${options_values['REMOVE']}" &&
+    -z "${options_values['RENAME']}" && -z "${options_values['LIST']}" &&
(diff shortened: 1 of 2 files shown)


rodrigosiqueira:
Very nice improvement! Thanks a lot.
PR merged in the unstable branch.

PR #771, opened 2023-02-05
src: kw_remote: Fix remove remote that is prefix of other remote

davidbtadokoro:
If there were two remotes, one with a name that is a prefix of the other (like "kw" and "kworkflow"), and we tried to remove the first one, the second one would have its first line removed from the remote.config. This resulted in this second remote being virtually also removed, although that was not the intention and the user was not warned.

The cause of the bug was a missing $ at the end of the pattern used by sed to remove the first line of the target remote in the remote.config (the 'Host <name>' line). This caused sed to not get only the exact match of the line and remove lines prefixed by the target one as well. As expected, the fix was just the inclusion of this missing $.


davidbtadokoro:
As context, below is a print of the reproduction of the bug

kw-remote_remove-remote-that-is-prefix-of-another

rodrigosiqueira:
Nice catch!
Change merged in the unstable branch.

Issue #501, opened 2021-10-27
Add autocompletion on ZSH


eshiraishi:
The Issue #458 suggested increasing the autocomplete feature for all features, however the way it is designed only supports bash, since it depends on some features exclusive to it (`compgen` and `complete` more specifically). To support other shells (like ZSH) another solution is needed.

PR #773, opened 2023-02-10
Add support for native Zsh completions

davidbtadokoro:
This commit adds a proof of concept of native kw completions for ZSH. The ZSH completions were adapted from the Bash completions using bashcompinit, but, as stated in issue #501, the adaptation (namely because of the mapfile command) was broken.

This initial implementation just adds completions for the 'kw build' command/module, but can be scaled to all other kw commands. The completion for the deploy module was made as a stub, just to have two top level kw commands for the first completion.

Keeping in mind that broken completions can degrade the user experience, it is worth nothing that this approach has some caveats:
- Althought the implementation facilitates the definition of rules for mutually exclusive options, demand of arguments for some options, and much more, there is no semantic checks (i.e. if we mix the option names or make typos, these errors are only caught at runtime);
- Could not think of a way to create automated tests;
- The native implementation requires some naming conventions like '_kw' for the name of the file holding the completion functions (besided not being a descriptive name it doesn't fit with the project);
- For it to work we need to append the fpath variable (functions path for the ZSH).

NOTE: This commit should be squashed in the future.

WIP #501


davidbtadokoro:
Below is a demo of the completions

kw-zsh-autocomplete

davidbtadokoro:
Also, these are the references used to make this initial implementation:
https://zsh.sourceforge.io/Guide/zshguide06.html
https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Completion-Functions
https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#writing-your-own-completion-functions

setup.sh
```diff
@@ -383,6 +381,15 @@ function append_bashcompletion()
   safe_append "source ${libdir}/${BASH_AUTOCOMPLETE}.sh" "${HOME}/${shellrc}"
 }
 
+function append_zshcompletion()
+{
+  local shellrc="$1"
```
rodrigosiqueira, line 386: Can we just hardcode `.zshrc` here and avoid this extra parameter?

rodrigosiqueira (changes requested):
@kwy95 could you try this?

@davidbtadokoro is it possible to expand the _kw to be used with bash autocomplete?

setup.sh
davidbtadokoro: > @davidbtadokoro is it possible to expand the _kw to be used with bash autocomplete?

Do you mean adapt the ZSH completion for Bash? If I understood it right, maybe there is a way to do this and I am going to look it up.

However, before this PR, the ZSH was adapted from a Bash native completion implementation and some incompatibilities arose from this non-native approach. Did not studied the Bash completion system, but the ZSH one is really complex - the official documentation even states that "_that there are probably lots of bugs around_" [(Section 10)](https://zsh.sourceforge.io/Guide/zshguide06.html) - so this adaptation may have some caveats.

davidbtadokoro: > Can we just hardcode `.zshrc` here and avoid this extra parameter?

Yes you are right! Maybe we can also hardcode `.bashrc` in the `append_bashcompletion` function, as fwiu the `shellrc` parameter existed for the function handle both Bash and ZSH (now each has its own function). What do you think?

rodrigosiqueira: > > @davidbtadokoro is it possible to expand the _kw to be used with bash autocomplete?
> 
> Do you mean adapt the ZSH completion for Bash? If I understood it right, maybe there is a way to do this and I am going to look it up.
> 
> However, before this PR, the ZSH was adapted from a Bash native completion implementation and some incompatibilities arose from this non-native approach. Did not study the Bash completion system, but the ZSH one is really complex - the official documentation even states that "_there are probably lots of bugs around_" [(Section 10)](https://zsh.sourceforge.io/Guide/zshguide06.html) - so this adaptation may have some caveats.

yeah... ignore my previous comment, just focus on zsh.

davidbtadokoro:
@rodrigosiqueira here is a demo of the `kw config` completion 

kw_config-completion-demo

setup.sh
```diff
@@ -359,9 +359,7 @@ function synchronize_files()
   if command_exists 'zsh'; then
     # Add tabcompletion to zshrc
     if [[ -f "${HOME}/.zshrc" || -L "${HOME}/.zshrc" ]]; then
-      safe_append '# Enable bash completion for zsh' "${HOME}/.zshrc"
-      safe_append 'autoload bashcompinit && bashcompinit' "${HOME}/.zshrc"
-      append_bashcompletion '.zshrc'
+      append_zshcompletion
```
kwy95, line 362: We need to remove the unnecessary legacy lines from the `.zshrc` file.

kwy95 (changes requested):
While trying to test this PR I could not make it work, after installing it and attempting to use the autocomplete feature I receive the following error:
```
kw bui…(anon):4: command not found: _kw_autocomplete                                                                                                      
(anon):4: command not found: _kw_autocomplete
```

My version of zsh is 5.9

setup.sh
davidbtadokoro: Thanks a lot for the feedback! Strangely I could not reproduce the bug and my zsh version is also 5.9 (however my setup is fairly fresh and maybe that is why).

@kwy95 removing the unnecessary legacy lines from `.zshrc` did the trick?

davidbtadokoro:
@rodrigosiqueira, aside from the bug reported by @kwy95 the feature in itself is finished and ready for the review.

src/_kw
```diff
@@ -22,4 +24,17 @@ _kw() {
   fi
 }
 
+_kw_build() {
```
rodrigosiqueira: Any specific reason for keeping the `{` in this line instead of a new line? Check: https://kworkflow.org/content/codingstyle.html#functions

src/_kw
```diff
@@ -0,0 +1,25 @@
+#compdef kw
```
rodrigosiqueira, line 1: Is this file specific to zsh? If so, I recommend you add a small piece of documentation on top of this file that clarifies this file target.

src/_kw
```diff
@@ -24,6 +25,18 @@ _kw() {
   fi
 }
 
+_kw_backup() {
+  local force=
+  if (( words[(I)-r|--restore] )); then
```
rodrigosiqueira: Any specific reason to use this `((` instead of `[[`?

src/_kw
```diff
@@ -24,6 +25,18 @@ _kw() {
   fi
 }
 
+_kw_backup() {
+  local force=
```
rodrigosiqueira: Can you set `force=''`?

src/_kw
```diff
@@ -1,7 +1,8 @@
 #compdef kw
 
 # Main dispatcher
-_kw() {
+_kw()
+{
```
rodrigosiqueira, line 5: I add a comment about this in the previous commit; this commit should not exist. Go back to the old commits, and make a good change there.
Take a look at the command: `git rebase -i`.

src/_kw
```diff
@@ -18,10 +18,14 @@ _kw()
     # Set the context for the command
     curcontext="${curcontext%:*:*}:kw-${command}"
     # Narrow the range of words we are looking at to exclude 'kw'
-    (( CURRENT-- ))
+    ((CURRENT--))
```
rodrigosiqueira, line 44: This change doesn't belong to this commit. Please, move it to where this line was introduced the first time.

src/_kw
```diff
@@ -18,10 +18,14 @@ _kw()
...
     shift words
-    # Run the completion for the command
-    _kw_${command}
+    # Run the completion for the command if it exists
+    if [[ $(type "_kw_${command}") =~ 'not found' ]]; then
+      _nothing
+    else
+      _kw_${command}
```
rodrigosiqueira, line 50: This eval automatically? Don't you need `eval`?

src/_kw
```diff
@@ -31,7 +35,7 @@ _kw()
 _kw_backup()
 {
   local force=
-  if (( words[(I)-r|--restore] )); then
+  if ((words[(I)-r|--restore])); then
```
rodrigosiqueira, line 62: This change doesn't belong to this commit.

setup.sh
```diff
@@ -359,7 +359,7 @@ function synchronize_files()
   if command_exists 'zsh'; then
     # Add tabcompletion to zshrc
     if [[ -f "${HOME}/.zshrc" || -L "${HOME}/.zshrc" ]]; then
-      append_zshcompletion '.zshrc'
+      append_zshcompletion
```
rodrigosiqueira, line 362: I guess this commit should be part of the first one.

src/_kw
```diff
@@ -4,12 +4,13 @@
 _kw()
 {
   local -a commands
-  
+
```
rodrigosiqueira, line 7: This change does not belong to this commit.

src/_kw
```diff
@@ -111,4 +112,36 @@ _kw_config()
...
+  if ((words[(I)-e|--event|-t|--ftrace|-g|--dmesg])); then
+    history='--history[create debug directory to keep track of all debugs made]'
+    follow='--follow[real time output]'
+    cmd='--cmd[run command]: : '
+  fi
+
+  if ((words[(I)-e|--event|-t|--ftrace])); then
+    disable='--disable[disable all events]'
+  fi
+
```
rodrigosiqueira, line 154: Drop this extra space.

src/_kw
```diff
@@ -111,4 +112,36 @@ _kw_config()
...
+    '(-l --list -r --reset -g --dmesg)'{-l-,--list=-}'[list debug options]: : ' \
+    '(-e --event -r --reset -l --list -t --ftrace -g --dmesg)'{-e,--event}'[events trace]: : ' \
+    '(-t --ftrace -r --reset -l --list -e --event -g --dmesg)'{-t,--ftrace}'[ftrace trace]: : ' \
+    '(-g --dmesg -r --reset -l --list -e --event -t --ftrace)'{-g,--dmesg}'[dmesg log]' \
+    '(--remote)--local[set local machine as target]' \
+    '(--local)--remote[set remote machine as target]: : ' \
+    $disable \
+    $history \
+    $follow \
+    $cmd
```
rodrigosiqueira, line 166: Don't we need double quotes here?

src/_kw
```diff
@@ -254,4 +255,52 @@ _kw_init()
...
+    '(-i --interactive -s --send -t --setup -l --list --verify --template)'{-i,--interactive}'[interactively prompt the user for the values of the options]' \
+    '(-l --list -s --send -t --setup -i --interactive --verify --template)'{-l,--list}'[lists the settings that mail uses]' \
+    '(--verify -s --send -t --setup -i --interactive -l --list --template)--verify[verify that all the settings needed are set and valid]' \
+    '(--template -s --send -t --setup -i --interactive -l --list --verify)--template=[load the default configuration values based on a given template]' \
+    $send_options \
+    $global \
+    $local \
+    $force \
+    $no_interactive \
+    $args
```
rodrigosiqueira, line 353: Double-quotes?

src/_kw
```diff
@@ -321,7 +322,7 @@ _kw_man()
 {
   _values 'kw commands' \
     'backup' 'build' 'codestyle' 'config' 'debug' 'deploy' 'device' 'diff' 'drm' 'env' \
-    'explore' 'init' 'mail' 'maintainers' 'pomodoro' 'report' 'self-update' 'ssh' 'vars'
+    'explore' 'init' 'mail' 'maintainers' 'pomodoro' 'remote' 'report' 'self-update' 'ssh' 'vars'
```
rodrigosiqueira, line 369: Is this change really part of this commit?

rodrigosiqueira (changes requested):
Very nice change! See my comments, but keep in mind that I review commits in order.

Also, how about writing a blog post in the kw blog about this?
* Blog: https://blog.kworkflow.org/
* Repo: https://github.com/kworkflow/blog

davidbtadokoro:
Hi @rodrigosiqueira ! 

Sorry for the delay. Thanks a lot for the feedback. Going to work on those. Also, it would be really nice to write on the kw blog!

davidbtadokoro:
This first force-push rebases the branch against the latest unstable changes and fix some problematic commits. They were problematic because:
- They should not exist (as in they should be part of older commits)
- They made changes that broke the atomicity of a commit, i.e., change something that doesn't belong to that commit scope

Working on the other requested changes ASAP.

src/_kw
davidbtadokoro: At first, I used this approach because the completion functions I was using as reference (the git ones) did the check this way and I could not make it work with `[[` instead of `((`.

After trying a bit, I came up with an alternative way that uses `[[`:
```bash
if [[ "${words[@]}" =~ ' -r | --restore' ]]; then
  ...
fi
```

This approach has the drawback of being more verbose (there are some checks that consider many options) and of using a not so reliable "work around" of transforming the `words` array in a string and comparing it with a regex expression. Fwiu, the former way did this in a efficient and elegant manner, however that is just a hunch and I cannot confirm this.

With those things in mind, what approach should I choose?

src/_kw
davidbtadokoro: Same as last comment

src/_kw
davidbtadokoro: Unfortunately, using double quotes here generates errors. But I get that this looks awful 😨

src/_kw
davidbtadokoro: It does eval automatically, but, as we are doing an expansion, I think it is better to use `eval` 👍

src/_kw
davidbtadokoro: Good point! Added a small description explaining to the purpose of the file.

davidbtadokoro:
@rodrigosiqueira  this force-push addresses the changes requested, but there are two points in which I'd like your opinion. Both are in the unresolved conversations.

Nevertheless, going to start writing the blog post.

src/_kw
```diff
@@ -0,0 +1,41 @@
...
+{
+  local -a commands
+
+  commands=()
+
+  if [[ "$CURRENT" -gt 2 ]]; then
+    # Remember the command name
+    local command="${words[2]}"
+    # Set the context for the command
+    curcontext="${curcontext%:*:*}:kw-${command}"
```
rodrigosiqueira, line 53: Just to try to keep the snake case pattern, rename this `curcontext` to `current_context`.

rodrigosiqueira (changes requested):
That is a very nice update!

Btw, update this part of the doc:

https://kworkflow.org/content/project_structure.html#adding-a-new-feature

Expand the description to tell that new features should also update the zsh and bash autocomplete.

davidbtadokoro:
> That is a very nice update!

Thanks a lot! 😄 

> Btw, update this part of the doc:
> 
> https://kworkflow.org/content/project_structure.html#adding-a-new-feature
> 
> Expand the description to tell that new features should also update the zsh and bash autocomplete.

Nice! Gonna work on that.

src/_kw
davidbtadokoro: Unfortunately, `curcontext` is a special parameter in the Zsh completion system and (I think) there is no way to change its name 😞

aquilamacedo:
> @kwy95 could you try this?
> 
> @davidbtadokoro is it possible to expand the _kw to be used with bash autocomplete?

i also use zsh and and it's working fine for me

@rodrigosiqueira @davidbtadokoro @kwy95 

2023-02-23_14-45-11_AdobeExpress (1)

davidbtadokoro:
@kwy95 says:
> We need to remove the unnecessary legacy lines from the `.zshrc` file.

This last force-push adds this change to the first commit

davidbtadokoro:
@rodrigosiqueira says:
> Btw, update this part of the doc:
> 
> https://kworkflow.org/content/project_structure.html#adding-a-new-feature
> 
> Expand the description to tell that new features should also update the zsh and bash autocomplete.

This last added commit implements this.

documentation/content/project_structure.rst
```diff
@@ -117,6 +117,31 @@ need to follow this template:
 3. Implement ``<your_feature_name>_main()`` function in the file you created in
    step 1.
 
+These next steps relate to Bash and Zsh completions to kw. They are not necessary
+for the feature to work, but they are important to maintain the project updated.
```
rodrigosiqueira, line 121: Delete `They are not [..]`. They are necessary; it is part of the new feature.

documentation/content/project_structure.rst
```diff
@@ -117,6 +117,31 @@ need to follow this template:
...
+   to the `commands` array inside the `_kw` function.
+2. Add a function in the `src/_kw` named ``_kw_<your_feature_name>``
+   and implement the completions for your feature. The
+   Zsh completion system is really complex (but powerful), so, aside from
+   referring to the already implemented completions in the file, you might
+   also want to look at these references: `short tutorial <https://github.com
+   /zsh-users/zsh-completions/blob/master/zsh-completions-howto.org
+   #writing-your-own-completion-functions/>`_, `long tutorial <https://zsh.sourceforge.
+   io/Guide/zshguide06.html>`_ and `man page for Zsh completion system
+   <https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Completion-Functions>`_
```
rodrigosiqueira, line 138: Add those references as bullets.

rodrigosiqueira (changes requested):

rodrigosiqueira:
Merged! Thanks!

PR #776, opened 2023-02-12
added support for gentoo

siddhanthrathod:


siddhanthrathod:
Local deployment works

PR #777, opened 2023-02-13
Added support for gentoo

siddhanthrathod:
Still WiP. only local deployment works for now.

### DIFF
--- a/src/build.sh
+++ b/src/build.sh
@@ -141,9 +141,8 @@ function kernel_build()
   fi
 
   if [[ -n "$doc_type" ]]; then
-    command="make ${optimizations} ${doc_type}${output_path}${output_kbuild_flag}"
-    cmd_manager "$flag" "$command"
-    return
+    build_doc "$flag" "$output_kbuild_flag" "$optimizations" "$doc_type" "$output_path"
+    return "$?"
   fi
 
   command="make ${optimizations} ${llvm}ARCH=${platform_ops}${warnings}${output_path}${output_kbuild_flag}"
@@ -180,6 +179,20 @@ function build_menu_config()
   local cmd
 
   cmd="make -j ${llvm}ARCH=${platform_ops} ${menu_config}${env_path}"
+  cmd_manager "$flag" "$cmd"
+}
+
+# This function builds kernel-doc, by default it will create htmldocs.
+function build_doc()
+{
+  local flag="$1"
+  local env_path="$2"
+  local optimizations="$3"
+  local doc_type="$4"
+  local output_path="$5"
+  local cmd
+
+  cmd="make ${optimizations} ${doc_type}${output_path}${env_path}"
   cmd_manager "$flag" "$cmd"
 }
 


rodrigosiqueira (changes requested):
Hi,

First of all, thanks a lot for your contribution; it is really nice to see kw growing to support other OSes families.

I added comments in some specific parts, and aside from that, I also want to add the following suggestions:

1. Keep your commits organized, so each commit has a clear goal. For example, for this PR, I would recommend something like this:
 * Commit 1: Add required setup support for Gentoo, maybe only change in the setup and something related to that.
 * Commit 2: Add support for local deploy.
 * Commit 3: Add support for remote deploy.
2. Take care of your commit message; they are really important. See: https://cbea.ms/git-commit/
3. Add unit test per commit.

setup.sh
```diff
@@ -56,6 +56,12 @@ function check_dependencies()
       [[ "$?" != 0 ]] && package_list="$package $package_list"
     done < "$DOCUMENTATION/dependencies/arch.dependencies"
     cmd="pacman -S $package_list"
+  elif [[ "$distro" =~ 'gentoo' ]]; then
+    while IFS='' read -r package; do
+      installed=$(qlist -I "$package" &> /dev/null)
+      [[ "$?" != 0 ]] && package_list="$package $package_list"
+    done < "$DOCUMENTATION/dependencies/gentoo.dependencies"
+    cmd="emerge $package_list"
```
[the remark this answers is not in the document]
siddhanthrathod: You mean the whole thing right ? (including all the other distros in one function)

src/plugins/kernel_install/gentoo.sh
```diff
@@ -0,0 +1,51 @@
...
+  'app-arch/xz-utils'
+  'app-arch/lzop'
+  'app-arch/zstd'
+  'sys-boot/os-prober'
+  'sys-kernel/dracut'
+)
+
+# Gentoo package manager command
+cmd="" 
+declare -g package_manager_cmd='echo ">=sys-boot/grub-2.06-r5 mount" | sudo tee -a /etc/portage/package.use/grub > /dev/null && sudo emerge --noreplace'
```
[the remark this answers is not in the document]
siddhanthrathod: Yes we need that part if we need os-prober as gentoo doesn't compile grub with mount flag which is needed for os-prober. Yes I'll do it in pre-setup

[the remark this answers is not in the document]
siddhanthrathod: Coincidentally I was just reading like an hour ago about how printf is better than bash echo 🙈

src/plugins/kernel_install/gentoo.sh
```diff
@@ -1,4 +1,3 @@
-
```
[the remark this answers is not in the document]
siddhanthrathod: Yes I'll do it

siddhanthrathod:
Thanks for the reply I'll improve in all the things mentioned above and improve the support by the end of this week and if possible I'll make an ebuild so that kw can be emerged directly in gentoo as in Debian in the future.

setup.sh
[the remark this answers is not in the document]
siddhanthrathod: Does this implementation work @rodrigosiqueira ?

src/plugins/kernel_install/gentoo.sh
```diff
@@ -0,0 +1,48 @@
+# Gentoo package names
```
rodrigosiqueira, line 1: This change does not belong to this commit, fwiu, in this commit, you are focusing on the Gentoo basic setup, meaning you need to move all changes in the `src` folder to its specific commit. In this commit, you should only care about the setup.

setup.sh
```diff
@@ -47,39 +47,24 @@ function check_dependencies()
   local pip_package_list=''
   local cmd=''
   local distro
-
+  debian=("apt list --installed | cut -d '/' -f1" "apt install -y")
+  arch=("pacman -Ql " "pacman -S")
+  gentoo=("qlist -I " "emerge")
+  fedora=("rpm -q " "dnf install -y")
+  declare -A distro_commands=(['arch']=arch ['gentoo']=gentoo ['fedora']=fedora ['debian']=debian)
+  declare -n c
```
rodrigosiqueira, line 55: `c`? You can use a more meaningful name.

setup.sh
```diff
@@ -47,39 +47,24 @@ function check_dependencies()
...
-      [[ "$?" -ne 0 ]] && package_list="$package $package_list"
-    done < "$DOCUMENTATION/dependencies/fedora.dependencies"
-    cmd="dnf install -y $package_list"
-  else
-    warning 'Unfortunately, we do not have official support for your distro (yet)'
-    warning 'Please, try to find the following packages:'
-    warning "$(cat "$DOCUMENTATION/dependencies/arch.dependencies")"
-    return 0
-  fi
+  for c in ${distro_commands[$distro]}; do
```
rodrigosiqueira, line 58: Is this `c` the same as the global `c`? Anyway, could you use a more meaningful name?
Finally, add double-quotes around `$distro`.

setup.sh
```diff
@@ -47,39 +47,24 @@ function check_dependencies()
...
+    pfix=${c[1]}
```
rodrigosiqueira, line 59: Could you use a better name for `pfix`?

setup.sh
```diff
@@ -47,39 +47,24 @@ function check_dependencies()
...
+    command=${c[0]}
+  done
+
+  while IFS='' read -r package; do
+    installed=$($command "$package" &> /dev/null)
+    [[ $? -ne 0 ]] && package_list="$package $package_list"
+  done < "$DOCUMENTATION/dependencies/$distro.dependencies"
```
rodrigosiqueira, line 66: About string concatenation, refer to: https://kworkflow.org/content/codingstyle.html#string-concatenation

Review all your PR for those string concatenations.

setup.sh
```diff
@@ -47,39 +47,24 @@ function check_dependencies()
   local pip_package_list=''
   local cmd=''
   local distro
-
+  debian=("apt list --installed | cut -d '/' -f1" "apt install -y")
```
rodrigosiqueira, line 50: btw, this commit sequence is better:

- Commit 1: This code refactor.
- Commit 2: Add Gentoo part.

rodrigosiqueira (changes requested):
You are on the right track; just some comments:

- Please review your code to use the string concatenation as described at: https://kworkflow.org/content/codingstyle.html#string-concatenation
- Try to better organize your commit sequence in an atomic way. Look for the `git rebase -i` command; mastering this command will be a powerful tool for you in any project.

src/plugins/kernel_install/gentoo.sh
siddhanthrathod: Okay I'll re base accordingly

setup.sh
siddhanthrathod: Sorry I thought I'll do that after testing but I forgot 🙈 I'll do it !

setup.sh
siddhanthrathod: Yes yes

setup.sh
siddhanthrathod: I'll sir

setup.sh
siddhanthrathod: I'll do it

setup.sh
siddhanthrathod: The curly braces like {$DOCUMENTATION} right ?

rodrigosiqueira: like this: `${DOCUMENTATION}`

siddhanthrathod: > like this: `${DOCUMENTATION}` 

Yes I meant this 🙈

setup.sh
```diff
@@ -47,33 +47,24 @@ function check_dependencies()
...
```
rodrigosiqueira, line 50: Use `--delimiter=DELIM` instead of `-d`. Additionally, if you don't have a string concatenation, I recommend using single-quotes (review this in this entire commit).

setup.sh
```diff
@@ -47,33 +47,24 @@ function check_dependencies()
...
+  arch=("pacman -Ql " "pacman -S")
+  gentoo=("qlist -I " "emerge")
+  fedora=("rpm -q " "dnf install -y")
+  declare -A distro_commands=(['arch']=arch ['gentoo']=gentoo ['fedora']=fedora ['debian']=debian)
+  declare -n c
```
rodrigosiqueira: I guess you did not update the `c` variable yet...

setup.sh
```diff
@@ -47,33 +47,24 @@ function check_dependencies()
...
-    cmd="dnf install -y $package_list"
-  else
-    warning 'Unfortunately, we do not have official support for your distro (yet)'
-    warning 'Please, try to find the following packages:'
-    warning "$(cat "$DOCUMENTATION/dependencies/arch.dependencies")"
-    return 0
-  fi
+  for c in ${distro_commands[$distro]}; do
+    pfix=${c[1]}
+    command=${c[0]}
```
rodrigosiqueira: Any reason not to use double quotes around `${c[0]}`. Replace the command variable with `cmd` since `command` is built-in.

setup.sh
rodrigosiqueira, line 1: Btw, add a meaningful commit message.

setup.sh
```diff
@@ -52,19 +52,19 @@ function check_dependencies()
...
-  declare -n c
+  declare -n distro_cmd
   distro=$(detect_distro '/')
 
-  for c in ${distro_commands[$distro]}; do
-    pfix=${c[1]}
-    command=${c[0]}
+  for distro_cmd in ${distro_commands[$distro]}; do
+    install_cmd=${distro_cmd[1]}
+    command=${distro_cmd[0]}
```
rodrigosiqueira: Ok, now I can see that you stack some of my comments from the previous review in this new commit (I review PR from the first commit to the last one). Changing the original commit is the right way to do this sort of modification. You can achieve it by using `git rebase -i SHA_COMMIT_BEFORE_YOUR_FIRST_COMMIT`, and in the interactive menu, you should use the `e`  option for each commit that you want to make a change. After ensuring each commit is self-contained, you finish the rebase and force push to your branch.

Check this page; it might help you:
https://flusp.ime.usp.br/git/feed-english/rewriting-history-101/

rodrigosiqueira (changes requested):

PR #779, opened 2023-02-16
src: kw_remote: Add support for global remote.config

### LINKED ISSUE #775: kw remote add/remove dos not work in the global scope
**Description:**

Even though the `remote.config` is available in the global config (`~/.config/kw/remote.config`), the `kw remote` options are not working on this file. Let's ensure that kw remote is also able to change this file.

Kamorst:
`kw remote` checks for the presence of `$PWD/.config/kw/remote.conf` and use it if a local one is not found.

Fixes #775.

### DIFF
--- a/src/kw_remote.sh
+++ b/src/kw_remote.sh
@@ -3,6 +3,7 @@ include "$KW_LIB_DIR/remote.sh"
 
 declare -gA options_values
 declare -g local_remote_config_file="${PWD}/.kw/remote.config"
+declare -g global_remote_config_file="${KW_ETC_DIR}/remote.config"
 
 function remote_main()
 {
@@ -43,24 +44,60 @@ function remote_main()
   fi
 }
 
+# Returns the local or global remote.config location.
+# The local has precedence.
+#
+# Returns:
+# The remote config file if found and 22 otherwise.
+function choose_correct_remote_config_file()
+{
+  local has_local_config='false'
+  local has_global_config='false'
+
+  if [[ -d "${PWD}/.kw" ]]; then
+    has_local_config='true'
+  fi
+
+  if [[ -d "${KW_ETC_DIR}" ]]; then
+    has_global_config='true'
+  fi
+
+  if [[ "${has_local_config}" == 'true' && -z "${options_values['GLOBAL']}" ]]; then
+    printf '%s' "${local_remote_config_file}"
+  elif [[ "${has_global_config}" == 'true' ]]; then
+    printf '%s' "${global_remote_config_file}"
+  else
+    return 22 # EINVAL
+  fi
+}
+
 function list_remotes()
 {
   local remote_config
   local str_process
+  local remote_config_file
+  local ret
 
-  # Check if .kw folder exits
-  if [[ ! -d "${PWD}/.kw" ]]; then
-    complain 'You do not have .kw folder, try kw init first'
-    return 22 # EINVAL
+  remote_config_file="$(choose_correct_remote_config_file)"
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain 'No local or global config found. Please reinstall kw or run "kw init" to create a .kw in this dir.'
+    return "$ret"
   fi
 
-  # If we don't have a remote.config file yet, let the user know
-  if [[ ! -f "${local_remote_config_file}" ]]; then
-    complain 'Did you run kw init? It looks like that you do not have remote.config'
-    return 22 # EINVAL
+  if [[ ! -f "$remote_config_file" ]]; then
+    # If we don't have a remote.config file yet, let the user know
+    if [[ "$remote_config_file" == "$local_remote_config_file" ]]; then
+      complain 'Did you run kw init? It looks like that you do not have remote.config.'
+      return 22 # EINVAL
+    fi
+    if [[ "$remote_config_file" == "$global_remote_config_file" ]]; then
+      complain 'Did you install kw correctely? It looks like that you do not have a global remote.config.'
+      return 22 # EINVAL
+    fi
   fi
 
-  remote_config=$(< "${local_remote_config_file}")
+  remote_config=$(< "${remote_config_file}")
 
   while read -r line; do
     grep --quiet "^#kw-default=" <<< "$line"
@@ -89,6 +126,14 @@ function add_new_remote()
   local host_ssh_config
   local user_ssh_config
   local port_ssh_config
+  local remote_config_file
+
+  remote_config_file="$(choose_correct_remote_config_file)"
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain 'No local or global config found. Please reinstall kw or run "kw init" to create a .kw in this dir.'
+    return "$ret"
+  fi
 
   read -ra add_parameters <<< "${options_values['PARAMETERS']}"
 
@@ -108,12 +153,12 @@ function add_new_remote()
   fi
 
   # If we don't have a remote.config file yet, let's create it
-  if [[ ! -f "${local_remote_config_file}" ]]; then
+  if [[ ! -f "${remote_config_file}" ]]; then
     if [[ ! -d "${PWD}/.kw" ]]; then
       complain 'Did you run kw init? It looks like that you do not have the .kw folder'
       exit 22 # EINVAL
     fi
-    touch "$local_remote_config_file"
+    touch "$remote_config_file"
     first_time='yes'
   fi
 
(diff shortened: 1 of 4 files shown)


Kamorst:
TODO: add some more tests.

davidbtadokoro:
> TODO: add some more tests.

Implemented single tests for add/remove/rename/list/set-default global remote.

src/kw_remote.sh
```diff
@@ -43,24 +44,59 @@ function remote_main()
...
 
-function list_remotes()
+# Returns the local or global remote.config location.
+# The local has precedence.
+function choose_correct_remote_config_file()
 {
-  local remote_config
-  local str_process
+  local has_local_config="false"
+  local has_global_config="false"
```
rodrigosiqueira, line 52: Uses single-quotes.

src/kw_remote.sh
```diff
@@ -43,24 +44,59 @@ function remote_main()
...
 
-  # Check if .kw folder exits
-  if [[ ! -d "${PWD}/.kw" ]]; then
-    complain 'You do not have .kw folder, try kw init first'
-    return 22 # EINVAL
+  if [[ -d "${PWD}/.kw" ]]; then
+    has_local_config="true"
```
rodrigosiqueira, line 55: Use single quotes around true.

src/kw_remote.sh
```diff
@@ -43,24 +44,59 @@ function remote_main()
...
+  fi
+
+  if [[ "${has_local_config}" = "false" && -d "${KW_ETC_DIR}" ]]; then
```
rodrigosiqueira, line 58: Use single quotes and `-eq` instead of `=`.

src/kw_remote.sh
```diff
@@ -43,24 +44,59 @@ function remote_main()
...
+    has_global_config="true"
   fi
 
-  # If we don't have a remote.config file yet, let the user know
-  if [[ ! -f "${local_remote_config_file}" ]]; then
-    complain 'Did you run kw init? It looks like that you do not have remote.config'
+  if [[ "${has_local_config}" = "true" ]]; then
+    #say "Using local remote.config." 1>&2
+    printf '%s' "${local_remote_config_file}"
```
rodrigosiqueira, line 64: `say` or `print`?

src/kw_remote.sh
```diff
@@ -43,24 +44,59 @@ function remote_main()
...
+    return 0
+  elif [[ "${has_global_config}" = "true" ]]; then
+    #say "Using global remote.config." 1>&2
+    printf '%s' "${global_remote_config_file}"
+    return 0
```
rodrigosiqueira, line 69: Maybe move these two `return 0` after the `if` condition.

src/kw_remote.sh
```diff
@@ -43,24 +44,59 @@ function remote_main()
...
+  else
+    complain 'No local or global config found. Please reinstall kw or run "kw init" to create a .kw in this dir.' 1>&2
```
rodrigosiqueira, line 71: Why `1>&2` at the end of the complain?

src/kw_remote.sh
```diff
@@ -43,24 +44,59 @@ function remote_main()
...
+function list_remotes()
+{
+  local remote_config
+  local str_process
+  local remote_config_file
+  local ret
+
+  remote_config_file="$(choose_correct_remote_config_file)"
+  ret="$?"
+  [[ "$ret" != "0" ]] && return "$ret"
```
rodrigosiqueira, line 85: You don't need quotes around 0.

src/kw_remote.sh
```diff
@@ -43,24 +44,59 @@ function remote_main()
...
+
+  if [[ ! -f "$remote_config_file" ]]; then
+    # If we don't have a remote.config file yet, let the user know
+    if [[ "$remote_config_file" = "$local_remote_config_file" ]]; then
```
rodrigosiqueira, line 89: replace `=` with `-eq`.

src/kw_remote.sh
```diff
@@ -89,6 +125,11 @@ function add_new_remote()
   local host_ssh_config
   local user_ssh_config
   local port_ssh_config
+  local remote_config_file
+
+  remote_config_file="$(choose_correct_remote_config_file)"
+  ret="$?"
+  [[ "$ret" != "0" ]] && return "$ret"
```
rodrigosiqueira, line 132: You don't need quotes around 0.

rodrigosiqueira (changes requested):

src/kw_remote.sh
Kamorst: i'm getting some strange error when making this alteration.
`$> ./run_tests.sh test kw_remote >/dev/null`
`./src/kw_remote.sh: line 87: [[: /tmp/shunit.IQPPtY/tmp/.kw/remote.config: syntax error: operand expected (error token is "/tmp/shunit.IQPPtY/tmp/.kw/remote.config")`
`./src/kw_remote.sh: line 97: /tmp/shunit.IQPPtY/tmp/.kw/remote.config: No such file or directory`
`./src/kw_remote.sh: line 87: [[: /tmp/shunit.IQPPtY/tmp/.config/kw/remote.config: syntax error: operand expected (error token is "/tmp/shunit.IQPPtY/tmp/.config/kw/remote.config")`

rodrigosiqueira: I reread the bash documentation, and I realized that -eq is only for arithmetic cooperation. We can use `=`, but tbh, I think `==` is more friendly since other languages use it; could you replace `=` with `==`?

src/kw_remote.sh
```diff
@@ -43,24 +44,57 @@ function remote_main()
   fi
 }
 
+# Returns the local or global remote.config location.
+# The local has precedence.
```
rodrigosiqueira, line 48: Document this function return.

src/kw_remote.sh
```diff
@@ -127,11 +169,11 @@ function add_new_remote()
   fi
 
   # Check if remote name already exists
-  grep -xq "^Host ${name}$" "$local_remote_config_file"
+  grep -xq "^Host ${name}$" "$remote_config_file"
```
rodrigosiqueira, line 172: Since you are changing this file, can you add one extra commit where you replace:

`grep -xq ..` with `grep --line-regexp --quiet`.

Using the full parameter name in the bash code makes it easier to understand what's going on in the line.

P.s.: I'm not sure if we documented this idea, but if not we should.

src/kw_remote.sh
```diff
@@ -127,11 +169,11 @@ function add_new_remote()
...
   if [[ "$?" == 0 ]]; then
-    sed -i -r "/^Host ${name}$/{n;s/Hostname.*/Hostname ${remote_parameters['REMOTE_IP']}/}" "$local_remote_config_file"
-    sed -i -r "/^Host ${name}$/{n;n;s/Port.*/Port ${remote_parameters['REMOTE_PORT']}/}" "$local_remote_config_file"
-    sed -i -r "/^Host ${name}$/{n;n;n;s/User.*/User ${remote_parameters['REMOTE_USER']}/}" "$local_remote_config_file"
+    sed -i -r "/^Host ${name}$/{n;s/Hostname.*/Hostname ${remote_parameters['REMOTE_IP']}/}" "$remote_config_file"
+    sed -i -r "/^Host ${name}$/{n;n;s/Port.*/Port ${remote_parameters['REMOTE_PORT']}/}" "$remote_config_file"
+    sed -i -r "/^Host ${name}$/{n;n;n;s/User.*/User ${remote_parameters['REMOTE_USER']}/}" "$remote_config_file"
```
rodrigosiqueira, line 176: same as my previous comment.

rodrigosiqueira (changes requested):
That is a very nice update!

I added a few comments inline.

Btw, we need one option to indicate kw remote if it should update the global config or the local one where local is the default if nothing is used.

For example:
```
kw remote add origin root@something --global # update global config
kw remote add origin root@other # Add to the local remote
kw remote remove origin --global
```

Could you also add a new commit for that?

Kamorst:
> That is a very nice update!
> 
> I added a few comments inline.
> 
> Btw, we need one option to indicate kw remote if it should update the global config or the local one where local is the default if nothing is used.
> 
> For example:
> 
> ```
> kw remote add origin root@something --global # update global config
> kw remote add origin root@other # Add to the local remote
> kw remote remove origin --global
> ```
> 
> Could you also add a new commit for that?

Added `--global` support to all remote sub-commands instead. I hope its okay.

I had forgotten to add this in a new commit, is that necessary?

rodrigosiqueira:
Merged! thanks

Issue #774, opened 2023-02-11
kw throws an error message if env.current does not exist when trying to destroy.


rodrigosiqueira:
**Description:**

If you create a new env and right after that try to destroy it, you will see this error:

```bash
kw env --destroy TEST-PT1
Are you sure you want to delete this environment? [y/N]: y
/home/siqueira/.local/lib/kw/kw_env.sh: line 221: /home/siqueira/Documents/Repositories/kernels/amd-next_dal-dev/.kw/env.current: No such file or directory
The "$env_name" environment has been destroyed.
```

I think `kw env --create` is not creating `env.current` after your change.

**How to Reproduce:**

1. Make sure you have a clean setup, without env.
2. Create a new env: kw env --create TEST-PT1
3. Destroy env: kw env --destroy TEST-PT1

I guess we just need a simple check in the kw_env.sh, before the line:

```
current_env=$(< "${local_kw_configs}/${ENV_CURRENT_FILE}")
```

**Expected behavior:**

No error message

aquilamacedo:
I'm working on it.

PR #780, opened 2023-02-20
src: kw_env: fixing the "env.current does not exist" bug.

aquilamacedo:
The destroy_env() function had a bug where it looked for the env.current file, and that file didn't exist. This bug occurred when the user created an env and then destroyed that same env.

This commit checks if the env.current file exists to see if the user is using the environment they want to be destroyed.

Closes: #774

### DIFF
--- a/src/kw_env.sh
+++ b/src/kw_env.sh
@@ -216,11 +216,12 @@ function destroy_env()
   if [[ $(ask_yN 'Are you sure you want to delete this environment?') =~ 0 ]]; then
     return 22 # EINVAL
   fi
-  # Here it is checked if the current env is equal to the env we want to
-  # destroy, if so, we call the exit_env function to exit the environment mode.
-  current_env=$(< "${local_kw_configs}/${ENV_CURRENT_FILE}")
-  if [[ "$current_env" == "$env_name" ]]; then
-    exit_env
+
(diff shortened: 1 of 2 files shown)


src/kw_env.sh
```diff
@@ -216,11 +216,15 @@ function destroy_env()
...
-  # destroy, if so, we call the exit_env function to exit the environment mode.
-  current_env=$(< "${local_kw_configs}/${ENV_CURRENT_FILE}")
-  if [[ "$current_env" == "$env_name" ]]; then
-    exit_env
+  # Here it is checked if the env.current file exists,
+  # that is, if the user uses a current env.
+  if [[ -f "${local_kw_configs}/${ENV_CURRENT_FILE}" ]]; then
+    current_env=$(< "${local_kw_configs}/${ENV_CURRENT_FILE}")
+    # Here it is checked if the current env is equal to the env we want to
+    # destroy, if so, we call the exit_env function to exit the environment mode.
```
rodrigosiqueira, line 224: You don't need all of these comments; just a single comment + a good commit message should be enough.

rodrigosiqueira (changes requested):

src/kw_env.sh
aquilamacedo: ok!!

rodrigosiqueira:
Merged! Thanks

Issue #787, opened 2023-03-07
standardize the code structure of the build.sh file


aquilamacedo:
**Description:**

The `build.sh` file needs to be organized to follow the pattern of the other files

**How is the code:**

![image](https://user-images.githubusercontent.com/43859386/223567447-05f14ce3-834a-4e2b-a2ac-d1670c3f4e5d.png)

**What is expected:**

Following the kw code structure pattern we should start with a function like `kernel_build_main()`

PR #788, opened 2023-03-07
src: build: standardize build.sh file.

aquilamacedo:
In order to follow the pattern of other kw files, this commit
organizes the order of functions in the `build.sh` file.

Closes: #787

### DIFF
--- a/kw
+++ b/kw
@@ -81,7 +81,7 @@ function kw()
       (
         include "$KW_LIB_DIR/build.sh"
 
-        kernel_build '' "$@"
+        build_kernel_main '' "$@"
         local ret="$?"
         alert_completion 'kw build' "$1"
         return "$ret"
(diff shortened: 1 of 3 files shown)


rodrigosiqueira (changes requested):
lgtm, but fix the CI failures.

aquilamacedo:
> lgtm, but fix the CI failures.

I'm trying to fix this error in the CI. According to review-dog, the problem is on line 743, but I don't understand the reason for the error.

image

rodrigosiqueira:
Please, check:
https://github.com/koalaman/shellcheck/wiki/SC2317

I guess `build_kernel_main` has an exit.

aquilamacedo:
Hi @rodrigosiqueira. I figured out what the problem was and fixed it. It had a duplicate function and two functions with the same name; I don't know how this code got past CI before. It is now working correctly!

rodrigosiqueira:
Merged, thanks.

PR #789, opened 2023-03-08
src: build: Adding verbose mode.

aquilamacedo:
This commit adds support for the verbose parameter within `kw build`. The verbose parameter gives details of the commands that are executed behind the scenes.

Note: This is part of the issue: #179


aquilamacedo:
Hi @rodrigosiqueira. 

I've started implementing the verbose parameter in `kw build`, do you have any recommendations for something I can change/add?

src/build.sh
```diff
@@ -102,7 +106,7 @@ function kernel_build()
   fi
 
   if [[ -n "$clean" ]]; then
-    build_clean "$flag"
+    build_clean "$flag" "$output_kbuild_flag"
```
rodrigosiqueira, line 109: This change and the modification to `build_clean` should be in its own commit.

src/build.sh
```diff
@@ -118,8 +122,7 @@ function kernel_build()
   fi
 
   if [[ -n "$menu_config" ]]; then
-    command="make -j ${llvm}ARCH=${platform_ops} ${menu_config}${output_kbuild_flag}"
-    cmd_manager "$flag" "$command"
+    build_menu_config "$flag" "$output_kbuild_flag" "$menu_config" "$platform_ops" "$llvm"
```
rodrigosiqueira, line 125: This modification should have a dedicated commit to it.

src/build.sh
```diff
@@ -142,11 +145,12 @@ function kernel_build()
   fi
 
   if [[ -n "$doc_type" ]]; then
-    command="make ${optimizations} ${doc_type}${output_path}${output_kbuild_flag}"
-    cmd_manager "$flag" "$command"
-    return
+    build_doc "$flag" "$output_kbuild_flag" "$optimizations" "$doc_type" "$output_path"
```
rodrigosiqueira, line 148: This change should have a dedicated commit.

src/build.sh
```diff
@@ -168,6 +172,41 @@ function kernel_build()
...
+# This function builds kernel-doc, by default it will create htmldocs.
+function build_doc()
+{
+  local flag="$1"
+  local env_path="$2"
+  local optimizations="$3"
+  local doc_type="$4"
+  local output_path="$5"
+  local cmd
+  say "$doc_type"
```
rodrigosiqueira, line 203: Add space between the variable declaration and `say`.

src/build.sh
```diff
@@ -352,6 +402,10 @@ function parse_build_options()
       --)
         shift
         ;;
+      TEST_MODE)
+        options_values['TEST_MODE']='TEST_MODE'
+        shift
+        ;;
```
rodrigosiqueira, line 408: I don't think we need this.

src/build.sh
```diff
@@ -376,6 +430,7 @@ function build_help()
     '  build (-S | --cpu-scaling) <percentage> - Scale CPU usage by factor' \
     '  build (--ccache) - Enable use of ccache' \
     '  build (-w | --warnings) [warning_levels] - Enable warnings' \
+    '  build (-v | --verbose) - Show a detailed output' \
```
rodrigosiqueira, line 433: You also need to update the man doc.

rodrigosiqueira (changes requested):

aquilamacedo:
Hi @rodrigosiqueira. 

I dedicated the commits to another pull request. When you merge, I'll update that pull request to continue the verbose option.

src/build.sh
aquilamacedo: We really don't need that, I ended up making a mistake.

aquilamacedo:
Hi @rodrigosiqueira. :)

I started to add verbose mode in some parts of the build, but I'm still implementing it for other parameters.

aquilamacedo:
Hi @rodrigosiqueira.

The `--verbose` option works for all **build** options except in the `--info` parameter. I wondered if it is worth implementing, as the only commands running in the build_info function are the commands created by **get_kernel_release()** and **get_kernel_versions()**. What do you think?

documentation/man/features/build.rst
```diff
@@ -16,6 +16,8 @@ SYNOPSIS
 | *kw* (*b* | *build*) [\--llvm] [\--alert=(s | v | (sv | vs) | n)]
 | *kw* (*b* | *build*) [(-c | \--clean)] [\--alert=(s | v | (sv | vs) | n)]
 | *kw* (*b* | *build*) [(-f | \--full-cleanup)] [\--alert=(s | v | (sv | vs) | n)]
+| *kw* (*b* | *build*) [(-v | \--verbose)] [\--alert=(s | v | (sv | vs) | n)]
```
rodrigosiqueira, line 19: Add `--verbose` in the `bash_autocomplete` and `_kw`.

documentation/man/features/env.rst
```diff
@@ -11,6 +11,7 @@ SYNOPSIS
 | *kw* (*env*) [(-d | \--destroy <NAME>)]
 | *kw* (*env*) [(-l | \--list)]
 | *kw* (*env*) [(-e | \--exit-env)]
+| *kw* (*env*) [(-v | \--verbose)]
```
rodrigosiqueira, line 14: Add `--verbose` in the bash_autocomplete and `_kw`.

rodrigosiqueira (changes requested):

rodrigosiqueira:
> Hi @rodrigosiqueira.
> 
> The `--verbose` option works for all **build** options except in the `--info` parameter. I wondered if it is worth implementing, as the only commands running in the build_info function are the commands created by **get_kernel_release()** and **get_kernel_versions()**. What do you think?

I think it is worth it. Check get_kernel_* function, and you will see that we are forcing SILENT.

rodrigosiqueira:
Hi @aquilamacedo , could you finish this PR?

aquilamacedo:
> Hi @aquilamacedo , could you finish this PR?

yes, I'm almost done updating _kw.

documentation/man/features/build.rst
magalilemes: It seems like `--alert` is a legacy option, which no longer works now. So no need to add it here.

src/build.sh
```diff
@@ -31,6 +31,7 @@ function build_kernel_main()
   local llvm
   local env_name
   local clean
+  local flag
```
magalilemes, line 34: This variable has already been declared.

src/kw_env.sh
magalilemes, line 1: Just so that I can understand this better: is `flag`, in this case, a global variable? If so, perhaps declare it that way.

aquilamacedo: Yes, the variable is being used globally as it is used in several functions. do you think it would be necessary to declare it as global?

src/build.sh
aquilamacedo: ok, i'm fixing this.

src/build.sh
```diff
@@ -119,6 +121,8 @@ function build_kernel_main()
     return "$?"
   fi
 
+  flag=${flag:-'SILENT'}
```
rodrigosiqueira, line 124: This should be at the beginning of this function.

src/build.sh
```diff
@@ -353,6 +366,10 @@ function parse_build_options()
         options_values['FULL_CLEANUP']=1
         shift
         ;;
+      --verbose | -v)
+        options_values['VERBOSE']=1
+        shift
```
rodrigosiqueira, line 371: Add a simple unit test for this new parameter (same idea for all the commits in this PR).

documentation/man/features/build.rst
```diff
@@ -83,6 +84,9 @@ OPTIONS
   build process, including configuration files and script output. This option also
   considers whether or not the user is using kw env.
 
+-v, \--verbose:
+  Verbose mode allows the user to see the commands executed under the hood.
```
rodrigosiqueira, line 88: Let's use something more formal in the documentation. Something like this:

Verbose mode. Causes kw to print debugging messages about its progress. This is helpful in debugging issues.

(Same idea for all commits)

src/kw_env.sh
rodrigosiqueira: flag should not be global. It should be local.

src/kw_env.sh
```diff
@@ -76,14 +81,17 @@ function use_target_env()
       # Check if the config file exists before trying to remove it.
       [[ ! -f "${local_kw_configs}/${config}.config" ]] && continue
 
-      mv "${local_kw_configs}/${config}.config" "$tmp_trash"
+      cmd="mv ${local_kw_configs}/${config}.config $tmp_trash"
```
rodrigosiqueira, line 84: since you are making this change, please add `{}` around `tmp_trash`.

src/backup.sh
```diff
@@ -14,19 +14,23 @@ decompress_path="$KW_CACHE_DIR/tmp-kw-backup"
 # recieves an argument and call other functions based on that.
 function backup()
 {
+  local flag
+
```
rodrigosiqueira, line 18: Add `flag=${flag:-'SILENT'}`.

documentation/man/features/mail.rst
```diff
@@ -12,6 +12,7 @@ SYNOPSIS
 | *kw mail* (-l | \--list)
 | *kw mail* \--verify [\--local | \--global]
 | *kw mail* \--template[=<template>] [-n | \--no-interactive] [\--local | \--global] [-f | \--force] [(<config> <value>)...]
+| *kw mail* (-V | \--verbose)
```
rodrigosiqueira, line 15: In this case, let's drop the -V option

src/mail.sh
```diff
@@ -23,6 +23,8 @@ declare -gr email_regex='[A-Za-z0-9_\.-]+@[A-Za-z0-9_-]+(\.[A-Za-z0-9]+)+'
 #shellcheck disable=SC2119
 function mail_main()
 {
+  local flag
+
```
rodrigosiqueira, line 27: Add `flag=${flag:-'SILENT'}` or similar

rodrigosiqueira (changes requested):

documentation/man/features/mail.rst
aquilamacedo: What do you mean when you say 'drop the -V option'? Are you suggesting removing the short option (-V) for verbose, leaving only the long option (--verbose)?

rodrigosiqueira: just remove the short option `-V` for the mail feature.

rodrigosiqueira:
Merged! Thanks

PR #790, opened 2023-03-08
run_tests.sh: Fix name collision when specifying test name

Kamorst:
When specifying the test (`run_tests.sh test <test>`) if the shunit2 repo where cloned inside the kworkflow repo `tests` folder and there is a file which start with the same name inside `tests/shunit2` the test might fail. As an example `run_tests.sh test init` fails because it finds `tests/shunit2/init_githooks.sh` and tries to run `tests/init_githooks.sh`, instead of `tests/init_test.sh`, which doesn't exists.

This commit fix this issue.


rodrigosiqueira:

Issue #792, opened 2023-03-10
`kw upstream-patches-ui` missing `xpath` and `curl` dependencies


davidbtadokoro:
**Description:**

Because the `xpath` and `curl` dependencies aren't checked when installing kw, setups that doesn't include those will result in a `command not found` error.

**How to Reproduce:**

If you don't have any of those dependencies installed on your machine, try to execute the `upstream-patches-ui` command and you shall see an error either immediately or when listing patches.

Otherwise, uninstall either of those and try executing the same as above.

**Expected behavior:**

The installation should check those dependencies, even though they are fairly common in most setups.

**System information:**

 - OS: Tested on both Arch and Debian
 - kw Version: 89dc843
 - Bash Version 5.1.16 (on Arch) and 5.1.4 (on Debian)

davidbtadokoro:
@rodrigosiqueira could you add the lore label, please?

PR #794, opened 2023-03-10
documentation: dependencies: Add curl and xpath dependencies

davidbtadokoro:
The commands curl and xpath are used by the kw upstream-patches-ui command, but those dependencies are not checked when kw is installed. Although they are really common commands in almost every linux installation, if the setup doesn't have these dependencies installed, a command not found error occurs.

This commit fix this problem by adding both dependencies to the dependencies files of Arch, Debian and Fedora distros.

Closes: #792

### DIFF
--- a/documentation/dependencies/arch.dependencies
+++ b/documentation/dependencies/arch.dependencies
@@ -26,3 +26,5 @@ rsync
 ccache
 python-setuptools
 dialog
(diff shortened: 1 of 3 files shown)


rodrigosiqueira:
Nice PR!
Merged in the unstable branch.

Issue #793, opened 2023-03-10
`kw upstream-patches-ui` add an `-h` option


davidbtadokoro:
**Description:**

`kw upstream-patches-ui` should have an `-h` option

davidbtadokoro:
@rodrigosiqueira can you add the lore label, please?

PR #795, opened 2023-03-10
src: upstream_patches_ui: Add help option

davidbtadokoro:
The kw upstream_patches_ui doesn't have a help option to inform the user about the command usage.

This commit adds a -h option that superficially tells what the feature is about.

Note: This commit also adds a --help option to show the man page of the feature, but, as there isn't a man page yet, using this option results in nothing. Also, the implementation follows the one done in the kw mail feature, as it is the most up-to-date one.

Closes: #793


rodrigosiqueira:
Merged. Thanks

Issue #797, opened 2023-03-10
Add notification of loading


rodrigosiqueira:
**Description:**

When using `upstream-patches-ui` to access patches from the public mailing list or to download them, there is a delay in completing this action. We should try to add some sort of visual notification of loading/waiting.

rodrigosiqueira:
@davidbtadokoro

PR #806, opened 2023-03-11
src: upstream_patches_ui: Add loading screen for delayed actions

davidbtadokoro:
When the user tries to access patches from the public mailing list or download a given patch (or series) there is a possible delay between the requisiton and the completion (i.e. the display of the new screen). There is no visual notification to the user (for the dialog view there is only a black screen with a cursor) which does not indicate whether an error happened or if it is just a delay.

This commit adds this visual notification with the addition of the create_loading_screen function that accepts a custom message to be displayed to the user. In case of the dialog view, this message is displayed inside an info box.

Note that this function may be expanded to display a progress bar or gauge.

Closes: #797

### DIFF
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -187,6 +187,57 @@ function create_simple_checklist()
   return "$ret"
 }
 
+# Create simple loading screen notification for delayed actions.
+#
+# @loading_message: The message to be displayed to the user while loading.
+# @height: Menu height in lines size, the default value is 8.
+# @width: Menu width in column size, the default value is 60.
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/kwlib.sh` function `cmd_manager`
+#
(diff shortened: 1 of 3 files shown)


rodrigosiqueira (changes requested):
Hi @davidbtadokoro ,
Thanks a lot for this PR; I added some inline comments.

It would be nice if we could attach a progress bar to the loading page by using some sort of; howeverowever, for now, we can move on with this solution.

Btw, there are multiple samples that came with the dialog project. For example:

https://github.com/tolik-punkoff/dialog-examples
(you probably have this somewhere in your filesystem since this is part of the dialog package)

There are multiple gauge examples, and an interesting one is named wheel.

davidbtadokoro:
Hey @rodrigosiqueira ,

Thanks for the feedback! Going to solve them asap.

rodrigosiqueira:
Merged! Thanks.

PR #807, opened 2023-03-13
setup.sh: Download pip packages inside virtualenv

Kamorst:
Do not litter the `$HOME` of the user installing pip packageds inside the home user directory. Instead download pip packages inside a virtualenv only when building the man pages and documentation.

### DIFF
--- a/documentation/dependencies/arch.dependencies
+++ b/documentation/dependencies/arch.dependencies
@@ -1,14 +1,11 @@
 bash
 git
 tar
-python-docutils
 pulseaudio
 libpulse
 dunst
-python-sphinx
 imagemagick
 graphviz
-python-virtualenv
 texlive-bin
 librsvg
 bzip2
@@ -16,7 +13,6 @@ lzip
 lzop
 zstd
 xz
-python-pip
 bc
 perl-authen-sasl
 perl-io-socket-ssl
--- a/setup.sh
+++ b/setup.sh
@@ -41,10 +41,11 @@ declare -r DOCUMENTATION='documentation'
 
 declare -r CONFIGS_PATH='configs'
 
+declare -r DOCS_VIRTUAL_ENV='docs_virtual_env'
+
 function check_dependencies()
 {
   local package_list=''
-  local pip_package_list=''
   local cmd=''
   local distro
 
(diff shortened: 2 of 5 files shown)


setup.sh
```diff
@@ -90,24 +91,46 @@ function check_dependencies()
...
 
+function generate_documentation()
+{
+  local ret
+
+  python3 -m venv "$DOCS_VENV"
+
+  source "${DOCS_VENV}/bin/activate"
+  install_pip_dependencies
+  ret=$?
```
rodrigosiqueira, line 121: Add double quotes around "$?".

setup.sh
```diff
@@ -331,8 +354,18 @@ function synchronize_files()
 
   # man file
   mkdir -p "$mandir"
+
+  python3 -m venv "$DOCS_VENV"
+
+  source "${DOCS_VENV}/bin/activate"
+  install_pip_dependencies
+  ret=$?
```
rodrigosiqueira, line 362: Double-quotes.
Do you need it? If so, you must add a check for the return.

siddhanthrathod:
This is great for operating systems as gentoo which don't allow pip to install packages system wide and --user flag or a --target is needed !

documentation/dependencies/arch.dependencies
```diff
@@ -1,22 +1,18 @@
...
 graphviz
-python-virtualenv
 texlive-bin
 librsvg
 bzip2
 lzip
 lzop
 zstd
 xz
-python-pip
```
rodrigosiqueira, line 19: Maybe I'm missing something, but if we remove `python-pip` how can you call `pip` in a fresh env that does not have `pip` installed by default?

setup.sh
```diff
@@ -41,10 +41,11 @@ declare -r DOCUMENTATION='documentation'
 
 declare -r CONFIGS_PATH='configs'
 
+declare -r DOCS_VENV='docs_venv'
```
rodrigosiqueira, line 44: What the `V` in `VENV`  means? Can we have a more readable name?

setup.sh
```diff
@@ -90,22 +91,32 @@ function check_dependencies()
...
+
+  python3 -m venv "$DOCS_VENV"
+
+  # Activate python virtual env
+  source "${DOCS_VENV}/bin/activate"
+  say 'Creating python virtual env...'
+  cmd="pip --quiet --require-virtualenv install --requirement \"${DOCUMENTATION}/dependencies/pip.dependencies\""
+  eval "$cmd"
+  ret="$?"
+  if [[ "$ret" == 0 ]]; then
```
rodrigosiqueira, line 108: Add a new line before the `if` condition.

setup.sh
```diff
@@ -90,22 +91,32 @@ function check_dependencies()
...
+    complain 'Could not install pip dependencies'
+  fi
+  # Deactivate python virtual env
+  deactivate
 
-    # Install pip packages
-    cmd="pip install $pip_package_list"
-    eval "$cmd"
+  if [[ -d "$DOCS_VENV" ]]; then
+    rm -r "$DOCS_VENV"
```
rodrigosiqueira, line 117: Use `--recursive` instead of `-r`.

rodrigosiqueira (changes requested):

documentation/dependencies/arch.dependencies
Kamorst: `python3 -m venv` will install pip inside the virtual env. So the host do not need to install it system-wide.

setup.sh
Kamorst: `V` stands for "virtual". I will change it.

Issue #796, opened 2023-03-10
Implement the bookmark feature


rodrigosiqueira:
**Description:**

In the `upstream-patches-ui`, there is the `bookmarked patches`, which is not implemented yet; it just has some dummy data. Ideally, this feature should work like this:

* User navigates through patches.
* When users select one specific patch, the bookmark option is available.
* If the user marks this option, kw must download the patch in the local machine to be accessed later on.
* Users who access the bookmark option should see the previously saved patches.
* In the bookmark screen, users should also be able to remove a patch from the bookmarked list.

rodrigosiqueira:
@davidbtadokoro

PR #808, opened 2023-03-14
src: upstream_patches_ui: Add bookmark feature

davidbtadokoro:
This commit adds a barebones implementation of the bookmark feature. This feature refers to the ability to choose a certain patch series from a public mailing list and save it to later consultation. After bookmarking a patch series, it will become available in the 'Bookmarked Patches' screen. At the moment, the only action available is to 'Unbookmark' a patch series, but it should be kind of simple to plug in new actions.

From the implementation point of view, kw needs to maintain a local database of all bookmarked patch series and this is done by creating and managing a file inside the kw data directory (much like kw kernel-config-manager). Although efficient and simple to manage, if the feature needs scaling (change/add entities/fields), a real DBMS should be considered. Also worth mentioning that, under the hood, bookmarking a patch series downloads it first in the same manner of the 'Download' action, i.e., saves the .mbox files inside a directory defined by the user in lore.config. Unbookmarking patches series also removes them first from local storage.

Closes: #796

### DIFF
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -680,3 +680,21 @@ function get_current_env_name()
 
   return "$ret"
 }
+
+# A common task is to remove files/directories. This function is a predicate
+# to check if a given path is safe to remove (e.g. is not the '/')
+#
+# @path: Path to be checked
+#
+# Return:
+# Returns 0 if the path is safe to remove, 1 if it is not and 2 if it doesn't
+# exists.
+function is_safe_path_to_remove()
+{
+  local path="$1"
+
+  [[ ! -e "$path" ]] && return 2   # ENOENT
+  [[ "$path" == '/' ]] && return 1 # EPERM
+
+  return 0
+}
--- a/src/lib/lore.sh
+++ b/src/lib/lore.sh
@@ -679,3 +679,33 @@ function parse_raw_series()
   _series['patch_id']="${columns[7]}"
   _series['timestamp']="${columns[8]}"
 }
+
+# This function is a predicate about the existence of given patch in the local
+# bookmark database.
+#
+# @target_patch: The patch metadata. Used to get the target patch id.
+#
+# Return:
+# Returns 0 if given patch is present in local database, 1 if it is not and 2 if
+# there is no local database.
+#
+# TODO:
+# - Revise the return value of 1.
+function is_bookmarked()
+{
+  local target_patch="$1"
+  local patch_id
+  local count
+
+  if [[ ! -f "${BOOKMARKED_SERIES_PATH}" ]]; then
+    return 2 # ENOENT
+  fi
+
+  patch_id=$(printf '%s' "${target_patch}" | sha256sum | cut -d ' ' -f1)
+  count=$(grep --count "${patch_id}" "${BOOKMARKED_SERIES_PATH}")
+  if [[ "$count" != 0 ]]; then
+    return 0
+  fi
+
+  return 1
+}
--- a/src/lib/lore.sh
+++ b/src/lib/lore.sh
@@ -18,6 +18,15 @@ declare -gr MAILING_LISTS_PAGE='lore_main_page.html'
 # Path to mailing list file to be parsed
 declare -g LIST_PAGE_PATH="${CACHE_LORE_DIR}/${MAILING_LISTS_PAGE}"
 
+# Directory for storing every data related to lore
+declare -g LORE_DATA_DIR="${KW_DATA_DIR}/lore"
+
+# File name for the lore bookmarked series
+declare -gr LORE_BOOKMARKED_SERIES='lore_bookmarked_series'
+
+# Path to bookmarked series file
+declare -g BOOKMARKED_SERIES_PATH="${LORE_DATA_DIR}/${LORE_BOOKMARKED_SERIES}"
+
 # List of lore mailing list tracked by the user
 declare -gA available_lore_mailing_lists
 
@@ -35,6 +44,26 @@ declare -gr SEPARATOR_CHAR='Æ'
 # can be a ',' but by default, we use 'Æ'. We used ',' in the example for make
 # it easy to undertand.
 declare -ag list_of_mailinglist_patches
+
+# This function creates the directory used by kw for any lore related data.
+#
+# Return:
+# Returns 0 if the lore data directory was created successfully and the failing
+# status code otherwise (probably 111 EACCESS).
+function create_lore_data_dir()
+{
+  local ret
+
+  [[ -d "${LORE_DATA_DIR}" ]] && return
+
+  mkdir -p "${LORE_DATA_DIR}"
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain "Could not create lore data dir in ${LORE_DATA_DIR}"
+  fi
+
+  return "$ret"
+}
 
 function setup_cache()
 {
@@ -360,6 +389,7 @@ function get_patches_from_mailing_list()
 {
   local target_mailing_list="$1"
   local -n _dialog_array="$2"
+  declare -A series
   local raw_string
   local count=1
   local index=0
@@ -373,15 +403,9 @@ function get_patches_from_mailing_list()
   processing_new_patches "$target_mailing_list"
 
   # Format data for printing
-  for element in "${list_of_mailinglist_patches[@]}"; do
-    IFS="${SEPARATOR_CHAR}" read -r -a columns <<< "$element"
-    patch_version="${columns[2]}"
-    total_patches="${columns[3]}"
-    patch_title="${columns[4]}"
-
-    # convert_title_to_patch_name "$patch_title"
-
-    tmp_data=$(printf 'V%-2s |#%-3s| %-100s' "$patch_version" "$total_patches" "$patch_title")
+  for raw_series in "${list_of_mailinglist_patches[@]}"; do
+    parse_raw_series "${raw_series}" 'series'
+    tmp_data=$(printf 'V%-2s |#%-3s| %-100s' "${series['patch_version']}" "${series['total_patches']}" "${series['patch_title']}")
     _dialog_array["$index"]="$tmp_data"
     ((index++))
   done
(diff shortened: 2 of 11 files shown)


src/lib/lore.sh
```diff
@@ -18,6 +18,12 @@ declare -gr MAILING_LISTS_PAGE='lore_main_page.html'
 # Path to mailing list file to be parsed
 declare -g LIST_PAGE_PATH="${CACHE_LORE_DIR}/${MAILING_LISTS_PAGE}"
 
+# File name for the lore bookmarked series
+declare -gr LORE_BOOKMARKED_SERIES='lore_bookmarked_series'
+
+# Path to bookmarked series file
+declare -g BOOKMARKED_SERIES_PATH="${KW_DATA_DIR}/${LORE_BOOKMARKED_SERIES}"
```
rodrigosiqueira, line 25: Add a new folder named `upstream_patches_ui`, i.e, `"${KW_DATA_DIR}/upstream_patches_ui/${LORE_BOOKMARKED_SERIES}"`.

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+# @patch_title: The title of the patch
+# @flag: Flag to control function output
+function delete_series_from_local_storage()
+{
+  local download_dir_path="$1"
+  local patch_title="$2"
+  local flag="$3"
+  local patch_file_name
+  local file_pattern
+  local command
```
rodrigosiqueira, line 480: `command` is a bash built-in; I recommend you use something like `cmd` instead.

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+
+  flag=${flag:-'SILENT'}
+
+  patch_file_name=$(convert_title_to_patch_name "$patch_title")
+
+  file_pattern=$(printf '%s/*%s' "${download_dir_path}" "${patch_file_name}")
+
+  ls "${file_pattern}" 1> /dev/null 2>&1
```
rodrigosiqueira, line 488: `ls` command behavior may vary across distros; replace it with the `find` command or another option.

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+
+  if [[ "$ret" == 0 ]]; then
+    command="rm ${file_pattern}"
```
rodrigosiqueira, line 491: Always add a check before trying to run the `rm` command; that way you can ensure that the bash variable does not have anything like `/` or `/home/something`.

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+
+  flag=${flag:-'SILENT'}
+
+  patch_file_name=$(convert_title_to_patch_name "$patch_title")
+
+  file_pattern=$(printf '%s/*%s' "${download_dir_path}" "${patch_file_name}")
+
+  ls "${file_pattern}" 1> /dev/null 2>&1
+
+  if [[ "$ret" == 0 ]]; then
```
rodrigosiqueira, line 490: `ret`? You did not initialize this variable. Did I miss something?

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+    command="rm ${file_pattern}"
+    cmd_manager "$flag" "$command"
+  else
+    return 2 # ENOENT
```
rodrigosiqueira, line 494: You forgot to document this function return.

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+  fi
+}
+
+# This function creates the lore bookmarked series file if it doesn't
+# already exists
+function create_lore_bookmarked_file()
+{
+  if [[ ! -f "${BOOKMARKED_SERIES_PATH}" ]]; then
+    touch "${BOOKMARKED_SERIES_PATH}"
+  fi
```
rodrigosiqueira, line 504: Maybe you can use something like this:

```bash
  [[ -f "${BOOKMARKED_SERIES_PATH}" ]] && return

  touch "${BOOKMARKED_SERIES_PATH}"
```

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+}
+
+# This function adds a given series to a local bookmark database for
+# series managed by kw. To achieve uniqueness for each series the function
+# assigns an id to it. Also a timestamp for informational purposes is also
```
rodrigosiqueira, line 509: /an id to it. Also a timestamp/an ID and timestamp/

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+  local download_dir_path="$2"
+  local patch_id
+  local timestamp
+
+  create_lore_bookmarked_file
+
+  patch_id=$(printf '%s' "${target_patch}" | sha256sum | cut -d ' ' -f1)
+  timestamp=$(date '+%Y/%m/%d %H:%M')
+
+  if [[ $(grep -c "^${patch_id}" "${BOOKMARKED_SERIES_PATH}") == 0 ]]; then
```
rodrigosiqueira, line 527: Use `--count` instead of `-c`.
Additionally, I would remove the `grep` command from the `if` condition, and I would just use `"$?" == 0`.

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+}
+
+# This function removes a series from the local bookmark database by its index
+# in the database.
+#
+# @series_index: The index in the local bookmark database
+#
+# TODO:
+# - Find an alternative way to identify a series, this one may not be the most
+#   reliable.
```
rodrigosiqueira, line 544: The SQLite work will be very helpful for this.

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+function remove_series_from_bookmark_by_index()
+{
+  local series_index="$1"
+
+  if [[ ! -f "${BOOKMARKED_SERIES_PATH}" ]]; then
+    return 2 # ENOENT
```
rodrigosiqueira, line 550: Don't forget to document returns.

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+  fi
+
+  sed -i "${series_index}d" "${BOOKMARKED_SERIES_PATH}"
```
rodrigosiqueira, line 553: Use `--in-place` instead of `-i`.

src/lib/lore.sh
```diff
@@ -458,3 +464,152 @@ function download_series()
...
+    return 2 # ENOENT
+  fi
+
+  _bookmarked_series=()
+
+  while IFS='' read -r series; do
+    IFS="${SEPARATOR_CHAR}" read -r -a columns <<< "$series"
+    timestamp="${columns[1]}"
+    patch_title="${columns[6]}"
+    patch_author="${columns[2]}"
```
rodrigosiqueira, line 584: At some point, we need to replace this `${columns[1]}` in the entire code with something more meaningful. Using an index is really confusing... maybe we can have a set of functions for that?

tests/lib/lore_test.sh
```diff
@@ -14,6 +14,31 @@ function oneTimeSetUp()
...
+  cd "${SHUNIT_TMPDIR}" || {
+    fail "($LINENO): setUp(): It was not possible to move into ${SHUNIT_TMPDIR}"
+    return
+  }
+}
+
+function tearDown()
+{
+  # Empty file
+  : > "${BOOKMARKED_SERIES_PATH}"
```
rodrigosiqueira, line 34: You can delete this file in the `tearDown` instead of empty it; I believe this way is more readable.

tests/lib/lore_test.sh
```diff
@@ -170,4 +195,102 @@ function test_extract_metadata_from_patch_title_rfc()
   assertEquals "($LINENO)" "$output" "$expected"
 }
 
+function test_create_lore_bookmarked_file()
+{
+  rm "${BOOKMARKED_SERIES_PATH}"
```
rodrigosiqueira, line 200: Make sure `"${BOOKMARKED_SERIES_PATH}"` is the path that you want to remove before calling `rm`.

tests/lib/lore_test.sh
```diff
@@ -170,4 +195,102 @@ function test_extract_metadata_from_patch_title_rfc()
...
+{
+  local target_patch1="someseries1"
+  local target_patch2="someseries2"
+  local target_patch3="someseries3"
+  local download_dir_path="somedir"
+  local output
+  local count
+
+  add_series_to_bookmark "${target_patch1}" "${download_dir_path}"
+  output=$(cat "${BOOKMARKED_SERIES_PATH}")
```
rodrigosiqueira, line 217: You can use something more optimal than `cat` here; check https://kworkflow.org/content/codingstyle.html#command-substitution-and-arithmetic-expression .

tests/lib/lore_test.sh
```diff
@@ -170,4 +195,102 @@ function test_extract_metadata_from_patch_title_rfc()
...
+
+  assertTrue "($LINENO) - Local bookmark database was not created" "[[ -f ${BOOKMARKED_SERIES_PATH} ]]"
+}
+
+function test_add_series_to_bookmark()
+{
+  local target_patch1="someseries1"
+  local target_patch2="someseries2"
+  local target_patch3="someseries3"
+  local download_dir_path="somedir"
```
rodrigosiqueira, line 212: Single-quotes.

tests/lib/lore_test.sh
```diff
@@ -170,4 +195,102 @@ function test_extract_metadata_from_patch_title_rfc()
...
+  assertTrue "($LINENO) - Should add series to local bookmark database" "[[ \"$?\" == 0 ]]"
+
+  add_series_to_bookmark "${target_patch1}" "${download_dir_path}"
+  output=$(cat "${BOOKMARKED_SERIES_PATH}")
+  count=$(printf '%s' "$output" | grep --count "${target_patch1}${SEPARATOR_CHAR}${download_dir_path}")
+  assertEquals "($LINENO) - Should not duplicate entries " 1 "$count"
+
+  add_series_to_bookmark "${target_patch2}" "${download_dir_path}"
+  add_series_to_bookmark "${target_patch3}" "${download_dir_path}"
+  count=$(wc -l "${BOOKMARKED_SERIES_PATH}" | cut -d ' ' -f1)
```
rodrigosiqueira, line 228: /-l/--lines/ and /-d/--delimiter/

rodrigosiqueira (changes requested):
Hi,

Thanks a lot for this PR. I added some inline comments and follows some general comments:

* The UI has some hang if I try the bookmarked option without anything (e.g., the first time after installing kw). You need to verify if backport list is empty and just add a notification to the user about that and an option to return to the previous menu.
* After I bookmark a patch, it will have some loading period, and the bookmarked option will be empty again. After bookmarking, a patch kw must keep that option marked; otherwise, users may need clarification later by thinking they did not bookmark the specific path.

davidbtadokoro:
Hi @rodrigosiqueira ,
Sorry for the delay. Working on it right now

src/lib/lore.sh
davidbtadokoro: You are right. Even more considering that even in this same file they don't follow a pattern...

By a set a functions you are talking about something like `get_timestamp`, `get_patch_title` and so on to encapsulate this explicit `${columns[i]}`, right?

tests/lib/lore_test.sh
davidbtadokoro: Right! I chose this approach because an error was thrown when trying to delete it (cause the file didn't exist). Thinking now, it is most likely because of the command on the comment right below 🤣

tests/lib/lore_test.sh
davidbtadokoro: Nice!

davidbtadokoro:
This force push addresses the inline comments and the addition of a screen to handle the case when there were no bookmarked patches to be displayed.

Working on making the Bookmark check persistent.

davidbtadokoro:
This last commit addresses the persistence of checks of a patch (series).

src/lib/lore.sh
```diff
@@ -36,6 +45,13 @@ declare -gr SEPARATOR_CHAR='Æ'
 # it easy to undertand.
 declare -ag list_of_mailinglist_patches
 
+function create_lore_data_dir()
```
rodrigosiqueira, line 48: Add documentation.

src/lib/lore.sh
```diff
@@ -36,6 +45,13 @@ declare -gr SEPARATOR_CHAR='Æ'
...
+{
+  [[ -d "${LORE_DATA_DIR}" ]] && return
+
+  mkdir -p "${LORE_DATA_DIR}"
+}
```
rodrigosiqueira, line 53: I was wondering about this function... what if we cannot create LORE_DATA_DIR? Maybe we should add some extra checks for `mkdir` and handle the rare case where we cannot create this folder.

src/lib/lore.sh
```diff
@@ -458,3 +474,169 @@ function download_series()
...
+{
+  local target_patch="$1"
+  local download_dir_path="$2"
+  local patch_id
+  local timestamp
+  local count
+
+  create_lore_bookmarked_file
+
+  patch_id=$(printf '%s' "${target_patch}" | sha256sum | cut -d ' ' -f1)
```
rodrigosiqueira, line 546: Don't we need to add any specific package in the setup phase to ensure that sha256sum is available?

src/upstream_patches_ui.sh
```diff
@@ -115,8 +113,73 @@ function dashboard_entry_menu()
...
+  local message_box
+  local columns
+  local patch_title
+  local download_dir_path
+
+  action_list=('Unbookmark')
+
+  # The local bookmark database starting index is 1 and the index
+  # passed as argument starts at 0.
+  ((series_index++))
```
rodrigosiqueira, line 150: Just to be safe and based on your comment, I recommend you to check `series_index` before trying to increment it.

src/upstream_patches_ui.sh
```diff
@@ -204,7 +271,9 @@ function show_series_details()
       for option in "${selected_options[@]}"; do
         case "$option" in
           'Bookmark')
-            printf 'TODO' # TODO
+            create_loading_screen_notification "Bookmarking patch(es)"$'\n'"- ${patch_title}"
+            download_series "$total_patches" "$patch_url" "${lore_config['download_to']}" "$patch_title"
+            add_series_to_bookmark "${target_patch}" "${lore_config['download_to']}"
```
rodrigosiqueira, line 292: If any of these functions fails, you should check and provide a message to the user that lets them know that the operation failed for some reason.

src/upstream_patches_ui.sh
```diff
@@ -113,14 +113,19 @@ function dashboard_entry_menu()
...
+
   if [[ -z "${screen_sequence['RETURNING']}" ]]; then
     get_bookmarked_series bookmarked_series
   fi
 
   # Avoiding stale value
   screen_sequence['RETURNING']=''
 
-  list_patches 'Bookmarked patches' bookmarked_series "${screen_sequence['SHOW_SCREEN']}"
+  fallback_message='kw could not find any bookmarked patches.'$'\n'
```
rodrigosiqueira, line 125: Could you add another `\n` to add a blank line between this message and the next one?

src/upstream_patches_ui.sh
```diff
@@ -113,14 +113,19 @@ function dashboard_entry_menu()
...
+  fallback_message+='Try bookmarking patches in the menu "Registered mailing list"'
+  list_patches 'Bookmarked patches' bookmarked_series "${screen_sequence['SHOW_SCREEN']}" \
+    "${fallback_message}"
```
rodrigosiqueira, line 128: Please make this a single-line command; it is more readable.

rodrigosiqueira (changes requested):
Very nice PR!
I added some inline comments.

src/lib/lore.sh
```diff
@@ -458,3 +474,169 @@ function download_series()
...
+  filename_pattern=$(convert_title_to_patch_name "$patch_title")
+
+  # Find doesn't return a non-zero status code when not matching any file.
+  # We use grep to work-around it.
+  files_to_remove=$(find "${download_dir_path}" -iname "*${filename_pattern}" | grep '.')
+  ret="$?"
+
+  if [[ "$ret" == 0 ]]; then
+    while IFS=' ' read -r file; do
+      # Just to avoid trying to remove fs root
```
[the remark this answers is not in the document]
davidbtadokoro: Nice! Yes, this is a fairly common task in the project 😆

src/kwlib.sh
```diff
@@ -680,3 +680,21 @@ function get_current_env_name()
...
+#
+# Return:
+# Returns 0 if the path is safe to remove, 1 if it is not and 2 if it doesn't
+# exists.
+function is_safe_path_to_remove()
+{
+  local path="$1"
+
+  [[ ! -e "$path" ]] && return 2 # ENOENT
+  [[ "$path" == '/' ]] && return 1
```
rodrigosiqueira, line 697: Please, document 1 as EPERM.

src/lib/lore.sh
```diff
@@ -458,3 +487,167 @@ function download_series()
...
+  else
+    return 2 # ENOENT
+  fi
+}
+
+# This function creates the lore bookmarked series file if it doesn't
+# already exists
+function create_lore_bookmarked_file()
+{
+  create_lore_data_dir
```
rodrigosiqueira, line 534: In this new implementation, the above function can fail. Please check if it succeeds before creating a new file.

src/upstream_patches_ui.sh
```diff
@@ -115,8 +113,78 @@ function dashboard_entry_menu()
...
+  ret="$?"
+
+  case "$ret" in
+    0) # OK
+      IFS=' ' read -ra selected_options <<< "$menu_return_string"
+      for option in "${selected_options[@]}"; do
+        case "$option" in
+          'Unbookmark')
+            delete_series_from_local_storage "${download_dir_path}" "${patch_title}"
+            remove_series_from_bookmark_by_index "${series_index}"
```
rodrigosiqueira, line 179: Correct me if I'm wrong, but these functions can fail. If so, could you please check the return value? I think we need to notify users in case of an error.

src/upstream_patches_ui.sh
rodrigosiqueira: +1 again in this comment.

src/upstream_patches_ui.sh
```diff
@@ -276,13 +282,19 @@ function show_series_details()
...
-            create_loading_screen_notification "Bookmarking patch(es)"$'\n'"- ${patch_title}"
-            download_series "$total_patches" "$patch_url" "${lore_config['download_to']}" "$patch_title"
-            add_series_to_bookmark "${target_patch}" "${lore_config['download_to']}"
+            create_loading_screen_notification 'Bookmarking patch(es)'$'\n'"- ${patch_title}"
+            {
+              download_series "$total_patches" "$patch_url" "${lore_config['download_to']}" "$patch_title" &&
+                add_series_to_bookmark "${target_patch}" "${lore_config['download_to']}"
+            } || {
+              create_message_box 'Error' 'Could not bookmark patch(es)'$'\n'"- ${patch_title}"
+            }
```
rodrigosiqueira, line 300: Please ignore my comments; now I can see that you are checking the return values. However, could you use a simple `if` condition here? It should be more readable and aligned with the code style.

src/upstream_patches_ui.sh
```diff
@@ -276,13 +282,19 @@ function show_series_details()
...
             ;;
           'Download')
-            create_loading_screen_notification "Downloading patch(es)"$'\n'"- ${patch_title}"
-            download_series "$total_patches" "$patch_url" "${lore_config['download_to']}" "$patch_title"
+            create_loading_screen_notification 'Downloading patch(es)'$'\n'"- ${patch_title}"
+            download_series "$total_patches" "$patch_url" "${lore_config['download_to']}" "$patch_title" || {
+              create_message_box 'Error' 'Could not download patch(es)'$'\n'"- ${patch_title}"
+            }
```
rodrigosiqueira, line 306: Same here... can we use `if`?

src/upstream_patches_ui.sh
```diff
@@ -305,15 +317,24 @@ function show_series_details()
...
+  local fallback_message="$4"
   local selected_patch
   local ret
 
-  create_menu_options "${menu_title}" '' '_target_array_list' 1
+  if [[ -z "${_target_array_list}" ]]; then
+    create_message_box "${menu_title}" "${fallback_message}"
+    ret="$?"
+    screen_sequence['SHOW_SCREEN']='dashboard'
+    return "$ret"
```
rodrigosiqueira, line 343: Move `secreen_sequence`  right after the `if` condition, and remove the `ret` variable part in favor of just returning `$?`.

src/upstream_patches_ui.sh
```diff
@@ -305,15 +317,24 @@ function show_series_details()
...
+  else
+    create_menu_options "${menu_title}" '' '_target_array_list' 1
+  fi
```
rodrigosiqueira, line 346: Since you added a return in the `if` condition, you don't need this else. You can call `create_menu_option` directly.

rodrigosiqueira (changes requested):

src/lib/lore.sh
davidbtadokoro: For sure. The issue regarding the integration BD. Going to implement these functions.

src/upstream_patches_ui.sh
davidbtadokoro: Sorry, I didn't understand the move of `screen_sequence`. When `list_patches` doesn't have patches to list, after the notification screen, the next screen is Dashboard. Otherwise, the next screen will be determined in the switch-case below. Moving `screen_sequence` after the `if` wouldn't be redundant?

rodrigosiqueira:
Very nice set of changes! Thanks a lot for all of these improvements.

Issue #805, opened 2023-03-11
Wrong message box on Dashboard screen


davidbtadokoro:
**Description:**

The message box on the Dashboard screen of `kw upstream-patches-ui` seems to be same of the mailing list registering screen.

**Expected behavior:**

The message box should be related with Dashboard screen.

**Screenshots:**

<img width="1195" alt="Screen Shot 2023-03-11 at 16 43 08" src="https://user-images.githubusercontent.com/95055784/224508571-3cfde4d5-8ab2-4d51-8434-74b52b1366f4.png">

PR #812, opened 2023-03-16
src: upstream_patches_ui: Fix Dashboard screen message box

davidbtadokoro:
The Dashboard screen message box isn't related to the screen. This commit fixes this by making the message box empty (no message), as the title 'Dashboard' is already indicative and this is a generic screen that assumes the menu options are self-descriptive. Also declared the variable used 'message_box' for safety keeping.

Closes: #805


rodrigosiqueira:
Merged in the unstable branch.
Thanks

PR #813, opened 2023-03-16
Fix kw init wrong template name

### LINKED ISSUE #748: kw init does not handle wrong template name correctly
**Description:**

If users try the wrong template name, it fails with a misleading message.

**How to Reproduce:**

1. Cd to a kernel repo
2. Try the below command and see the error:
```
kw init --template="x86_64"
No such: ~/.config/kw/init_templates/:x86_64/kworkflow_template.config
```

**Expected behavior:**

Do not show the wrong path because it adds an `:` to it, which can mislead users. Show all the available template names.

Note: I'm not sure why we add `:`; we need to investigate it and see if we can find a better solution.

Kamorst:
Do not prepend ':' to the template name. It was necessary to set a
default template if the option `--template` is not used. This is not
necessary anymore.

Closes #748

### DIFF
--- a/src/init.sh
+++ b/src/init.sh
@@ -47,6 +47,11 @@ function init_kw()
     return 22 # EINVAL
   fi
 
+  if [[ -n "${options_values['VERBOSE']}" ]]; then
+    flag='VERBOSE'
+  fi
+  flag=${flag:-'SILENT'}
+
   config_file_already_exist_question
 
   get_template_name ''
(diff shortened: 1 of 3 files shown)


src/init.sh
```diff
@@ -68,57 +68,56 @@ function init_kw()
...
-        set_config_value 'arch' "${options_values['ARCH']}" "${PWD}/${KW_DIR}/${build_name}"
-      elif [[ -z "${options_values['FORCE']}" ]]; then
-        complain 'This arch was not found in the arch directory'
-        complain 'You can use --force next time if you want to proceed anyway'
-        say 'Available architectures:'
-        find "$PWD/arch" -maxdepth 1 -mindepth 1 -type d -printf '%f\n' | sort -d
-      fi
-    fi
+  if [[ ! -f "$config_file_template" || ! -f "$build_config_file_template" ]]; then
+    complain "No such: $config_file_template"
```
rodrigosiqueira, line 68: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/init.sh
```diff
@@ -68,57 +68,56 @@ function init_kw()
...
-        set_config_value 'ssh_port' "${remote_parameters['REMOTE_PORT']}"
-      fi
+  mkdir -p "$PWD/$KW_DIR"
+  cp "$config_file_template" "$PWD/$KW_DIR/$name"
+  cp "$vm_config_file_template" "${PWD}/${KW_DIR}/${vm_name}"
+  cp "$build_config_file_template" "${PWD}/${KW_DIR}/${build_name}"
+  cp "$deploy_config_file_template" "${PWD}/${KW_DIR}/${deploy_name}"
+  cp "$mail_config_file_template" "${PWD}/${KW_DIR}/${mail_name}"
+  cp "$notification_config_file_template" "${PWD}/${KW_DIR}/${notification_name}"
+  cp "$remote_file_template" "${PWD}/${KW_DIR}/${remote_name}"
```
rodrigosiqueira, line 79: Could you create another commit that addresses these comments:
1. Use string concatenation.
2. Use cmd_manager instead of the command.
3. Enable verbose mode.

Or, at least, create an issue for tracking these requests.

P.s.: This comment is valid for the entire file.

src/init.sh
```diff
@@ -68,57 +68,56 @@ function init_kw()
...
+
+  sed -i -e "s/USERKW/$USER/g" -e '/^#?.*/d' "$PWD/$KW_DIR/${vm_name}"
+  sed -i -e "s,SOUNDPATH,$KW_SOUND_DIR,g" -e '/^#?.*/d' "$PWD/$KW_DIR/${notification_name}"
```
rodrigosiqueira, line 82: Use the full parameter name.

rodrigosiqueira (changes requested):

src/init.sh
```diff
@@ -47,17 +47,18 @@ function init_kw()
     return 22 # EINVAL
   fi
 
+  if [[ -n "${options_values['VERBOSE']}" ]]; then
+    flag='VERBOSE'
+  fi
+  flag=${flag:-'SILENT'}
```
rodrigosiqueira, line 53: This should be set before the condition that checks if the use used the `-h` option.

src/init.sh
```diff
@@ -47,17 +47,18 @@ function init_kw()
     return 22 # EINVAL
   fi
 
+  if [[ -n "${options_values['VERBOSE']}" ]]; then
+    flag='VERBOSE'
+  fi
```
rodrigosiqueira, line 52: How about just:
``[[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'``

src/init.sh
```diff
@@ -68,60 +69,69 @@ function init_kw()
...
+  cmd="cp \"$build_config_file_template\" \"${PWD}/${KW_DIR}/${build_name}\""
+  cmd_manager "$flag" "$cmd"
+  cmd="cp \"$deploy_config_file_template\" \"${PWD}/${KW_DIR}/${deploy_name}\""
+  cmd_manager "$flag" "$cmd"
+  cmd="cp \"$mail_config_file_template\" \"${PWD}/${KW_DIR}/${mail_name}\""
+  cmd_manager "$flag" "$cmd"
+  cmd="cp \"$notification_config_file_template\" \"${PWD}/${KW_DIR}/${notification_name}\""
+  cmd_manager "$flag" "$cmd"
+  cmd="cp \"$remote_file_template\" \"${PWD}/${KW_DIR}/${remote_name}\""
+  cmd_manager "$flag" "$cmd"
```
rodrigosiqueira, line 92: You don't need the `cmd` variable here; add the string directly to the `cmd_manager` parameter.

src/init.sh
```diff
@@ -68,60 +69,69 @@ function init_kw()
...
+
+  cmd="sed --in-place --expression \"s/USERKW/${USER}/g\" --expression '/^#?.*/d' \"${PWD}/${KW_DIR}/${vm_name}\""
+  cmd_manager "$flag" "$cmd"
+  cmd="sed --in-place --expression \"s,SOUNDPATH,${KW_SOUND_DIR},g\" --expression '/^#?.*/d' \"${PWD}/${KW_DIR}/${notification_name}\""
+  cmd_manager "$flag" "$cmd"
```
rodrigosiqueira, line 97: Same as the previous comment.

src/init.sh
```diff
@@ -68,60 +69,69 @@ function init_kw()
...
-            "${PWD}/${KW_DIR}/${deploy_name}"
-          ;;
-        *)
-          complain 'Target can only be vm, local or remote.'
-          ;;
-      esac
+  if [[ -n "${options_values['REMOTE']}" ]]; then
+    populate_remote_info "${options_values['REMOTE']}"
+    if [[ "$?" == 22 ]]; then
+      complain 'Invalid remote:' "${options_values['REMOTE']}"
```
rodrigosiqueira, line 113: Add space after `:`.

rodrigosiqueira (changes requested):

PR #814, opened 2023-03-16
src: build: refactoring some options.

aquilamacedo:
This Pull Request refactors the `build --doc` and `build --menuconfig` options
from the build. The env path problem in `build --clean` is also fixed


rodrigosiqueira:
Merged,
Thanks

Issue #554, opened 2022-01-29
Improve assert_equal_help output


rodrigosiqueira:
**Description:**

Let's make `assert_equal_help` display the failure case as:

 Expected Result: <something>
 ->Actual Result: <something>

Maybe we can show the expected result in blue colors and the failure in red.

**Type:** Refactor

**Why:**

Currently, `assert_equal_help` shows the actual output and the expected result next to each other, which is not friendly.

RaviRahar:
@rodrigosiqueira This is my first contribution here. does this seem alright?

PR #818, opened 2023-03-18
tests: utils: Improve assert_equal_help output

RaviRahar:
Earlier, assert_equal_help showed the actual output and expected result next to each other:

ASSERT: line (128): expected:<31> but was:<32>

Now it prints in new lines. New format is:

ASSERT: line (128): Use this year
Expected Result: 31
-> Actual Result: 32

Also the color of Expected result is set to blue and Actual Result to Red

Closes: #554

### DIFF
--- a/tests/utils.sh
+++ b/tests/utils.sh
@@ -376,18 +376,25 @@ function assert_no_line_match()
 # @result_to_compare Raw output to be compared
 function assert_equals_helper()
 {
+  # colors to use to print error
+  local color_none='\033[0m'
+  local color_red='\033[1;31m'
+  local color_green='\033[1;32m'
+
   local msg="$1"
   local line="$2"
-  # See bugs section in github.com/koalaman/shellcheck/wiki/SC2178
-  # shellcheck disable=SC2178
   local expected="$3"
   local result_to_compare="$4"
 
   line=${line:-'Unknown line'}
 
-  # See bugs section in github.com/koalaman/shellcheck/wiki/SC2178
-  # shellcheck disable=2128
-  assertEquals "line $line: $msg" "$result_to_compare" "$expected"
+  if ! assertEquals "$expected" "$result_to_compare" &> /dev/null; then
+    printf '%bASSERT:%b line %s: %s\n  %bExpected Result:%b %b%s%b\n  %b-> Actual Result:%b %b%s%b\n' \
+      "${color_red}" "${color_none}" "$line" "$msg" \
+      "${color_green}" "${color_none}" "${color_green}" "$expected" "${color_none}" \
+      "${color_red}" "${color_none}" "${color_red}" "$result_to_compare" "${color_none}"
+    return "$?"
+  fi
 }
 
 # Create an invalid file path


tests/utils.sh
```diff
@@ -376,18 +376,27 @@ function assert_no_line_match()
 # @result_to_compare Raw output to be compared
 function assert_equals_helper()
 {
+  # colors to use to print error
+  local _none='\033[0m'
+  local _red='\033[1;31m'
+  local _blue='\033[1;34m'
```
rodrigosiqueira, line 382: Let's use green for the expected result.

tests/utils.sh
```diff
@@ -376,18 +376,27 @@ function assert_no_line_match()
...
 
   line=${line:-'Unknown line'}
 
-  # See bugs section in github.com/koalaman/shellcheck/wiki/SC2178
-  # shellcheck disable=2128
-  assertEquals "line $line: $msg" "$result_to_compare" "$expected"
+  if ! assertEquals "$expected" "$result_to_compare" &> /dev/null; then
+    printf '%bASSERT:%b line %s: %s
+%bExpected Result:%b %b%s%b 
+%b-> Actual Result:%b %b%s%b\n' \
```
rodrigosiqueira, line 394: Add two extra spaces, something like this:
```
ASSERT: line (372): Default ARCH did not match expectation
  Expected Result: x86_4 
  -> Actual Result: x86_66
```

rodrigosiqueira: Also, could you keep the first parameter of `printf` in the same line?

tests/utils.sh
```diff
@@ -376,18 +376,27 @@ function assert_no_line_match()
...
+      "${_red}" "${_none}" "$line" "$msg" \
+      "${_blue}" "${_none}" "${_blue}" "$expected" "${_none}" \
+      "${_red}" "${_none}" "${_red}" "$result_to_compare" "${_none}"
+    return $?
```
rodrigosiqueira, line 398: Add double-quotes around $?

rodrigosiqueira (changes requested):
Very nice PR!
Could you also add colors to `compare_command_sequence` in another PR?

tests/utils.sh
RaviRahar: I thought it would be more readable. I saw it in compare_command_sequence<br>  Changed it now to be in one line.

RaviRahar:
> Very nice PR! Could you also add colors to `compare_command_sequence` in another PR?

done #820

tests/utils.sh
```diff
@@ -376,18 +376,25 @@ function assert_no_line_match()
 # @result_to_compare Raw output to be compared
 function assert_equals_helper()
 {
+  # colors to use to print error
+  local _none='\033[0m'
+  local _red='\033[1;31m'
+  local _green='\033[1;32m'
```
rodrigosiqueira, line 382: I forgot to ask, but why add `_`? We usually use it when dealing with references (i.e., `local -n _variable`). If there is no particular reason, please, rename these variables to not use `_`.

RaviRahar: I myself was going to rename them later, just forgot. Corrected.

rodrigosiqueira:
Merged in the unstable branch! Thanks!

RaviRahar:
@rodrigosiqueira I can see that the changes have been merged, but why does github still show unmerged though?

rodrigosiqueira:
@RaviRahar , if you see your change in the unstable branch, it was merged.

I don't like the GitHub workflow, and for this reason, I don't follow it. The closer/merge PR status does not mean anything... if your change is in the right branch, don't worry. Please, refer to https://kworkflow.org/content/howtocontribute.html

PR #820, opened 2023-03-18
tests: utils: Improve compare_command_sequence output

RaviRahar:
Earlier, compare_command_sequence showed the actual output and expected result
in following format:

        ASSERT:line 140, statement 1:
        Expected: ""
        but got:  "Build             19 00:03:40 00:00:06 00:00:23"

Now it prints in slightly changed format:

        ASSERT: line 140, statement 1:
          Expected: ""
          -> but got: "Build             19 00:03:40 00:00:06 00:00:23"

Also the color of Expected result is set to green and the line with actual
result is set to red

### DIFF
--- a/tests/utils.sh
+++ b/tests/utils.sh
@@ -33,6 +33,11 @@ LOCAL_TO_DEPLOY_DIR="to_deploy"
 LOCAL_REMOTE_DIR="remote"
 KERNEL_INSTALL_PLUGIN_PATH="src/plugins/kernel_install/"
 
+# Colors
+readonly KW_COLOR_NONE='\033[0m'
+readonly KW_COLOR_RED='\033[1;31m'
+readonly KW_COLOR_GREEN='\033[1;32m'
+
 function init_env()
 {
   unset -v KW_LIB_DIR KWORKFLOW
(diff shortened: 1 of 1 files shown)


tests/utils.sh
```diff
@@ -295,6 +295,11 @@ function mk_fake_kw_env()
 # @result_to_compare A raw output from the string
 function compare_command_sequence()
 {
+  # colors to use to print failure
+  local color_none='\033[0m'
+  local color_red='\033[1;31m'
+  local color_green='\033[1;32m'
```
rodrigosiqueira, line 301: You already declared these variables in the `assert_equals_helper`; it is time to make them read-only and global to this file to avoid code duplication.

rodrigosiqueira (changes requested):

tests/utils.sh
```diff
@@ -33,6 +33,11 @@ LOCAL_TO_DEPLOY_DIR="to_deploy"
 LOCAL_REMOTE_DIR="remote"
 KERNEL_INSTALL_PLUGIN_PATH="src/plugins/kernel_install/"
 
+# Colors
+KW_COLOR_NONE='\033[0m'
+KW_COLOR_RED='\033[1;31m'
+KW_COLOR_GREEN='\033[1;32m'
```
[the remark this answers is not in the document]
RaviRahar: which is preferred declare -r or readonly. I know there is no difference when using in global scope.

tests/utils.sh
```diff
@@ -304,10 +309,12 @@ function compare_command_sequence()
 
   while read -r f; do
     if [[ "${expected_res[$count]}" != "${f}" ]]; then
-      fail "line $line, statement $count: $msg
-Expected: \"${expected_res[$count]}\"
-but got:  \"${f}\"
-"
+      if ! fail &> /dev/null; then
+        printf '%bASSERT:%b line %s, statement %d: %s\n  %bExpected:%b %b\"%s\"%b\n  %b-> but got:%b %b\"%s\"%b\n' \
+          "${KW_COLOR_RED}" "${KW_COLOR_NONE}" "$line" "$count" "${msg}" \
```
kwy95, line 314: You don't need to use curly braces in these variables that have simple names, prefer not using them in this case.

kwy95 (changes requested):

tests/utils.sh
RaviRahar: you're right. removed.

rodrigosiqueira:
Hi @RaviRahar I forced some errors in the deploy test to see the `compare_command_sequence` and `assert_equals_helper` fails; however, it does not get red in my terminal. Could you check that?

Screenshot from 2023-04-01 10-34-22

RaviRahar:
> Hi @RaviRahar I forced some errors in the deploy test to see the `compare_command_sequence` and `assert_equals_helper` fails; however, it does not get red in my terminal. Could you check that?
> 
> Screenshot from 2023-04-01 10-34-22

This was because I used `declare -r` and not `declare -rg`. Declare makes variables to be of local scope if sourced inside a function and because we use include to source files, the variables became local. Now I have used readonly and it keeps scope of variables to be global.

Issue #801, opened 2023-03-10
Integrate kw build with upstream-patches-ui


rodrigosiqueira:
kw already supports build; let's add integration between upstream-patches-ui where developers can build an upstream patch.

PR #823, opened 2023-03-20
Integrate `kw upstream-patches-ui` with `kw build`

davidbtadokoro:
This PR is for implementing the issue #801 , i.e., integrate `upstream-patches-ui` with `build` to let the user apply and build patches (series) from the lore interface.

At the moment this is just a draft and only implements a straightfoward way to apply a patch to a given kernel tree source and target branch (also, without tests, documentation, error handling and safety checks).

### DIFF
--- a/etc/lore.config
+++ b/etc/lore.config
@@ -3,3 +3,5 @@ default_ui=dialog
 dialog_layout=black_and_white
 lists=
 download_to=/tmp/kw_download
+kernel_tree=
+target_branch=
--- a/src/lib/lore.sh
+++ b/src/lib/lore.sh
@@ -709,3 +709,41 @@ function is_bookmarked()
 
   return 1
 }
+
+function apply_patch()
+{
+  local total_patches="$1"
+  local patch_url="$2"
+  local patch_title="$3"
+  local kernel_tree="$4"
+  local target_branch="$5"
+  local flag="$6"
+  local path_to_tmp_dir
+  local filename_pattern
+  local files_to_remove
+  local cmd
+  local ret
+
+  path_to_tmp_dir=$(mktemp --directory)
+
+  download_series "${total_patches}" "${patch_url}" "${path_to_tmp_dir}" "${patch_title}"
+
+  cmd="git -C ${kernel_tree} switch ${target_branch}"
+  cmd_manager "$flag" "$cmd"
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain "Could not switch kernel tree ${kernel_tree} to branch ${target_branch}"
+    return "$ret"
+  fi
+
+  filename_pattern=$(convert_title_to_patch_name "${patch_title}")
+  files_to_apply=$(find "${path_to_tmp_dir}" -iname "*${filename_pattern}" | sort)
+  while IFS=' ' read -r file; do
+    cmd="git -C ${kernel_tree} am ${file}"
+    cmd_manager "$flag" "$cmd"
+    if [[ "$ret" != 0 ]]; then
+      complain "Could not apply patch ${file} kernel tree ${kernel_tree}"
+      return "$ret"
+    fi
+  done <<< "${files_to_apply}"
+}
--- a/src/upstream_patches_ui.sh
+++ b/src/upstream_patches_ui.sh
@@ -255,9 +255,10 @@ function show_series_details()
   local patch_metadata
   local raw_series
   local message_box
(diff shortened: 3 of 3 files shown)


etc/lore.config
```diff
@@ -3,3 +3,5 @@ default_ui=dialog
 dialog_layout=black_and_white
 lists=
 download_to=/tmp/kw_download
+kernel_tree=
+target_branch=
```
rodrigosiqueira, line 7: For some reason, when I introduced this file, I did not add proper documentation. Could you create one commit, adding the documentation to the previous and new options? See the below link as a reference:

https://github.com/kworkflow/kworkflow/blob/unstable/etc/mail.config

rodrigosiqueira: Also, let's call `kernel_tree_path`.

src/lib/lore.sh
```diff
@@ -458,3 +458,41 @@ function download_series()
   done
   wait
 }
+
```
rodrigosiqueira, line 461: Remember to document this function.

src/lib/lore.sh
```diff
@@ -458,3 +458,41 @@ function download_series()
...
+  local flag="$6"
+  local path_to_tmp_dir
+  local filename_pattern
+  local files_to_remove
+  local cmd
+  local ret
+
+  path_to_tmp_dir=$(mktemp --directory)
+
+  download_series "${total_patches}" "${patch_url}" "${path_to_tmp_dir}" "${patch_title}"
```
rodrigosiqueira, line 478: The download can fail; add a check for this case.

src/lib/lore.sh
```diff
@@ -458,3 +458,41 @@ function download_series()
...
+  cmd="git -C ${kernel_tree} switch ${target_branch}"
+  cmd_manager "$flag" "$cmd"
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain "Could not switch kernel tree ${kernel_tree} to branch ${target_branch}"
+    return "$ret"
+  fi
+
+  filename_pattern=$(convert_title_to_patch_name "${patch_title}")
+  files_to_apply=$(find "${path_to_tmp_dir}" -iname "*${filename_pattern}" | sort)
```
rodrigosiqueira, line 489: The default behavior of `sort` may vary across distros, using an explicit order, e.g.; `--dictionary-order`.

src/lib/lore.sh
```diff
@@ -458,3 +458,41 @@ function download_series()
...
+
+  cmd="git -C ${kernel_tree} switch ${target_branch}"
+  cmd_manager "$flag" "$cmd"
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain "Could not switch kernel tree ${kernel_tree} to branch ${target_branch}"
+    return "$ret"
+  fi
+
+  filename_pattern=$(convert_title_to_patch_name "${patch_title}")
```
rodrigosiqueira, line 488: `download_series` already call `convert_title_to_path_name`, we can refactor `download_series` to return the `converted_title_to_path_name` and remove this step. I.e., we would have `filename_path=$(download_series ...)`

src/lib/lore.sh
```diff
@@ -458,3 +458,41 @@ function download_series()
...
+  files_to_apply=$(find "${path_to_tmp_dir}" -iname "*${filename_pattern}" | sort)
+  while IFS=' ' read -r file; do
+    cmd="git -C ${kernel_tree} am ${file}"
```
rodrigosiqueira, line 491: Since `git am` can fail, we could use `--reject` to generate the `.rej` file and allow the developer to fix the issue manually. We can try our best to make this `am` very robust. One idea is to use the `wiggle` in case of git am failure for trying to apply the failed patch again.

We should update the target branch before trying to merge it. This can increase the chance of successfully applying a patch.

Another thing to consider is to create a new branch based on the `target_branch` to apply the new patch. We don't want to 'contaminate' the main branch with unmerged patches. You can create a pattern for those branches, maybe something like:

```
YYYY-MM-DD-HH-MM-SS-PATCH-TITLE-SHORT-SHA
```

Remember that applying patches can easily fail, and we must define some strategies. We could create a status system in the updtream-patches-ui to track if a patch failed to be applied.

1. https://linux.die.net/man/1/wiggle

rodrigosiqueira (changes requested):

davidbtadokoro:
Closing this draft to not clutter the PRs tab. Hope to reopen it In the near future.

PR #825, opened 2023-03-22
documentation: man: features: env: fixing typo in exit-env option name.

aquilamacedo:
The previous commit mistakenly referred to the exit-env option as `-n` instead of `-e` This commit corrects the error by renaming the option to `-e`.


rodrigosiqueira:
Merged! Thanks.

PR #838, opened 2023-04-26
Fix kw init wrong template name 

aquilamacedo:
Do not prepend ':' to the template name. Previously, it was required
to set a default template when the `--template` option was not used.
However, this is no longer necessary.

Closes https://github.com/kworkflow/kworkflow/issues/748

### DIFF
--- a/documentation/man/features/init.rst
+++ b/documentation/man/features/init.rst
@@ -6,8 +6,8 @@ kw-init
 
 SYNOPSIS
 ========
-| *kw* *init* [\--template[=name]]
-| *kw* *init* [\--arch <name>][\--remote <user>@<ip>:<port>][\--target <target>]
+| *kw* *init* [\--verbose] [\--template[=name]]
+| *kw* *init* [\--verbose] [\--arch <name>][\--remote <user>@<ip>:<port>][\--target <target>]
 
 DESCRIPTION
 ===========
@@ -36,6 +36,9 @@ OPTIONS
 \--target <target>:
   Set the variable `default_deploy_target` from **kworkflow.config** to
   *<target>*, which can be local or remote.
+
+\--verbose:
+  Verbose mode allows the user to see the commands executed under the hood.
 
 EXAMPLES
 ========
--- a/src/init.sh
+++ b/src/init.sh
@@ -68,57 +68,56 @@ function init_kw()
   notification_config_file_template="${KW_ETC_DIR}/notification_template.config"
   remote_file_template="${KW_ETC_DIR}/remote.config"
 
-  if [[ -f "$config_file_template" && -f "$build_config_file_template" ]]; then
-    mkdir -p "$PWD/$KW_DIR"
-    cp "$config_file_template" "$PWD/$KW_DIR/$name"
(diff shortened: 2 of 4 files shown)


documentation/man/features/init.rst
```diff
@@ -37,6 +37,9 @@ OPTIONS
   Set the variable `default_deploy_target` from **kworkflow.config** to
   *<target>*, which can be local or remote.
 
+-v, \--verbose:
```
rodrigosiqueira: We decided not to support `-v`, please remove it.

src/init.sh
```diff
@@ -47,6 +47,11 @@ function init_kw()
     return 22 # EINVAL
   fi
 
+  if [[ -n "${options_values['VERBOSE']}" ]]; then
+    flag='VERBOSE'
+  fi
+  flag=${flag:-'SILENT'}
```
rodrigosiqueira: How about:
```bash
flag=${flag:-'SILENT'}
[[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
```

src/init.sh
```diff
@@ -69,17 +74,27 @@ function init_kw()
...
-  cp "$vm_config_file_template" "${PWD}/${KW_DIR}/${vm_name}"
-  cp "$build_config_file_template" "${PWD}/${KW_DIR}/${build_name}"
-  cp "$deploy_config_file_template" "${PWD}/${KW_DIR}/${deploy_name}"
-  cp "$mail_config_file_template" "${PWD}/${KW_DIR}/${mail_name}"
-  cp "$notification_config_file_template" "${PWD}/${KW_DIR}/${notification_name}"
-  cp "$remote_file_template" "${PWD}/${KW_DIR}/${remote_name}"
-
-  sed --in-place --expression "s/USERKW/${USER}/g" -e '/^#?.*/d' "${PWD}/${KW_DIR}/${vm_name}"
-  sed --in-place --expression "s,SOUNDPATH,${KW_SOUND_DIR},g" -e '/^#?.*/d' "${PWD}/${KW_DIR}/${notification_name}"
+  cmd="mkdir --parents \"${PWD}/${KW_DIR}\""
```
rodrigosiqueira: I don't think you need to add `"` inside `"`. In other words, drop the internal quotes for command and the other.

documentation/man/features/init.rst
```diff
@@ -6,8 +6,8 @@ kw-init
 
 SYNOPSIS
 ========
-| *kw* *init* [(-v | \--verbose)] [\--template[=name]]
-| *kw* *init* [(-v | \--verbose)] [\--arch <name>][\--remote <user>@<ip>:<port>][\--target <target>]
```
rodrigosiqueira: You don't need this extra commit for that; go back to the previous commit and just remove `-v` from there.

rodrigosiqueira (changes requested):

aquilamacedo:
Hi @rodrigosiqueira. Actually, those changes you asked for were already made in the last commit.

rodrigosiqueira:
@aquilamacedo  squash your last commit with the previous one.

rodrigosiqueira:
Merged!
Thanks a lot

PR #840, opened 2023-05-13
src: build: Fix failure in the save-log option

rodrigosiqueira:
Due to the recent set of updates in the env feature, kw got a regression in the save-log-to option since it tries to concatenate the 'O=...' parameter with the 'tee' command. This commit fixes this issue by adding the 'O=...' option in the right place.

### DIFF
--- a/src/build.sh
+++ b/src/build.sh
@@ -122,7 +122,7 @@ function build_kernel_main()
     return "$?"
   fi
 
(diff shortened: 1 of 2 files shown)


davidbtadokoro (commented):
Didn't test it, but looks legit to me!

rodrigosiqueira:
Thanks!
I already merged this change.

Issue #842, opened 2023-05-18
Use `b4` to download patches from the public mailing lists


davidbtadokoro:
**Description:**

At the moment, in `upstream-patches-ui`, kw implements a way to directly download patches from lore. Use [b4](https://b4.docs.kernel.org/en/latest/index.html) instead.

**Type:**

Performance and robustness.

**Why:**

Although consistent, the current implementation lacks capabilities, such like:
- Download a full thread
- Download in a `git am` ready format (separate the cover letter)

`b4` is a dedicated tool for these tasks, having extra capabilities that (as `upstream-patches-ui` matures) may come in handy, like sending automated contributor feedback. Also, the delegation of these tasks makes the feature more robust considering that `b4` is stable and so on.

PR #843, opened 2023-05-19
src: lib: lore: Use b4 tool for downloading patch series

davidbtadokoro:
The download of a patch series in `upstream-patches-ui` was done using an in-house implementation. Although functional, the `b4` tool is better suited for this task, as this is a dedicated tool, and its capabilities are more vast and robust. Also, considering `b4` reliable, this results in less code to maintain.

This commits makes `upstream-patches-ui` use `b4` to do the task of downloading a series. The downloaded file is in a .mbx format and ready to be applied to a git tree, i.e., the patches are in order, there are no discussion messages and there is no cover letter. The in-house implementation did this, but there was a file for each patch and it considered the cover letter as a patch (which resulted in a bug).

Note: The in-house implementation is still used for fetching metadata from the lore.kernel.org API (e.g. for checking available mailing list and patch series). `b4` doesn't seem to handle this task, but it may be beneficial to find a way to also delegate it.

Closes: https://github.com/kworkflow/kworkflow/issues/842

### DIFF
--- a/src/lib/lore.sh
+++ b/src/lib/lore.sh
@@ -411,112 +411,90 @@ function get_patches_from_mailing_list()
   done
 }
 
-function title_to_path()
-{
-  local title="$1"
-  local file_name
-
-  # Replace space in favor of _
-  file_name="${title// /_}"
-
-  # Replace .
-  file_name="${file_name//./_}"
-
-  # Replace special character
-  file_name="${file_name//[&*+%!:,]/}"
-
-  # Replace /
-  file_name="${file_name//[\/\\]/-}"
-
-  printf '%s' "$file_name"
-}
-
-function convert_title_to_patch_name()
-{
-  local title="$1"
-  local index="$2"
-  local file_name
-
-  file_name=$(title_to_path "$title")
-
-  if [[ -n "$index" ]]; then
-    printf '%04d-%s\n' "$index" "${file_name}.mbox"
-    return
-  fi
-
-  printf '%s\n' "${file_name}.mbox"
-}
-
-function url_update_patch_number()
-{
-  local url="$1"
-  local new_number="$2"
-
-  url="${url/-[0-9]*-/-$link_ref-}"
-
-  printf '%s' "$url"
-}
-
+# This function downloads a patch series in a .mbx format to a given directory
+# using a series URL. The series URL should be the concatenation of the lore URL, the
+# target mailing list, and the message ID. Below is an example of such a URL:
+#   https://lore.kernel.org/some-list/2367462342.4535535-1-email@email.com/
+# The output filename is '<message ID>.mbx'. This function uses the `b4` tool underneath
+# to delegate the process of fetching and downloading the series. The file generated is
+# ready to be applied into a git tree, containing just the patches in the right order,
+# without the cover letter.
+#
+# @series_url: The URL of the series
+# @save_to: Path to the target output directory
+# @flag: Flag to control function output
+#
+# Return:
+# Return 0 if the thread was successfully downloaded, 22 if the series URL or the output
+# directory passed as arguments is empty and the error code of `b4` in case it fails.
 function download_series()
 {
-  local total_patches="$1"
-  local first_message_id="$2"
-  local save_to="$3"
-  local title="$4"
-  local total=0
-  local link_ref=0
-  local url
-  local patch_file_name
+  local series_url="$1"
+  local save_to="$2"
+  local flag="$3"
+  local series_filename
+  local cmd
+  local ret
 
-  mkdir -p "$save_to"
+  flag=${flag:-'SILENT'}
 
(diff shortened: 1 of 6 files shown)


src/lib/lore.sh
```diff
@@ -411,82 +411,52 @@ function get_patches_from_mailing_list()
...
+  series_filename=$(printf '%s' "$series_url" | rev | cut --delimiter '/' -f2 | rev)
+  url=$(replace_http_by_https "$series_url")
 
-    url=$(url_update_patch_number "$url" "$link_ref")
-    patch_file_name=$(convert_title_to_patch_name "$title" "$link_ref")
-    download "${url}raw" "$patch_file_name" "${save_to}" &
-  done
-  wait
+  # Issue the command to download the series
+  cmd="b4 --quiet mbox '${url}' --outdir '${save_to}' --mbox-name '${series_filename}.mbox'"
```
davidbtadokoro, line 449: @rodrigosiqueira the `mbox` subcommand results in the downloading of the whole thread. If we use `am` instead, everything not essential for using `git am` is trimmed and the patches are ordered. The previous version of `download_series` practically used this second approach, but:
1. It didn't differ the cover letter from the patches (resulted in errors)
2. The downloaded file was broken down into `n` files (`n` being the number of patches + cover letter).

What do you think?

src/lib/lore.sh
```diff
@@ -411,112 +411,77 @@ function get_patches_from_mailing_list()
...
+  # Create the output directory if it doesn't exists
+  cmd_manager "$flag" "mkdir --parents '${save_to}'"
+  ret="$?"
+  [[ "$ret" != 0 ]] && return "$ret"
 
-  until ! is_the_link_valid "$url"; do
-    ((total++))
-    ((link_ref++))
+  # Although, by default, b4 uses the message-ID as the output name, we should assure it
+  series_filename=$(printf '%s' "$series_url" | rev | cut --delimiter '/' -f2 | rev)
```
rodrigosiqueira, line 447: I'm not sure If I understood this line; could you help me to understand it?

For example, if I have this URL: 'https://lore.kernel.org/amd-gfx/20230520092518.1993-1-Arunpravin.PaneerSelvam@amd.com/T/#u'

`series_filename` will be `T`?

Anyway, maybe you should consider creating a function for the above operation to make it easy to read.

src/lib/lore.sh
```diff
@@ -411,112 +411,77 @@ function get_patches_from_mailing_list()
...
-  done
-  wait
+  # For safe-keeping, check the protocol used
+  url=$(replace_http_by_https "$series_url")
+
+  # Issue the command to download the series
+  cmd="b4 --quiet am '${url}' --no-cover --outdir '${save_to}' --mbox-name '${series_filename}.mbx'"
+  cmd_manager "$flag" "$cmd"
+  ret="$?"
+  [[ "$ret" != 0 ]] && return "$ret"
```
rodrigosiqueira, line 456: Do you know which type of error b4 can return? If so, we should consider a better return log with a reasonable error message.

rodrigosiqueira (changes requested):

src/lib/lore.sh
```diff
@@ -411,112 +411,77 @@ function get_patches_from_mailing_list()
...
+  # Safety checking
+  if [[ -z "$series_url" || -z "$save_to" ]]; then
+    return 22 # EINVAL
+  fi
 
-  url=$(replace_http_by_https "$first_message_id")
+  # Create the output directory if it doesn't exists
+  cmd_manager "$flag" "mkdir --parents '${save_to}'"
+  ret="$?"
+  [[ "$ret" != 0 ]] && return "$ret"
```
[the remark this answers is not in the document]
davidbtadokoro: Nice. I didn't return a message to the user, because any output to the stdin in this feature is overwritten by dialog screens, resulting in a flickering message. However, your message made me think that the view can capture this message and display it via dialog. This makes the error thrown when `download_series` fails more fine-grained.

src/lib/lore.sh
davidbtadokoro: Sorry for that. My intent was to get the last string delimited by `/`s, so I inverted the string and got the second item. However, your example made me think that this method is not stable as just getting the 5th item delimited in the straight order 😭

src/lib/lore.sh
davidbtadokoro: Nice catch.
Could reproduce 2 kinds of errors returned by `b4` with this template of execution:

- Invalid arguments returns 2 ENOENT (strange, should return 22 EINVAL)
- Unknown URL (message-ID) and invalid `save_to` returns 1 EPERM. Also strange. Why the same errno, if the error messages displayed are different?

Either way, two types of errors are better than just a generic one 😆

rodrigosiqueira:
Merged.
Thanks

Issue #841, opened 2023-05-14
Add Bash and Zsh completions for `upstream-patches-ui`


davidbtadokoro:
**Description:**

Although not having any option, there are no Bash nor Zsh completions for the feature.

davidbtadokoro:
Remind to fix the top level completion commands for Bash

<img width="1131" alt="Screen Shot 2023-05-19 at 13 15 58" src="https://github.com/kworkflow/kworkflow/assets/95055784/899ce743-8864-44ba-81b8-38dbdfc4e35a">

PR #844, opened 2023-05-19
Add Bash and Zsh completions for `upstream-patches-ui`

davidbtadokoro:
This PR just adds Bash and Zsh completions for the `upstream-patches-ui` feature.

Closes: #841


rodrigosiqueira:
Merged!
Thanks

PR #850, opened 2023-05-28
src: kw_remote: Fix set-default option

rodrigosiqueira:
The set-default option is completely broken, and it does not work when using the 'add' option and also does not work when used alone. This commit addresses those issues by adjusting the code sequence and fixing some of the checks.

### DIFF
--- a/src/kw_remote.sh
+++ b/src/kw_remote.sh
@@ -165,12 +165,6 @@ function add_new_remote()
   remote="${remote_parameters['REMOTE_USER']}@${remote_parameters['REMOTE_IP']}"
   remote+=":${remote_parameters['REMOTE_PORT']}"
 
-  # Check if user request to set this new entry a default
-  if [[ -z "$first_time" && -n ${options_values['DEFAULT_REMOTE']} ]]; then
-    options_values['DEFAULT_REMOTE']="$name"
-    set_default_remote
-  fi
-
   # Check if remote name already exists
   grep --line-regexp --quiet "^Host ${name}$" "$remote_config_file"
   if [[ "$?" == 0 ]]; then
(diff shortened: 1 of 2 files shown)


tests/kw_remote_test.sh
```diff
@@ -164,6 +165,45 @@ function test_add_new_remote_multiple_entry_with_duplication()
...
+    '  User juca'
+  )
+
+  options_values['PARAMETERS']='origin root@test-debian:3333'
+  output=$(add_new_remote)
+
+  options_values['DEFAULT_REMOTE_USED']=1
+  options_values['DEFAULT_REMOTE']='debian-machine'
+  options_values['PARAMETERS']='debian-machine root@test-debian'
+  output=$(add_new_remote)
```
davidbtadokoro, line 195: Shouldn't there be an assertion here?

src/kw_remote.sh
```diff
@@ -188,6 +182,12 @@ function add_new_remote()
     printf '  Port %s\n' "${remote_parameters['REMOTE_PORT']}"
     printf '  User %s\n' "${remote_parameters['REMOTE_USER']}"
   } >> "$remote_config_file"
+
+  # Check if user request to set this new entry a default
```
davidbtadokoro, line 186: Sorry if I'm missing the point, but there is only one default per context, so shouldn't it be '(...) new entry **_as_** default' instead of '(...) new entry **_a_** default'?

davidbtadokoro (changes requested):
Nice change @rodrigosiqueira ! Saw this bug the other day, but thought the problem was me not using the feature correctly 😭. Left some comments, but overall it looks legit to me 👍

src/kw_remote.sh
```diff
@@ -188,6 +182,12 @@ function add_new_remote()
...
+  if [[ -z "$first_time" && ${options_values['DEFAULT_REMOTE_USED']} -eq 1 ]]; then
+    options_values['DEFAULT_REMOTE']="$name"
+    set_default_remote
+  fi
```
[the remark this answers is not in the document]
rodrigosiqueira: When you set a new remote for the first time, the file `remote.config` might not be available, and the highlighted code will fail because the target file does not exist. That's why I moved this file below to the file creation.

tests/kw_remote_test.sh
rodrigosiqueira: We don't need any assertion because this test just populates the `remote.config` file.

davidbtadokoro (approved):
Thanks for the update and for answering my questions!

rodrigosiqueira:
Thanks!

Issue #851, opened 2023-05-28
verbose mode for kw ssh


aquilamacedo:
**Description:**

Currently, it is not possible to view the commands being executed by `kw ssh`.

**Why:**

To improve the user experience and allow better troubleshooting, this issue proposes to implement support for verbose mode in SSH. By enabling verbose mode, users will have the ability to preview running commands, making it possible to identify errors.

**API draft:**

```
kw ssh  --verbose
kw ssh --command "whoami" --verbose
kw ssh --script some_script.sh --verbose
```

PR #852, opened 2023-05-28
src: kw_ssh: adding verbose mode

aquilamacedo:
This commit adds support for the verbose parameter within `kw ssh`. The verbose parameter gives details of the commands that are executed behind the scenes.

Closes: #851

### DIFF
--- a/documentation/man/features/ssh.rst
+++ b/documentation/man/features/ssh.rst
@@ -6,9 +6,9 @@ kw-ssh
 
 SYNOPSIS
 ========
-| *kw* (*s* | *ssh*) [(-s | \--script) <script-path>]
-| *kw* (*s* | *ssh*) [(-c | \--command) <string-command>]
-| *kw* (*s* | *ssh*) [(-r | \--remote) <USER@IP:PORT | CONFIG_HOST_NAME>]
+| *kw* (*s* | *ssh*) [(-s | \--script) <script-path>] [\--verbose]
+| *kw* (*s* | *ssh*) [(-c | \--command) <string-command>] [\--verbose]
+| *kw* (*s* | *ssh*) [(-r | \--remote) <USER@IP:PORT | CONFIG_HOST_NAME>] [\--verbose]
 
 DESCRIPTION
 ===========
@@ -28,6 +28,11 @@ OPTIONS
 
 -r <USER@IP:PORT | CONFIG_HOST_NAME>, \--remote <USER@IP:PORT | CONFIG_HOST_NAME>:
   Uses parameter to ssh into the target machine
+
+\--verbose:
+  Verbose mode is an option that causes the kw program to display debug messages to track
+  its progress. This functionality is very useful during the debugging process, allowing
+  you to identify possible errors more easily.
 
 EXAMPLES
 ========
--- a/src/_kw
+++ b/src/_kw
@@ -456,6 +456,7 @@ _kw_s() { _kw_ssh }
 _kw_ssh()
 {
   _arguments : \
+    '(--verbose)--verbose[enable verbose mode]' \
     '(-r --remote)'{-r,--remote}'[explicitly tell which target machine to ssh into]: : ' \
     '(-s --script -c --command)'{-s,--script}'[execute a bash command in the target machine]: : ' \
     '(-c --command -s --script)'{-c,--command}'[execute a bash script in the target machine]: : '
--- a/src/kw_ssh.sh
+++ b/src/kw_ssh.sh
@@ -14,16 +14,18 @@ declare -gA options_values
 #   attempt to execute a command or script on the remote host.
 function kw_ssh_main()
 {
+  local flag="$1"
   local port
   local script_path
   local cmd
   local ssh_cmd
-  local flag
   local user
   local remote
   local remote_file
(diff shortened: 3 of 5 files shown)


davidbtadokoro (commented):
Really nice change @aquilamacedo ! Couldn't find any apparent change needed in my review. Also, tested it locally and it looks legit to me!

src/kw_ssh.sh
```diff
@@ -72,14 +74,13 @@ function handle_ssh()
...
+  flag=${flag:-'SILENT'}
+
   user="${remote_parameters['REMOTE_USER']}"
   remote="${remote_parameters['REMOTE_IP']}"
   port="${remote_parameters['REMOTE_PORT']}"
   remote_file="${remote_parameters['REMOTE_FILE']}"
   remote_file_host="${remote_parameters['REMOTE_FILE_HOST']}"
-  flag=${options_values['VERBOSE']}
-
-  [[ -z "$flag" ]] && ssh_compose+=' -v'
```
rodrigosiqueira, line 82: I did not get why the verbose worked if you dropped this line.

documentation/man/features/ssh.rst
```diff
@@ -9,6 +9,8 @@ SYNOPSIS
 | *kw* (*s* | *ssh*) [(-s | \--script) <script-path>]
 | *kw* (*s* | *ssh*) [(-c | \--command) <string-command>]
 | *kw* (*s* | *ssh*) [(-r | \--remote) <USER@IP:PORT | CONFIG_HOST_NAME>]
+| *kw* (*s* | *ssh*) [\--verbose]
```
rodrigosiqueira, line 12: `--verbose` can be used in all options; please, add it to all options.

src/_kw
```diff
@@ -456,6 +456,7 @@ _kw_s() { _kw_ssh }
 _kw_ssh()
 {
   _arguments : \
+    '(--verbose)--verbose[enable verbose mode]' \
```
rodrigosiqueira, line 459: Verbose can be used in all options.

rodrigosiqueira (changes requested):
Add verbose to `is_ssh_connection_configured` as well.

src/_kw
aquilamacedo: In this case the verbose is encompassing all options

aquilamacedo:
Verbose has also been added to the is_ssh_connection_configured function. @rodrigosiqueira

rodrigosiqueira (changes requested):
Hi @aquilamacedo ,

This change is good, but I noticed that you are not passing the verbose option to the ssh commands. In the process of building the ssh command, you should check if the user specifies the `--verbose` option and compose the ssh command with the flag `-v`. In other words, we want the kw command output and also the ssh output.

aquilamacedo:
I'm breaking my head on this pull request because my changes with the "-v" of ssh are conflicting with the unit tests for some reason, I'm still debugging to try to understand what's going on.

image

aquilamacedo:
Hi @rodrigosiqueira. For some reason the tests are entering the conditional `[[ "$flag" == 'VERBOSE' ]] && ssh_cmd+=' -v'` in the `remote.sh` file and adding the `-v` in almost all the tests, I believe that this is because of `test_mode`, should I remove that `test_mode` parameter?

src/kw_ssh.sh
```diff
@@ -39,7 +39,9 @@ function kw_ssh_main()
 
   script_path=${options_values['SCRIPT']}
   cmd=${options_values['CMD']}
-  flag=${options_values['VERBOSE']}
+
+  [[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
+  flag=${flag:-'SILENT'}
```
rodrigosiqueira, line 44: These lines, together with the `parser_ssh_options` are the root cause of your test failures. For example, try this:

```diff
 
-  [[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
   flag=${flag:-'SILENT'}
+  [[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
+  [[ "${options_values['VERBOSE']}" -eq 'TEST_MODE' ]] && flag='TEST_MODE'
```

Please, review your code carefully and do not use the above suggestion in your final PR since this is a workaround to highlight the issue.

aquilamacedo:
Hi @rodrigosiqueira. Your suggestion made me understand the problem better, I adapted the code and now it's working. Could you review my implementation and give some feedback? Thx :)

src/kw_ssh.sh
```diff
@@ -39,7 +39,9 @@ function kw_ssh_main()
 
   script_path=${options_values['SCRIPT']}
   cmd=${options_values['CMD']}
-  flag=${options_values['VERBOSE']}
+
+  flag="${options_values['TEST_MODE']:-'SILENT'}"
+  [[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
```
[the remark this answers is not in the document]
aquilamacedo: You're right, I ended up adding some unnecessary parts, but I fixed that ;)

rodrigosiqueira:
Merged! Thanks

Issue #829, opened 2023-03-29
Add 'Settings' menu to Dashboard screen


davidbtadokoro:
**Description:**

In `kw upstream-patches-ui`, settings like the default kernel source tree and branch, dialog layout and default directory to save patches are important and are only configurable through `kw config` or by manually changing the `lore.config` file. We should add a menu in the Dashboard screen to allow the user to configure those through the feature itself.

**Type:**

Functionality.

**Why:**

This will improve the user experience as configurations can be changed on the fly without needing to exit the feature.

Issue #811, opened 2023-03-16
Add 'Manage Mailing Lists' menu to Dashboard screen


davidbtadokoro:
**Description:**

In `kw upstream-patches-ui`, the only moment when the user can register/unregister mailing lists is when first using the feature. Add a menu in the Dashboard that allows the user to manage mailing lists. When accessing this menu, the user should see the registered mailing lists checked.

**Type:**

Functionality.

**Why:**

We shouldn't expect the user to know beforehand each mailing list he wants to keep track of.

PR #853, opened 2023-05-30
Add 'Settings' menu for `upstream-patches-ui`

davidbtadokoro:
This PR has the purpose of adding a 'Settings' menu for the `upstream-patches-ui` feature. A 'Settings' menu refers to an interface inside the feature to all its settings, mostly located in the `lore.config` file.

The 'Settings' menu is located in the 'Dashboard' screen, and, inside it, there is a submenu for each setting. At the moment, only the 'lists' and 'download_path' settings are configurable via the feature.

It is worth noting that any future setting (be it inside `lore.config` or not) should have a dedicated sub-menu inside the 'Settings' menu.

Closes: #829 
Closes: #811

### DIFF
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -373,6 +373,79 @@ function create_directory_selection_screen()
   return "$ret"
 }
 
+# Create a screen with a list of choices.
+#
+# @box_title: Title of the box
+# @message_box: The message to be displayed
+# @_choices: Associative array reference containing all the
+#   `<choice><choice_description>` pairs to be displayd
+# @_check_statuses: An array reference containing all the statuses of the checks
+#   (if they are on/off).
+# @height: Menu height in lines size
+# @width: Menu width in column size
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/kwlib.sh` function `cmd_manager`
+#
+# Return:
+# Returns 0 if the 'OK' button is pressed and 1 if the 'Cancel' button is pressed.
+# The key of the chosen pair (i.e. the `<choice>`) is returned via the
+# `menu_return_string` variable.
+function create_choice_list_screen()
+{
+  local box_title="$1"
+  local message_box="$2"
+  local -n _choices="$3"
+  local -n _check_statuses="$4"
+  local height="$5"
+  local width="$6"
+  local flag="$7"
+  local choice_description
+  local index=0
+  local cmd
+  local ret
+
+  flag=${flag:-'SILENT'}
+  height=${height:-$DEFAULT_HEIGHT}
+  width=${width:-$DEFAULT_WIDTH}
+  back_title=${back_title:-"${KW_UPSTREAM_TITLE}"}
+
+  # Escape all single quotes to avoid breaking arguments
+  back_title=$(str_escape_single_quotes "$back_title")
+  box_title=$(str_escape_single_quotes "$box_title")
+  message_box=$(str_escape_single_quotes "$message_box")
+
+  # Add layout to dialog
+  if [[ -n "$DIALOG_LAYOUT" ]]; then
+    cmd="DIALOGRC=${DIALOG_LAYOUT}"
+  fi
+
+  # Add general information to the dialog box
+  cmd+=" dialog --backtitle $'${back_title}' --title $'${box_title}' --clear --colors"
+  # Add radiolist screen
+  cmd+=" --radiolist $'${message_box}'"
+  # Set height and width
+  cmd+=" '${height}' '${width}' '0'"
+  # Add choices (${choices["${choice}"]} refers to the choice's description)
+  for choice in "${!_choices[@]}"; do
+    choice_description=$(str_escape_single_quotes "${_choices["${choice}"]}")
+    choice=$(str_escape_single_quotes "$choice")
+    if [[ "${_check_statuses["$index"]}" == 1 ]]; then
+      cmd+=" $'${choice}' $'${choice_description}' 'on'"
+    else
+      cmd+=" $'${choice}' $'${choice_description}' 'off'"
+    fi
+    ((index++))
+  done
+
+  [[ "$flag" == 'TEST_MODE' ]] && printf '%s' "$cmd" && return 0
+
+  exec 3>&1
+  menu_return_string=$(cmd_manager "$flag" "$cmd" 2>&1 1>&3)
+  ret="$?"
+  exec 3>&-
+  return "$ret"
+}
+
 # Creates a help message box for a dialog's screen. There must be a file
 # that follows the pattern of `load_module_text`, for reference see `src/kwio.sh`.
 #
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -318,6 +318,92 @@ function create_message_box()
   return "$ret"
 }
 
+# Create directory selection screen.
+#
+# @starting_path: Path that is in the input box when screen starts
+# @box_title: Title of the box
+# @height: Menu height in lines size
+# @width: Menu width in column size
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/kwlib.sh` function `cmd_manager`
+#
+# Return:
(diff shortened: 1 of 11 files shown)


src/upstream_patches_ui.sh
```diff
@@ -193,6 +200,35 @@ function show_bookmarked_series_details()
...
+  menu_list_string_array=('Mailing Lists')
+  message_box=''
+  create_menu_options 'Settings' "$message_box" 'menu_list_string_array' 1
+  ret="$?"
+
+  case "$ret" in
+    0) # OK
+      case "$menu_return_string" in
+        1) # Mailing Lists
+          screen_sequence['SHOW_SCREEN']='register'
```
rodrigosiqueira: If I follow: Setting > Mailing Lists, I'll only have `OK` and `EXIT` options which make sense for the first screen. If I choose the OK option, it will send me back to the dashboard, which is not intuitive. It is time to parameterize the `register` option to enable it to return to the previous menu.

src/upstream_patches_ui.sh
```diff
@@ -193,6 +200,35 @@ function show_bookmarked_series_details()
...
 }
 
+# Screen that shows all types of settings available.
+function show_settings_screen()
+{
+  local -a menu_list_string_array
+  local message_box
+  local ret
+
+  menu_list_string_array=('Mailing Lists')
```
rodrigosiqueira: I think the name `Mailing List` is not good since it does not give the action idea. Maybe we can try something like `Follow Mailing List` or similar.

src/lib/dialog_ui.sh
```diff
@@ -178,7 +178,7 @@ function create_simple_checklist()
   cmd+=" '${height}' '${width}' '${list_height}'"
 
   for item in "${_menu_list_string_array[@]}"; do
-    if [[ -n "${_check_statuses}" && -n ${_check_statuses["$index"]} ]]; then
+    if [[ "${_check_statuses["$index"]}" == 1 ]]; then
```
rodrigosiqueira, line 181: Can we have a test that exposes this bug?

src/upstream_patches_ui.sh
```diff
@@ -402,31 +401,46 @@ function list_patches()
...
+  local index=0
   local lore_config_path="${PWD}/.kw/lore.config"
   local ret
 
   if [[ ! -f "${lore_config_path}" ]]; then
     lore_config_path="${KW_ETC_DIR}/lore.config"
   fi
 
-  retrieve_available_mailing_lists "$flag"
+  create_loading_screen_notification "Retrieving available mailing lists from lore.kernel.org"
```
rodrigosiqueira, line 447: Single-quotes.

src/upstream_patches_ui.sh
```diff
@@ -402,31 +401,46 @@ function list_patches()
...
   # shellcheck disable=SC2207
   IFS=$'\n' menu_list_string_array=($(sort <<< "${!available_lore_mailing_lists[*]}"))
   unset IFS
 
-  message_box="It looks like that you don't have any lore list registered; please"
-  message_box+=" select one or more of the below list:"
+  # Put check marks on mailing lists already registered
+  for list in "${menu_list_string_array[@]}"; do
+    check_statuses["$index"]=0
+    # TODO: This may be inefficient, but we can't use partial matches as lists that are
```
rodrigosiqueira, line 457: I guess this is not a big deal since the number of lists is very small.

src/upstream_patches_ui.sh
```diff
@@ -443,6 +457,9 @@ function register_mailing_list()
     1) # Exit
       handle_exit "$ret"
       ;;
+    3) # Return
+      screen_sequence['SHOW_SCREEN']='dashboard'
```
rodrigosiqueira, line 489: I guess that my comment in the second commit makes sense here.

src/lib/dialog_ui.sh
```diff
@@ -295,6 +295,81 @@ function create_message_box()
...
+  flag=${flag:-'SILENT'}
+
+  box_title='Directory Selection Help'
+  message_box='There are 3 regions in the Directory Selection screen:'$'\n'
+  message_box+='- [Upper Box]: list of directories in the current path.'$'\n'
+  message_box+='- [Lower Box]: the current path (input box).'$'\n'
+  message_box+='- [Buttons]: "OK" to confirm directory path, "Cancel" to cancel selection and "Help" to show this screen.'$'\n'$'\n'
+  message_box+='To move between the regions, use the <TAB> key.'$'\n'
+  message_box+='To complete the current path with the highlighted directory, use the <SPACE> key.'$'\n'
+  message_box+='Typing while in the Upper Box or the Lower Box alters the current path.'
```
rodrigosiqueira, line 368: We have a better way to deal with this :) Check `load_module_text`.

rodrigosiqueira (changes requested):

rodrigosiqueira:
@melissawen

rodrigosiqueira (changes requested):

src/upstream_patches_ui.sh
```diff
@@ -202,11 +202,17 @@ function show_bookmarked_series_details()
 # Screen that shows all types of settings available.
 function show_settings_screen()
 {
+  local lore_config_path="${PWD}/.kw/lore.config"
```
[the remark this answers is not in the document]
davidbtadokoro: Hi, @rodrigosiqueira it doesn't create. Already made an issue for it, but I think is time to make a PR.
Thanks for the reminder!

src/lib/dialog_ui.sh
```diff
@@ -365,6 +365,62 @@ function create_directory_selection_help_screen()
...
+
+  # Add general information to the dialog box
+  cmd+=" dialog --backtitle $'${back_title}' --title $'${box_title}' --clear --colors"
+  # Add radiolist screen
+  cmd+=" --radiolist $'${message_box}'"
+  # Set height and width
+  cmd+=" '${height}' '${width}' '0'"
+  # Add choices (${choices["${choice}"]} refers to the choice's description)
+  for choice in "${!_choices[@]}"; do
+    cmd+=" $'${choice}' $'${_choices["${choice}"]}' off"
```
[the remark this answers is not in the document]
davidbtadokoro: You are right. This change request slipped through in my last force-push. Going to work on it ASAP.

tests/lib/lore_test.sh
```diff
@@ -421,4 +421,22 @@ function test_save_new_lore_config()
...
+
+  # Mock for command 'git branch --verbose'
+  # shellcheck disable=SC2317
+  function git()
+  {
+    printf ' branch1  \t cafebabe commit_subject1\n  branch2 \t\t deadbeef commit_subject2\n\t\t branch3   c00010ff commit_subject3'
+  }
+  export -f git
+
+  get_kernel_tree_branches "$kernel_tree_path" 'branches'
```
rodrigosiqueira, line 436: Where this `$kernel_tree_branches` came from?
Also, I think that a test that creates a fake repository with multiple branches is more valuable for this case; in other words, I recommend you not use the `git` mock and use the function in a fake repository created by the test. I think we have some examples of that in the kw code.

rodrigosiqueira (changes requested):

tests/lib/lore_test.sh
davidbtadokoro: Hi @rodrigosiqueira 
That was my first approach as I agree that mocking the `git` in this superficial way is not ideal. The tests worked perfectly but I was met with this message when trying to the sample repository of the project:

Screen Shot 2023-06-10 at 20 18 41

My approach was to create a sample git repository and copy it to the project's tree. From what I understand this warning means that a submodule (a git repo insider another) must be declared together with a remote. Is it okay to ignore this message or am I making it wrong?

rodrigosiqueira: In the setup phase, you can create a temporary folder followed by `git init`; that way, you will have a valid git repository in a temporary folder. In the tear_down phase, you can delete the temporary folder.

davidbtadokoro: Thanks a lot! I'm sure that this approach will do 😄

rodrigosiqueira:
Thanks for all the improvements in this PR, merged into the unstable branch.

PR #855, opened 2023-06-04
`upstream-patches-ui`: dialog's severe bugs with certain arguments

davidbtadokoro:
I noted that some severe bugs happen when we don't consider the expansions when building a dialog command. For example, the command `(...) dialog --menu "${item}"` when `item='something<double_quote>something'` or `item='something<apostrophe>something'` results in wrong expansions when passing it to `cmd_manager` (and `eval` in turn). These bugs result in crashes that potentially can drive users off the feature.

This PR adds two commits:

1. Adds mocks to reproduce some of the bugs (possibly there are plenty more variations). Instructions to reproduce the bugs are in the commit message.
2. Adds a proof-of-concept fix that pragmatically solves the problem. However, we need to take a careful and robust approach to solve those, and this fix is far from ideal.

**[UPDDATE]**
Transformed this PR in a possible solution to the problem.


davidbtadokoro:
@rodrigosiqueira Here is the PR with mocks to reproduce the bugs that I told you.
To be sincere, although I guess I got to the root of the problem, I think we should carefully plan how to solve it and would appreciate your insights on the matter.
Thanks a lot in advance (and sorry for the desperation 😭 ).

rodrigosiqueira:
Very nice catch! We must convert your mocks into real tests.

About this problem, how about a solution like this:

```diff
diff --git a/src/lib/dialog_ui.sh b/src/lib/dialog_ui.sh
index 96150dc..3146f06 100644
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -100,13 +100,12 @@ function create_menu_options()
 
   for item in "${_menu_list_string_array[@]}"; do
     # Work-around to prevent bugs
-    item=$(printf '%s' "$item" | sed -e "s/\"/'/g" -e "s/\`/'/g")
     if [[ -n "$no_index" ]]; then
-      cmd+=" $\"${item}\" ''"
+      cmd+=" $'${item}' ''"
       continue
     fi
 
-    cmd+=" '${index}' $\"${item}\""
+    cmd+=" '${index}' $'${item}'"
     ((index++))
   done
```

src/lib/dialog_ui.sh
```diff
@@ -99,6 +99,8 @@ function create_menu_options()
   cmd+=" '${height}' '${width}' '${max_elements_displayed_in_the_menu}'"
 
   for item in "${_menu_list_string_array[@]}"; do
+    # Work-around to prevent bugs
+    item=$(printf '%s' "$item" | sed -e "s/\"/'/g" -e "s/\`/'/g")
```
rodrigosiqueira, line 103: With my previous comments, we could simplify this part by using something like `sed -e "s/'/\\\'/g"`. I guess this address the single-quotes problem. Ideally, we don't want this change here; we want to move it to the lore file. Or we can create another dedicated function to process the item just for the dialog code.

davidbtadokoro:
@rodrigosiqueira Made this PR into a proposed solution. If you can, take a look 👍

rodrigosiqueira:
Very nice fix! Merged into the unstable branch.
Thanks!

Issue #849, opened 2023-05-21
`upstream-patches-ui` wrong title in 'new patches' screen after returning from 'series details' screen


davidbtadokoro:
**Description:**

After returning from a series details screen, the 'new patches' screen has the wrong title (which, at first was correct).

**How to Reproduce:**

1. Navigate to a mailing list's new patches screen
2. Select a series to see its details and return back

**Expected behavior:**

The title should be correct no matter how many series we see.

**Screenshots:**

<img width="1437" alt="Screen Shot 2023-05-21 at 11 40 03" src="https://github.com/kworkflow/kworkflow/assets/95055784/7627878d-fc4d-4cf5-b453-d4697c42d2f6">

<img width="1437" alt="Screen Shot 2023-05-21 at 11 40 11" src="https://github.com/kworkflow/kworkflow/assets/95055784/381ab593-2356-42ce-ad12-7c8220258ed3">

<img width="1438" alt="Screen Shot 2023-05-21 at 11 40 21" src="https://github.com/kworkflow/kworkflow/assets/95055784/e440ffdc-f23a-47bd-ae10-fb2a9460c95a">

**System information:**

- OS: Manjaro
- kw Version: unstable branch eba1443
- Bash Version: 5.1.16

PR #861, opened 2023-06-13
src: upstream_patches_ui: Fix 'New Patches' screen title bug

davidbtadokoro:
In the `upstream-patches-ui` feature, the title of the screen where new patches from a given mailing list are shown is incorrectly displayed when returning from a series details screen. In other words, when we first visit the 'New Patches' screen, the title is correct (something like 'Patches from amd-gfx)'. But, if we select a patch series to see its details and return to the previous screen ('New Patches' screen), the title will be wrong (something like 'Patches from 0').

This commit fixes this by adding a global variable named `current_mailing_list` to keep track of the latest mailing list queried, i.e., the latest mailing list that the feature has fetched new patches. We already checked to see if `screen_sequence['RETURNING']` was set, signaling that we are (supposedly) returning from a series details screen and we don't need to query the mailing list again, but the list name displayed was resetted (with `screen_sequence['SCREEN_PARAMETER']` value) regardless. Now, the list name displayed is defined by the `current_mailing_list` value and it is only set when not returning from a (supposed) series details screen.

Closes: #849

### DIFF
--- a/src/upstream_patches_ui.sh
+++ b/src/upstream_patches_ui.sh
@@ -13,6 +13,7 @@ include "${KW_LIB_DIR}/kwlib.sh"
 declare -ga registered_lists
 declare -ga patches_from_mailing_list
 declare -ga bookmarked_series
+declare -g current_mailing_list
 
 declare -gA screen_sequence=(
   ['SHOW_SCREEN']='manage_mailing_lists'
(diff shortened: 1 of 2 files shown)


rodrigosiqueira:
Merged! Thanks

PR #862, opened 2023-06-15
Adding 'Apply' action and 'Kernel Config File' setting menu to `kw upstream-patches-ui`

davidbtadokoro:
This PR has the objective of integrating the `kw upstream-patches-ui` feature with `kw build`, i.e., allowing the user to build a patch version of the kernel through the feature's UI. The steps involved are:

- [x] Add the 'Apply' action to apply the patch version of the kernel to a given kernel tree and a target branch
- [x] Add the 'Kernel Config File' setting to let the user define which .config file to use
- ~~Add the 'Build' action to build the patch version of the kernel to a given kernel tree and a target branch~~

It is expected that 'Apply' triggers the 'Download' action and 'Build' triggers 'Apply'.

**[UPDATE]**
This PR was reduced to not integrate the feature with `kw build`, as the feature needs to be cleaned and refactored. Integrating `kw build` before it will only increase the mess.

### DIFF
--- a/etc/dialog_help/file_selection_help.txt
+++ b/etc/dialog_help/file_selection_help.txt
@@ -1,0 +1,12 @@
+[file_selection_help_box_title]:
+File Selection Help
+[file_selection_help_message_box]:
+There are 4 regions in the Directory Selection screen:
+- [Upper Left Box]: list of directories in the current path.
+- [Upper Right Box]: list of files in the current path.
+- [Lower Box]: the current path (input box).
+- [Buttons]: "OK" to confirm file path, "Cancel" to cancel selection and "Help" to show this screen.
+
+To move between the regions, use the <TAB> key.
+To complete the current path with the highlighted directory/file, use the <SPACE> key.
+Typing while in the Upper Box or the Lower Box alters the current path.
--- a/src/kwlib.sh
+++ b/src/kwlib.sh
@@ -717,3 +717,60 @@ function get_git_repository_branches()
     _branches["$branch"]="$branch_metadata"
   done <<< "$output"
 }
+
+# This function gets metadata about the kernel config files (.config files) managed
+# by kw. The data is transmitted using an associative array reference passed as
+# argument. Each key-value pair of the array is like:
+#  `array_reference[<kernel_config_name>]='<kernel_config_description>'`
+#
+# @_kernel_configs: Associative array reference where data will be trasmitted
+#
+# Return:
+# Returns data regarding the kernel config files managed by kw through the array
+# reference passed as argument if successful. In case the `select_from` function
+# returns an error code, return this code.
+function get_kernel_configs_managed_by_kw()
+{
+  local -n _kernel_configs="$1"
+  local query_output
+  local kernel_config_name
+  local kernel_config_description
+  local ret
+
+  # Resetting associative array reference to prevent false configs
+  _kernel_configs=()
+
+  query_output=$(select_from 'kernel_config' '"name","description"')
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    return "$ret"
+  fi
+
+  # Return, in case there are no kernel config files under kw management
+  [[ -z "$query_output" ]] && return
+
+  while IFS=$'\n' read -r entry; do
+    kernel_config_name=$(printf '%s' "$entry" | cut --delimiter='|' -f1)
+    kernel_config_description=$(printf '%s' "$entry" | cut --delimiter='|' -f2)
+    _kernel_configs["$kernel_config_name"]="$kernel_config_description"
+  done <<< "$query_output"
+}
+
+# This function gets the path of a kernel config file managed by kw given a kernel
+# config name.
+#
+# @kernel_config_name: Name of the kernel config file managed by kw
+#
+# Return:
+# Returns 2 (ENOENT) in case there is no kernel config file managed by kw with the name
+# `@kernel_config_name`. Otherwise, it returns 0 and the path to the kernel config file.
+function get_kernel_config_path_managed_by_kw()
+{
+  local kernel_config_name="$1"
+
+  kernel_config_path=$(select_from "kernel_config WHERE name IS '${kernel_config_name}'" '"path"')
+
+  [[ -z "$kernel_config_path" ]] && return 2 # ENOENT
+
+  printf '%s' "$kernel_config_path"
+}
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -373,6 +373,69 @@ function create_directory_selection_screen()
   return "$ret"
 }
 
+# Create file selection screen.
+#
+# @starting_path: Path that is in the input box when screen starts
+# @box_title: Title of the box
+# @extra_label: Label to override 'Extra' button. If empty, 'Extra' button
+#   is not displayed.
+# @height: Menu height in lines size
+# @width: Menu width in column size
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/kwlib.sh` function `cmd_manager`
+#
+# Return:
+# Returns 0 if the 'OK' button is pressed, 1 if the 'Cancel' button is pressed
+# and 2 if the 'Help' button is pressed. The string in the input box (current
+# path) is returned via the `menu_return_string` variable.
+function create_file_selection_screen()
+{
+  local starting_path="$1"
+  local box_title="$2"
+  local extra_label="$3"
+  local height="$4"
+  local width="$5"
+  local flag="$6"
+  local cmd
+  local ret
+
+  flag=${flag:-'SILENT'}
+  height=${height:-'15'}
+  width=${width:-'80'}
+  back_title=${back_title:-"${KW_UPSTREAM_TITLE}"}
+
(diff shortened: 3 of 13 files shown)


rodrigosiqueira (changes requested):

src/lib/lore.sh
```diff
@@ -732,3 +735,93 @@ function save_new_lore_config()
...
+          complain 'Application of rejected file failed...'
+          return "$ret"
+        else
+          success 'Application of rejected file was a success!'
+        fi
+      done <<< "$output"
+      warning 'Application of rejected patch was a success!'
+      warning 'Recommended examining each replaced file for unresolved conflicts and semantic.'
+    else
+      complain "Could not apply series ${series_path}"
```
magalilemes, line 820: I wonder if, after reaching this point,  kw should abort the `git am` operation it started, or even inform the user that they are in the middle of a `git am` and should themselves deal with this, i.e. abort, skip or solve the issue.
I do miss the error message, gone with `/dev/null 2>&1`, that would help understanding why the series could not be applied. I'm not sure, though, how this output could be presented to the user, as it doesn't seem suitable for the UI.

davidbtadokoro: Thanks for the review @magalilemes and nice catch!
You are right, we should abort or inform the user that there is a `git am` in progress. I opted for the latter thinking, as the application is encapsulated in a new branch, it is good to present the user with the option to deal with it but completely forgot to warn the user about it 😞. Because of the way dialog is implemented, any output to the stdout is not displayed correctly, which is why I used `/dev/null 2>&1` to not "pollute" the dialog's screen. Have the feeling that there is a better way to do this and I'm going to look at it.

davidbtadokoro:
@rodrigosiqueira Sorry for the delay. Working on integrating with `kw build` (hoping that it is less complicated than adding the 'Apply' action 😆). If you can, please review my new changes and look at the comment that @magalilemes made above (would appreciate your insight on the matter).

src/upstream_patches_ui.sh
```diff
@@ -187,6 +189,34 @@ function show_bookmarked_series_details()
...
+            create_yes_no_prompt 'Warning' "$message_box"
+            case "$?" in
+              0) # Yes
+                create_loading_screen_notification 'Applying patch(es)'$'\n'"- ${series['patch_title']}"
+                output=$(download_series "${series['patch_url']}" "${lore_config['save_patches_to']}")
+                if [[ "$?" != 0 ]]; then
+                  create_message_box 'Error' 'Could not download patch(es):'$'\n'"- ${series['patch_title']}"$'\n'"[error message] ${output}"
+                  return
+                fi
+                output=$(apply_patchset "${series['patch_title']}" "$output" "${lore_config['kernel_tree_path']}" "${lore_config['kernel_tree_branch']}")
```
rodrigosiqueira, line 208: Add a newline after each `fi`.

src/upstream_patches_ui.sh
```diff
@@ -432,6 +462,34 @@ function show_series_details()
...
+                if [[ "$?" != 0 ]]; then
+                  create_message_box 'Fail' 'Could not apply patch(es):'$'\n'"- ${series['patch_title']}"$'\n'$'\n'"[error message] ${output}"
+                  return
+                fi
+                create_message_box 'Success' "'${series['patch_title']}' applied."$'\n'"${output}"
+                ;;
+              1) # No
+                create_message_box '' 'Patch application aborted!'
+                ;;
+            esac
```
rodrigosiqueira, line 559: Is this the same code as the previous one? Can you create a common function for that?

etc/dialog_help/file_selection_help.txt
```diff
@@ -0,0 +1,12 @@
...
+[file_selection_help_message_box]:
+There are 4 regions in the Directory Selection screen:
+- [Upper Left Box]: list of directories in the current path.
+- [Upper Right Box]: list of files in the current path.
+- [Lower Box]: the current path (input box).
+- [Buttons]: "OK" to confirm file path, "Cancel" to cancel selection and "Help" to show this screen.
+
+To move between the regions, use the <TAB> key.
+To complete the current path with the highlighted directory/file, use the <SPACE> key.
+Typing while in the Upper Box or the Lower Box alters the current path.
```
rodrigosiqueira, line 12: Maybe create a single commit for this string change.

src/kwlib.sh
```diff
@@ -717,3 +717,54 @@ function get_git_repository_branches()
...
+{
+  local -n _kernel_configs="$1"
+  local query_output
+  local kernel_config_name
+  local kernel_config_description
+
+  # Resetting associative array reference to prevent false configs
+  _kernel_configs=()
+
+  query_output=$(select_from 'kernel_config' '"name","description"')
```
rodrigosiqueira, line 741: `select_from` may fail; you must handle does cases.

src/lib/dialog_ui.sh
```diff
@@ -373,6 +373,61 @@ function create_directory_selection_screen()
...
+# @height: Menu height in lines size
+# @width: Menu width in column size
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/kwlib.sh` function `cmd_manager`
+#
+# Return:
+# Returns 0 if the 'OK' button is pressed, 1 if the 'Cancel' button is pressed
+# and 2 if the 'Help' button is pressed. The string in the input box (current
+# path) is returned via the `menu_return_string` variable.
+function create_file_selection_screen()
```
rodrigosiqueira, line 389: This function should have its own commit.

src/lib/dialog_ui.sh
```diff
@@ -373,6 +373,61 @@ function create_directory_selection_screen()
...
+{
+  local starting_path="$1"
+  local box_title="$2"
+  local height="$3"
+  local width="$4"
+  local flag="$5"
```
rodrigosiqueira, line 395: fselect does not support an extra button and label rename? If so, add those extra parameters.

src/upstream_patches_ui.sh
```diff
@@ -338,6 +340,72 @@ function show_settings_screen()
...
+          message_box+='1) Choose from your .config files managed by kw'$'\n'
+          message_box+='2) Choose from your filesystem'
+          create_yes_no_prompt 'Kernel Config File' "$message_box" 'Option 1' 'Option 2' 'Return'
+          case "$?" in
+            0) # Choose from your .config files managed by kw
+              get_kernel_configs_managed_by_kw 'kernel_configs'
+              if [[ "${#kernel_configs[@]}" == 0 ]]; then
+                create_message_box '' 'There are no kernel configs under kw management'
+                return
+              fi
```
rodrigosiqueira, line 357: newline.

rodrigosiqueira (changes requested):

davidbtadokoro:
@rodrigosiqueira Thanks for the comments! Going to work on them 👍

src/lib/lore.sh
```diff
@@ -732,3 +736,180 @@ function save_new_lore_config()
...
+    complain "'${kernel_tree_path}' is not a Linux kernel tree"
+    return 22 # EINVAL
+  fi
+
+  cmd="git -C ${kernel_tree_path} switch ${base_kernel_tree_branch} --quiet"
+  cmd_manager "$flag" "$cmd"
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain "Could not switch kernel tree ${kernel_tree_path} to branch ${base_kernel_tree_branch}"
+    return "$ret"
```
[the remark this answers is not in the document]
davidbtadokoro: Don't know all but, for example, trying `git switch` whilst rebasing you get the message
```
fatal: cannot switch branch while rebasing
```
and 128 as return.

src/lib/lore.sh
```diff
@@ -732,3 +736,180 @@ function save_new_lore_config()
...
+
+  new_branch_name=$(date +'%Y-%m-%d-%H-%M-%S-')
+  new_branch_name+=$(string_to_unix_filename "$series_title")
+  cmd="git -C ${kernel_tree_path} checkout -b ${new_branch_name} --quiet"
+  cmd_manager "$flag" "$cmd"
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain "Could not create new branch ${new_branch_name} based on ${base_kernel_tree_branch}"
+    return "$ret"
+  fi
```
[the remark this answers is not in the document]
davidbtadokoro: Hmm, here I can't think of an example of creating a new branch on top of HEAD and failing. Any ideas?

src/lib/lore.sh
davidbtadokoro: Good idea, going to work on that. I was thinking of not using
```bash
return "$ret"
```
throughout this function and returning static numbers for each error handling to make it predictable and possible to return a specific number here to signal the caller that a Yes/No screen needs to be created to prompt the user between these 3 options. What do you guys think?

src/upstream_patches_ui.sh
davidbtadokoro: Nice catch, this is also really bothering me as it is happening throughout the whole file. Going to work on that in a separate PR for refactoring this whole file.

rodrigosiqueira:
3 patches merged. Thanks

PR #866, opened 2023-06-17
Fix variable name in remote2host

magalilemes:
One of the variables passed to the rsync command had a typo, so operations like fetching a config file from a remote machine would not succeed. Replace `dsc` with `dst` in the rsync command.

### DIFF
--- a/src/remote.sh
+++ b/src/remote.sh
@@ -182,7 +182,7 @@ function remote2host()
 
   if [[ -z "$remote" && -z "$port" && -z "$user" ]]; then
     if [[ -n ${remote_parameters['REMOTE_FILE']} ]]; then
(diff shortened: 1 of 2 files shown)


davidbtadokoro (approved):
@magalilemes, great catch!
Didn't test it, but it seems really cut-and-dried.

rodrigosiqueira:
@magalilemes very nice catch! This change also highlighted a gap in our test coverage. Could you also add a test for this function in this PR?

magalilemes:
> @magalilemes very nice catch! This change also highlighted a gap in our test coverage. Could you also add a test for this function in this PR?

@rodrigosiqueira just added some tests for the function that should cover the usage of a remote file as well as passing of parameters from the command line.

rodrigosiqueira:
Thanks a lot for this change!

PR #867, opened 2023-06-17
src: upstream_patches_ui: Replace undefined help function call

davidbtadokoro:
When calling a help screen for an arbitrary dialog screen, we use a generic function named `create_help_screen` passing the name of the screen in question. In the 'Save Patches To' menu, the wrong function is called, named `create_directory_selection_help_screen` (a function that existed when there was no generic help function).

This commit fixes this by replacing this undefined function call for the right one.


rodrigosiqueira:
Merged! Thanks

Issue #865, opened 2023-06-17
`upstream-patches-ui` 'Kernel Tree Path' configuration registers relative paths


davidbtadokoro:
**Description:**

In the `kw upstream-patches-ui`, a really important configuration in `lore.config` is `kernel_tree_path`. It is possible to set it through the feature, but it checks to see if the path passed corresponds to a Linux kernel tree before registering it in the config file. It accepts relative paths to kernel trees (which is desirable), but registers this relative path in the config file. This can potentially break some parts of the feature.

**How to Reproduce:**

1. Install kw
2. `cd` into a kernel tree
3. Run `kw upstream-patches-ui`
4. Go to `Dashboard -> Settings -> Kernel Tree Path`
5. Input `.` (dot) as a path
6. Check your `lore.config` to see that the path registered is `.`

**Expected behavior:**

Relative paths to Linux kernel trees can be **accepted**, but the **registered** path should be absolute.

**Screenshots:**

<img width="1435" alt="Screen Shot 2023-06-17 at 00 09 49" src="https://github.com/kworkflow/kworkflow/assets/95055784/3943f4f7-288e-4512-a4af-8df8c3895255">
<img width="880" alt="Screen Shot 2023-06-17 at 00 10 31" src="https://github.com/kworkflow/kworkflow/assets/95055784/6e99798a-2473-4949-89d1-ff5b5f8e750b">

**Additional context:**

I think this isn't a complicated fix, as using the `realpath` command to the accepted paths should do.

PR #868, opened 2023-06-17
src: upstream_patches_ui: Fix relative paths in 'Kernel Tree Path'

davidbtadokoro:
When selecting a Linux kernel source tree through the `upstream-patches-ui` feature, relative paths are accepted when they refer to a kernel tree, but the relative path is registered in the `lore.config` file, which can break parts of the feature, depending where it is executed.

This commit fixes this by still allowing relative paths but converting them to absolute paths using the `realpath` command.

Closes: #865


rodrigosiqueira:
Thanks! Patch merged into the unstable branch

PR #875, opened 2023-06-24
tests: utils: Remove extra char

rodrigosiqueira:
When assert_equals_helper helper fails, the failed and expected strings are misaligned for a single character, making it hard to compare the final result. This commit just removes the extra character, that way, it will be visually evident the fail pattern.


davidbtadokoro (approved):
Nicely done @rodrigosiqueira!
This misalignment also bothered me and, although simple, this change will help a lot when debugging!

rodrigosiqueira:
Thanks!

PR #877, opened 2023-06-24
src: patch_hub: Rename `upstream-patches-ui` feature to `patch-hub`

davidbtadokoro:
The `upstream-patches-ui` feature has a name that can be improved, both in its meaning and size.

As voted in https://github.com/kworkflow/kworkflow/discussions/872, the new name chosen was `patch-hub`. It is shorter, catchier (in my opinion) and holds the meaning of a centralized place (hence, 'hub') to work with patches.

This commit consolidates this renaming. It only changes instances of `upstream-patches-ui` to `patches-hub` and `upstream_patches_ui` to `patch_hub` in file/directory names, function names, and documentation (be it the project's documentation or in comments).

The only change that isn't this straight renaming is in `src/lib/dialog_ui.sh` where we renamed the global variable `KW_UPSTREAM_TITLE` to `KW_PATCH_HUB_TITLE` and changed its value from `kw upstream patches manager` to `kw patch-hub`.


melissawen:
LGTM. I opened some issues that is not related to this PR, but general improvements for the feature.

Issue #883, opened 2023-07-20
Remote machine ssh warning follows inconsistent path to output information


melissawen:
**Description:**

I have this configuration in my `remote.config` file:

```
#kw-default=origin
Host origin
  Hostname 192.168.31.178
#  Hostname 192.168.28.214
#  Hostname 192.168.203.135
  Port 22
  User root
```

When something goes wrong, KW warns the user by showing the current settings but the way we are collecting this information is quite wrong. It is just reading lines and ignoring keywords. Considering my example above, if the machine is not available it shows the following warning:

```
$ kw s
We could not reach the remote machine by using:
 IP: 192.168.31.178
 User: 192.168.203.135
 Port: 192.168.28.214
Please ensure that the above info is correct.
Suggestion: Check if your remote machine permits root login via ssh
or check if your public key is in the remote machine.
```
So, it doesn't check the first two lines are comments. Probably it's a wrong usage of comments, but also we need to improve it on KW side. I can use this configuration with comments without issues when the machine is available.

**How to Reproduce:**

Just put in the `remote.config` file the example above.

**Expected behavior:**

I was expecting this output:

```
$ kw s
We could not reach the remote machine by using:
 IP: 192.168.31.178
 User: root
 Port: 22
Please ensure that the above info is correct.
Suggestion: Check if your remote machine permits root login via ssh
or check if your public key is in the remote machine.
```

**System information:**

 - OS: Debian Bookworm
 - kw Version: kw unstable

melissawen:
Fixed by 9483f4a

PR #887, opened 2023-07-29
src: lib: remote: Fix ssh connection fail message with remote.config

davidbtadokoro:
When a ssh connection using a `remote.config` file fails, the message shown about the failed attempt displays the wrong IP, User and Port values, if the `remote.config` file doesn't follows this pattern for each host:

Host \<host\>
&nbsp;&nbsp;Hostname \<IP address\>
&nbsp;&nbsp;Port \<port number\>
&nbsp;&nbsp;User \<username\>

For example, if there are commented lines inbetween, or the order is different, the values displayed on a failed ssh connection will be messed.

This happens because the implementation to build the fail message considered that the file would strictly follow the pattern above and picked the Hostname by the line after the 'Host' line, the Port by the second line after the 'Host' line and the User by the third line after the 'Host' line.

This commit fixes this bug by fetching the IP, User and Port values in the `remote.config` file more reliably, using pattern matching with grep instead of assuming a "well-formatted" file. It also adds an unit test to reproduce a "bad-formatted" `remote.config` file case.

Closes: #883

### DIFF
--- a/src/lib/remote.sh
+++ b/src/lib/remote.sh
@@ -43,16 +43,37 @@ function ssh_connection_failure_message()
   local user=${remote_parameters['REMOTE_USER']}
   local remote_file=${remote_parameters['REMOTE_FILE']}
   local remote_file_host=${remote_parameters['REMOTE_FILE_HOST']}
-  local host_line_number
+  local line_number
+  local trimmed_remote_file
+  local line
 
   # If a config file is being used, we need to set the IP, port
   # and user from the config file.
   if [[ -z "$remote" && -z "$port" && -z "$user" ]]; then
     if [[ -n "${remote_parameters['REMOTE_FILE']}" ]]; then
-      host_line_number=$(grep -n "Host ${remote_file_host}" "${remote_file}" | cut -d: -f1)
-      remote=$(sed -n "$((host_line_number + 1))p" "${remote_file}" | cut -d\  -f4)
-      port=$(sed -n "$((host_line_number + 2))p" "${remote_file}" | cut -d\  -f4)
-      user=$(sed -n "$((host_line_number + 3))p" "${remote_file}" | cut -d\  -f4)
+      # Getting the line number of the target host
+      line_number=$(grep --line-number "^Host ${remote_file_host}\$" "$remote_file" | cut --delimiter=':' -f1)
+      ((line_number++))
+      # Trimming remote file from target host line upward
+      trimmed_remote_file=$(tail --lines "+${line_number}" "${remote_file}")
+      # Getting the line number of next host from target host, if it exists
+      line_number=$(printf '%s' "$trimmed_remote_file" | grep --line-number --max-count=1 '^Host' | cut --delimiter=':' -f1)
+      # In case next host exists, trim remote file again from next host line downward
+      if [[ -n "$line_number" ]]; then
+        ((line_number--))
+        trimmed_remote_file=$(printf '%s' "$trimmed_remote_file" | head --lines "$line_number")
+      fi
+
+      # Here, we are matching lines with Hostname/Port/User occurrences that happen between the target
+      # host line and the next host line (if it exists) in the remote config file and only have leading
+      # spaces (excludes commented lines). We then cut the resulting string to get only the value of
+      # the Hostname/Port/User line.
+      line=$(printf '%s' "$trimmed_remote_file" | grep --max-count=1 '^[[:blank:]]*Hostname')
+      remote=$(printf '%s' "$line" | rev | cut --delimiter=' ' -f1 | rev)
+      line=$(printf '%s' "$trimmed_remote_file" | grep --max-count=1 '^[[:blank:]]*Port')
+      port=$(printf '%s' "$line" | rev | cut --delimiter=' ' -f1 | rev)
+      line=$(printf '%s' "$trimmed_remote_file" | grep --max-count=1 '^[[:blank:]]*User')
+      user=$(printf '%s' "$line" | rev | cut --delimiter=' ' -f1 | rev)
     else
       complain 'Could not find remote config file.'
       complain 'Suggestion: check if there is a remote.config or try using'
--- a/tests/lib/remote_test.sh
+++ b/tests/lib/remote_test.sh
@@ -37,6 +37,7 @@ function oneTimeSetUp()
   cp -f 'tests/samples/kworkflow.config' "$TEST_PATH/.kw/"
   cp -f 'tests/samples/dmesg' "$TEST_PATH"
   cp -f "${KW_REMOTE_SAMPLES_DIR}/remote.config" "${TEST_PATH}/.kw/"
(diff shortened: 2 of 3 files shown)


tests/lib/remote_test.sh
```diff
@@ -264,6 +265,29 @@ function test_ssh_connection_failure_message()
   assertEquals "($LINENO):" "$no_config_file_failure_message" "$ret"
 }
 
+test_ssh_connection_failure_message_with_bad_formatted_remote_config()
```
magalilemes, line 268: For consistency, use the function keyword to define a function

src/lib/remote.sh
```diff
@@ -43,16 +43,21 @@ function ssh_connection_failure_message()
...
-      host_line_number=$(grep -n "Host ${remote_file_host}" "${remote_file}" | cut -d: -f1)
-      remote=$(sed -n "$((host_line_number + 1))p" "${remote_file}" | cut -d\  -f4)
-      port=$(sed -n "$((host_line_number + 2))p" "${remote_file}" | cut -d\  -f4)
-      user=$(sed -n "$((host_line_number + 3))p" "${remote_file}" | cut -d\  -f4)
+      target_host_line_number=$(grep -n "Host ${remote_file_host}" "${remote_file}" | cut -d: -f1)
+      trimmed_remote_file=$(tail --lines "+${target_host_line_number}" "${remote_file}")
+      # Here, we are matching the first line with Hostname/Port/User occurrence that happens after
+      # the `target_host_line_number` line and only have leading spaces (excludes commented lines).
+      # We then cut the resulting string to get only the value of the Hostname/Port/User line.
+      remote=$(printf '%s' "$trimmed_remote_file" | grep --max-count=1 '^[[:blank:]]*Hostname' | cut -d\  -f4)
```
magalilemes, line 58: Come to think of it, would this regex (and the subsequent ones too) work if tabs were used, instead of white spaces, to indent the remote config file?

tests/lib/remote_test.sh
davidbtadokoro: Thanks a lot for the catch, @magalilemes!
Now that you mentioned it, it would be nice to have a ShellCheck rule for this one. However, I only found the opposite 🤣 https://www.shellcheck.net/wiki/SC2112.

src/lib/remote.sh
davidbtadokoro: Yes, I'm pretty sure that `[[:blank:]]` captures any whitespace (at least in my tests it worked 😆). Modified the sample file `remote_4.config` (used in the added unit test) to have lines with tabs, just to be sure. Thanks for the heads up!

magalilemes: Interesting. It seems like the sample file only contains whitespaces still. I added a tab and the test fails. Could you double check that please?
image

davidbtadokoro: Hmm, interesting, this sample file does contains tabs, but the test does fails if we add more tabs or no tabs (or any number of whitespace that isn't 2). The problem actually is the `cut -d\  -f4` part of the command, as it assumes that there are exactly 2 leading whitespaces in the line. Going to fix that, thanks for digging deep in this matter 👍

davidbtadokoro: Hi @magalilemes fixed it. If possible, can you double check this for me?

magalilemes: @davidbtadokoro Just tested and it works now.
However, what I've been meaning to say with tab is the tab character (ascii = 9). For instance, using vscode, if I select the lines I can see the difference between "real" tabs and whitespaces. I added a tab right before "Port 123" for demonstrating:
image

rodrigosiqueira:
lgtm, @melissawen could you check this on your side?

melissawen:
There is one tricky thing happening. It is not collecting information defined by Host (and stop the data collection when the next Host definition starts). It's currently looking for the next 3 valid keywords (Hostname, Port and User) regardless its Host definition boundaries. One example:

```
$ cat .kw/remote.config 
#kw-default=origin
Host origin
# Port 33
# Hostname 192.168.31.173
# Hostname 192.168.31.178
# Ppp 22
  User root

Host melissawen
  Hostname 192.168.31.178
  Port 22
  User melissawen

$ kw s
 IP: 192.168.31.178
 User: root
 Port: 22
```

It may lead to misunderstanding any configuration issue.
In this example, my machine is up but the `Host root` data is wrong, so I can't access it. But the warning output wrongly says it's correctly configured.

davidbtadokoro:
Hi @melissawen.
First, thanks a lot for pointing out this error about Host boundaries. Fixed it in the latest force-push. If possible, see if it is ok.

src/lib/remote.sh
```diff
@@ -43,16 +43,37 @@ function ssh_connection_failure_message()
...
   # If a config file is being used, we need to set the IP, port
   # and user from the config file.
   if [[ -z "$remote" && -z "$port" && -z "$user" ]]; then
     if [[ -n "${remote_parameters['REMOTE_FILE']}" ]]; then
-      host_line_number=$(grep -n "Host ${remote_file_host}" "${remote_file}" | cut -d: -f1)
-      remote=$(sed -n "$((host_line_number + 1))p" "${remote_file}" | cut -d\  -f4)
-      port=$(sed -n "$((host_line_number + 2))p" "${remote_file}" | cut -d\  -f4)
-      user=$(sed -n "$((host_line_number + 3))p" "${remote_file}" | cut -d\  -f4)
+      # Getting the line number of the target host
+      line_number=$(grep -n "Host ${remote_file_host}" "$remote_file" | cut -d: -f1)
```
rodrigosiqueira, line 55: `--line-number` instead of `-n` and `--delimiter=DELIM` instead of `-d`.

src/lib/remote.sh
```diff
@@ -43,16 +43,37 @@ function ssh_connection_failure_message()
...
+      ((line_number++))
+      # Trimming remote file from target host line upward
+      trimmed_remote_file=$(tail --lines "+${line_number}" "${remote_file}")
+      # Getting the line number of next host from target host, if it exists
+      line_number=$(printf '%s' "$trimmed_remote_file" | grep -n --max-count=1 '^Host' | cut -d: -f1)
```
rodrigosiqueira, line 60: Check my previous comment.

src/lib/remote.sh
```diff
@@ -43,16 +43,37 @@ function ssh_connection_failure_message()
...
+      # In case next host exists, trim remote file again from next host line downward
+      if [[ -n "$line_number" ]]; then
+        ((line_number--))
+        trimmed_remote_file=$(printf '%s' "$trimmed_remote_file" | head -n "$line_number")
```
rodrigosiqueira, line 64: Check my previous comment.

rodrigosiqueira (changes requested):

melissawen:
I just tested and LGTM. Thanks.

davidbtadokoro:
Hi @rodrigosiqueira.
I have addressed your comments. Thanks for the feedback.
Also, thanks @melissawen for testing it (again 😆)

rodrigosiqueira:
Thanks!

PR #889, opened 2023-08-02
`kw patch-hub`: Add reliable fetch of latest patchsets from mailing list

davidbtadokoro:
This PR implement the most critical point in `kw patch-hub` development, which is the interaction with the Lore API to fetch the latest patchsets from a given mailing list.

It introduces a realiable way to navigate through the whole history of a mailing list.

It can also be used as a base for making more specific queries to lore archives.

### DIFF
--- a/.github/workflows/kcov.yml
+++ b/.github/workflows/kcov.yml
@@ -20,7 +20,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          sudo apt install -y kcov shunit2 bc sqlite3 bsdmainutils
+          sudo apt install -y kcov shunit2 bc sqlite3 bsdmainutils libxml-xpath-perl
 
       - name: Prepare for tests
         run: |
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -36,14 +36,24 @@ function ui_setup()
   fi
 }
 
-# This function is responsible for creating dialog menus.
+# Create menu with options that the user can select. This function is flexible enough
+# to allow the menu to have from 2 to 4 buttons with custom labels, options displayed
+# with or without indexes and to display just a range of options from the array passed
+# as reference.
 #
-# @box_title: This is the menu title used on the top left of the dialog screen.
-# @message_box: The instruction text used for this menu.
-# @_menu_list_string_array: An array reference containing all the strings to be used in the menu.
-# @cancel_label: Cancel label. If not set, the default is 'Exit']
+# @box_title: This is the menu title used on the top left of the dialog screen
+# @message_box: The instruction text used for this menu
+# @_list_of_options_array: An array reference with the options to be displayed
+# @starting_index: Starting index of range of options to be displayed. If not set,
+#   the default is 0.
+# @ending_index: Ending index of range of options to be displayed. If not set, the
+#   default is the size of `_list_of_options_array` minus 1 (last index of array).
+# @extra_label: Extra label. If not set, the 'Extra' button won't be displayed
+# @cancel_label: Cancel label. If not set, the default is 'Exit'
+# @help_label: Help label. If not set, the 'Help' button won't be displayed
 # @height: Menu height in lines size
 # @width: Menu width in column size
+# @no_index: Boolean to control display of index. If set, indexes won't be displayed
 # @flag How to display a command, the default value is
 #   "SILENT". For more options see `src/lib/kwlib.sh` function `cmd_manager`
 #
@@ -55,18 +65,20 @@ function create_menu_options()
 {
   local box_title="$1"
   local message_box="$2"
-  local -n _menu_list_string_array="$3"
-  local back_button_label="$4"
-  local cancel_label="$5"
-  local height="$6"
-  local width="$7"
-  local max_elements_displayed_in_the_menu="$8"
-  local no_index="$9"
-  local flag="${10}"
-  local index=1
+  local -n _list_of_options_array="$3"
+  local starting_index="$4"
+  local ending_index="$5"
+  local extra_label="$6"
+  local cancel_label="$7"
+  local help_label="$8"
+  local height="$9"
+  local width="${10}"
+  local no_index="${11}"
+  local flag="${12}"
+  local option
   local cmd
 
-  if [[ "${#_menu_list_string_array[@]}" -eq 0 ]]; then
+  if [[ "${#_list_of_options_array[@]}" -eq 0 ]]; then
     return 22 # EINVAL
   fi
 
@@ -74,33 +86,39 @@ function create_menu_options()
   height=${height:-$DEFAULT_HEIGHT}
   width=${width:-$DEFAULT_WIDTH}
   cancel_label=${cancel_label:-'Exit'}
-  max_elements_displayed_in_the_menu=${max_elements_displayed_in_the_menu:-'0'}
+  starting_index=${starting_index:-0}
+  ending_index=${ending_index:-$((${#_list_of_options_array[@]} - 1))}
 
   # Escape all single quotes to avoid breaking arguments
   box_title=$(str_escape_single_quotes "$box_title")
-  cancel_label=$(str_escape_single_quotes "$cancel_label")
   message_box=$(str_escape_single_quotes "$message_box")
+  cancel_label=$(str_escape_single_quotes "$cancel_label")
+  extra_label=$(str_escape_single_quotes "$extra_label")
+  help_label=$(str_escape_single_quotes "$help_label")
 
   cmd=$(build_dialog_command_preamble "$box_title")
   # Change cancel label
   cmd+=" --cancel-label $'${cancel_label}'"
   # Add extra button
-  if [[ -n "$back_button_label" ]]; then
-    cmd+=" --extra-button --extra-label 'Return'"
+  if [[ -n "$extra_label" ]]; then
+    cmd+=" --extra-button --extra-label $'${extra_label}'"
+  fi
+  # Add help button
+  if [[ -n "$help_label" ]]; then
+    cmd+=" --help-button --help-label $'${help_label}'"
   fi
   # Add Menu screen
   cmd+=" --menu $'${message_box}'"
   # Set height, width, and max display itens
-  cmd+=" '${height}' '${width}' '${max_elements_displayed_in_the_menu}'"
+  cmd+=" '${height}' '${width}' '0'"
   # Add each menu option with or without index
-  for item in "${_menu_list_string_array[@]}"; do
-    item=$(str_escape_single_quotes "$item")
+  for i in $(seq "$starting_index" "$ending_index"); do
+    option=$(str_escape_single_quotes "${_list_of_options_array["$i"]}")
     if [[ -n "$no_index" ]]; then
-      cmd+=" $'${item}' ''"
-      continue
+      cmd+=" $'${option}' ''"
+    else
+      cmd+=" '${i}' $'${option}'"
     fi
-    cmd+=" '${index}' $'${item}'"
-    ((index++))
   done
 
   [[ "$flag" == 'TEST_MODE' ]] && printf '%s' "$cmd" && return 0
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -589,6 +589,79 @@ function create_form_screen()
   run_dialog_command "$cmd" "$flag"
 }
 
+# Create a rangebox to collect an integer from the user. This rangebox can be
+# used to guaranteedly an integer from a restricted range.
+#
+# @box_title: Title of the box
+# @message_box: The message to be displayed
+# @min_value: Minimum value of the range
+# @max_value: Maximum value of the range
+# @default_value: Starting value of the rangebox
+# @extra_label: Extra label. If not set, the 'Extra' button won't be displayed
+# @cancel_label: Cancel label. If not set, the default is 'Exit'
+# @help_label: Help label. If not set, the 'Help' button won't be displayed
+# @height: Menu height in lines size
+# @width: Menu width in column size
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/lib/kwlib.sh` function `cmd_manager`
+#
+# Return:
+# Returns 0 if the 'Ok' button is pressed, 1 if the 'Cancel' button is pressed,
+# 2 if the 'Help' button is pressed, and 3 if the 'Extra' button is pressed.
+# The menu_return_string will store the user selected value.
+function create_rangebox_screen()
+{
+  local box_title="$1"
+  local message_box="$2"
+  local min_value="$3"
+  local max_value="$4"
+  local default_value="$5"
+  local extra_label="$6"
+  local cancel_label="$7"
+  local help_label="$8"
+  local height="$9"
+  local width="${10}"
+  local flag="${11}"
+  local cmd
+
+  min_value=${min_value:-'0'}
+  max_value=${max_value:-'100'}
+  default_value=${default_value:-$(((min_value + max_value) / 2))}
+  cancel_label=${cancel_label:-'Exit'}
+  flag=${flag:-'SILENT'}
+  height=${height:-'10'}
+  width=${width:-'60'}
+
+  # Escape all single quotes to avoid breaking arguments
+  box_title=$(str_escape_single_quotes "$box_title")
+  message_box=$(str_escape_single_quotes "$message_box")
+  cancel_label=$(str_escape_single_quotes "$cancel_label")
+  help_label=$(str_escape_single_quotes "$help_label")
+  extra_label=$(str_escape_single_quotes "$extra_label")
+
+  cmd=$(build_dialog_command_preamble "$box_title")
+  # Override 'Cancel' button label
+  cmd+=" --cancel-label $'${cancel_label}'"
+  # Add 'Extra' button
+  if [[ -n "${extra_label}" ]]; then
+    cmd+=" --extra-button --extra-label $'${extra_label}'"
+  fi
+  # Add 'Help' button
+  if [[ -n "${help_label}" ]]; then
+    cmd+=" --help-button --help-label $'${help_label}'"
+  fi
+  # Add Rangebox screen
+  cmd+=" --rangebox $'${message_box}'"
+  # Set height and width
+  cmd+=" '${height}' '${width}'"
+  # Set range and default value
+  cmd+=" '${min_value}' '${max_value}' '${default_value}'"
+
+  [[ "$flag" == 'TEST_MODE' ]] && printf '%s' "$cmd" && return 0
+
+  run_dialog_command "$cmd" "$flag"
+}
+
 # Creates a help message box for a dialog's screen. There must be a file
 # that follows the pattern of `load_module_text`, for reference see `src/lib/kwio.sh`.
 #
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -372,12 +372,16 @@ function create_file_selection_screen()
   run_dialog_command "$cmd" "$flag"
 }
 
-# Create a screen with a list of choices.
+# Create a screen with a list of choices. This function supports both choices with descriptions
+# and without description. The first behaviour happens by passing an associative array as
+# `_choices`, but, as associative arrays in bash are hashed, the order of display may not be the
+# order of definition. The second behaviour happens by passing an indexed array as `_choices` and,
+# in this case, the order of definition of the elements is the same as the displayed one.
 #
 # @box_title: Title of the box
 # @message_box: The message to be displayed
-# @_choices: Associative array reference containing all the
-#   `<choice><choice_description>` pairs to be displayd
+# @_choices: Reference to an associative array with `<choice><choice_description>`
+#   pairs to be displayed, or to an indexed array with a list of choices.
 # @_check_statuses: An array reference containing all the statuses of the checks
 #   (if they are on/off).
 # @ok_label: Override the default 'Ok' button label option (Default is 'Yes').
@@ -433,17 +437,30 @@ function create_choice_list_screen()
   cmd+=" --radiolist $'${message_box}'"
   # Set height and width
   cmd+=" '${height}' '${width}' '0'"
-  # Add choices (${choices["${choice}"]} refers to the choice's description)
-  for choice in "${!_choices[@]}"; do
-    choice_description=$(str_escape_single_quotes "${_choices["${choice}"]}")
-    choice=$(str_escape_single_quotes "$choice")
-    if [[ "${_check_statuses["$index"]}" == 1 ]]; then
-      cmd+=" $'${choice}' $'${choice_description}' 'on'"
-    else
-      cmd+=" $'${choice}' $'${choice_description}' 'off'"
-    fi
-    ((index++))
-  done
+  # Add choices without description if `_choices` is an indexed array
+  if [[ "$(typeset -p "$3")" =~ 'declare -a' ]]; then
+    for choice in "${_choices[@]}"; do
+      choice=$(str_escape_single_quotes "$choice")
+      if [[ "${_check_statuses["$index"]}" == 1 ]]; then
+        cmd+=" $'${choice}' '' 'on'"
+      else
+        cmd+=" $'${choice}' '' 'off'"
+      fi
+      ((index++))
+    done
+  # Add choices with description if `_choices` is an associative array
+  else
+    for choice in "${!_choices[@]}"; do
+      choice_description=$(str_escape_single_quotes "${_choices["${choice}"]}")
+      choice=$(str_escape_single_quotes "$choice")
+      if [[ "${_check_statuses["$index"]}" == 1 ]]; then
+        cmd+=" $'${choice}' $'${choice_description}' 'on'"
+      else
+        cmd+=" $'${choice}' $'${choice_description}' 'off'"
+      fi
+      ((index++))
+    done
+  fi
 
   [[ "$flag" == 'TEST_MODE' ]] && printf '%s' "$cmd" && return 0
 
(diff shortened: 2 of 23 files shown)


src/lib/lore.sh
```diff
@@ -312,67 +320,75 @@ function processing_new_patches()
...
-    if [[ "$line" =~ href ]]; then
-      message_id_link=$(str_get_value_under_double_quotes "$line")
-
-      if is_introduction_patch "$message_id_link"; then
-        # Process each patch in parallel
-        thread_for_process_patch "$index" "$shared_dir_for_parallelism" \
-          "$processed_line" "$message_id_link" "$title" &
-        ((index++))
+  while [[ "$index" -lt "$NUMBER_OF_PATCHSETS_FETCHED" ]]; do
+    query_filter="?q=rt:${days}.day.ago..${last_patch_timestamp}+AND+NOT+s:Re&x=A"
```
davidbtadokoro, line 333: Another possibility for `query_filter` is `"?q=s:PATCH+AND+NOT+s:Re&o=${inferior_limit}&x=A"` where the **patch** of number `inferior_limit` + 1 sent to the mailing list is the first one in the response list.

Go to https://lore.kernel.org/amd-gfx/?q=s:PATCH+AND+NOT+s:Re&o=0 and https://lore.kernel.org/amd-gfx/?q=s:PATCH+AND+NOT+s:Re&o=200 for examples of the 1-200 last **patches** sent to the amd-gfx list and of the 201-400 last **patches**, respectively.

src/lib/lore.sh
```diff
@@ -34,6 +34,14 @@ declare -gA available_lore_mailing_lists
 # Special character used for separate data.
 declare -gr SEPARATOR_CHAR='Æ'
 
+# TODO: Remove hardcode and add setting for this
+# Number of patchsets needed to fetch when querying a lore mailing list
+declare -gr NUMBER_OF_PATCHSETS_FETCHED=100
```
rodrigosiqueira, line 39: Let's use 30 by default.

src/lib/lore.sh
rodrigosiqueira: We need to document all of your work somewhere, something that highlights explanations of the above queries and their difference. Maybe a new kw blog post rich in details about this work.

src/lib/lore.sh
```diff
@@ -312,67 +320,75 @@ function processing_new_patches()
...
-      message_id_link=$(str_get_value_under_double_quotes "$line")
-
-      if is_introduction_patch "$message_id_link"; then
-        # Process each patch in parallel
-        thread_for_process_patch "$index" "$shared_dir_for_parallelism" \
-          "$processed_line" "$message_id_link" "$title" &
-        ((index++))
+  while [[ "$index" -lt "$NUMBER_OF_PATCHSETS_FETCHED" ]]; do
+    query_filter="?q=rt:${days}.day.ago..${last_patch_timestamp}+AND+NOT+s:Re&x=A"
+    query_url="${LORE_URL}/${target_mailing_list}/${query_filter}"
```
rodrigosiqueira, line 334: I was wondering if we should create a new function to build the URL and ensure it is valid inside this new function (e.g., `compose_url_with_verification()`).

src/lib/lore.sh
```diff
@@ -312,67 +320,75 @@ function processing_new_patches()
...
+
+    download "$query_url" "$list_patches_file_name" "$CACHE_LORE_DIR" "$flag" || return "$?"
```
rodrigosiqueira, line 336: Since this is a large line, I'm under the impression that using a normal `if` will make this code easier to read.

src/lib/lore.sh
```diff
@@ -312,67 +320,75 @@ function processing_new_patches()
...
+
+    xpath_query="${NAME_EXP}|${EMAIL_EXP}|${TITLE_EXP}|${LINK_EXP}"
+    raw_xml=$(< "${CACHE_LORE_DIR}/${list_patches_file_name}")
+    pre_processed=$(printf '%s' "$raw_xml" | xpath -q -e "$xpath_query")
```
rodrigosiqueira, line 340: This is another code that might be better if it has a function.

P.s.: Document the use of XPath in the lore context; it will also be good.

src/lib/lore.sh
```diff
@@ -312,67 +320,75 @@ function processing_new_patches()
...
+
+    # Converting to:
+    #  Author, email, version, total patches, patch title, link
+    shared_dir_for_parallelism=$(mktemp -d)
```
rodrigosiqueira, line 344: If you want to avoid access to the file system, you can try to replace `mktemp -d` with `/dev/shm` since this is basically a folder in the RAM system. Before you try this, I recommend you check how fast you will use/dev/shm compared with /tmp to ensure that accessing the disk is part of the bottleneck.

rodrigosiqueira (changes requested):

davidbtadokoro:
Hi @rodrigosiqueira.
Thanks a lot for the many invaluable feedbacks and comments. Going to refine this draft ASAP.

src/lib/lore.sh
davidbtadokoro: Nice idea. This is also an opportunity to sturdy further the lore API.

src/lib/lore.sh
davidbtadokoro: Thanks a lot for the idea. This may help with the performance issues. Anyway, going to keep this in mind for when we optimize the feature. Thanks again!

davidbtadokoro:
Hi, @rodrigosiqueira.
I updated the commit with the changes you commented and some more. Sorry for the big commit.
Also, I'm working on a commit to follow it that adds pagination (i.e., the user can see the N older/earlier patchsets).

davidbtadokoro:
Really strange, the CI/CD fails because the `xpath` command is not found.

Screen Shot 2023-08-10 at 00 34 11

Tried some fixes but with no avail 😞

Any ideas?

rodrigosiqueira:
> Really strange, the CI/CD fails because the `xpath` command is not found.
> 
> Screen Shot 2023-08-10 at 00 34 11
> 
> The test runs fine in local and it doesn't seems to be a network environment issue in the runner side.
> 
> Tried some fixes but with no avail disappointed
> 
> Any ideas?

@davidbtadokoro 

If CI is missing some package, check https://github.com/kworkflow/kworkflow/blob/unstable/.github/workflows/unit_tests.yml . I guess you probably need to update it.

davidbtadokoro:
> > Really strange, the CI/CD fails because the `xpath` command is not found.
> > 
> > Screen Shot 2023-08-10 at 00 34 11
> > 
> > The test runs fine in local and it doesn't seems to be a network environment issue in the runner side.
> > 
> > Tried some fixes but with no avail disappointed
> > 
> > Any ideas?
> 
> @davidbtadokoro 
> 
> If CI is missing some package, check https://github.com/kworkflow/kworkflow/blob/unstable/.github/workflows/unit_tests.yml . I guess you probably need to update it.

Oh, that must be it! I tried changing https://github.com/kworkflow/kworkflow/blob/master/.github/workflows/test_setup_and_docs.yml instead of unit_tests.yml ;)

davidbtadokoro:
> I ran a quick test on `ath10k`, and I noticed it took a long time. Is it possible to let the user know that some lists will take more time? Also, I really want to see `next` and `previous` pages as a way to validate your new approach.

For sure, I can add a message for it. But I think that the performance issue (besides the fact that we may need to add some optimizations in the future and that the list may be less "busy" than others) may be influenced by some patches that are discarded by `kw patch-hub` ([draft issue](https://github.com/orgs/kworkflow/projects/2/views/1?pane=issue&itemId=35278244)).

And sorry for the delay on the `next` and `previous` pages. I'm working on it and it is looking good.

src/lib/lore.sh
```diff
@@ -276,72 +287,151 @@ function reset_list_of_mailinglist_patches()
...
+  # timestamp when the message arrived at the lore.kernel.org servers, although its
+  # description is really confusing (see the reference link). We need to check if there
+  # aren't any queries that are more correct and, maybe, more efficient.
+  query_filter="?q=rt:${lower_end_in_days_ago}.day.ago..${upper_end_in_timestamp}+AND+NOT+s:Re&x=A"
+  query_url="${LORE_URL}/${target_mailing_list}/${query_filter}"
+  printf '%s' "$query_url"
+}
+
+# This function pre-processes an XML file containing a list of patches, extracting
+# just the metadata needed to process a XML element representing a patch. The `xpath`
```
rodrigosiqueira, line 369: /a XML/an XML/

src/lib/lore.sh
```diff
@@ -276,72 +287,151 @@ function reset_list_of_mailinglist_patches()
...
+# command is used to capture the desired fields for each patch. A simplified example
+# of a XML element representing a patch is:
```
rodrigosiqueira, line 371: /a XML/an XML/

src/lib/lore.sh
```diff
@@ -276,72 +287,151 @@ function reset_list_of_mailinglist_patches()
...
+#	  	<author>
+#	  		<name>Jay Cornwall</name>
+#	  		<email>jay.cornwall@amd.com</email>
+#	  	</author>
+#	  	<title>[PATCH] drm/amdkfd: Add missing tba_hi programming on aldebaran</title>
+#	  	<updated>2023-08-09T21:27:00Z</updated>
+#	  	<link href="http://lore.kernel.org/amd-gfx/20230809212615.137674-1-jay.cornwall@amd.com/"/>
+#	  	<id></id>
+#	  	<content></content>
+#	  </entry>
```
rodrigosiqueira, line 382: Reduce the indentation to align with `<entry> `

src/lib/lore.sh
```diff
@@ -352,29 +442,102 @@ function processing_new_patches()
...
+    if is_html_file "${CACHE_LORE_DIR}/${xml_result_file_name}"; then
+      days="$((days + TIMEFRAME_IN_DAYS))"
+      continue
+    fi
+
+    raw_xml=$(< "${CACHE_LORE_DIR}/${xml_result_file_name}")
+
+    # Processing patches into patchsets that will be stored in `list_of_mailinglist_patches`.
+    pre_processed_patches=$(pre_process_xml_result "$raw_xml")
+    # TODO: Is passing `pre_processed_patches` (huge string) as argument a possible bottleneck?
```
rodrigosiqueira, line 548: What do you consider huge?

src/lib/lore.sh
```diff
@@ -352,29 +442,102 @@ function processing_new_patches()
...
+    process_patchsets "$pre_processed_patches"
+
+    # Update lower and upper ends of time period to query.
+    days="$((days + TIMEFRAME_IN_DAYS))"
+    last_patch_timestamp=$(printf '%s' "$raw_xml" | xpath -q -e '//entry[last()]/updated/text()')
+    last_patch_timestamp=$(TZ=UTC date --date "${last_patch_timestamp} - 1 seconds" '+%Y-%m-%dT%H:%M:%SZ')
```
rodrigosiqueira: Do we have any helpers for that date manipulation?

src/lib/web.sh
```diff
@@ -49,3 +49,26 @@ function replace_http_by_https()
...
+# decided it isn't, and 2 (ENOENT) if `@file_path` doesn't correspond to a file.
+function is_html_file()
+{
+  local file_path="$1"
+  local first_line_of_file
+
+  [[ ! -f "$file_path" ]] && return 2 # ENOENT
+
+  first_line_of_file=$(head -n 1 "$file_path" | tr '[:upper:]' '[:lower:]')
+  if [[ "$first_line_of_file" =~ ^(<html|<\!doctype html>) ]]; then
```
rodrigosiqueira, line 69: This is usually true; I guess this will always work for the lore use case 100%. However, in terms of URLs, this is not always true. I would add more checks if the first check fails to make this function more robust. For example, you can look for `<header>` and `<body>`; if this function does not find anything like that, then you return false.

tests/lib/lore_test.sh
```diff
@@ -440,4 +441,114 @@ function test_save_new_lore_config()
   assert_equals_helper 'Wrong lore.config contents' "$LINENO" "$expected" "$output"
 }
 
+function test_compose_lore_query_url_with_verification()
```
rodrigosiqueira, line 444: I suggest you split these tests by cases. For example, move the invalid option to its own test and the good cases as well.

tests/lib/web_test.sh
```diff
@@ -50,4 +73,35 @@ function test_replace_http_by_https()
   assert_equals_helper 'Expected https' "($LINENO)" "$output" "$expected"
 }
 
+function test_is_html_file()
```
rodrigosiqueira, line 76: split this test. See my previous comment about that.

.github/workflows/kcov.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          sudo apt install -y kcov shunit2 bc sqlite3 bsdmainutils
+          sudo apt install -y kcov shunit2 bc sqlite3 bsdmainutils libxml-xpath-perl
```
rodrigosiqueira, line 23: You can move this change to the first commit.

src/lib/dialog_ui.sh
```diff
@@ -108,6 +108,85 @@ function create_menu_options()
...
+# @width: Menu width in column size
+# @flag How to display a command, the default value is
+#   "SILENT". For more options see `src/lib/kwlib.sh` function `cmd_manager`
+#
+# Return:
+# The index in `_options_array` of the option selected is stored in the `menu_return_string`
+# variable. If the user hit the 'OK' button 0 is returned, if hit the 'Exit' button 2
+# is returned, if hit the `Previous` button 3 is returned and if hit the `Next` button 1
+# is returned.
+function create_menu_options_with_pagination()
```
rodrigosiqueira, line 133: Do we really need this function? This looks too specific to be part of a lib. Maybe I'm missing something but it looks like you can change `create_menu_options()` to make it more flexible to support the behavior that you implemented here.

src/lib/lore.sh
```diff
@@ -35,16 +35,24 @@ declare -gA available_lore_mailing_lists
 declare -gr SEPARATOR_CHAR='Æ'
 
 # TODO: Remove hardcode and add setting for this
-# Number of patchsets needed to fetch when querying a lore mailing list
-declare -gr NUMBER_OF_PATCHSETS_PER_FETCH=30
+# Number of patchsets per page when querying a lore mailing list
+declare -g NUMBER_OF_PATCHSETS_PER_PAGE=30
```
rodrigosiqueira, line 39: As part of this PR, I would like to see these configuration variables in the config file.

src/lib/dialog_ui.sh
rodrigosiqueira: Additionally, after passing through this commit, I believe this change does not belong to this commit. You need to create a specific commit only for this.

rodrigosiqueira (changes requested):

src/lib/lore.sh
davidbtadokoro: Don't know if this is huge, but in the ten thousands lines and one hundred thousands words mark.

src/lib/lore.sh
```diff
@@ -276,72 +287,151 @@ function reset_list_of_mailinglist_patches()
...
+  fi
+
+  # Verifying if lower end is valid, i.e., is an integer
+  if [[ ! "$lower_end_in_days_ago" =~ ^[0-9]+$ ]]; then
+    return 22 # EINVAL
+  fi
+
+  # Verifying if upper end is valid, i.e., is a real date in the correct format
+  if [[ -n "$upper_end_in_timestamp" ]]; then
+    date '+%Y-%m-%dTH:M:SZ' --date "$upper_end_in_timestamp" > /dev/null 2>&1
```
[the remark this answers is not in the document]
davidbtadokoro: Agree, we need to test this. I was thinking if this helper isn't too specific to lore, so I'm going to add it to `lore.sh`. What do you think?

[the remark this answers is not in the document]
davidbtadokoro: Actually, do you know if the `YYYY-mm-ddTHH:MM:SSZ` is used anywhere or is a known format at all? In that case, we can add to helper to `kw_time_and_date` with a name like `is_in_<name_of_format>_timestamp_format`

davidbtadokoro:
Hey @rodrigosiqueira, updated the PR, but I'm still working on the last commit to add the settings to the config file (and maybe to the 'Settings' screen). Doing it ASAP.

src/lib/dialog_ui.sh
```diff
@@ -55,52 +65,60 @@ function create_menu_options()
...
+  local help_label="$8"
+  local height="$9"
+  local width="${10}"
+  local no_index="${11}"
+  local flag="${12}"
+  local option
   local cmd
 
-  if [[ "${#_menu_list_string_array[@]}" -eq 0 ]]; then
+  if [[ "${#_options_array[@]}" -eq 0 ]]; then
```
rodrigosiqueira, line 81: How about changing `_options_array` to `_list_of_options_array`?

src/ui/patch_hub/latest_patchsets_from_mailing_list.sh
```diff
@@ -1,24 +1,58 @@
...
 function show_latest_patchsets_from_mailing_list()
 {
-  local -a new_patches
-  local fallback_message
+  local starting_index
+  local ending_index
+  local box_title
+  local extra_label='Previous'
+
+  create_loading_screen_notification "Loading patchsets from ${current_mailing_list} list (page ${PAGE})"
```
rodrigosiqueira, line 15: I'm unsure if I like the `page ${PAGE}` part... to me, it is a little bit weird that we see "page 1" and after two next, we see "page 3". If we can show something like "pages 1-2", it should be fine. If it is not possible, maybe we should not display any number to avoid misleading users.

src/lib/dialog_ui.sh
```diff
@@ -433,17 +437,30 @@ function create_choice_list_screen()
...
-    else
-      cmd+=" $'${choice}' $'${choice_description}' 'off'"
-    fi
-    ((index++))
-  done
+  # Add choices without description if `_choices` is an indexed array
+  if [[ "$(typeset -p "$3")" =~ 'declare -a' ]]; then
+    for choice in "${_choices[@]}"; do
+      choice=$(str_escape_single_quotes "$choice")
+      if [[ "${_check_statuses["$index"]}" == 1 ]]; then
```
rodrigosiqueira, line 444: I believe that the reason that the default option was not working correctly came from the fact that you mapped one array (`_check_statuses`) against an associative array (`_choices`), and associative arrays do not follow the same order as a regular array. Before your change, you were trying to use an array (index) to match an associative array order that can not be guaranteed. If you change `_check_status` to an associative array, and require the user to use the same key for `_check_statuses` and `_choices`, you can map the default options.

src/lib/lore.sh
```diff
@@ -318,13 +310,16 @@ function process_name()
...
+#
+# @lore_requests_timeframe: Size of the timeframe considered in days.
 function reset_current_lore_fetch_session()
 {
+  local lore_requests_timeframe="$1"
+
   list_of_mailinglist_patches=()
   PATCHSETS_PROCESSED=0
-  DAYS="$TIMEFRAME_SIZE_IN_DAYS"
+  DAYS="$lore_requests_timeframe"
```
rodrigosiqueira: btw, you can still have a fallback option in case we have any parser issue or if the user set it to a weird value. I.e., set a default case if this variable is invalid.

rodrigosiqueira (changes requested):

src/lib/lore.sh
```diff
@@ -276,72 +312,146 @@ function reset_list_of_mailinglist_patches()
...
+#	    <title>[PATCH] drm/amdkfd: Add missing tba_hi programming on aldebaran</title>
+#     <updated>2023-08-09T21:27:00Z</updated>
+#	    <link href="http://lore.kernel.org/amd-gfx/20230809212615.137674-1-jay.cornwall@amd.com/"/>
+#   </entry>
+#
+# The pre-processed version of this example element would be:
+#   Jay Cornwall
+#   jay.cornwall@amd.com
+#   [PATCH] drm/amdkfd: Add missing tba_hi programming on aldebaran
+#    href="http://lore.kernel.org/amd-gfx/20230809212615.137674-1-jay.cornwall@amd.com/"
```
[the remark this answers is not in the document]
davidbtadokoro: Yes, `//entry/link/@href` in the xpath query results in this leading space. We can remove it, but don't know if this extra processing is worth (maybe I'm underthinking it, but it seems that we would have to add another `sed` in the pipe). What do you think?

src/lib/lore.sh
```diff
@@ -276,72 +312,146 @@ function reset_list_of_mailinglist_patches()
...
+function process_patchsets()
+{
+  local pre_processed_patches="$1"
+  local shared_dir_for_parallelism
+  local starting_index
+  local line
+  local processed_patchset
+  local patch_url
+  local patch_title
+  local count
```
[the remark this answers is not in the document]
davidbtadokoro: Cool! Didn't know this pattern of declaration.

src/lib/lore.sh
```diff
@@ -276,72 +312,146 @@ function reset_list_of_mailinglist_patches()
...
 
-      if is_introduction_patch "$message_id_link"; then
+      if is_introduction_patch "$patch_url"; then
         # Process each patch in parallel
-        thread_for_process_patch "$index" "$shared_dir_for_parallelism" \
-          "$processed_line" "$message_id_link" "$title" &
-        ((index++))
+        thread_for_process_patch "$PATCHSETS_PROCESSED" "$shared_dir_for_parallelism" \
+          "$processed_patchset" "$patch_url" "$patch_title" &
+        ((PATCHSETS_PROCESSED++))
```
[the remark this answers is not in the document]
davidbtadokoro: In a later commitl, this becomes more evident, but my intetion was to use this global variable to hold the "state" across many calls of `fetch_latest_patchsets_from` (and its subroutines). We could shift the responsability to the caller (`PATCHSETS_PROCESSED` becomes an argument for `fetch_latest_patchsets_from` and its subroutines), but I thought this broke the MVC pattern that is loosely implemented and because I think the caller would have to use a global variable also.

src/lib/lore.sh
```diff
@@ -352,29 +462,102 @@ function processing_new_patches()
...
+  last_patch_timestamp=''
+  xml_result_file_name="${target_mailing_list}-patches.xml"
+  PATCHSETS_PROCESSED=0
+  flag=${flag:-'SILENT'}
+
+  while [[ "$PATCHSETS_PROCESSED" -lt "$NUMBER_OF_PATCHSETS_PER_FETCH" ]]; do
+    # Building URL for querying lore servers for a xml file with patches.
+    lore_query_url=$(compose_lore_query_url_with_verification "$target_mailing_list" "$days" "$last_patch_timestamp")
+    ret="$?"
+    [[ "$ret" != 0 ]] && return "$ret"
```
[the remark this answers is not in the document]
davidbtadokoro: `compose_lore_query_url_with_verification` returning something other than 0 means that either one of its three arguments is broken (empty or invalid), which would happen in any next iteration of the while loop. Maybe `fetch_latest_patchsets_from` "knowing" this is a sign of high coupling, though...

src/lib/lore.sh
```diff
@@ -352,29 +462,102 @@ function processing_new_patches()
...
+
+    # Request xml file with patches.
+    download "$lore_query_url" "$xml_result_file_name" "$CACHE_LORE_DIR" "$flag"
+    ret="$?"
+    [[ "$ret" != 0 ]] && return "$ret"
```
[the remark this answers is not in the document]
davidbtadokoro: In the same line of my last response, a bad URL means bad URLs for any of the next iterations of the while loop. However, I agree that if a download failed because of network issues, for example, we should retry to download it. Do you have an idea to fine grain those errors to treat them accordingly?

src/lib/lore.sh
```diff
@@ -352,29 +462,102 @@ function processing_new_patches()
...
+    # In this case, we expand the timeframe and make another fetch.
+    if is_html_file "${CACHE_LORE_DIR}/${xml_result_file_name}"; then
+      days="$((days + TIMEFRAME_IN_DAYS))"
+      continue
+    fi
+
+    raw_xml=$(< "${CACHE_LORE_DIR}/${xml_result_file_name}")
+
+    # Processing patches into patchsets that will be stored in `list_of_mailinglist_patches`.
+    pre_processed_patches=$(pre_process_xml_result "$raw_xml")
```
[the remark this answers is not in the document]
davidbtadokoro: I was wrong in my previous guess of the size of `raw_xml`. At most, it has 800 lines, relating to the maximum 200 patches in lore's response, but I do agree on passing the path instead.

src/ui/patch_hub/latest_patchsets_from_mailing_list.sh
davidbtadokoro: Hmm, you are right. But I don't know if we can know in advance the number of patchsets fetched, as this number fluctuates across queries. Think the best approach for now is to not display the number.

src/lib/dialog_ui.sh
davidbtadokoro: Sorry for the messy commit message, but the issue wasn't that `_choices` and `_check_statuses` were not in sync, as I make the caller populate `_check_statuses` in the same manner that we do the matching in this function, i.e., with a foreach loop which has a consistent order when looping through keys of an associative array across multiple "calls". The problem was that this order of keys isn't the same as the order of declaration, neither the lexicographic order, like you can see in this screenshot:

Screen Shot 2023-08-20 at 18 34 34

This made it virtually impossible for the caller to define an order of displaying the choices with an associative array, which was the problem that I tried tackling with this commit.

rodrigosiqueira:
Thanks for this improvement. Change merged into the unstable branch.

Issue #897, opened 2023-08-16
Unexpected output shown in kw ssh


aquilamacedo:
**Description:**

When running the `kw ssh --command 'ls'` command, I notice that it is displaying information that should only appear when using the `--verbose` option.

![image](https://github.com/kworkflow/kworkflow/assets/43859386/b7918f49-6dad-476a-a504-82ac1f01c6e1)

While analyzing and running the code involving the `kw_ssh_main` function, I've encountered an issue when setting the value of the flag variable using `local flag="$1"`. This occurs when a command is executed in the terminal with arguments that include command-line options, such as `kw ssh --command 'ls -la'`. In this scenario, the command interpreter treats `--command` as the first argument of the function, resulting in incorrect behavior and breaking the flag logic.

![image](https://github.com/kworkflow/kworkflow/assets/43859386/accc4e2b-e4fc-4da4-bbb4-7c8cecafb81e)

**Expected behavior:**

The absence of any verbose mode display, just plain command execution.

PR #898, opened 2023-08-16
src: kw_ssh: Fixing unexpected output shown in kw ssh.

aquilamacedo:
This commit addresses the problem where the `kw ssh` command was incorrectly displaying verbose mode output even when the --verbose option was not used.

Closes: #897

### DIFF
--- a/src/kw_ssh.sh
+++ b/src/kw_ssh.sh
@@ -14,7 +14,7 @@ declare -gA options_values
 #   attempt to execute a command or script on the remote host.
 function kw_ssh_main()
 {
-  local flag="$1"
+  local flag
   local port
   local script_path
   local cmd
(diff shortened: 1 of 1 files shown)


melissawen:
From the meeting Aug 18th:
- It seems an issue around capturing `flag` as the first parameter and shift doesn't seem the right solution. You have to investigate better the current behaviour. It mat be related to setting `flag` as `silent` in a different point of the workflow.
- Also, improve commit message to describe better the issue and change

aquilamacedo:
Hi @rodrigosiqueira @melissawen. I found a simpler solution to this problem, setting `flag=${options_values['TEST_MODE']}` for the test mode, I believe it is simpler this way. I also improved the issue message, to better detail this problem.

rodrigosiqueira:
Thanks

Issue #857, opened 2023-06-08
Verbose mode for kw options


aquilamacedo:
**Description:**

Verbose mode is an option in kw that allows the user to see the commands executed behind the scenes. This functionality is very useful during the debugging process, allowing you to identify possible errors more easily.

**TODO list:**

- [x] kw backup verbose (60eda4404d99b12e7602086c537e23c8a21a07fc)
- [x] kw env verbose (c0efb7fb25a2dbe3938a571e718dd10b8d900045)
- [x] kw drm verbose (a42592a5fa7c6704d62f3d08f2486d1964223887)
- [x] kw explore verbose (7f34cac9c1ee6cb7c0c5f02aed2233910e820454)
- [x] kw init verbose (7c4cfbe60bbd98a35a337c702e29b08703a9d2ca)
- [x] kw mail verbose (dabdbcc6e04eee03b90dcebeae1dd6bb0c0cb32f)
- [x] kw build verbose (06e21a09e7d2b88b61d9d51295aecd151133cc5c)
- [x] kw deploy verbose
- [x] kw ssh verbose
- [x] kw sef-update verbose
- [x] kw diff verbose
- [x] kw config verbose
- [x] kw kernel-config-manager verbose
- [x] kw pomodoro verbose
- [x] kw report verbose
- [x] kw device verbose
- [x] kw maintainers verbose
- [x] kw bd verbose
- [x] src/lib/kw_db.sh
- [x] kw codestyle verbose
- [x] kw debug verbose
- [x] kw clear-cache verbose

**Targeted users:**

This verbose functionality can be useful for users as well as maintainers and developers.

**API draft:**

Usage example:

```
kw ssh --verbose
```
```
kw env -c MACHINE_X86 --verbose
```

fernandohjunqueira:
Me and @Andregnl are working on the `kw config` verbose mode.

aquilamacedo:
I'm working in verbose mode in `kw db`

PR #903, opened 2023-08-21
src: self_update: Adding verbose mode.

aquilamacedo:
This commit adds support for the verbose parameter within `kw self-update`. The verbose parameter gives details of the commands that are executed behind the scenes.

Note: This is part of the issue: #857


rodrigosiqueira:
Nice change! Merged into the unstable branch.

Issue #800, opened 2023-03-10
Enable query for specific string from lore


rodrigosiqueira:
Lore provides some API for search in the public repository; let's add this option to upstream-patches-ui.

PR #904, opened 2023-08-25
`patch-hub`: Add query based on string functionality

davidbtadokoro:
This PR adds queries based on a string to `kw patch-hub`. Lore API provides a great number of filters to match strings against specific fields of the message, like 'Subject', 'To', 'Body', and even the whole message.

The PR adds queries for strings that are match against the whole message. The query is made in the 'Search String in Lore' menu inside 'Dashboard' through an input box.

Closes: #800

### DIFF
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -27,6 +27,8 @@ function ui_setup()
 
   DEFAULT_WIDTH="$columns"
   DEFAULT_HEIGHT="$lines"
+  DEFAULT_SMALL_WIDTH="$((columns / 2))"
+  DEFAULT_SMALL_HEIGHT="$((lines / 3))"
 
   # Set sefault layout
   if [[ -n "$default_layout" ]]; then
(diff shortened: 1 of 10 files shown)


davidbtadokoro:
Hi @rodrigosiqueira.

Based on our offline conversation, I remade this PR focusing on a search in the whole message first.

src/lib/dialog_ui.sh
```diff
@@ -679,6 +679,71 @@ function create_rangebox_screen()
...
+  local help_label="$5"
+  local height="$6"
+  local width="$7"
+  local flag="$8"
+  local cmd
+
+  cancel_label=${cancel_label:-'Exit'}
+  flag=${flag:-'SILENT'}
+  height=${height:-'10'}
+  width=${width:-'60'}
```
rodrigosiqueira, line 717: We should start to use `DEFAULT_WIDTH` and `DEFAULT_HEIGHT`.

src/lib/web.sh
```diff
@@ -76,3 +76,50 @@ function is_html_file()
...
+# @string: String to be encoded
+#
+# Return:
+# The function outputs the encoded string and returns 0 in any case. An empty `string`
+# value is a valid argument.
+#
+# CREDITS:
+# This function was adapted to follow kw's coding style, but it was implemented
+# by meleu (https://github.com/meleu) and it was taken from his blog
+# (https://meleu.sh/urlencode/).
```
rodrigosiqueira, line 99: Where is the source code? Is this code explicitly under GPL or a compatible license?

src/lib/web.sh
```diff
@@ -76,3 +76,50 @@ function is_html_file()
...
+      encoded_string+="$char"
+    else
+      # %% - Literal '%'
+      # %02X - Two digit hexadecimal with 0 preceding, if necessary
+      # "'$char" - ASCII number of "$char"
+      encoded_string+=$(printf '%%%02X' "'$char")
+    fi
+  done
+
+  printf '%s' "$encoded_string"
```
rodrigosiqueira, line 124: I noticed that if I try to search for "Something 3.1" the search does not work, is it possible that this function does not convert the `.` correctly?

src/lib/lore.sh
```diff
@@ -334,7 +335,8 @@ function compose_lore_query_url_with_verification()
   # TODO: We need to use the query prefix 's:Re:' to filter out replies and match
   # only real patches. Are we filtering possible patches? If no, can we filter more
   # messages to obtain a lighter response file?
-  query_filter="?q=rt:..+AND+NOT+s:Re&x=A&o=${min_index}"
+  query_filter="?x=A&o=${min_index}&q=rt:..+AND+NOT+s:Re"
+  [[ -n "$additional_filters" ]] && query_filter+="+AND+$additional_filters"
```
rodrigosiqueira, line 339: add `{}` around `additional_filters`.

src/ui/patch_hub/patch_hub_core.sh
```diff
@@ -94,7 +99,12 @@ function dashboard_entry_menu()
   local message_box
   local ret
 
-  menu_list_string_array=('Registered mailing list' 'Bookmarked patches' 'Settings')
+  menu_list_string_array=(
+    'Registered mailing list'
+    'Bookmarked patches'
+    'Search String in Lore'
```
rodrigosiqueira, line 105: Remove `String`.

src/ui/patch_hub/search_string_in_lore.sh
```diff
@@ -0,0 +1,46 @@
...
+
+  case "$ret" in
+    0) # OK
+      search_string_in_lore "$menu_return_string"
+      ;;
+    1) # Exit
+      handle_exit "$ret"
+      ;;
+    3) # Return
+      ;;
```
rodrigosiqueira, line 24: I think you forgot to add the return action.

rodrigosiqueira (changes requested):

src/lib/web.sh
davidbtadokoro: Found that the source reference is https://github.com/dylanaraps/pure-bash-bible#percent-encode-a-string.

src/lib/web.sh
davidbtadokoro: Found that the source reference is https://github.com/dylanaraps/pure-bash-bible#percent-encode-a-string.

src/lib/web.sh
davidbtadokoro: Actually, the problem wasn't in this commit, but in the next one. The function `search_string_in_lore` didn't call `url_encode` before passing the string as argument. FIxed that and added test cases to account for that. Sorry for the lack of attention here.

rodrigosiqueira (changes requested):
Overall, the code lgtm.

About this feature, it really bothers me the repetition and the long time it takes to complete. I can merge this change soon, but the condition will create an issue for removing the duplications (it should be a high priority) and another for improving the delay (maybe for this search, we can have a simple list that just shows the patch's name).

davidbtadokoro:
Hi @rodrigosiqueira, thanks a lot for the feedback! Going to update it ASAP.

> About this feature, it really bothers me the repetition and the long time it takes to complete. I can merge this change soon, but the condition will create an issue for removing the duplications (it should be a high priority) and another for improving the delay (maybe for this search, we can have a simple list that just shows the patch's name).

I'm working on a way to optimize the overall fetch mechanism of `patch-hub`, and I think that the duplication will be solved as a side-effect. In any case, I agree with you, we need to address these 2 points with high priority.

rodrigosiqueira:
Hi,
I merged this change. Please create the two issues that I told you in the review.

Thanks a lot for all your effort!

PR #906, opened 2023-09-02
Initially block LKML

### LINKED ISSUE #755: Remove `linux-kernel<at>vger` from `kw mail` recipients
**Description:**

Unfortunately, the kernel community watching `linux-kernel@vger.kernel.org` is too broad where people with toxic behaviors may participate in discussions. I was thinking of removing the automatic addition of this list in generate_kernel_recipients(), since many don't want this kind of interaction.

**Type:** Refactor

**Why:** even though the ML is listed in the get maintainers script, we usually send patches to specific subsystems that provide relevant feedback on the proposed changes. Many don't want to participate in discussions with other subsystem communities that don't follow a code of conduct and being exposed to toxic behaviors.

melissawen:
`kw mail` automatically adds LKML as a receipt of all patches since this is recommended by Kernel Doc, but Kernel Doc also requires to not send more than 15 patches at once to the vger mailing lists and states that the volume on that list has caused a number of developers to tune it out. To prevent users from being surprised by this automatic addition, initialize `mail.config` with LKML blocked. Users can restore the default behavior once they are more familiar with the `kw mail` feature and its options.

Closes #755

### DIFF
--- a/etc/mail.config
+++ b/etc/mail.config
@@ -2,5 +2,7 @@
 send_opts=--annotate --cover-letter --no-chain-reply-to --thread
 
 # You can choose to block certain emails from being added to the recipients
-# list of your patches, separate values with commas
-#blocked_emails=
+# list of your patches, separate values with commas. Linux Kernel ML is
+# initially blocked to avoid mistakenly send more than 15 patches at once to
+# the vger mailing lists. Configure it according to your needs.
+blocked_emails=linux-kernel@vger.kernel.org


davidbtadokoro (commented):
Hi @melissawen, thanks for the change! Can't judge on the motivation of the change, but, by the commit message, it make sense to me. In any case, tested it with `--simulate` and it doesn't add LKML to CC. Also, don't know if we should add a test for this, as a bad implementation may lead to actually sending mails by mistake, but, if not, maybe we should do it.

Teste-by: David Tadokoro \<davidbtadokoro@usp.br\>

rodrigosiqueira:
Hi,
I'll merge this change since it is a simple change in the config file. It would be nice if we open an issue to make kw aware of patches with more than 15 patches and based on that, drop lkml. For now, let's block it every time.

PR #907, opened 2023-09-03
src: plugins: fix local check for command existence

### LINKED ISSUE #713: kw deploy fails to update grub config
**Description:**

Running `kw bd` for local deployment (default_deploy_target=local), the kernel deployment fails in the last step of updating grub config that is expected to run the `grub-mkconfig -o /boot/grub/grub.cfg` command. In addition, we can see this output  `There is no grub-mkconfig command in the system.` The issue seems to be around the command_exists() implementation that doesn't include a `sudo -E` prefix for checking command existence. command_exists() is called by  define_grub_cmd_update() that returns in run_bootloader_update() and aborts the deployment execution.

**How to Reproduce:**

1. set `default_deploy_target=local` in .kw/deploy.config (using --local is likely the same)
2. run kw bd

**Expected behavior:**

grub-mkconfig exists in a x86 machine, so it's expected the command being detected and deployment completes clearly.
In fact, I manually included `sudo` prefix in the grub_cmd of define_grub_cmd_update() and it fixed the issue.

**System information:**

 - OS: Debian GNU/Linux bookworm/sid
 - kw Version: alpha (I'm on top of steamos PR #618) 
 - Bash Version: version 5.2.0(1)-rc2 (x86_64-pc-linux-gnu)

melissawen:
define_grub_cmd_update() fails to check any grub-update commands in local deployment. This happens because grub commands should be run by a privilege user (root) and also because command_exists() is not performing correctly for unknown reason. To overcome this issue, check command as a root user by running `sudo -u root -i command -v`. It works correctly for both local/sudo and remote/root.

Closes #713


rodrigosiqueira:
Hi @melissawen , thanks a lot for this PR. We talked off-line, and from our conversation, I created this PR:

https://github.com/kworkflow/kworkflow/pull/912

I'll close this PR since the above PR addresses the issue that you fixed in this PR.

PR #909, opened 2023-09-06
src: mail: Add 'additional_emails' configuration

davidbtadokoro:
`kw mail` already has a configuration to automatically block emails as recipients named 'blocked_emails'. An opposite configuration, to automatically add emails as recipients doesn't exist, though.

In this context, add the configuration 'additional_emails' to `kw mail` to automatically add pre-determined recipients.

### DIFF
--- a/etc/mail.config
+++ b/etc/mail.config
@@ -6,3 +6,8 @@ send_opts=--annotate --cover-letter --no-chain-reply-to --thread
 # initially blocked to avoid mistakenly send more than 15 patches at once to
 # the vger mailing lists. Configure it according to your needs.
 blocked_emails=linux-kernel@vger.kernel.org
+
+# You can add emails to always be included as recipients. Emails should be
+# separated by comma. There is a separate config for the To: and CC: recipients.
+default_to_recipients=
+default_cc_recipients=
--- a/documentation/man/features/mail.rst
+++ b/documentation/man/features/mail.rst
@@ -32,6 +32,11 @@ the union of the recipients of each patch as the recipients of the cover-letter.
   *kworkflow.config* file.
 
 .. note::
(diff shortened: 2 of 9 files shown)


etc/mail.config
```diff
@@ -6,3 +6,7 @@ send_opts=--annotate --cover-letter --no-chain-reply-to --thread
 # initially blocked to avoid mistakenly send more than 15 patches at once to
 # the vger mailing lists. Configure it according to your needs.
 blocked_emails=linux-kernel@vger.kernel.org
+
+# You can add emails to always be included as recipients. Emails should be
+# separated by comma
+additional_emails=
```
rodrigosiqueira, line 12: I don't think this field expresses the idea behind it really well. Maybe we can try something like `demails_always_included` (it is still a bad name, it is just an example)...

We also should have one for `to` and another one for `cc`.

rodrigosiqueira (changes requested):
In general, it lgtm. Remember to add a test and reference @melissawen issue in the commit message. Also, expand the man documentation.

davidbtadokoro:
Hey @rodrigosiqueira, thanks for the feedback!
I updated the PR addressing your comments.
In respect to the configurations name, I used `default_to_recipients` and `default_cc_recipients`. What do you think?

melissawen:
names sounds good to me.

rodrigosiqueira:
Overall, it lgtm. @melissawen I'll wait for your okay for this change before I merge it.

melissawen:
Just tested and it works as expected. I noticed that when I add the same e-mail address to `default_to_recipients` and `default_cc_recipients` it doesn't repeat and the list of e-mails in `To:` is prioritized. All good, you can add my t-b:
melissa.srw \<at\> gmail.com

PR #913, opened 2023-09-09
Fix device feature

rodrigosiqueira:


src/device_info.sh
```diff
@@ -607,40 +575,63 @@ function show_data()
...
+      --local)
+        options_values['TARGET']="$LOCAL_TARGET"
+        shift
+        ;;
+      --) # End of options, beginning of arguments
+        shift
+        ;;
+      *)
+        complain "Invalid option: ${1}"
+        exit 22 # EINVAL
```
davidbtadokoro, line 626: I think you just need to assign the invalid option to `options_values['ERROR']` and return 22.
Because of your last commit, the `complain` line is redundant and the `exit` line halts the execution without the chance to throw an error message and call `device_info_help`.

src/device_info.sh
```diff
@@ -341,9 +320,9 @@ function get_gpu()
       done
       ;;
     3) # REMOTE_TARGET
-      pci_addresses=$(cmd_remotely "$cmd_pci_address" "$flag" "$ip" "$port")
+      pci_addresses=$(cmd_remotely "$cmd_pci_address" "$flag")
       for g in $pci_addresses; do
-        gpu_info=$(cmd_remotely "lspci -v -s $g" "$flag" "$ip" "$port")
+        gpu_info=$(cmd_remotely "lspci -v -s $g" "$flag")
```
davidbtadokoro, line 320: Use `{}` aroung `g`.

src/device_info.sh
```diff
@@ -451,9 +422,9 @@ function get_chassis()
       fi
       ;;
     3) # REMOTE_TARGET
-      cmd_remotely "test -f $dmi_file_path" "$flag" "$ip" "$port"
+      cmd_remotely "test -f $dmi_file_path" "$flag"
```
davidbtadokoro, line 420: Use `{}` around `dmi_file_path`

davidbtadokoro (changes requested):
Hi @rodrigosiqueira, thanks for this change!
Left some comments just on the second commit.

src/device_info.sh
```diff
@@ -607,40 +575,63 @@ function show_data()
...
-      exit 0
-      ;;
-    *)
-      complain "Invalid option: $option"
-      return 22 # EINVAL
-      ;;
-  esac
+  remote_parameters['REMOTE_IP']=''
+  remote_parameters['REMOTE_PORT']=''
+  remote_parameters['REMOTE_USER']=''
```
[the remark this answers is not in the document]
rodrigosiqueira: It is, the function `populate_remote_info ''` will use those variables.

src/device_info.sh
```diff
@@ -607,40 +575,63 @@ function show_data()
...
+  fi
+
+  populate_remote_info ''
+  if [[ "$?" == 22 ]]; then
+    options_values['ERROR']="Invalid remote: $remote"
+    return 22 # EINVAL
+  fi
+
+  eval "set -- $options"
+  while [[ "$#" -gt 0 ]]; do
```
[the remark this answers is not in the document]
rodrigosiqueira: Nice catch. I handled it by adding the help check in the device_info_main like we do on other parts of the code.

src/device_info.sh
rodrigosiqueira: nice catch!

rodrigosiqueira:
@davidbtadokoro thanks a lot for your feedback, I updated my branch based on your comments.

src/device_info.sh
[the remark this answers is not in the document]
davidbtadokoro: Oh, I see it. Thanks!

src/device_info.sh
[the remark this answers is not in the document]
davidbtadokoro: Nice!

src/device_info.sh
davidbtadokoro: Think that you forgot to update this line.

davidbtadokoro (changes requested):
Thanks for the update @rodrigosiqueira!
Apart from a minor nitpick everything looks legit.

src/device_info.sh
davidbtadokoro: @rodrigosiqueira here is the nitpick : )

rodrigosiqueira:
Thanks a lot for your feedback and review.

PR #915, opened 2023-09-09
src: kernel_install: add pacman-key init to arch pre-setup

melissawen:
Initizalizing keyring fits pacman setup guideline at: https://wiki.archlinux.org/title/Pacman/Package_signing

Add the missing steps to steamOS dev environment configuration.

Note: a previous step of restarting pacman-init service fails on steamOS, since there is no pacman-init service there. I'm not sure about other arch-base OSes and it's harmless on steamOS, so I didn't remove it. I don't have another arch-based system, better double-check with someone that has it.


rodrigosiqueira:
Thanks a lot for your patch. I already merged it into the unstable branch.

PR #918, opened 2023-09-16
src: _kw: Add Zsh completion for `kw build` `--full-cleanup` option

davidbtadokoro:
Add Zsh completion for the `--full-cleanup` option of the `kw build` feature.


davidbtadokoro:
@aquilamacedo I assumed that `--full-cleanup` should be mutually excludent to all other options besides `--verbose`, in other words, when using `--full-cleanup` no other option should be displayed aside from `--verbose`.

Is this correct? And if so, can you check if this PR adds this behaviour correctly?

davidbtadokoro:
Thanks for the change! Merged into the unstable.

rodrigosiqueira:
@davidbtadokoro feel free to merge this change in the unstable branch.

davidbtadokoro:
After feedback approving the behaviour from @aquilamacedo, I merged it on unstable.

PR #926, opened 2023-09-25
`run_tests`: fix tests on non-english machines

### LINKED ISSUE #916: Tests fail on non-english locales
When tests are ran with a different locale, the tests fail because of the language. Examples of failing tests are below. Locale is pt_BR.UTF-8.

**How to Reproduce:**

Set locale to pt_BR.UTF-8. Run ./run_tests.sh. Some tests will fail with translations of expected output. Examples:

test_kw_parse_get_errors
ASSERT:line 538, statement 0:
Expected: "kw: unrecognized option '--fee'"
but got:  "kw: opção não reconhecida “--fee”"

ASSERT:line 538, statement 1:
Expected: "kw: option '--xpto' requires an argument"
but got:  "kw: a opção “--xpto” requer um argumento"

**Expected behavior:**

Clearly, these tests should pass, since the error is the same as the expected one.

**System information:**

 - OS: Arch Linux
 - kw Version: fdf96de
 - Bash Version: 5.1.16(1)-release (x86_64-pc-linux-gnu)

**Additional context:**

jooag:
On machines with non-english locales, some error messages generated are printed in that language. The tests assertions, thus, fail inappropriately. In this context, set LANGUAGE environment variable before running the tests.

Closes #916.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -107,7 +107,8 @@ if [[ "$#" -eq 0 ]]; then
   # Note: Usually we want to use double-quotes on bash variables, however,
   # in this case we want a set of parameters instead of a single one.
   strip_path $files_list
-  run_tests
+  # Note: LANGUAGE environment variable must be set in order to avoid error messages in other languages.
+  LANGUAGE=en_US.UTF_8 run_tests
 elif [[ "$1" == 'list' ]]; then
   index=0
   files_list=$(find ./tests/ -name '*_test.sh')
@@ -118,7 +119,7 @@ elif [[ "$1" == 'list' ]]; then
   done
 elif [[ "$1" == 'test' ]]; then
   strip_path "${@:2}"
-  run_tests
+  LANGUAGE=en_US.UTF_8 run_tests
 else
   show_help
 fi


rodrigosiqueira:
@davidbtadokoro

run_tests.sh
```diff
@@ -107,7 +107,8 @@ if [[ "$#" -eq 0 ]]; then
   # Note: Usually we want to use double-quotes on bash variables, however,
   # in this case we want a set of parameters instead of a single one.
   strip_path $files_list
-  run_tests
+  # Note: LANGUAGE environment variable must be set in order to avoid error messages in other languages.
+  LANGUAGE=en_US.UTF_8 run_tests
```
davidbtadokoro, line 111: You also need to do this in line 122 to avoid single test runs like

```bash
./run_tests.sh test kwlib_test
```

from failing.

davidbtadokoro (changes requested):
This is really a great catch! Thanks @jooag and @GuilhermeMoreno.

Aside from my inline comment that I left, in the commit message header, you used \` in \`run_tests\`: fix [...]. See if you can remove it.

At last, don't forget to add the `Signed-off-by` tag (see https://kworkflow.org/content/howtocontribute.html#certify-your-work-by-adding-your-signed-off-by-line)

Nevertheless, tested it and it worked! Although it is a small change it has a great impact 👍

davidbtadokoro (approved):
Hey @jooag and @GuilhermeMoreno, and thanks for the update.

I have just 3 minor nitpicks about the commit message:
1. Try breaking line in the commit message body to avoid surpassing 72 characters;
2. Don't add a blank line between the `Co-authored-by` and `Signed-off-by`;
3. I think (@rodrigosiqueira can verify this) you don't need to use `Co-authored-by` for the same person that is referenced in the `Signed-off-by`.

Fixed those as they are really small nitpicks, but don't forget to mind those in the future.

Nevertheless, everything LGTM, so, merged into the unstable : )

PR #929, opened 2023-09-28
Standardization and introducing a verbose mode for kw diff

aquilamacedo:
This commit standardizes the structure of the `diff.sh`
file and adds the verbose mode option to `kw diff`

### DIFF
--- a/src/diff.sh
+++ b/src/diff.sh
@@ -1,17 +1,15 @@
-include "${KW_LIB_DIR}/lib/kwio.sh"
 include "${KW_LIB_DIR}/lib/kwlib.sh"
+include "${KW_LIB_DIR}/lib/kwio.sh"
 
-# Hash containing user options
-declare -gA diff_options
+declare -gA options_values
 
-function diff_manager()
+function diff_main()
 {
-  local files_paths=${*: -2}
-  local interactive
-  local target_1
-  local target_2
+  local target_a="$1"
+  local target_b="$2"
+  local flag
 
-  IFS=' ' read -r -a files <<< "$files_paths"
+  IFS=' ' read -r -a files <<< "${target_a} ${target_b}"
   for file in "${files[@]}"; do
     if [[ "$file" =~ -h|--help ]]; then
       diff_help "$file"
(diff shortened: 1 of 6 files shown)


src/diff.sh
```diff
@@ -1,98 +1,49 @@
...
-  local files_paths=${*: -2}
-  local interactive
-  local target_1
-  local target_2
+  local target_a="$1"
+  local target_b="$2"
+  local flag
 
-  IFS=' ' read -r -a files <<< "$files_paths"
+  IFS=' ' read -r -a files <<< "$target_a $target_b"
```
rodrigosiqueira, line 12: String concatenation.

src/diff.sh
```diff
@@ -1,98 +1,49 @@
...
-  fi
-
-  interactive="${diff_options['INTERACTIVE']}"
-  test_mode="${diff_options['TEST_MODE']}"
+  interactive="${options_values['INTERACTIVE']}"
+  test_mode="${options_values['TEST_MODE']}"
 
   if [[ "$test_mode" == 'TEST_MODE' ]]; then
-    printf '%s\n' "$target_1 $target_2 $interactive"
+    printf '%s\n' "$target_a $target_b $interactive"
```
rodrigosiqueira, line 40: Use one `%s` per variable, and separate those variables from the double-quotes.

src/diff.sh
```diff
@@ -144,30 +95,81 @@ function diff_side_by_side()
...
   if [[ ! -f "$file_1" || ! -f "$file_2" ]]; then
-    complain "Make sure that $file_1 and $file_2 are a valid files"
+    diff_help
     return 22 # EINVAL
   fi
 
   diff_cmd="$diff_cmd $file_1 $file_2"
   if [[ "$interactive" == 1 ]]; then
-    diff_cmd="$diff_cmd | less -R"
+    diff_cmd="${diff_cmd} | less -R"
```
rodrigosiqueira, line 111: Use `--RAW‐CONTROL‐CHARS` instead of `-R`

src/diff.sh
```diff
@@ -144,30 +95,81 @@ function diff_side_by_side()
...
   fi
 
   cmd_manager "$flag" "$diff_cmd"
 }
 
+# This function gets raw data and based on that fill out the options values to
+# be used in another function.
+#
+# @raw_options String with all user options
```
rodrigosiqueira, line 120: fwiu, you dropped this variable.

rodrigosiqueira (changes requested):

src/diff.sh
davidbtadokoro: @aquilamacedo, considering the latest force-push, I think you forgot to also replace `'%s\n'` for `'%s %s %s\n'`

davidbtadokoro (changes requested):
Hi @aquilamacedo, thanks so much on the change. Besides the comment that I left on the conversation that I reopened, you should remove the full stop in both commit messages headers.

Apart from that, everything looks good to me 👍

src/diff.sh
```diff
@@ -33,17 +33,20 @@ function diff_main()
   interactive="${options_values['INTERACTIVE']}"
   test_mode="${options_values['TEST_MODE']}"
 
+  [[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
+  flag=${flag:-'SILENT'}
```
davidbtadokoro, line 37: Maybe add this line as the first command of the function after local variables declaration to follow the "pattern" you used in other verbose introductions.

aquilamacedo:
> Hi @aquilamacedo, thanks so much on the change. Besides the comment that I left on the conversation that I reopened, you should remove the full stop in both commit messages headers.
> 
> Apart from that, everything looks good to me 👍

Thanks for the review @davidbtadokoro, I just made this changes! :)

davidbtadokoro (approved):
Hey @aquilamacedo and thank you so much for the update.
Everything looks legit to me!
@rodrigosiqueira I'll wait for your greenlight

davidbtadokoro:
Actually, there's another issue with our old friend Zsh completions system 😪

rodrigosiqueira:
@aquilamacedo could you address David's comment?

aquilamacedo:
> @aquilamacedo, actually, there's another issue with our old friend Zsh completions system 😪
> 
> Because the way `diff_main` is implemented, arguments 1 and 2 **must** be paths. Currently, Zsh suggest completions for `--verbose` and `--no-interactive` for args 1 and 2 (type `kw diff -` and hit tab to see). Below is the diff to fix this:
> 
> ```
> diff --git a/src/_kw b/src/_kw
> index ca531b1..d133339 100644
> --- a/src/_kw
> +++ b/src/_kw
> @@ -228,9 +228,16 @@ _kw_device()
>  _kw_df() { _kw_diff }
>  _kw_diff()
>  {
> +  local verbose=''
> +  local no_interactive=''
> +  if [[ "$CURRENT" -gt 3 ]]; then
> +    verbose='(--verbose)--verbose[enable verbose mode]'
> +    no_interactive='--no-interactive[displays all diff in two columns at once]'
> +  fi
> +
>    _arguments : \
> -    '(--verbose)--verbose[enable verbose mode]' \
> -    '--no-interactive[displays all diff in two columns at once]' \
> +    $verbose \
> +    $no_interactive \
>      '1: :_files' \
>      '2: :_files'
>  }
> ```
> 
> Just as a curiosity, what I'm doing here is displaying `--verbose` and `--no-interactive` only if the global variable `CURRENT` (from Zsh native completions system) is greater than 3. Just don't ask me why it needs to be greater than 3 rather than 2 🤣 (my hunch is that the buffer has `diff <target_a> <target_b>`, when `CURRENT = 3`).

I just made the changes. I also admit that the zsh code is a bit confusing 🤣

aquilamacedo:
> @aquilamacedo could you address David's comment?

Hi @rodrigosiqueira. I just addressed his comment :)

rodrigosiqueira:
Nice! @davidbtadokoro , if you are okay with this final version, could you merge this change?

Reviewed-by: Rodrigo Siqueira <siqueirajordao@riseup.net>

davidbtadokoro:
Hey @aquilamacedo and @rodrigosiqueira. Everything was okay to me!
Merged into unstable branch 👍

PR #930, opened 2023-09-30
Fix kernel-config-manager to work with env

rodrigosiqueira:


src/kernel_config_manager.sh
```diff
@@ -734,6 +747,10 @@ function parse_kernel_config_manager_options()
         options_values['TARGET']="$REMOTE_TARGET"
         shift 2
         ;;
+      --verbose)
+        options_values['VERBOSE']=1
+        shift
+        ;;
```
davidbtadokoro, line 753: You forgot to add `verbose` to the `long_options` definition. Without it, the `--verbose` option gets denied by `kw_parse` before getting here.

davidbtadokoro (changes requested):
Hi @rodrigosiqueira and thanks for the contribution!
Overall it looks very nice. Left some comments, see if you can address those.
Also, going to add a comment below with the diff of both Bash and Zsh completions.

src/kernel_config_manager.sh
```diff
@@ -43,6 +43,8 @@ function kernel_config_manager_main()
     exit 22 # EINVAL
   fi
 
+  [[ -n "${options_values['VERBOSE']}" ]] && flag='HIGHLIGHT_CMD'
```
davidbtadokoro, line 46: `flag` is setted by default to `SILENT` at line 22. Because of the "pattern" used by @aquilamacedo when [introducing verbose to the features](https://github.com/kworkflow/kworkflow/blob/b8129dc38e8c8f1ca1a3e29e787f9fa4c69d641a/src/self_update.sh#L14C1-L14C1), I think you could change line 22 to be only
```bash
local flag
```
and add the following line as the first command of the function
```bash
flag=${flag:-'SILENT'}
```

davidbtadokoro:
@rodrigosiqueira, below is the diff to add completions for both Bash and Zsh

```
diff --git a/src/_kw b/src/_kw
index 3a47db1..cbeea6b 100644
--- a/src/_kw
+++ b/src/_kw
@@ -317,6 +317,7 @@ _kw_kernel-config-manager()
   fi
 
   _arguments : \
     '(-s --save -r --remove -l --list --get --fetch)'{-s,--save}'[create a snapshot of the .config file inside the current dir in kw management]: : ' \
     '(-r --remove -s --save -l --list --get --fetch)'{-r,--remove}'[remove an existing .config file in kw management]: : ' \
     '(-l --list -s --save -r --remove --get --fetch)'{-l,--list}'[lists all the .config file versions available]' \
diff --git a/src/bash_autocomplete.sh b/src/bash_autocomplete.sh
index 97104b2..6da600f 100644
--- a/src/bash_autocomplete.sh
+++ b/src/bash_autocomplete.sh
@@ -45,7 +45,8 @@ function _kw_autocomplete()
   kw_options['m']="${kw_options['mantainers']}"
 
   kw_options['kernel-config-manager']='--force --save --description --list --get
-                                       --remove --fetch --output --optimize --remote'
+                                       --remove --fetch --output --optimize --remote
+                                       --verbose'
   kw_options['k']="${kw_options['kernel-config-manager']}"
 
   kw_options['config']='--local --global --show --help --verbose'
```

rodrigosiqueira:
Hi @davidbtadokoro , thanks a lot for your feedback. I address all your comments, let me know if it looks good to you.

davidbtadokoro:
Thanks a lot for the update @rodrigosiqueira!
Merged into the unstable branch 👍

PR #936, opened 2023-10-10
src: pomodoro: Add verbose mode

aquilamacedo:
Add support for the verbose parameter within `kw pomodoro`. The verbose parameter gives details of the commands that are executed behind the scenes.

Note: This is part of the issue: #857


rodrigosiqueira:
I guess the pomodoro feature will depend on the verbose in the db feature to be more helpful. Anyway, it merged into the unstable branch.

Issue #596, opened 2022-04-24
Installation fails if package databases aren't up to date


isinyaaa:
**Description:**

Even though the setup fails to install some dependencies, the `kw` setup script continues installing.

**How to Reproduce:**

1. Attempt to install `kw` on an out of date system (maybe if you're offline that'd work as well
2. Then some packages won't install as they can't be found
3. setup continues as if everything were alright

**Expected behavior:**

`kw` should either stop the setup as soon as some error occurs or attempt some fix before quitting.

**Screenshots:**

<img width="1503" alt="image" src="https://user-images.githubusercontent.com/39812919/164953678-28792986-e745-45f7-bfb5-d5aa041cf962.png">

**System information:**

 - OS: Arch Linux ARM (arch-based)
 - Kw Version: unstable (dcfc948)
 - Bash Version: 5.1.16

PR #940, opened 2023-10-17
setup.sh: stop kw installation if dependencies cannot be installed.

jppaulo06:
We perform a check to see if the installation of the dependencies was successful, which exits the program if not.

For example, the user may lack sudo privileges or answer NO when asked to install the dependencies.

Closes: #596

### DIFF
--- a/.github/workflows/test_setup_and_docs.yml
+++ b/.github/workflows/test_setup_and_docs.yml
@@ -12,6 +12,12 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
+      # This is here just because git-email cannot be installed in the 
+      # Github CI otherwise.
+      - name: Add git-core PPA for git-email
+        run: |
+          sudo add-apt-repository ppa:git-core/ppa -y
+
       - name: Update the system
         run: |
           sudo apt update -y
--- a/setup.sh
+++ b/setup.sh
@@ -51,6 +51,8 @@ function check_dependencies()
   local package_list=''
   local cmd=''
   local distro
+  local ret
+  local error_message=''
 
   distro=$(detect_distro '/')
 
@@ -81,8 +83,9 @@ function check_dependencies()
 
   if [[ -n "$package_list" ]]; then
     if [[ "$FORCE" == 0 ]]; then
-      if [[ $(ask_yN "Can we install the following dependencies $package_list?") =~ '0' ]]; then
-        return 0
+      if [[ $(ask_yN "Can we install the following dependencies? ${package_list}") =~ '0' ]]; then
+        complain 'Aborting kw installation...'
+        exit 125 # ECANCELED
       fi
     fi
 
@@ -92,8 +95,19 @@ function check_dependencies()
     else
       eval "sudo $cmd"
     fi
-  fi
+    ret="$?"
 
+    # Installation failed...
+    if [[ "$ret" -ne 0 ]]; then
+      error_message='[ERROR]: Dependency installation failure: '
+      [[ "$ret" -eq 1 ]] && error_message+='Lacked sudo privileges to install the packages:'
+      [[ "$ret" -ne 1 ]] && error_message+='Something went wrong when installing the packages:'
+      error_message+=$'\n'"$package_list"
+      complain "$error_message"
+      exit "$ret"
+    fi
+
+  fi
 }
 
 function generate_documentation()


rodrigosiqueira:
@jppaulo06 could you investigate why this PR failed in the CI?

davidbtadokoro:
> @jppaulo06 could you investigate why this PR failed in the CI?

Hi @rodrigosiqueira. I was discussing this with @jppaulo06 and @uwla earlier today. We could pin down to the fact that the below error

Screen Shot 2023-10-17 at 19 41 17

always happens when running this job on the CI, as you can see in any run of https://github.com/kworkflow/kworkflow/actions/workflows/test_setup_and_docs.yml (for example, you can see it [here](https://github.com/kworkflow/kworkflow/actions/runs/6553481859/job/17798880575) by expanding the 'Check Installation' tab).

The thing is, with this PR, the installation stops if a dependency could not be resolved (which I think is the correct behavior) and exits with a non-zero code, which causes the job to fail. In other words, this PR exposed what it seems to be a long-lasting bug in the CI.

setup.sh
```diff
@@ -92,8 +93,20 @@ function check_dependencies()
...
     fi
-  fi
 
+    # installation failed...
+    local ret="$?"
+    if [[ "$ret" -ne 0 ]]; then
+      local errmsg=''
+      errmsg="[ERROR]:Dependency installation failure: "
+      [[ "$ret" -eq 1 ]] && errmsg+="lacked sudo privileges to install the packages "
+      [[ "$ret" -ne 1 ]] && errmsg+="something went wrong when installing "
```
davidbtadokoro, line 103: Use single-quotes in all these 3 lines. Also, don't forget to double check if these return codes are consistent, if you didn't already.

setup.sh
```diff
@@ -92,8 +93,20 @@ function check_dependencies()
...
+      errmsg+=$package_list
```
davidbtadokoro, line 104: Use double-quotes around `$package_list`. 

**Suggestion**: add a line break before `$package_list`, like

```bash
errmsg+=$'\n'"$package_list"
```

to make a more organized output to the user.

davidbtadokoro (changes requested):
Hey @jppaulo06  and @uwla, and thank you guys so much for this change! Making kw more robust is always welcomed and all the more for a critical part like the installation script.

Left some comments, so see if you can address them (don't hesitate to ping me for any doubt).

Besides those, please address these 3 points related to the commit message:

1. Remove the full stop at the end of the subject line;
2. Make one of the `Co-authored-by` lines a `Signed-off-by` line (see why [here](https://kworkflow.org/content/howtocontribute.html#certify-your-work-by-adding-your-signed-off-by-line));
3. Nice use of imperative form by using '**perform**', but maybe make it `Perform a check (...)` instead of `We perform a check (...)`.

**NOTE:** This PR exposed a bug on the CI, as mentioned in my comment above. I see two approachs: fix the bug as a first commit to this PR, or fix the bug in a future PR (we would have to settle with the CI fail). @rodrigosiqueira, what's your opinion on this?

rodrigosiqueira:
@davidbtadokoro @jppaulo06 let's fix the CI as part of this PR.

rodrigosiqueira:
@jppaulo06 any progress?

uwla:
We have fixed the missing git-email dependency in the CI.

But, before we accept this PR, we would like to improve the error message displayed by KW if the dependencies cannot be installed.

We will do it pretty soon.

uwla:
I think we are done here @davidbtadokoro @rodrigosiqueira. What do you guys think?

.github/workflows/test_setup_and_docs.yml
```diff
@@ -12,6 +12,10 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
+      - name: Add PPA
+        run: |
+          sudo add-apt-repository ppa:git-core/ppa -y
```
davidbtadokoro, line 17: Good and simple approach to solve this CI problem!!! 😄

However, I'm going to have to be a little more nitpicky, as I consider the bar higher for CI changes 😬.

3 things:

1. Please, add a succinct comment above the job to explain the reasoning behind adding this PPA and change the name of the job to something more specific (`Add PPA` is too broad). If you can achieve a self-explanatory job name, you can drop the comment part.
2. Just to keep the commit history consistent (no commit fails in the CI) move this commit to be the first in the series, as the CI bug you detected is independent from the current first commit.
3. I know that being a CI bug, you are limited in term of fixes, but is this fix "weak" (i.e., it may stop working in the future)? @rodrigosiqueira, what do you think?

setup.sh
```diff
@@ -100,9 +100,9 @@ function check_dependencies()
     # Installation failed...
```
davidbtadokoro, line 100: Squash the current third commit with the current first commit of the series (the one that introduced these lines).

setup.sh
```diff
@@ -100,9 +100,9 @@ function check_dependencies()
...
     if [[ "$ret" -ne 0 ]]; then
       error_message='[ERROR]: Dependency installation failure:'
-      [[ "$ret" -eq 1 ]] && error_message+='\nLacked sudo privileges to install the packages'
-      [[ "$ret" -ne 1 ]] && error_message+='\nSomething went wrong when installing the packages'
-      error_message+='\n'"$package_list"
+      [[ "$ret" -eq 1 ]] && error_message+=' Lacked sudo privileges to install the packages:'
+      [[ "$ret" -ne 1 ]] && error_message+=' Something went wrong when installing the packages:'
+      error_message+="\n$package_list"
```
davidbtadokoro, line 105: The assignment `error_message+="\n$package_list"` doesn't add a newline before the value of `package_list`. `complain "$error_message"` would result in something like

```bash
[ERROR]: Dependency installation failure: Lacked sudo privileges to install the packages:\n<list of packages>
```

Instead, you should use `error_message+=$'\n'"$package_list"` to add this leading newline.

To better understand the why of it see this excerpt from the Bash reference manual below:

Screen Shot 2023-10-30 at 22 49 58

davidbtadokoro (changes requested):
Hi @uwla and @jppaulo06, and thanks for update! Really happy you guys could manage to fix this CI problem.

We are almost at the perfect point, just some more details and we are good to go 😃

Please, see if you can address the comments that I left, and feel free to ask me for any doubts.

setup.sh
```diff
@@ -81,8 +83,9 @@ function check_dependencies()
 
   if [[ -n "$package_list" ]]; then
     if [[ "$FORCE" == 0 ]]; then
-      if [[ $(ask_yN "Can we install the following dependencies $package_list?") =~ '0' ]]; then
-        return 0
+      if [[ $(ask_yN 'Can we install the following dependencies $package_list?') =~ '0' ]]; then
```
rodrigosiqueira, line 86: https://kworkflow.org/content/codingstyle.html#string-concatenation

.github/workflows/test_setup_and_docs.yml
rodrigosiqueira: When we migrate to Sourcehut, I think this will not be an issue. However, for now, it is acceptable to use this PPA.

rodrigosiqueira (changes requested):

.github/workflows/test_setup_and_docs.yml
jppaulo06: > Good and simple approach to solve this CI problem!!! 😄
> 
> However, I'm going to have to be a little more nitpicky, as I consider the bar higher for CI changes 😬.
> 
> 3 things:
> 
> 1. Please, add a succinct comment above the job to explain the reasoning behind adding this PPA and change the name of the job to something more specific (`Add PPA` is too broad). If you can achieve a self-explanatory job name, you can drop the comment part.
> 2. Just to keep the commit history consistent (no commit fails in the CI) move this commit to be the first in the series, as the CI bug you detected is independent from the current first commit.
> 3. I know that being a CI bug, you are limited in term of fixes, but is this fix "weak" (i.e., it may stop working in the future)? @rodrigosiqueira, what do you think?

Just made some changes in the pull request, but forgot 1. and 2.
I will change it in a moment again...

davidbtadokoro:
Hey @uwla and @jppaulo06, and thank you so much for updating the PR with great attention to detail.

Retested the PR and reviewed it again, and everything looks legit to me. So merged into the unstable branch 👍

PR #941, opened 2023-10-17
Fix missing git credentials in tests

uwla:
The test suite `kwklib_test` required the git credentials `user.name` and `user.email`, but these were not defined when creating ephemeral git repositories for the tests. Rather, the tests relied on global git credentials, which if not set would cause the test to fail because it is not possible to make git commits without the mentioned credentials. This PR adds local git credentials when creating temporary git repos. That way, the test suite will pass even if no global git credentials are defined.


rodrigosiqueira:
Thanks!

PR #942, opened 2023-10-17
Fix typo in device module

uwla:
Replace `--eliminate` with `--delimiter` in cut command options in deviceinfo module, which was due to a typo.


rodrigosiqueira:
Thanks
Merged

Issue #934, opened 2023-10-05
Can't set cflags with a string that starts with - (minus)


melissawen:
**Description:**

Related to #839 

Unable to set this string as cflags "-DAMD_PRIVATE_COLOR".
I need to escape the minus symbol with a bar "\-DAMD_PRIVATE_COLOR" to get it working as expected

**How to Reproduce:**

Run kw b --cflags  "-DAMD_PRIVATE_COLOR"

PR #943, opened 2023-10-23
src: build: Fix cflags formatting issue 

aquilamacedo:
This commit solves the issue where setting the string as cflags "-DAMD_PRIVATE_COLOR" was not functioning as expected. The code has been updated the code to properly escape the minus symbol in cflags, making it work as expected

Closes: #934

### DIFF
--- a/src/build.sh
+++ b/src/build.sh
@@ -127,7 +127,7 @@ function build_kernel_main()
   command="make ${optimizations} ${llvm}ARCH=${platform_ops}${warnings}"
 
   if [[ -n "$cflags" ]]; then
-    command+=" KCFLAGS=${cflags}"
+    command+=" KCFLAGS=\"${cflags}\""
   fi
   command+="${output_kbuild_flag}${output_path}"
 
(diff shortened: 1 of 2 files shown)


aquilamacedo:
@melissawen, I made a PR related to the issue you had opened, could you review it for me?

davidbtadokoro:
Hey @melissawen, if you want, you can leave the code review to me. However, I feel like we need your greenlight for the behavior, as you opened the issue and surely knows much more about the domain.

aquilamacedo:
@davidbtadokoro and @rodrigosiqueira, if you have any comments, your feedback would be more than welcome too. :)

davidbtadokoro (changes requested):
Hi @aquilamacedo, and thanks for tackling this issue.

I have no inline comments. However, you claim to be escaping the minus `-` symbol, but it seems to me that what you are actually doing is surrounding the `cflags` value with double-quotes (e.g., `KCFLAGS="-DAMD_PRIVATE_COLOR"`). In case in not misunderstanding things, it may be best to change the commit message to express this.

Other than that, I tried reproducing the bug reported in #934, and this PR seems to fix it 👍

In any case, I feel that the greenlight of @melissawen is a must.

aquilamacedo:
Hi @davidbtadokoro. Thx for your review :). I've updated the commit message to provide a more accurate description of the changes. Could you please verify if everything is in order now?

davidbtadokoro (commented):
Hi @aquilamacedo! Read you updated commit message and it looks great to me 👍 

As I tested the behaviour in the last review and the code looks good, lets just wait for Melissa's greenlight to merged it 😬

melissawen:
Just tested it. All good! Thank you guys for fixing the issue and reviewing.

davidbtadokoro:
Hi @melissawen and thanks for testing it! @aquilamacedo, since everything seems OK, I merged your PR into the unstable. Thanks 👍

Issue #880, opened 2023-06-29
Improve UX of kw patch-hub


melissawen:
**Description:**

This is the list of things that needs improvement for someone unfamiliar with the tool (maybe we can make it an umbrella)

- [x] Documentation says `COMMAND TO INTERACT WITH THE PUBLIC MAILING LISTS` and state the same in the description. I would prefer something clearer about the limits of this. AFAIK, it only interacts with patches in the public ML, not with every kind of e-mail there: `to interact with patches in public ML`.
- [x] Homogeneous statements for settings options: or everything as an action (verb, i.e.: Register, Save, Select, etc) or everything as a noun.
- [x] `Invalid option` doesn't list `patch-hub` neither `upstream-patch-ui` as possible commands (note: `Invalid options` shows up when you type a unknown command)
- [x] Advise users about where they can find the patch after download it. Maybe the same for bookmarked patches
- [x] The way that Bookmark works in a patch window is a bit confusing. When I check the bookmark checkbox in a patch, I now it's bookmarked, but when I uncheck the checkbox, I was expected it's unbookmarked, but it's not. Currently I need to leave the window, go to the main menu and to the Bookmarked patches management window to do it.

**Type:**

Documentation and UX

**Why:**

Users are not familiar with this kind of tool, hard to predict what is going on. Needs a better explanation of each action.

PR #944, opened 2023-10-25
Improve kw patch hub ux

JGBSouza:
resolve some of the issues mentioned at #880, except for the fourth improvement

### DIFF
--- a/src/ui/patch_hub/patch_hub_core.sh
+++ b/src/ui/patch_hub/patch_hub_core.sh
@@ -193,7 +193,7 @@ function list_patches()
   local ret
 
   if [[ -z "${_target_array_list}" ]]; then
-    create_message_box "${menu_title}" "${fallback_message}"
+    create_message_box 'Error' "${fallback_message}"
     screen_sequence['SHOW_SCREEN']='dashboard'
     return "$?"
   fi
--- a/src/ui/patch_hub/patch_hub_core.sh
+++ b/src/ui/patch_hub/patch_hub_core.sh
@@ -152,9 +152,20 @@ function show_registered_mailing_lists()
   local -a registered_mailing_lists
   local message_box
   local ret
+  local fallback_message
 
   # Load registered mailing lists from config file into array
   IFS=',' read -r -a registered_mailing_lists <<< "${lore_config['lists']}"
+
+  if [[ "${#registered_mailing_lists[@]}" -eq 0 ]]; then
+    fallback_message='kw could not find any registered mailing list.'$'\n'$'\n'
+    fallback_message+='Try to register mailing lists in the menu "Settings" -> '
+    fallback_message+='"Register/Unregister Mailing Lists"'
+
+    create_message_box 'Error' "${fallback_message}"
+    screen_sequence['SHOW_SCREEN']='dashboard'
+    return
+  fi
 
   message_box='Below, you can see the lore.kernel.org mailing lists that you have registered.'$'\n'
   message_box+='Select a mailing list to see the latest patchsets sent to it.'
--- a/documentation/man/features/patch-hub.rst
+++ b/documentation/man/features/patch-hub.rst
@@ -11,7 +11,7 @@ SYNOPSIS
 DESCRIPTION
 ===========
 The `kw patch-hub` feature provides an interface with the public mailing
-lists archived on `https://lore.kernel.org`. The feature can be used to just consult
+lists archived on `https://lore.kernel.org`, focusing on patch series. The feature can be used to just consult
 patch series from a given list, but, as it integrates with other kw features, it
 simplifies the process of reviewing patch series, like:
 
--- a/src/ui/patch_hub/settings.sh
+++ b/src/ui/patch_hub/settings.sh
@@ -14,9 +14,9 @@ function show_settings_screen()
   menu_list_string_array=(
     'Register/Unregister Mailing Lists'
     'Save Patches To'
-    'Kernel Tree Path'
-    'Kernel Tree Target Branch'
-    'Patchsets Per Page'
+    'Edit Kernel Tree Path'
+    'Edit Kernel Tree Target Branch'
+    'Edit Patchsets Per Page'
   )
   create_menu_options 'Settings' '' 'menu_list_string_array' '' '' 'Return'
   ret="$?"


documentation/man/kw.rst
```diff
@@ -76,8 +76,8 @@ dmesg. All the debug options are intended to support remote and local targets.
 
   :ref:`kw-debug<debug-doc>`
 
-COMMAND TO INTERACT WITH THE PUBLIC MAILING LISTS
--------------------------------------------------
+COMMAND TO INTERACT WITH PATCHES IN THE PUBLIC MAILING LISTS
+------------------------------------------------------------
```
rodrigosiqueira, line 80: Squash this commit with the previous one.

src/ui/patch_hub/patch_hub_core.sh
```diff
@@ -159,6 +159,16 @@ function show_registered_mailing_lists()
...
   message_box+='Select a mailing list to see the latest patchsets sent to it.'
 
+  fallback_message='kw could not find any registered mailing list.'$'\n'$'\n'
+  fallback_message+='Try to register mailing lists in the menu "Settings" then ' 
+  fallback_message+='"Register/Unregister Mailing Lists"'
+
+  if [[ "${#registered_mailing_lists[@]}" -eq 0 ]]; then
+      create_message_box 'Error' "${fallback_message}"
+      screen_sequence['SHOW_SCREEN']='dashboard'
+      return "$?"
```
rodrigosiqueira: There is a good chance that `"$?"` is 0, which means success, and from this code, I don't think you intend to return a success code. Am I correct? If so, you probably need to return 22 here or any other error code.

src/ui/patch_hub/patch_hub_core.sh
```diff
@@ -203,7 +203,7 @@ function list_patches()
   local ret
 
   if [[ -z "${_target_array_list}" ]]; then
-    create_message_box "${menu_title}" "${fallback_message}"
+    create_message_box "Error" "${fallback_message}"
```
rodrigosiqueira, line 206: Use single-quotes around `Error`.

rodrigosiqueira (changes requested):
Overall, it looks good; I just left some comments.
@davidbtadokoro is the maintainer of the patch-hub; in this sense, he has the final say about this PR.

src/ui/patch_hub/patch_hub_core.sh
```diff
@@ -159,6 +159,16 @@ function show_registered_mailing_lists()
   message_box='Below, you can see the lore.kernel.org mailing lists that you have registered.'$'\n'
   message_box+='Select a mailing list to see the latest patchsets sent to it.'
 
+  fallback_message='kw could not find any registered mailing list.'$'\n'$'\n'
+  fallback_message+='Try to register mailing lists in the menu "Settings" then ' 
+  fallback_message+='"Register/Unregister Mailing Lists"'
```
davidbtadokoro: As this fallback message will one be used inside the `if` statement block, move these three lines to inside the `if` statemet block.

src/ui/patch_hub/patch_hub_core.sh
```diff
@@ -159,6 +159,16 @@ function show_registered_mailing_lists()
   message_box='Below, you can see the lore.kernel.org mailing lists that you have registered.'$'\n'
   message_box+='Select a mailing list to see the latest patchsets sent to it.'
 
+  fallback_message='kw could not find any registered mailing list.'$'\n'$'\n'
```
davidbtadokoro, line 162: Don't forget to declare this variable at the top of the `show_registered_mailing_lists()` function (below `local message_box`)

src/ui/patch_hub/patch_hub_core.sh
```diff
@@ -159,6 +159,16 @@ function show_registered_mailing_lists()
...
+  fallback_message+='Try to register mailing lists in the menu "Settings" then '
```
davidbtadokoro: Instead of ` then `, I think ` -> ` is a better approach.

src/ui/patch_hub/patch_hub_core.sh
```diff
@@ -159,6 +159,16 @@ function show_registered_mailing_lists()
...
 
+  fallback_message='kw could not find any registered mailing list.'$'\n'$'\n'
+  fallback_message+='Try to register mailing lists in the menu "Settings" then ' 
+  fallback_message+='"Register/Unregister Mailing Lists"'
+
+  if [[ "${#registered_mailing_lists[@]}" -eq 0 ]]; then
+      create_message_box 'Error' "${fallback_message}"
+      screen_sequence['SHOW_SCREEN']='dashboard'
+      return "$?"
+  fi
```
davidbtadokoro: For this whole `if` statement **block**, there are three things you should do:

- **Number 1**:

> There is a good chance that "$?" is 0, which means success, and from this code, I don't think you intend to return a success code. Am I correct? If so, you probably need to return 22 here or any other error code.

Actually, because of the FSM in `patch-hub`, returning a code 22 ends execution. Also, I feel that we can interpret that the "Error" is that there are no lists registered, and that this error is handled inside this `if` statement block, so we don't need to return an error code.

That being said, you should replace `return $?` by a simple `return`, like it is done across the feature for this handled "Errors".

- **Number 2**:

You used an indentation with four spaces in comparison with the `if` statement indentation, so you added one more indentation than needed. In other words,

```
if [[ "${#registered_mailing_lists[@]}" -eq 0 ]]; then
    create_message_box 'Error' "${fallback_message}"
    ...
```

should be

```
if [[ "${#registered_mailing_lists[@]}" -eq 0 ]]; then
  create_message_box 'Error' "${fallback_message}"
  ...
```

- **Number 3**:

Move this whole `if` statement block to between

```
  # Load registered mailing lists from config file into array
  IFS=',' read -r -a registered_mailing_lists <<< "${lore_config['lists']}"
```

and

```
  message_box='Below, you can see the lore.kernel.org mailing lists that you have registered.'$'\n'
  message_box+='Select a mailing list to see the latest patchsets sent to it.'
```

as, at this point, it is already possible to determine if there are no lists registered.

src/ui/patch_hub/patch_hub_core.sh
```diff
@@ -160,13 +160,13 @@ function show_registered_mailing_lists()
...
+  fallback_message+='Try to register mailing lists in the menu "Settings" then '
   fallback_message+='"Register/Unregister Mailing Lists"'
 
   if [[ "${#registered_mailing_lists[@]}" -eq 0 ]]; then
-      create_message_box 'Error' "${fallback_message}"
-      screen_sequence['SHOW_SCREEN']='dashboard'
-      return "$?"
+    create_message_box 'Error' "${fallback_message}"
+    screen_sequence['SHOW_SCREEN']='dashboard'
+    return "$?"
```
davidbtadokoro, line 169: All these changes don't belong here, but in the third commit in the PR (the one where you introduced these lines). I see that you noticed the indentation problem I mentioned in the third commit, but this should be updated there using interactive rebase.

davidbtadokoro (changes requested):
Hi @JBGSouza, and thanks so much for tackling this `patch-hub` related UX issues.

Left some comments, so please try to address them. Don't hesitate to ask me for any doubts about them or anything else.

davidbtadokoro:
Hi @JGBSouza, and thanks for the dedication on updating this PR!

Everything looked legit to me, so I merged it into the unstable branch 👍

Issue #705, opened 2022-09-23
kw env --use should alert users if their local kernel needs to be clean


rodrigosiqueira:
If a developer uses the current kernel root to develop something and later decides to create an env, they will see a makefile failure because of some old compiled files. To fix that, the developer must run `make distclean`, which might not be obvious to everyone. Let's make this requirement more explicit by:
1. Check if the current folder needs to be cleaned.
2. If so, alert the user and ask them if they want kw to do that.
3. If the user says no, make sure that we print a message with the required command.

PR #945, opened 2023-10-27
src: env: Block env switch if the kernel tree is not clean

rodrigosiqueira:
One of the requirements of using kw env is imposed by how the O= option works, which requires that the kernel tree is clean to work properly. Currently, the command kw env --use only allows users to switch if the env is clean, which can lead to build failures and consequently mislead users. This commit addresses this issue by adding a check phase before the user tries to switch to a new env; if the kernel tree is not clean, kw fails the operation with an informative message.

Closes: #705

### DIFF
--- a/src/kw_env.sh
+++ b/src/kw_env.sh
@@ -34,8 +34,12 @@ function env_main()
   fi
 
   if [[ -n "${options_values['USE']}" ]]; then
-    use_target_env
-    return "$?"
+    validate_env_before_switch
+    if [[ "$?" != 22 ]]; then
+      use_target_env
+      return "$?"
+    fi
+    return 22 # EINVAL
   fi
 
   if [[ -n "${options_values['DESTROY']}" ]]; then
@@ -52,6 +56,38 @@ function env_main()
     exit_env
     return "$?"
   fi
+}
+
+# If the env option is requested, it means the O= option from the kernel
+# makefile is used, and this option requires a clean kernel tree to work as
+# expected. This function checks if the current kernel tree is clean.
+#
+# Return:
+# Return 22 in case of error
+function validate_env_before_switch()
+{
+  local should_fail=0
+  local list_of_object_file
+
+  load_module_text "${KW_ETC_DIR}/strings/env.txt"
+
+  # Check if there is a .config file
+  [[ -f "${PWD}/.config" ]] && should_fail=1
+
+  # Check if there is any object file
+  list_of_object_file=$(find "${PWD}" -name '*.o')
+  [[ "$?" != 0 || -n ${list_of_object_file} ]] && should_fail=1
+
+  # Check for ko files
+  list_of_object_file=$(find "${PWD}" -name '*.ko')
+  [[ "$?" != 0 || -n ${list_of_object_file} ]] && should_fail=1
+
+  if [[ "$should_fail" == 1 ]]; then
+    complain "${module_text_dictionary[use_failure_explanation]}"
+    return 22 # EINVAL
+  fi
+
+  return 0
 }
 
 # When we switch between different kw envs we just change the symbolic links
--- a/tests/kw_env_test.sh
+++ b/tests/kw_env_test.sh
@@ -20,6 +20,9 @@ setUp()
   # Create fake .kw folder
   mk_fake_kw_folder "$TEST_PATH"
 
+  # Copy strings
+  cp "${ORIGINAL_PATH}/etc/strings/env.txt" "$TEST_PATH"
+
   # Create a dummy kernel .config file
   cp "$STD_CONFIG_FILE" ./
 }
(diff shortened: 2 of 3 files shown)


src/kw_env.sh
```diff
@@ -54,6 +58,34 @@ function env_main()
...
+  local should_fail=0
+  local list_of_object_file
+
+  load_module_text "${KW_ETC_DIR}/strings/env.txt"
+
+  # Check if there is a .config file
+  [[ -f "${PWD}/.config" ]] && should_fail=1
+
+  # Check if there is any object file
+  list_of_object_file=$(find "${PWD}" -name '*.o')
```
davidbtadokoro, line 78: Don't we need to also check for `.ko` files, `bzImage`, and the like? Maybe if there aren't `.o` files those I mentioned won't be present, but I confess I don't know for sure...

src/kw_env.sh
```diff
@@ -54,6 +58,34 @@ function env_main()
   fi
 }
 
+# If the env option is requested, it means the O= option from the kernel
+# makefile is used, and this option requires a clean kernel tree to work as
+# expected. This function checks if the current kernel tree is clean.
+#
+# Return:
+# Return 22 in case of error
+function validate_env_before_use()
```
davidbtadokoro, line 67: I don't know much about the `kw env` feature, but what do you think about the name `validate_env_before_switch`?

tests/kw_env_test.sh
```diff
@@ -20,6 +20,9 @@ setUp()
   # Create fake .kw folder
   mk_fake_kw_folder "$TEST_PATH"
 
+  # Copy strings
+  cp -r "${ORIGINAL_PATH}/etc/strings" "$TEST_PATH"
```
davidbtadokoro, line 24: Nitpick: maybe copy only the `etc/strings/env.txt` file instead of the whole directory.

tests/kw_env_test.sh
```diff
@@ -195,6 +198,40 @@ function test_use_target_env_invalid_env()
   assertEquals "($LINENO) Env does not exists" "$?" 22
 }
 
+function test_validate_env_before_use_invalid_case_with_config()
+{
+  local output
+
+  output=$(validate_env_before_use)
```
davidbtadokoro, line 205: You are not using the value of `output` for an assertion in this test case, and it seems that it will produce part of the text in `etc/strings/env.text`.

tests/kw_env_test.sh
```diff
@@ -195,6 +198,40 @@ function test_use_target_env_invalid_env()
...
+  assertEquals "(${LINENO}) it should not allow switch to a new env due to .config file." "$?" 22
```
davidbtadokoro, line 206: I think `assert_equals_helper` is a better function for this type of assertion, because of the indentation and color of what is the expected value and what was the actual value.

tests/kw_env_test.sh
```diff
@@ -195,6 +198,40 @@ function test_use_target_env_invalid_env()
...
+  local output
+
+  # Remove the .cofig file
+  mv '.config' 'config'
+
+  # Create a fake object file
+  mkdir -p 'drivers/gpu/drm/'
+  touch 'drivers/gpu/drm/drm_atomic.o'
+
+  output=$(validate_env_before_use)
```
davidbtadokoro, line 220: Same as the previous comment, you are not asserting that `output` has part of the text from `etc/strings/env.txt`.

tests/kw_env_test.sh
```diff
@@ -195,6 +198,40 @@ function test_use_target_env_invalid_env()
...
+  assertEquals "(${LINENO}) it should not allow switch to a new env due to object files." "$?" 22
```
davidbtadokoro, line 221: Same as the previous comment, maybe use `assert_equals_helper`.

tests/kw_env_test.sh
```diff
@@ -195,6 +198,40 @@ function test_use_target_env_invalid_env()
...
+}
+
+function test_validate_env_before_use_no_config_and_no_object_files()
+{
+  local output
+
+  # Remove the .cofig file
+  mv '.config' 'config'
+
+  output=$(validate_env_before_use)
```
davidbtadokoro, line 231: Same as the previous comment: you are not checking that `output` is empty.

tests/kw_env_test.sh
```diff
@@ -195,6 +198,40 @@ function test_use_target_env_invalid_env()
...
+  assertEquals "(${LINENO}) it should not allow switch to a new env due to object files." "$?" 0
```
davidbtadokoro, line 232: Same as the previous comment, maybe use `assert_equals_helper`.

davidbtadokoro, line 232: Also, the message 'it should not allow switch to a new env due to object files.' seems wrong. Shouldn't it be something like '**it should** allow switch to a new env.'?

davidbtadokoro (changes requested):
Hi @rodrigosiqueira, and thanks for the effort on attacking one of two issues needed for the beta release.

Left some comments, so see if you can address them. But mind the fact that I'm not acquainted with `kw env`, so feel free to ignore some of them if it doesn't make sense.

src/kw_env.sh
rodrigosiqueira: Nice catch! I'll update that.

src/kw_env.sh
rodrigosiqueira: I'll use that.

tests/kw_env_test.sh
rodrigosiqueira: Good idea.

davidbtadokoro:
Hey @rodrigosiqueira, and thanks a lot for the update!

LGTM, so merged into the unstable branch!

Issue #262, opened 2021-05-15
Add integration test


rodrigosiqueira:
Currently, we only have unit tests. Let's start to investigate how to add an integration test; some random ideas:

1. Create an integration test setup; maybe, we can use virtual machines for that;
2. Start with a simple feature like `kw b`;
3. Find a way to validate if something happened as expected;
4. Provide a report;
5. Can we integrate it with Travis or similar?

alan-barzilay:
Maybe we could use docker containers for the tests, I heard that they are easier to use with Travis than virtual machines

alan-barzilay:
Scratch that, docker uses the same kernel as the host system so it will probably have limited use for us. Bu since 2019 Vagrant seems to be supported in travis https://stackoverflow.com/questions/37307114/launching-a-vagrant-vm-inside-travis-ci

rodrigosiqueira:
Nice,
Maybe we can also use another type of CI that facilitates running an integration CI. It would be nice if we have a pipeline for these things.

rodrigosiqueira:
Hi @aquilamacedo,

After passing through your PR, you gave me a lot of ideas, and I'll try to summarize how we can approach this problem.

# Setup phase

Integration tests will require a lot of setup to work as we want. With this idea in mind, I suggested creating a prepare function in the run_integration test. You could add a `--prepare` option to the integration script, but you can also call it by default in case the user just calls `run_integration_test.sh`. This function should do the following:

1. Setup any folder used by the integration test;
2. Download any container or VM necessary for the test;
3. Conduct any setup necessary in the container or VM.

# Run phase

We should only try to run the integration tests if the setup criteria are fulfilled. For the run phase, we will need to create tools to work together with shunit. For example:

1. We will need a library to get data from the container/VM.
2. We will probably need ways to kill a container/VM in case of failure.

# Test method

We need an easy way to write tests, and we can have a good blend between in-house tools and shunit. Instead of trying to create this ideal scenario, let's only focus on test kw build. We could have something like this pseudo-code:

```sh
#!/bin/bash

. ../src/lib/kw_include.sh --source-only
include '../src/lib/kwio.sh'
include './utils.sh'

os='debian'

function setUp()
{
  start_clean_container
  configure_kernel_tree_in_the_container
  install_current_kw_version_in_the_container
  cd kernel_inside_container
  setup_kw_for_build_test
}

function tearDown()
{
  stop_container
}

function test_simple_build()
{
   local output

[...]

PR #946, opened 2023-10-27
integration_tests: Vagrantfile: Integration Testing to kw build

aquilamacedo:
This commit implements support for integration tests for the `kw build` command.

We use Vagrant to provision a virtual machine and Ansible to configure it. The purpose of this integration test is to create a virtual machine and execute the `kw build` command within it to verify that all steps occur as expected.

I still have many improvements to implement in this functionality, here is an example of some:

**TODOS**:

* It doesn't make sense to clone the kworkflow repository from scratch (this was just for testing because it's completely wrong). I'll still fix it to get the local kworkflow folder.

* I'm doing a lot of manual work in Ansible that can be automated, like fetching the dependencies file.


rodrigosiqueira:
Hi @aquilamacedo , check my latest post at:
https://github.com/kworkflow/kworkflow/issues/262

aquilamacedo:
Hi @rodrigosiqueira. I've already checked it out and I'm working on your review :)

PR #953, opened 2023-11-04
integration_tests: Containerfile: Podman integration testing

fernandohjunqueira:
This PR implements an initial file structure to start developing integration tests for kw.

It is using Podman to provide a debian container with kw installed in the PATH and the kernel repository cloned inside.

The purpose of this commit is to provide an interface inside the folder `integration_tests` to set up and manipulate the containers to write integration tests.

We can still improve this by a lot, but we'd like to hear some feedback to know how to keep going on. Feel free to comment on anything we've made.

What we intend to do next is:

- Keep tests inside /integration_tests/tests/ folder
- Create a script `run_integration_tests.sh` similar to the one provided in [this PR](https://github.com/kworkflow/kworkflow/pull/946), to automatically run all test files from a path.
- Write a simple integration test for kw config manager

### DIFF
--- a/documentation/content/tests.rst
+++ b/documentation/content/tests.rst
@@ -9,6 +9,11 @@ Kw's tests rely on `shunit2`. The `run_tests.sh` script automatically detects
 can have shunit2 source code in `tests/` (you can clone it from
 https://github.com/kward/shunit2).
 
+In order to run the integration tests, it is necessary to install Podman and  to
+configure it to run in `rootless` mode. Podman  is  available  via  the  default
+package manager of popular distros, such as Arch, Debian, Fedora and those based
+on them.
+
 If you want to run all the tests, try::
 
   ./run_tests.sh
@@ -20,6 +25,37 @@ List all available test files::
 Or run individual tests with::
 
   ./run_tests.sh test TESTFILE1 ...
+
+To limit the scope of the tests, pass the flag `--unit` or `--integration` as
+the first argument to any of the examples above. So, the syntax is::
+
+  ./run_tests.sh [scope] [command] [args]
+
+Where `[scope]` can be `--unit` or `--integration`. The placeholder  `[command]`
+can be either `list`, `test` or simply omited in order to run  all  tests.  Here
+are some examples:
+
+  ./run_tests.sh --unit                       # run all unit tests
+  ./run_tests.sh --unit list                  # list all unit tests
+  ./run_tests.sh --unit test device           # test device unit test
+  ./run_tests.sh --integration                # run all integration tests
+  ./run_tests.sh --integration list           # list all integration tests
+  ./run_tests.sh --integration test device    # test device integration test
+  ./run_tests.sh                              # run all tests
+  ./run_tests.sh list                         # list all tests
+  ./run_tests.sh test device                  # run all device tests
+
+The integration tests can take over 10 minutes to run in the first time  because
+podman is building the container images to be used in the tests, which  requires
+installing kw's dependencies in the supported distros.  After  the  images  have
+been built and cached, running the integration tests  should  take  only  a  few
+seconds each time.
+
+Then, the local kw repo is copied to the containers and installed  again,  which
+takes very few seconds. For optimization purposes,  the  containers  are  reused
+accross tests. If you add a new commit or checkout to another branch, such  that
+HEAD points to another commit, the containers  will  be  destroyed  and  created
+again in order to install the current local version of kw.
 
 Kw is already prepared to run tests, build the documentation and check the
 installation in the github workflow.
--- a/tests/integration/config_test.sh
+++ b/tests/integration/config_test.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+include './tests/unit/utils.sh'
+include './tests/integration/utils.sh'
+
+function oneTimeSetUp()
+{
+  local distro
+
+  # copy config files to the containers
+  for distro in "${DISTROS[@]}"; do
+    container_copy "kw-${distro}" "${SAMPLES_DIR}/config" '/tmp/.kw'
+  done
+}
+
+# The current workaround is this: we are using specially crafted config files
+# that overwrite all global options in the same order displayed by kw.
+function local_config_test_helper()
+{
+  local config="$1"
+  local config_file
+  local container
+  local distro
+  local expected_output
+
+  # make sure config is defined
+  if [[ -z "${config}" ]]; then
+    fail "(${LINENO}): 'config' is not defined."
+  fi
+
+  # the path to the config file
+  config_file="${SAMPLES_DIR}/config/${config}.config"
+
+  # the expected output is the same for all distros
+  expected_output=$(sort --dictionary-order "${config_file}")
+
+  for distro in "${DISTROS[@]}"; do
+    container="kw-${distro}"
+
+    # get the output inside the container
+    output=$(container_exec --workdir '/tmp' "${container}" "kw config --show --local ${config}")
+
+    if [[ "$?" -ne 0 ]]; then
+      fail "(${LINENO}) kw failed to show local variables in ${distro}"
+    fi
+
+    # remove prefix and N/A values from the output and sort them, before comparing
+    output=$(grep --invert-match 'N/A' <<< "${output}" | sed "s/^${config}.//" | sort --dictionary-order)
+
+    assertEquals "(${LINENO}): kw config failed for ${distro}" "${expected_output}" "${output}"
+  done
+}
+
+function test_build_config()
+{
+  local_config_test_helper 'build'
+}
+
+function test_deploy_config()
+{
+  local_config_test_helper 'deploy'
+}
+
+function test_mail_config()
+{
+  local_config_test_helper 'mail'
+}
+
+function test_notification_config()
+{
+  local_config_test_helper 'notification'
+}
+
+function test_vm_config()
+{
+  local_config_test_helper 'vm'
+}
+
+invoke_shunit
--- a/tests/integration/device_test.sh
+++ b/tests/integration/device_test.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+
+include './tests/unit/utils.sh'
+include './tests/integration/utils.sh'
+include './src/device_info.sh'
+
+# test if `kw device --local` in the container matches the host
+function device_info_test_helper()
+{
+  local distro="${1}"
+  local container
+  local buffer
+  local filter
+  local filter_regex
+  local fs_type
+  local fs_size
+  local output
+  local expected_output
+
+  # buffer with all information.
+  container="kw-${distro}"
+  buffer=$(container_exec "${container}" 'kw device --local')
+
+  # some fields must be ignored because they surely won't match.
+  filter=(
+    'Root filesystem' 'Size' 'Mounted on'                        # storage fields
+    'Distribution' 'Distribution version' 'Desktop environments' # desktop fields
+  )
+
+  # add | separator to each filter item.
+  filter_regex=$(sed 's/ /|/g' <<< "${filter[@]}")
+
+  # the actual perl regex.
+  filter_regex="(${filter_regex}):"
+
+  # get the output, after applying filter.
+  output=$(printf '%s' "${buffer}" | grep --invert-match --perl-regexp "${filter_regex}")
+
+  # get the expected output from the host machine, also filtering information.
+  expected_output=$(device_main --local | grep --invert-match --perl-regexp "${filter_regex}")
+
+  # TODO: remove this block of code after issue #1000 has been solved.
+  #
+  # deviceinfo shows GPU information for Arch, but not for Debian and Fedora due
+  # some libraries being present in Arch by default,  but  not  in  the  others.
+  # Therefore, if this information is not present, we must filter it out.
+  grep 'GPU:' <<< "${output}" > /dev/null
+  if [[ "$?" -ne 0 ]]; then
+    # we resort to SED because GREP can't filter out lines after context IF we
+    # use invert match. First, we tell sed to stop printing lines after matching
+    # the pattern GPU then delete the line containing that pattern.
+    expected_output=$(sed '/GPU:/q' <<< "${expected_output}" | sed '/GPU:/d')
+  fi
+
+  assertEquals "(${LINENO}): kw device failed for ${distro}" "${output}" "${expected_output}"
+
+  # check storage information
+  fs_type=$(container_inspect --format '{{.Driver}}' "${container}")
+  fs_size=$(container_inspect --format '{{.GraphDriver.Data.WorkDir}}' "${container}" |
+    xargs df -h | tail --lines 1 | awk '{ print $2 }')
+
+  assert_substring_match 'Wrong filesystem type.' "${LINENO}" "${buffer}" "Root filesystem: ${fs_type}"
+  assert_substring_match 'Wrong filesystem size.' "${LINENO}" "${buffer}" "Size: ${fs_size}"
+  assert_substring_match 'Wrong mounting point.' "${LINENO}" "${buffer}" "Mounted on: /"
+}
+
+function test_device_archlinux()
+{
+  device_info_test_helper 'archlinux'
+}
+
+function test_device_debian()
+{
+  device_info_test_helper 'debian'
+}
+
+function test_device_fedora()
+{
+  device_info_test_helper 'fedora'
+}
+
+invoke_shunit
--- a/tests/integration/kw_version_test.sh
+++ b/tests/integration/kw_version_test.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+include './src/lib/kwio.sh'
+include './tests/unit/utils.sh'
+include './tests/integration/utils.sh'
+
+declare -g expected_output
+
+function oneTimeSetUp()
+{
+  local kw_git_dir
+  local kw_dir
+  local head_hash
+  local branch_name
+  local base_version
+
+  # git directory path
+  kw_dir="${KWROOT_DIR}"
+  kw_git_dir="${kw_dir}/.git"
+
+  # In order to check correctness of `kw --version`, we collect some information
+  # from the git repo:
+  #
+  # - Base version (alpha, beta, or other)
+  # - Branch name
+  # - Commit sha
+  #
+  # Because the local KW repo is copied to the container, we run  the  following
+  # commands directly on the host instead of running in the container.
+  head_hash=$(git --git-dir "${kw_git_dir}" rev-parse --short HEAD)
+  branch_name=$(git --git-dir "${kw_git_dir}" rev-parse --short --abbrev-ref HEAD)
+  base_version=$(head --lines 1 "${kw_dir}/src/VERSION")
+
+  # using the gathered information, we build the expected output
+  expected_output=$(printf '%s\nBranch: %s\nCommit: %s' "$base_version" "$branch_name" "$head_hash")
+}
+
+function kw_version_test_helper()
+{
+  local distro="$1"
+  local container
+  local output
+
+  # collect the kw version in the container
+  container="kw-${distro}"
+  output=$(container_exec "${container}" 'kw --version')
+
+  assertEquals "(${LINENO}): kw version failed for ${distro}" "$expected_output" "$output"
+}
+
+function test_kw_version_on_archlinux()
+{
+  kw_version_test_helper 'archlinux'
+}
+
+function test_kw_version_on_debian()
+{
+  kw_version_test_helper 'debian'
+}
+
+function test_kw_version_on_fedora()
+{
+  kw_version_test_helper 'fedora'
+}
+
+invoke_shunit
--- a/tests/integration/podman/Containerfile_archlinux
+++ b/tests/integration/podman/Containerfile_archlinux
@@ -1,0 +1,7 @@
+FROM docker.io/library/archlinux
+
+RUN pacman -Syu --noconfirm git
+
+COPY ./clone_and_install_kw.sh .
+
+RUN bash ./clone_and_install_kw.sh
--- a/tests/integration/podman/clone_and_install_kw.sh
+++ b/tests/integration/podman/clone_and_install_kw.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# The primary purpose of this file is to install kw's dependencies in the  image
+# we will built. This will save up time when installing the local copy of KW  in
+# the running container.
+#
+# We clone kw and install the unstable branch in order for the base image of the
+# container to have all required dependencies. Later, the local KW repo will  be
+# copied to the container and installed, but the dependencies won't have  to  be
+# installed again, which saves a lot of time.
+
+kw_dir='/tmp/kw'
+
+git clone https://github.com/kworkflow/kworkflow "${kw_dir}"
+cd "$kw_dir" || exit 1
+git checkout unstable
+./setup.sh --install --force --skip-docs
--- a/tests/integration/utils.sh
+++ b/tests/integration/utils.sh
@@ -202,3 +202,14 @@ container_copy()
     fail "(${LINENO}): Failed to copy host files to the container."
   fi
 }
+
+# inspect the container
+container_inspect()
+{
+  # shellcheck disable=SC2068
+  podman container inspect $@
+
+  if [[ "$?" -ne 0 ]]; then
+    fail "(${LINENO}): Failed to inspect the container."
+  fi
+}
--- a/tests/integration/utils.sh
+++ b/tests/integration/utils.sh
@@ -188,3 +188,17 @@ function container_exec()
     fail "(${LINENO}): Failed to execute the command in the container."
   fi
 }
+
+# copy local file to container
+container_copy()
+{
+  local container="$1"
+  local src="$2"
+  local dst="$3"
+
+  podman container cp "${src}" "${container}":"${dst}"
+
+  if [[ "$?" -ne 0 ]]; then
+    fail "(${LINENO}): Failed to copy host files to the container."
+  fi
+}
--- a/tests/integration/utils.sh
+++ b/tests/integration/utils.sh
@@ -1,0 +1,190 @@
+#!/bin/bash
+
+include './src/lib/kwio.sh'
+
+declare -g CONTAINER_DIR # has the container files to build the container images
+declare -g SAMPLES_DIR   # has sample files used accross the integration tests
+declare -g KWROOT_DIR    # local kw dir to be copied to and installed in the containers
+declare -g DISTROS       # distributions we will run the integration tests
+
+# ensure path to directories is absolute
+script_dir=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
+CONTAINER_DIR="${script_dir}/podman"
+SAMPLES_DIR="${script_dir}/samples"
+KWROOT_DIR=$(realpath "${script_dir}/../..")
+
+# supported distros
+DISTROS=(
+  'archlinux'
+  'debian'
+  'fedora'
+)
+
+# Builds a container image for the given distro
+#
+# @param $1 The OS distribution of the target container image.
+# @return $? The status code of the command ran to build the image.
+function build_distro_image()
+{
+  local distro="${1}"
+  local file="${CONTAINER_DIR}/Containerfile_${distro}"
+
+  podman image build --file "$file" --tag "kw-${distro}" > /dev/null 2>&1
+
+  # Check if the command failed
+  if [[ "$?" -ne 0 ]]; then
+    fail "(${LINENO}): Error building the image for distribution ${distro}"
+    return "$?"
+  fi
+}
+
+# Builds container images and create containers used accross the tests.
+#
+# @param $1 log level to print info, debug or error
+function setup_container_environment()
+{
+  local container_img
+  local container_name
+  local distro            # current distro
+  local distros_ok        # array of distros whose setup succeed
+  local i                 # current step of the setup
+  local n                 # total number of steps
+  local working_directory # working directory in the container
+
+  # initialize some values
+  distros_ok=()
+  i=0
+  n=$((${#DISTROS[@]} * 2))
+  working_directory='/tmp/kw'
+
+  for distro in "${DISTROS[@]}"; do
+
+    # Only build the image if it does not exists. That's because trying to build
+    # the podman image takes a second or two even if it exists and is cached.
+    podman image exists "kw-${distro}"
+    if [[ "$?" -ne 0 ]]; then
+      # progress message:
+      i=$((i + 1))
+      say "[${i}/${n}] Building container image for ${distro}."
+
+      # Build the image or fail.
+      build_distro_image "$distro"
+      if [[ "$?" -ne 0 ]]; then
+        complain "failed to setup container environment for distro ${distro}"
+
+        # print we will skip creating the container image
+        i=$((i + 1))
+        say "[${i}/${n}] Skip creating ${distro} container."
+
+        # continue the setup for other distros.
+        continue
+      fi
+    else
+      # progress message:
+      i=$((i + 1))
+      say "[${i}/${n}] Using cached container image for ${distro}."
+    fi
+
+    # The name of the container and the container image are equal here,  but  it
+    # is useful to make a distinction that in some places we should use the image
+    # name and in others the container name (they are equal in value but are not
+    # equivalent in meaning).
+    container_img="kw-${distro}"
+    container_name="${container_img}"
+
+    # If container exists, we tear it down and create a new one in order
+    # ensure KW installation reflects the latest local changes.
+    podman container exists "${container_name}"
+    if [[ "$?" -eq 0 ]]; then
+      teardown_single_container "${container_name}"
+    fi
+
+    # progress message:
+    i=$((i + 1))
+    say "[${i}/${n}] Creating ${distro} container."
+
+    # containers are isolated environments designed to run a  process.  After  the
+    # process ends, the container is destroyed. In order execute multiple commands
+    # in the container, we need to keep the container,  which  means  the  primary
+    # process must not terminate. Therefore, we run a never-ending command as  the
+    # primary process,  so  that  we  can  execute  multiple  commands  (secondary
+    # processes) and get the output of each of them separately.
+    container_run \
+      --workdir "${working_directory}" \
+      --volume "${KWROOT_DIR}":"${working_directory}" \
+      --env PATH='/root/.local/bin:/usr/bin' \
+      --name "${container_name}" \
+      --detach \
+      "${container_img}" sleep infinity > /dev/null
+
+    if [[ "$?" -ne 0 ]]; then
+      fail "(${LINENO}): Failed to run the container ${container_name}"
+    fi
+
+    # install kw again
+    container_exec "${container_name}" \
+      ./setup.sh --install --force --skip-checks --skip-docs > /dev/null 2>&1
+
+    if [[ "$?" -ne 0 ]]; then
+      fail "(${LINENO}): Failed to install kw in the container ${container_name}"
+    else
+      # add distro to array of distros that worked.
+      distros_ok+=("$distro")
+    fi
+  done
+
+  # Update DISTRO so it only has distros whose setup succeed.
+  # Thus, the integration tests can run even if some distros failed to set up.
+  DISTROS=("${distros_ok[@]}")
+}
+
+# destroy a single container
+function teardown_single_container()
+{
+  local container="$1"
+
+  podman container exists "${container}"
+
+  if [[ "$?" -eq 0 ]]; then
+    # destroy the container, waiting 0 seconds to send SIGKILL
+    podman container rm --force --time 0 "${container}" > /dev/null 2>&1
+  fi
+}
+
+# destroy all containers used in the tests
+teardown_container_environment()
+{
+  local distro
+  local i=0                # current step of tear down
+  local n="${#DISTROS[@]}" # total number of steps of tear down
+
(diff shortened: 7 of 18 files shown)


integration_tests/Containerfile
```diff
@@ -0,0 +1,82 @@
...
+    && rm -rf /var/lib/apt/lists/*
+USER user
+
+
+
+# SET KERNEL AND WORKDIR
+
+WORKDIR ./linux
+# To execute commands inside kernel tree
+RUN zcat /proc/config.gz > .config
```
gustavofukunaga, line 77: The command can lead to compatibility issues when the host kernel version differs from the container kernel version. This is due to the fact that containers share the same kernel as the host system, and the container's behavior is dependent on the kernel of the host system.

For instance, running the above command within a Debian container hosted on Fedora may encounter problems. The `zcat /proc/config.gz` command may not find the `config.gz` in `/proc` because Fedora uses the `cp /boot/config-$(uname -r) .config` command to build the `.config`. Even if we manage to create the `.config` file, it may not be suitable for use in the Debian container due to differences in kernel configurations between Fedora and Debian.

## Potential Solution: Maintain Versioned .config Files

To address kernel version compatibility issues, we can consider maintaining versioned `.config` files within the project repository. This allows the selection of the appropriate `.config` file that matches the distribution of the container, ensuring compatibility.

I'm not sure if this is the best solution so feel free to correct me or add new informations!

integration_tests/Containerfile
```diff
@@ -0,0 +1,82 @@
+# Use the base Debian image from the Podman hub
+FROM docker.io/library/debian
+
+
+
```
rodrigosiqueira, line 5: Do we need all of these spaces?

integration_tests/Containerfile
```diff
@@ -0,0 +1,82 @@
...
+# INSTALLING DEPENDENCIES
+
+# Update package repositories and install necessary packages
+RUN apt update -y
+
+# Copy and install dependencies before setup to save build time 
+COPY ../documentation/dependencies/debian.dependencies /dependencies/debian.dependencies
+COPY ./integration_tests/dependencies/kernel.dependencies /dependencies/kernel.dependencies
+
+RUN DEBIAN_DEPENDENCIES=$(cat /dependencies/debian.dependencies) \
```
rodrigosiqueira, line 15: Check this: https://kworkflow.org/content/codingstyle.html#command-substitution-and-arithmetic-expression

integration_tests/Containerfile
```diff
@@ -0,0 +1,82 @@
...
+    && apt install -y $DEBIAN_DEPENDENCIES
+
+RUN KERNEL_DEPENDENCIES=$(cat /dependencies/kernel.dependencies) \
+    && apt install -y $KERNEL_DEPENDENCIES
+
+
+
```
rodrigosiqueira, line 22: Do we need all of these extra spaces?

integration_tests/Containerfile
```diff
@@ -0,0 +1,82 @@
...
+# CREATING USER
+
+# Create a new user
+RUN useradd -ms /bin/bash user
```
rodrigosiqueira, line 26: Prefer long options, e.g., --create-home instead of -m.

rodrigosiqueira: Also, it would be better to use a specific name, maybe kw_integration_test_user.

integration_tests/Containerfile
```diff
@@ -0,0 +1,82 @@
...
+
+# Set the working directory to the user's home directory
+WORKDIR /home/user
+
+
+
+# CLONE KERNEL TREE
+
+# TODO: Is cached. What if kernel is updated?
+RUN git clone --depth 1 https://github.com/torvalds/linux.git
```
rodrigosiqueira, line 39: Can we add a logic like:
1. If there is no repo, clone it.
2. If there is a repo, update it.

integration_tests/build_integration_test.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+. ./config.sh
+
+function setUp()
+{
+  # Builds or updates image
+  podman build -f ./Containerfile -t "$IMAGE" ../.
```
rodrigosiqueira, line 8: https://kworkflow.org/content/codingstyle.html#avoid-short-options

rodrigosiqueira: I think the build phase should be inside the run_integration_tests in the preparation phase.

integration_tests/config.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+CONTAINER="kw-integration"
+IMAGE="kw-raw"
```
rodrigosiqueira, line 4: Single-quotes. Btw, do we really need this file? Can we squeeze it into the util.sh and ensure we always include this file?

integration_tests/integration.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
rodrigosiqueira, line 1: I did not get why do we need this file.

integration_tests/utils.sh
```diff
@@ -0,0 +1,30 @@
...
+  # Run integration tests
+  podman exec "$CONTAINER" bash -c "./run_integration_tests.sh"
+}
+
+# Executes a command inside the container
+function run_kw_command()
+{
+  local command="$1"
+
+  podman exec "$CONTAINER" bash -c "kw $command"
```
rodrigosiqueira, line 22: https://kworkflow.org/content/codingstyle.html#string-concatenation

run_integration_tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
rodrigosiqueira, line 1: Am I missing something in this file? It does not look like it does something.

rodrigosiqueira (changes requested):
Hi,
I did not test in my local setup yet, but I left some comments. One important thing that you missed was the documentation; I want to read your doc and try this PR. Please try to address my comments and add documentation.

integration_tests/Containerfile
rodrigosiqueira: From our discussion, we'll not deploy a new kernel in the container; for that, we'll use Vagrant.

uwla:
@rodrigosiqueira, besides the requested improvements you mentioned, what do you think about eventually moving the existing tests under `tests/` directory to `tests/unit` and moving the `integration_tests` to `tests/integration`?

In my opinion, it makes more sense to keep all tests under `tests/` and to also make a distinction between unit and integration tests by keeping them in separated directories under `tests/`.

Let us know if eventually we can make that change as a part of this PR or if we just leave it as it is.

run_integration_tests.sh
```diff
@@ -0,0 +1,22 @@
...
+
+. ./src/lib/kw_include.sh --source-only
+include './integration_tests/utils.sh'
```
rodrigosiqueira, line 4: You should not use test features in this script. Here, you should only care about:
1. Setup
2. Run specific test
3. Run all tests

You should have something like:
```sh
function prepare()
{
# Check if the container is ready for use
# If not, initialize the container
}

function run_single_test()
{
   local test_name="$1"
  # Check if test_name exist in the integration_test/ folder
}

function run_all_tests()
{
 # Check all tests available in the integration_test folder, and run all of them.
}
...
```

integration_tests/utils.sh
```diff
@@ -0,0 +1,43 @@
...
+{
+  # Run unit tests
+  podman exec "$CONTAINER" bash -c "./run_tests.sh"
+}
+
+function run_integration_tests()
+{
+  # Run integration tests
+  podman exec "$CONTAINER" bash -c "./run_integration_tests.sh"
+}
```
rodrigosiqueira, line 16: tbh, I did not get the raison d'être of these functions... 

However, they gave me some ideas :)

We could organize the code as something like this:

```sh
run_integration_test.sh
integration_test/
 -> utils/
      -> Containerfile
      -> utils.sh
 kw_build_test.sh
 kw_version_test.sh
 ...
```

In the `run_integration_test.sh` you basically start the container before running any test and stop/clean after the test finishes, that way, each test does not need to deal with starting and stopping a VM; i.e., the test can assume the container is running and ready.

The `kw_version_test.sh` can take advantage of shunit and it can look like this:

```sh
include '../tests/utils.sh'

oneTimeSetUp()
{
  # Get the current branch name
  # Get the first line of the VERSION file
}

function test_check_if_kw_version_gives_the_correct_version()
{
   output=$(run_kw_command 'kw version')

   # compare $output with the expected value
}
```

integration_tests/utils.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
rodrigosiqueira, line 1: Maybe you can create a new file named `podman_lib.sh` and just include it in the test files. Also, keep in mind that some commands may fail; in this sense, add a proper error check. Also, consider using `cmd_manager`, that way, you can also create a verbose option for the integration test.

rodrigosiqueira (changes requested):
Hi,
First of all, thanks for the update.

Please take care of how you organize your PR. For example, you applied my previous comments in a new commit that undo many things from the last commit; there is a better way of doing it. When crafting your PR, keep in mind that each commit must make sense alone. You should try to tell a story with your commit sequence. Something like this:

1. Commit 1: In this commit, you introduce a basic code structure. Add some basic files and maybe some dummy functions.
2. Commit 2: Add this new cool library for dealing with XYZ.
3. Commit 3: Fill out some of the gaps from the commit 1 with things from the commit 2.

If you get a review and need to update it, do it in the proper commit; do not add a new commit addressing the review comments. If you don't know how to do that, check the command `git rebase -i SHA`.

In my review, you can see some of my comments. Overall, my suggestion is:

1. run_integration_test: This script should prepare the container so that tests can assume the container is already running and will stop at the end of the test. You should be able to request a single test or run all of them. In the case of running all tests, it will be better if you can randomly run them every time the user executes them.
2. integration_test folder: Follow the same pattern as we have in the test folders.
3. Tests: Take advantage of shunit.

Finally, about rename the `tests` folder, it would be better if:

1. In the tests folders, you create a new folder named `unit_tests` and move all unit tests to this folder.
2. In the tests folder, you create another folder named `integration_tests`.
3. Update the run_tests.sh script. Maybe you can have something like `run_tests.sh --integration <name>`, `run_test.sh --unit <name>`.

For this part, I recommend you open a new PR that makes these changes only for the unit test. In parallel, you keep your work in the integration test but with the idea that we will change the test organization soon.

uwla:
While I was working on more integration tests, I stumbled accross an strange behavior:

```
root@0cd127e7d8f7:/tmp# kw config --show --local build
build.arch=x86_64
build.kernel_img_name=bzImage
build.cross_compile=N/A
build.menu_config=nconfig
build.doc_type=htmldocs
build.cpu_scaling_factor=100
build.enable_ccache=N/A
build.warning_level=N/A
build.use_llvm=N/A
build.cflags=N/A
root@0cd127e7d8f7:/tmp# cat .kw/build.config 
build.arch=arm
build.kernel_img_name=linux
build.menu_config=nconfig
build.doc_type=man
build.cpu_scaling_factor=75
root@0cd127e7d8f7:/tmp# kw --version
alpha
Branch: unstable
Commit: 606ab65
```

Can someone explain how to proper use kw config? I thought creating a dir called `.kw` and adding a file named `build.config` would overwrite global configurations.

tests/integration/device_test.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+include './tests/unit/utils.sh'
+include './tests/integration/utils.sh'
+
```
rodrigosiqueira, line 5: I'm unsure if this test adds value because we are testing just one small part of the device, similar to what we do with the unit test. Consider dropping this test for now or make it like this:
1. In the container, get all the info that `kw device` should return, but manually.
2. run `kw device`
3. Compare the expected values with the ones that you manually collected.

rodrigosiqueira (changes requested):

tests/integration/kw_version_test.sh
```diff
@@ -1,44 +1,28 @@
...
-  output="$(podman container run kw-archlinux /root/.local/bin/kw --version)"
+  output="$(podman container run kw-archlinux kw --version)"
   assertEquals "$expected_kw_version" "$output"
 }
 
 test_installs_on_debian()
 {
   build_debian_image > /dev/null
-  output="$(podman container run kw-debian /root/.local/bin/kw --version)"
+  output="$(podman container run kw-debian kw --version)"
```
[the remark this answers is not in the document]
uwla: Because I added `ENV PATH="/root/.local/bin:$PATH"` in the containerfile, which adds `kw` to path by default.

tests/integration/podman/Containerfile_archlinux
```diff
@@ -8,4 +8,6 @@ WORKDIR /opt/kw
 
 RUN yes | ./setup.sh --install
 
+ENV PATH="/root/.local/bin:${PATH}"
```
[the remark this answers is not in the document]
uwla: Yes, that was why. I'll add a comment explaining it.

tests/integration/config_test.sh
```diff
@@ -0,0 +1,65 @@
...
+  local distro="$1"
+  local config="$2"
+  local container_image="kw-${distro}"
+  local expected_output=$(cat "${samples_dir}/${config}.config")
+
+  # build image
+  eval 'build_${distro}_image > /dev/null'
+
+  # get output
+  output=$(podman run -w /tmp -v "${samples_dir}":/tmp/.kw "$container_image" kw config --show "$2")
```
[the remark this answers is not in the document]
uwla: Ok, I'll create a small lib to avoid direct references to `podman` commands.

tests/integration/device_test.sh
uwla: Ok, this is a better approach.

uwla:
@fernandohjunqueira, @gustavofukunaga and I have made improvements in the PR, following your suggestions and requests.

Could you review it again @rodrigosiqueira ?

tests/integration/device_test.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+include './tests/unit/utils.sh'
+include './tests/integration/utils.sh'
+include './src/device_info.sh'
+
+# test if `kw device --local in the container matches the host
+device_info_tester_helper()
```
rodrigosiqueira, line 8: Please, add `function <name>`.

rodrigosiqueira (changes requested):

tests/integration/kw_version_test.sh
```diff
@@ -0,0 +1,65 @@
...
+  #
+  # - Base version (alpha, beta, or other)
+  # - Branch name
+  # - Commit sha
+  #
+  # Because the local KW repo is copied to the container, we run  the  following
+  # commands directly on the host instead of running in the container.
+  head_hash=$(git --git-dir "${kw_git_dir}" rev-parse --short HEAD)
+  branch_name=$(git --git-dir "${kw_git_dir}" rev-parse --short --abbrev-ref HEAD)
+  base_version=$(cat "${kw_dir}/src/VERSION" | head -n 1)
```
rodrigosiqueira, line 38: Check:
* https://kworkflow.org/content/codingstyle.html#command-substitution-and-arithmetic-expression
* https://kworkflow.org/content/codingstyle.html#avoid-short-options

tests/integration/kw_version_test.sh
```diff
@@ -0,0 +1,65 @@
...
+
+  # using the gathered information, we build the expected output
+  expected_output=$(printf '%s\nBranch: %s\nCommit: %s' "$base_version" "$branch_name" "$head_hash")
+
+  # collect the kw version in the container
+  container="kw-${distro}"
+  output=$(container_exec "${container}" kw --version)
```
rodrigosiqueira, line 45: To be safe, add single quotes around `'kw --version'`.

tests/integration/kw_version_test.sh
```diff
@@ -0,0 +1,65 @@
...
+
+  assertEquals "$expected_output" "$output"
```
rodrigosiqueira, line 47: `assertEquals` accepts messages, a message that displays the target distro (`distro`) and `$LINE`.

tests/integration/kw_version_test.sh
rodrigosiqueira, line 38: I also think you can move this entire part to the setup part since those will be constants. After that, you can also remove this function and execute the final parts in the functions that invoked this helper.

tests/integration/podman/clone_and_install_kw.sh
```diff
@@ -0,0 +1,17 @@
...
+# The primary purpose of this is to install kw's dependencies in  the  container
+# image. This will save up time when installing the local  copy  of  KW  in  the
+# running container.
+#
+# We clone kw and install the unstable branch in order for the base image of the
+# container to have all required dependencies. Later, the local KW repo will  be
+# copied to the container and installed, but the dependencies won't have  to  be
+# installed again, which saves a lot of time.
+
+kw_dir=/tmp/kw
```
rodrigosiqueira, line 12: Use single-quotes around '/tmp/kw'

tests/integration/podman/clone_and_install_kw.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# The primary purpose of this is to install kw's dependencies in  the  container
```
rodrigosiqueira, line 3: [..] of this **file** is [..]

tests/integration/utils.sh
```diff
@@ -0,0 +1,132 @@
...
+# supported distros
+DISTROS=('archlinux' 'debian' 'fedora')
+
+# Builds a container imagem for the given distro
+function build_distro_image()
+{
+  local distro="${1}"
+  local file="${CONTAINER_DIR}/Containerfile_${distro}"
+
+  podman image build --file "$file" --tag "kw-${distro}" 2> /dev/null
```
rodrigosiqueira, line 25: I suppose this command may fail, right? If so, you must handle the return (`$?`) value.

tests/integration/utils.sh
```diff
@@ -0,0 +1,132 @@
...
+      if [[ "${sha_kw}" == "${sha_container}" ]]; then
+        continue
+      else
+        teardown_single_container "${container_name}"
+      fi
+    fi
+
+    # Maybe the container is not running, but it does exist.
+    # In this case, we tear it down because it could be in a improper state.
+    if podman container exists "${container_name}"; then
```
rodrigosiqueira, line 82: Put this command outside the `if` condition, and just check `$?` (also use ``[[]]``).

tests/integration/utils.sh
```diff
@@ -0,0 +1,132 @@
...
+      teardown_single_container "${container_name}"
+    fi
+
+    # containers are isolated environments designed to run a  process.  After  the
+    # process ends, the container is destroyed. In order execute multiple commands
+    # in the container, we need to keep the container,  which  means  the  primary
+    # process must not terminate. Therefore, we run a never-ending command as  the
+    # primary process,  so  that  we  can  execute  multiple  commands  (secondary
+    # processes) and get the output of each of them separately.
+    working_directory=/tmp/kw
```
rodrigosiqueira, line 92: Add single quotes around the path.

tests/integration/utils.sh
```diff
@@ -0,0 +1,132 @@
...
+    podman run \
+      --workdir "${working_directory}" \
+      --volume "${KWROOT_DIR}":"${working_directory}" \
+      --env PATH='/root/.local/bin:/usr/bin' \
+      --name "${container_name}" \
+      --detach \
+      "${container_img}" sleep infinity > /dev/null
```
rodrigosiqueira, line 99: I guess this command can fail, right? If so, check `$?`.

tests/integration/utils.sh
```diff
@@ -0,0 +1,132 @@
...
+
+    # install kw again
+    podman exec "${container_name}" \
+      ./setup.sh -i --force --skip-docs > /dev/null 2>&1
```
rodrigosiqueira, line 103: Same thing, check the result. Also, if this command fails, it means one of the containers cannot be used, right? If that is true, you probably need a mechanism to deal with that case, otherwise, some tests will fail.

tests/integration/utils.sh
```diff
@@ -0,0 +1,132 @@
...
+  done
+}
+
+# destroy a single container
+function teardown_single_container()
+{
+  local container="$1"
+
+  if podman container exists "${container}"; then
```
rodrigosiqueira, line 112: move this command outside this `if`.

tests/integration/utils.sh
```diff
@@ -0,0 +1,132 @@
...
+
+  for distro in "${DISTROS[@]}"; do
+    teardown_single_container "kw-${distro}"
+  done
+}
+
+# execute a given command in the container
+function container_exec()
+{
+  podman container exec "$@" 2> /dev/null
```
rodrigosiqueira, line 131: This command can fail, right? If so, add a check to the result.

tests/integration/config_test.sh
```diff
@@ -0,0 +1,79 @@
...
+  local expected_output
+
+  # make sure config is defined
+  assertTrue 'config is not defined' "[[ -n '${config}' ]]"
+
+  # the path to the config file
+  config_file="${SAMPLES_DIR}/config/${config}.config"
+
+  # the expected output is the same for all distros
+  expected_output=$(sort "${config_file}")
```
rodrigosiqueira, line 36: Avoid default behavior since it can change across distro. I guess you want `--dictionary-order` behavior in this case.

tests/integration/config_test.sh
```diff
@@ -0,0 +1,79 @@
...
+function local_config_test_helper()
+{
+  local config="$1"
+  local config_file
+  local container
+  local distro
+  local expected_output
+
+  # make sure config is defined
+  assertTrue 'config is not defined' "[[ -n '${config}' ]]"
```
rodrigosiqueira, line 30: I think you can remove the `assertTrue` from here and move this check as a simple `if` in the setup phase.

tests/integration/config_test.sh
```diff
@@ -0,0 +1,79 @@
...
+  config_file="${SAMPLES_DIR}/config/${config}.config"
+
+  # the expected output is the same for all distros
+  expected_output=$(sort "${config_file}")
+
+  for distro in "${DISTROS[@]}"; do
+    container="kw-${distro}"
+
+    # get the output inside the container
+    output=$(container_exec --workdir /tmp "${container}" kw config --show --local "${config}")
```
rodrigosiqueira, line 42: Add single-quotes around `/tmp` and double quotes around the kw command.

tests/integration/config_test.sh
```diff
@@ -0,0 +1,79 @@
...
+
+    # assert command did not fail
+    assertTrue "kw failed show local variables in $distro" "[[ 0 -eq $? ]]"
```
rodrigosiqueira, line 45: https://kworkflow.org/content/codingstyle.html#string-concatenation

tests/integration/config_test.sh
```diff
@@ -0,0 +1,79 @@
...
+
+    # remove prefix and N/A values from the output and sort them, before comparing
+    output=$(printf '%s' "$output" | grep -v 'N/A' | sed "s/^${config}.//" | sort)
```
rodrigosiqueira, line 48: https://kworkflow.org/content/codingstyle.html#avoid-short-options

rodrigosiqueira: `sort --dictionary-order`.

tests/integration/config_test.sh
```diff
@@ -0,0 +1,72 @@
...
+  for distro in "${DISTROS[@]}"; do
+    container="kw-${distro}"
+
+    # get the output
+    output=$(container_exec --workdir /tmp "${container}" kw config --show --local "${config}")
+
+    # remove prefix and N/A values from the output and sort them, before comparing
+    output=$(printf '%s' "$output" | grep -v 'N/A' | sed "s/^${config}.//" | sort)
+
+    assertEquals "$expected_output" "$output"
```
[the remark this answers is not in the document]
rodrigosiqueira: Please address my previous comment.

tests/integration/device_test.sh
```diff
@@ -0,0 +1,74 @@
...
+  local container
+  local buffer
+  local filter
+  local filter_regex
+  local output
+  local expected_output
+
+  # buffer with all information
+  container="kw-${distro}"
+  buffer=$(container_exec "$container" kw device --local)
```
rodrigosiqueira, line 25: Add single-quotes around the kw command.

tests/integration/device_test.sh
```diff
@@ -0,0 +1,74 @@
...
+
+  # some fields must be ignored because they surely won't match
```
rodrigosiqueira, line 27: To solve this problem, you can just get that info directly from the container.

tests/integration/device_test.sh
```diff
@@ -0,0 +1,74 @@
...
+  # get the output, after applying filter
+  output=$(printf '%s' "${buffer}" | grep --invert-match --perl-regexp "${filter_regex}")
+
+  # get the expected output from the host machine, also filtering information
+  expected_output=$(device_main --local | grep --invert-match --perl-regexp "${filter_regex}")
+
+  # deviceinfo shows GPU information for Arch, but not for Debian and Fedora.
+  # This is due to some libraries being present in Arch, but not in the others.
+  # Therefore, if this information is not present, we must filter it out.
+  if ! grep 'GPU:' <<< "${output}" > /dev/null; then
```
rodrigosiqueira, line 48: Move this command outside this `if` and just check `$?`.

tests/integration/device_test.sh
```diff
@@ -0,0 +1,74 @@
...
+    # we resort to SED because GREP can't filter out lines after context IF we
+    # use invert match. First, we tell sed to stop printing lines after matching
+    # the pattern GPU then delete the line containing that pattern
+    expected_output=$(sed '/GPU:/q' <<< "${expected_output}" | sed '/GPU:/d')
+  fi
+
+  # compare
+  assertEquals "${output}" "${expected_output}"
```
rodrigosiqueira, line 56: Please, always add the `$LINE` and the distro in the message.

tests/integration/utils.sh
```diff
@@ -0,0 +1,132 @@
...
+  local sha_kw
+  local working_directory
+  local output # output of podman used to check if containers exist or not
+
+  for distro in "${DISTROS[@]}"; do
+    # Only build the image if it does not exists. That's because trying to build
+    # the podman image takes a half second even if it exists and is cached.
+    if ! podman image exists "kw-${distro}"; then
+      # Build the image or exit. The integration tests cannot continue otherwise
+      if ! build_distro_image "$distro" > /dev/null; then
```
rodrigosiqueira, line 44: Move all of those commands outside the `if` and check `$?` instead.

tests/integration/utils.sh
```diff
@@ -0,0 +1,132 @@
...
+      --volume "${KWROOT_DIR}":"${working_directory}" \
+      --env PATH='/root/.local/bin:/usr/bin' \
+      --name "${container_name}" \
+      --detach \
+      "${container_img}" sleep infinity > /dev/null
+
+    # install kw again
+    podman exec "${container_name}" \
+      ./setup.sh -i --force --skip-docs > /dev/null 2>&1
+  done
```
rodrigosiqueira, line 104: I think I'm missing something... Where do you copy the current branch to the container? I'm under the impression that we only run the kw from the unstable version.

rodrigosiqueira (changes requested):
Hi,
First of all, thanks a lot for keeping improving this PR. We are getting close to a very good base.

I added a lot of comments, and please address all of my comments to avoid many rounds of 'ping pong' review. If you disagree with some of my comments or want to dispute them, add a comment.

Finally, we need to manage two important things:
* The integration test dependencies: maybe we should consider creating a dependence file just for dev mode.
* Documentation: Expand our documentation page to describe this new integration test.

tests/integration/utils.sh
uwla: When we start the container in detached mode, we mount the `${KWROOT_DIR}`, which points to the root directory of the local kw repo, onto `/tmp/kw` in the container. We also set `/tmp/kw` as the working directory. Therefore, it does install the current version of `kw`. It installs the version that matches the latest commit in the local kw git repo.

tests/integration/device_test.sh
uwla: I think I did not understand  what you mean.

If I manually get the hardware information from the container, won't that be code duplication of the device module? I'd be running the same code of device module, but directly in the container. The purpose shouldn't be test that the output of `device --local` in the container matches the hardware information of the host?

tests/integration/device_test.sh
uwla: It is not possible to add single-quotes around the kw commands because podman will try to run a file/binary/program called `kw\ device\ --local` instad of running `kw` and passing two arguments: `device` and `--local`.

Therefore, we cannot add single-quotes here.
Also, the way it is now give us more flexibility to run `kw` commands with several flags.

tests/integration/kw_version_test.sh
uwla: It is not possible to add single-quotes around the kw commands because podman will try to run a file/binary/program called kw\ device\ --local instad of running kw and passing two arguments: device and --local.

uwla:
@rodrigosiqueira we have addressed all of your requested changes.

One of them we addressed via comments, explaining that it could not be done (the single quotes around kw commands when executing them in the container).

tests/integration/utils.sh
```diff
@@ -166,3 +166,12 @@ function container_exec()
...
   fi
 }
+
+# copy local file to container
+container_copy()
+{
+  local container="$1"
+  local src="$2"
+  local dst="$3"
+  podman container cp "${src}" "${container}":"${dst}"
```
rodrigosiqueira, line 176: Add a new line between the local and the podman command.

tests/integration/device_test.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
rodrigosiqueira, line 1: I think some dependency is missing for this test because when I run it, I see some log messages:

Screenshot from 2023-12-16 10-47-58

tests/integration/device_test.sh
```diff
@@ -0,0 +1,75 @@
...
+  local output
+  local expected_output
+
+  # buffer with all information
+  container="kw-${distro}"
+  buffer=$(container_exec "$container" kw device --local)
+
+  # some fields must be ignored because they surely won't match
+  filter=(
+    'Root filesystem' 'Size' 'Mounted on'                        # storage fields
```
rodrigosiqueira, line 29: I guess we can get this information from the container at run time instead of ignoring it.

documentation/content/tests.rst
```diff
@@ -9,6 +9,12 @@ Kw's tests rely on `shunit2`. The `run_tests.sh` script automatically detects
 can have shunit2 source code in `tests/` (you can clone it from
 https://github.com/kward/shunit2).
 
+In order to run the integration tests, it is necessary to install Podman and  to
+configure it to run in `rootless` mode. Podman  is  available  via  the  default
```
rodrigosiqueira, line 13: I think running as a rootless is already the default for podman. Am I wrong?

documentation/content/tests.rst
```diff
@@ -9,6 +9,12 @@ Kw's tests rely on `shunit2`. The `run_tests.sh` script automatically detects
...
+package manager of popular distributions, such  as  Arch  Linux  and  Arch-based
+distros like Manjaro, Debian and Debian-based distros like Ubuntu, Fedora, among
```
rodrigosiqueira, line 15: How about:
[..] such as Arch, Debian, and fedora based-distro.

documentation/content/tests.rst
```diff
@@ -21,5 +27,25 @@ Or run individual tests with::
...
+installing kw's dependencies in the supported distros.  After  the  images  have
+been built and cached, running the integration tests  should  take  only  a  few
+seconds each time.
+
+Then, the local kw repo is copied to the containers and installed  again,  which
+takes very few seconds. For optimization purposes,  the  containers  are  reused
+accross tests. If you add a new commit or checkout to another branch, such  that
+HEAD points to another commit, the containers  will  be  destroyed  and  created
+again in order to install the current local version of kw.
+
```
rodrigosiqueira, line 49: Add one example that shows how to run all integration tests and another one that shows how to run one specific test.

rodrigosiqueira (changes requested):

tests/integration/utils.sh
```diff
@@ -0,0 +1,168 @@
...
+  local working_directory
+  local output # output of podman used to check if containers exist or not
+
+  for distro in "${DISTROS[@]}"; do
+    # Only build the image if it does not exists. That's because trying to build
+    # the podman image takes a second or two even if it exists and is cached.
+    podman image exists "kw-${distro}"
+    if [[ $? -ne 0 ]]; then
+      # Build the image or exit. The integration tests cannot continue otherwise
+      build_distro_image "$distro" > /dev/null
```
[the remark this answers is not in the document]
uwla: Great idea! Will do.

documentation/content/tests.rst
uwla: Yes, but you have to configure rootless mode by adding 65536 subuids (sub-user ids) in order to properly run podman containers. Each one of the subuids will be mapped to a user uid in the containers; that's how podman works under the hood, mapping unprivileged user ids from the host to user ids in the container. If the user running podman does not have subuids, podman may not work properly.

tests/integration/device_test.sh
uwla: Yes, but I do not know why this is not working in your machine and is working on my machine because we install all kw's dependencies while building the container images...

I'll try to reproduce this behavior in a Debian VM. If the issue is due to missing dependencies in the dependency lists, I guess the proper action is to open a new issue and solve it there.

tests/integration/utils.sh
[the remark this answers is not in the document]
uwla: @rodrigosiqueira can you please detail how this verbose feature should work? 
For example, should podman's output be shown to the user while it builds the container images? Which information should be displayed (maybe all?) and which should be supressed?

Just a side note: Initially, I intetionally redirect all of podman's output to /dev/null because I do not like to see it mixed with the default standardized output of tests (that's my personal preference).

tests/integration/utils.sh
```diff
@@ -0,0 +1,168 @@
...
+
+  for distro in "${DISTROS[@]}"; do
+    teardown_single_container "kw-${distro}"
+  done
+}
+
+# execute a given command in the container
+function container_exec()
+{
+  podman container exec "$@" 2> /dev/null
```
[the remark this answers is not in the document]
uwla: I did it and it worked!

tests/integration/utils.sh
```diff
@@ -0,0 +1,168 @@
...
+  podman image build --file "$file" --tag "kw-${distro}" 2> /dev/null
+
+  # Check if the command failed
+  if [[ $? -ne 0 ]]; then
+    fail "($LINENO): Error building the image for distribution ${distro}"
+  fi
+}
+
+# Builds container images and create containers used accross the tests.
+function setup_container_environment()
```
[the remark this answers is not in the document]
uwla: Instead of adding a spinner, I guess we should show more relevant information, such as:

```
Setting up environment for integration tests.
[1/6] Building ArchLinux image...
[2/6] Creating ArchLinux container...
[3/6] Building Debian image...
[4/6] Creating Debian container...
[5/6] Building Debian image...
[6/6] Creating Debian container...
Environment for integration tests has been set up.
```

Later on:

```
Tearing down environment for integration tests.
[1/3] Removing ArchLinux container...
[2/3] Removing Debian container...
[3/3] Removing Fedora container...
```

What do you think?

I prefer this way because (1) it is more informative and (2) I prefer not implement shell spinners (I have never done it..).

tests/integration/utils.sh
```diff
@@ -0,0 +1,168 @@
...
+
+      # first, get the sha inside the container
+      sha_container=$(podman exec "${container_name}" kw --version | grep Commit | sed 's/Commit: //')
+
+      # get the sha of this repo
+      sha_kw=$(git --git-dir "${KWROOT_DIR}/.git" rev-parse --short HEAD)
+
+      # if container is up to date, we will reuse it.
+      # Otherwise, destroy the container and create a new one.
+      if [[ "${sha_kw}" == "${sha_container}" ]]; then
```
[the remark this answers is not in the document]
uwla: @rodrigosiqueira please take a look at my personal fork of this branch, in which I implement a better way to run tests:

https://github.com/uwla/kworkflow/tree/podman-integration-testing

I have done several improvements there. I have not incorporated those improvements in this PR because I have not finished addressing all requested changes yet.

tests/integration/device_test.sh
uwla: I can reproduce it locally.

tests/integration/utils.sh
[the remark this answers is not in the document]
rodrigosiqueira: I agree with you :)

tests/integration/utils.sh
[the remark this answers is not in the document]
rodrigosiqueira: For the verbose mode, let's show everything.

tests/integration/device_test.sh
uwla: PR #1001 fixes this

rodrigosiqueira:
Hi @uwla @fernandohjunqueira ,

I was reviewing this change, but at this point, most of the things that I have are small, and I think this PR is good enough to be merged. I merged it into the unstable branch.

Thanks a lot for your persistence and great work on this feature! I really appreciate all your effort and dedication.

Thanks

PR #954, opened 2023-11-04
src: device_info: Add verbose mode

rodrigosiqueira:
This commit enables the verbose option for the device info; unfortunately, to make these changes happen, I had to find a balance between updating legacy code and adding a new one. In particular, I had to create a new verbose function only for this feature because most of the code processes strings, and the verbose option should display them in the logs. Additionally, I decided to drop many tests that validate the command sequence, which, for the specific case of device_info, is not helpful because the potential of the errors are in the command result and test it is the correct approach.

Related-to:  #857

### DIFF
--- a/src/device_info.sh
+++ b/src/device_info.sh
@@ -31,6 +31,8 @@ declare -gA options_values
 # information of a target machine.
 function device_main()
 {
+  local flag
+
   if [[ "$1" =~ -h|--help ]]; then
     device_info_help "$1"
     exit 0
@@ -43,17 +45,28 @@ function device_main()
     exit 22 # EINVAL
   fi
 
+  [[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
+  flag=${flag:-'SILENT'}
+
   if [[ "${options_values['TARGET']}" == "$REMOTE_TARGET" ]]; then
     # Check connection before try to work with remote
-    is_ssh_connection_configured 'SILENT'
+    is_ssh_connection_configured "$flag"
     if [[ "$?" != 0 ]]; then
       ssh_connection_failure_message
       exit 101 # ENETUNREACH
     fi
   fi
 
-  learn_device "${options_values['TARGET']}"
-  show_data
+  learn_device "${options_values['TARGET']}" "$flag"
+  show_data "$flag"
+}
+
+function show_verbose()
+{
+  local flag="$1"
+  local cmd="$2"
+
+  [[ "$flag" == 'VERBOSE' ]] && printf '%s\n' "$cmd"
 }
 
 # This function populates the ram element from the device_info_data global
@@ -73,13 +86,17 @@ function get_ram()
   case "$target" in
     1) # VM_TARGET
       ram="$(printf '%s\n' "${vm_config[qemu_hw_options]}" | sed --regexp-extended 's/.*-m ?([0-9]+).*/\1/')"
-      ram="$(numfmt --from-unit=M --to-unit=K "$ram")"
+      cmd="numfmt --from-unit=M --to-unit=K ${ram}"
+      show_verbose "$flag" "$cmd"
+      ram=$(cmd_manager 'SILENT' "$cmd")
       ;;
     2) # LOCAL_TARGET
-      ram=$(cmd_manager "$flag" "$cmd")
+      show_verbose "$flag" "$cmd"
+      ram=$(cmd_manager 'SILENT' "$cmd")
       ;;
     3) # REMOTE_TARGET
-      ram=$(cmd_remotely "$cmd" "$flag")
+      show_verbose "$flag" "$cmd"
+      ram=$(cmd_remotely "$cmd" 'SILENT')
       ;;
   esac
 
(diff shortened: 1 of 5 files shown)


src/device_info.sh
```diff
@@ -73,13 +86,17 @@ function get_ram()
...
       ram="$(printf '%s\n' "${vm_config[qemu_hw_options]}" | sed --regexp-extended 's/.*-m ?([0-9]+).*/\1/')"
-      ram="$(numfmt --from-unit=M --to-unit=K "$ram")"
+      cmd="numfmt --from-unit=M --to-unit=K ${ram}"
+      show_verbose "$flag" "$cmd"
+      ram=$(cmd_manager 'SILENT' "$cmd")
       ;;
     2) # LOCAL_TARGET
-      ram=$(cmd_manager "$flag" "$cmd")
+      show_verbose "$flag" "$cmd"
+      ram=$(cmd_manager 'SILENT' "$cmd")
```
davidbtadokoro, line 95: Just to be sure, you are using `show_verbose "$flag" "$cmd"`, because `ram=$(cmd_manager "$flag" "$cmd")` (with `flag=VERBOSE`) outputs the command to the `ram` variable instead of to the standard output, right? I tested without `show_verbose` and it seems that's the case.

src/device_info.sh
```diff
@@ -129,9 +153,17 @@ function get_cpu()
...
+  show_verbose "$flag" "$cmd"
+  cpu_currently=$(cmd_manager 'SILENT' "$cmd")
+
+  cmd="printf '%s\n' '${cpu_frequency}' | grep 'CPU max MHz'"
+  show_verbose "$flag" "$cmd"
+  cpu_max=$(cmd_manager 'SILENT' "$cmd")
+
+  cmd="printf '%s\n' '${cpu_frequency}' | grep 'CPU min MHz'"
+  show_verbose "$flag" "$cmd"
+  cpu_max=$(cmd_manager 'SILENT' "$cmd")
```
davidbtadokoro, line 166: Think that it should be `cpu_min` here. This added a bug that made the max frequency displayed to have the min frequency value and, also, made the min frequency to not be displayed.

src/device_info.sh
```diff
@@ -156,18 +188,20 @@ function get_disk()
   local cmd
   local fs
 
-  flag=${flag:-'SILENT'}
```
davidbtadokoro, line 159: Is there a reason for deleting this line?

src/device_info.sh
```diff
@@ -202,31 +244,44 @@ function get_os()
   local os_id_like
 
   target=${target:-"${options_values['TARGET']}"}
-  flag=${flag:-'SILENT'}
```
davidbtadokoro, line 205: Is there a reason for deleting this line?

src/device_info.sh
```diff
@@ -247,13 +302,12 @@ function get_os()
 function get_desktop_environment()
 {
   local target="$1"
+  local flag="$2"
   local cmd
   local desktop_env
   local formatted_de='unidentified'
   local ux_regx="'gnome-shell$|kde|mate|cinnamon|lxsession|openbox$'"
 
-  flag=${flag:-'SILENT'}
```
davidbtadokoro, line 255: Is there a reason for deleting this line?

davidbtadokoro (changes requested):
Hi @rodrigosiqueira, thanks for keep tackling the verbose issue! Also, sorry for the delay in reviewing this PR.

Left some comments, but all-in-all, it works really nice! Tested both in the `--local` and `--remote` targets and it shows nice information for debug (specially for the remote). I didn't test in `--vm` targets, because this option is deprecated, as it doesn't even accepts the `--vm` flag (I'll open an issue for it).

Below, is the diff for the completions for Zsh. If you want, I can add this when the PR is updated, before me merging it.

```
diff --git a/src/_kw b/src/_kw
index e06049b..cdf1f63 100644
--- a/src/_kw
+++ b/src/_kw
@@ -220,6 +220,7 @@ _kw_deploy()
 _kw_device()
 {
   _arguments : \
     '(--remote --vm)--local[show hardware information from host machine]' \
     '(--local --vm)--remote[show hardware information from a remote machine]: :' \
     '(--remote --vm)--vm[show hardware information from a virtual machine]'
```

src/device_info.sh
rodrigosiqueira: Yeah, since we need to capture the output of the command, using the VERBOSE option will mess with the collected value. I added the `SILENT` flag and created the `show_verbose` option to workaround this problem. Maybe in the future, we can find a clever way to improve `cmd_manager` so that the VERBOSE flag is not captured in these cases. For now, the `show_verbose` is the easy solution.

src/device_info.sh
rodrigosiqueira: Nice catch!

src/device_info.sh
rodrigosiqueira: We should get `flag` as a parameter at this function level.

src/device_info.sh
rodrigosiqueira: same as my previous comment.

rodrigosiqueira:
Hi @davidbtadokoro , I believe I addressed all your comments.

davidbtadokoro:
Hey @rodrigosiqueira 

> Hi @davidbtadokoro , I believe I addressed all your comments.

Thanks! Going to look at it ASAP.

src/device_info.sh
davidbtadokoro: Sure, it makes sense 😄

davidbtadokoro:
Everything looks good to me! Merged into the unstable branch 👍

Issue #949, opened 2023-11-02
Misleading error message when `./setup.sh --install` fails because of unexistent dependency on Arch system


davidbtadokoro:
**Description:**

Alfter PR #940, recently merged by me, a misleading error message is shown when the dependency that it is trying to be install doesn't exist on an Arch system. The error message is

```
[ERROR]: Dependency installation failure: Lacked sudo privileges to install the packages:
```

even when the user didn't lack sudo privileges.

**How to Reproduce:**

1. Make sure you are on an Arch system
2. Checkout to the unstable branch 
3. Update this branch
4. Add a line to the file `documentation/dependencies/arch.dependencies` like `unexistent-dependency`
5. Run `./setup.sh --install`
6. Say yes to the installation of the dependency and input your admin password

**Expected behavior:**

This misleading error message shouldn't be displayed

**Screenshots:**

Bug in my Arch system

![Screen Shot 2023-11-02 at 02 26 11](https://github.com/kworkflow/kworkflow/assets/95055784/aa07de90-d1a0-4544-889b-d9c61364fe03)

Tested it on a Debian 11 Bullseye, and it doesn't seem to happen

![Screen Shot 2023-11-02 at 02 22 36](https://github.com/kworkflow/kworkflow/assets/95055784/c369cabe-077f-413a-b4e9-25f3593d355e)

**System information:**

 - OS: Manjaro 23.0.4 Uranos
 - kw Version: branch unstable at commit 223d131
 - Bash Version 5.1.16

**Additional Context**

This may be happening on Arch and not Debian, for example, because of the different codes returned by `pacman` and `apt`. The merged PR assumes error code 1 means lack of sudo privileges, which doesn't seem to be true to `pacman`

![Screen Shot 2023-11-02 at 02 42 55](https://github.com/kworkflow/kworkflow/assets/95055784/67227387-152e-4f96-8b69-91e402da4855)

davidbtadokoro:
@uwla and @jppaulo06, sorry for not noticing this "bug" before merging it into the unstable. I only tested it on a Debian 11 VM, and it worked as a charm. My bad, anyway 😞

jppaulo06:
Ohh, I understand. Apparently, pacman returns 1 when the package has not been found...
Maybe, then, we could just print the error given by the package manager. What do you think about it @davidbtadokoro?

jppaulo06:
[...]

PR #955, opened 2023-11-05
Change dependencies installation messages

jppaulo06:
This pull request enhances the messages printed when installing KW dependencies (usually executing `./setup --install`).

As referred in #949, error messages can become misleading when dependencies installation fails. This [pull request fixes this problem](https://github.com/GuilhermeMoreno/kworkflow/commit/f12b21c91c6c82c56cd4e94c69b071de4c3a4dbd) with a minimal solution.

Furthermore, it [introduces enhancements to the question prompt presented before starting the dependencies installation](https://github.com/kworkflow/kworkflow/commit/fee732ecb6fee9a432ff9c19546169f8e4d8e4db), both in terms of formatting and politeness.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -86,7 +86,7 @@ function check_dependencies()
 
   if [[ -n "$package_list" ]]; then
     if [[ "$FORCE" == 0 ]]; then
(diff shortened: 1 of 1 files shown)


davidbtadokoro:
Hey @jppaulo06, thanks for the change! Everything looked fine for me, so I merget it into the unstable 👍

PR #957, opened 2023-11-06
kw: show notice message if kw runs in developer mode

### LINKED ISSUE #956: Show notice message when running kw in dev mode
**Description:**

When running `kw` in devmode, show a notice message to warn the developer about it and its implications (such as possible changes in behavior).

**Type:**

Others

**Why:**

Some features of `kw` work different when in developer mode and it can cause confusion to developers.

Specifically, I was implementing some features in `kw` and added the following symlink: `path/to/kworkflow/kw -> ~/.local/bin/kw`.
When I ran `kw <cmd> --help` it showed me the warning:

```
Couldn't find the man page for config!
```

That is because in developer mode, the variable `KW_MAN_DIR` is set to `${KW_BASE_DIR}/documentation/man` instead of `${KW_SHARE_DIR}/man` and I did not have man pages at `${KW_BASE_DIR}/documentation/man` but I had at `${KW_SHARE_DIR}/man.

I thought my installation was wrong and I spent several minutes untill I figured out the source of the problem... Later, I thought there was a bug with the `manpage` module and almost opened a new issue, but fortunately I realized it was not a bug.

By telling the developer `kw` is running in developer mode, he/she can be aware of possible changes in the behavior of `kw` and not mistake some different behavior (due to change in the `env` variables) as if it were a bug.

This can save precious time of developers of `kw` in the future.

uwla:
When kw runs in developer mode, some env vars are changed to point the cache/data/db/doc/etc/lib/man directories to the repository in which the `kw` script is at.

This can cause confusion to developers, which may expect some files (such as database files, manual pages,  plugin files) to be somewhere else and may think kw has a bug (due to devmode) when it doesn't.

This PR basically shows a notice message to inform the developer that kw is running in developer mode and some envvars have been changed. The message is sent to stderror to prevent failing tests (because it would disrupt the output expected to the commands).

**OBS**: I made this commit because I myself ran into problems due to the variables being changed without my awareness.

Closes #956

### DIFF
--- a/kw
+++ b/kw
@@ -36,6 +36,19 @@ KW_STATE_DIR="${XDG_STATE_HOME:-"${HOME}/.local/state"}/${KWORKFLOW}"
 # ENV dir name
 ENV_DIR='envs'
 
+# Export external variables required by kworkflow
+export KWORKFLOW
+
+#INJECT_CODE_TRACING_SETUP
+
+# This is the one and only time a file will be sourced this way.
+# The include function (sourced from this file) should always be used for file sourcing.
+. "${KW_LIB_DIR}/lib/kw_include.sh" --source-only
+
+
+# Dev mode has to begin after sourcing lib/kw_include.sh
+# Otherwise, it is not possible to use helpers.
+
 ##BEGIN-DEV-MODE##
 KW_BIN="$(readlink -f "${BASH_SOURCE[0]}")"
 KW_BASE_DIR="$(dirname "${KW_BIN}")"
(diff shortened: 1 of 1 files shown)


uwla:
I realized I have wrongly named the branch from which I made the PR.
I renamed the branch on Github, but this caused the PR to be closed.
I wish the PR would not have been closed but the branch name would have been updated.

uwla:
@davidbtadokoro , what should I do?
Should I restore the old branch (which has the wrong name) ?
Or should I create a new PR from the branch with the correct name?

uwla:
By the way, in the renamed branch, I have fixed the code that failed the CI (it was an extra empty line which caused the shell formater check to fail).

PR #958, opened 2023-11-10
Add verbose to debug

rodrigosiqueira:


src/lib/kwlib.sh
```diff
@@ -118,6 +118,14 @@ function cmd_manager()
   eval "$command_for_eval"
 }
 
+function show_verbose()
+{
+  local flag="$1"
+  local cmd="$2"
+
+  [[ "$flag" == 'VERBOSE' ]] && printf '%s\n' "$cmd"
+}
```
davidbtadokoro, line 127: Think it is worth adding a simple unit test for this.

src/debug.sh
```diff
@@ -88,7 +88,7 @@ function debug_main()
   fi
 
   if [[ -n "$reset" ]]; then
-    reset_debug "$target"
+    reset_debug "$target" "$flag"
```
davidbtadokoro, line 91: Can't we squash this commit with the previous one?

davidbtadokoro (changes requested):
Hey @rodrigosiqueira, thanks so much for this verbose change!

Left some comments so see if you can address them.

As a side note, I tested the verbose behaviour, but don't really know how to use the `kw debug` feature to test it further than `kw debug --list`. Can you mention some commands for me to test?

rodrigosiqueira:
> Hey @rodrigosiqueira, thanks so much for this verbose change!
> 
> Left some comments so see if you can address them.
> 
> As a side note, I tested the verbose behaviour, but don't really know how to use the `kw debug` feature to test it further than `kw debug --list`. Can you mention some commands for me to test?

Thanks, I'll update the PR.

Btw, we have a tutorial for this:

https://kworkflow.org/tutorials/kernel-debug.html

src/lib/kwlib.sh
```diff
@@ -118,6 +118,14 @@ function cmd_manager()
   eval "$command_for_eval"
 }
 
+function show_verbose()
+{
+  local flag="$1"
+  local cmd="$2"
+
+  [[ "$flag" == 'VERBOSE' ]] && printf '%s\n' "$cmd"
```
aquilamacedo, line 126: Wouldn't it be better to use `say "$cmd"` here? It would ensure that the output adheres to the blue color standard of the verbose mode.

rodrigosiqueira:
@davidbtadokoro @aquilamacedo Thanks for your feedback. I updated this PR based on your suggestions.

davidbtadokoro:
Thanks for the update @rodrigosiqueira! Merged into the unstable branch.

PR #959, opened 2023-11-11
`patch-hub`: Move function `handle_exit` and update documentation

davidbtadokoro:
The function `handle_exit` used by `patch-hub` Controller component to stop the execution of the FSM is:

1. Implemented in a out-of-place way, inside `src/lib/dialog_ui.sh`. It fits better inside the file `src/ui/patch_hub/patch_hub_core.sh`;
2. Has a misleading documentation stating it is used to handle the exit of Dialog (View component).

This PR introduces two commits that try to address these issues.

### DIFF
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -911,22 +911,6 @@ function run_dialog_command()
   return "$ret"
 }
 
-# This function is responsible for handling the dialog exit.
-#
-# @exit_status: Exit code
-function handle_exit()
-{
-  local exit_status="$1"
-
-  # Handling stop
-  case "$exit_status" in
-    1 | 22 | 255) # Exit
-      clear
-      exit 0
-      ;;
-  esac
-}
-
 function prettify_string()
 {
   local fixed_text="$1"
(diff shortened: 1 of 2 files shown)


src/lib/dialog_ui.sh
rodrigosiqueira, line 1: hmmm... I don't agree with this change... I was working on a draft of the visual setup of `kw init` which uses dialog. With this change, I'll need to include `patch_hub_core.sh`, which does not make sense to me.

rodrigosiqueira (changes requested):

src/lib/dialog_ui.sh
davidbtadokoro: After reading your comment, I'm with you on that. `handle_exit` is a function to be used in other UIs. That being said, I don't think the function fits `src/lib/dialog_ui.sh` neither. What do you say to moving it to `kwlib.sh` or createing a lib function in `src/ui` for UIs?

rodrigosiqueira: > That being said, I don't think the function fits src/lib/dialog_ui.sh neither.

But I think the error codes inside this function and the `clear` are particular behaviors associated with the dialog.

PR #960, opened 2023-11-11
`patch-hub`: Update handler function of `dashboard` state

davidbtadokoro:
The function `dashboard_entry_menu` was one of the first functions that handles a state in the FSM, and it is outdated, as the function name does not follow the pattern of `show_<name_of_state>`, and prints a message in the standard output that won't be displayed to the user.

This PR introduces two commits to address both issues.


rodrigosiqueira:
Merged! Thanks

PR #963, opened 2023-11-15
src: lib: lore: Change /tmp/ to /dev/shm dir

### LINKED ISSUE #902: `kw patch-hub`: Use `/dev/shm` instead of `/tmp` for better performance
**Description:**

When fetching patchsets, `kw patch-hub` uses files created in `/tmp` to process them. I suspect that there is a bottleneck in the I/O operations that can be mitigated using `/dev/shm` instead.

**Type:**

Enhancement.

**Why:**

`kw patch-hub` performance for fetching patchsets is poor and we should optimize it.

JGBSouza:
To enable parallel processing of the patches, kw used the slower option, /tmp/ directory, instead of /dev/shm/ directory.

 In this case, to speed up the process, change the directory to /dev/shm.

This PR resolves #902

### DIFF
--- a/src/lib/kwlib.sh
+++ b/src/lib/kwlib.sh
@@ -4,6 +4,7 @@ include "${KW_LIB_DIR}/lib/kw_db.sh"
 include "${KW_LIB_DIR}/lib/kw_time_and_date.sh"
 
 ENV_DIR='envs'
+KW_SHARED_MEMORY_DEFAULT_DIR='/dev/shm'
 
 # Array with compression programs accepted by tar
 declare -ga compression_programs=('gzip' 'bzip2' 'lzip' 'lzma' 'lzop' 'zstd'
@@ -124,6 +125,24 @@ function show_verbose()
   local cmd="$2"
 
   [[ "$flag" == 'VERBOSE' ]] && say "$cmd"
+}
+
+# This function create the shared memory directory, if
+# @KW_SHARED_MEMORY_DEFAULT_DIR is not available, create it on '/tmp/' instead
+#
+# @KW_SHARED_MEMORY_DEFAULT_DIR is a global variable that sets the target
+# '/dev/shm' directory
+#
+# Returns:
+# Returns the path for the new directory
+function create_shared_memory_dir()
+{
+  if [[ -d "$KW_SHARED_MEMORY_DEFAULT_DIR" && -w "$KW_SHARED_MEMORY_DEFAULT_DIR" ]]; then
+    printf '%s' "$(mktemp --tmpdir="$KW_SHARED_MEMORY_DEFAULT_DIR" --directory)"
+    return
+  fi
+
+  printf '%s' "$(mktemp --directory)"
 }
 
 # Checks if a directory is a kernel tree root
--- a/src/lib/lore.sh
+++ b/src/lib/lore.sh
@@ -433,7 +433,8 @@ function process_patchsets()
   local pids
   local i
 
-  shared_dir_for_parallelism=$(mktemp --directory)
+  shared_dir_for_parallelism=$(create_shared_memory_dir)
+
   starting_index="$PATCHSETS_PROCESSED"
   count=0
   i=0
--- a/tests/unit/lib/kwlib_test.sh
+++ b/tests/unit/lib/kwlib_test.sh
@@ -95,6 +95,22 @@ function teardownGitRepository()
   fi
 }
 
+function setupKwSharedMemoryDefaultDir()
+{
+  KW_SHARED_MEMORY_DEFAULT_DIR="${SHUNIT_TMPDIR}/test/tmp_dir"
+  mkdir --parents "$KW_SHARED_MEMORY_DEFAULT_DIR"
+}
+
+function teardownSharedMemoryDefaultDir()
+{
+  is_safe_path_to_remove "$KW_SHARED_MEMORY_DEFAULT_DIR"
+  if [[ "$?" == 0 ]]; then
+    rm --recursive "$KW_SHARED_MEMORY_DEFAULT_DIR"
+  else
+    fail 'It was not possible to remove TMP_DIR in the specified directory'
+  fi
+}
+
 function test_is_kernel_root()
 {
   is_kernel_root "$SHUNIT_TMPDIR"
@@ -963,6 +979,28 @@ function test_show_verbose()
 
   output=$(show_verbose 'VERBOSE' "$cmd")
   assert_equals_helper 'Expected value of command' "$LINENO" "$cmd" "$output"
+}
+
+function test_check_if_create_shared_memory_dir_creates_dir_in_kw_shared_memory_default_dir()
+{
+  local output
+
+  setupKwSharedMemoryDefaultDir
+
+  output=$(create_shared_memory_dir)
+  assertTrue "${LINENO}: Did not create directory at ${output}" "[[ -d ${output} ]]"
+  assertTrue "${LINENO}: Did not create directory at KW_SHARED_MEMORY_DEFAULT_DIR" '[[ $output =~ $KW_SHARED_MEMORY_DEFAULT_DIR ]]'
+
+  teardownSharedMemoryDefaultDir
+}
+
+function test_check_if_create_shared_memory_dir_creates_dir_in_default_mktemp_dir()
+{
+  local output
+
+  output=$(create_shared_memory_dir)
+  assertTrue "${LINENO}: Did not create directory at ${output}" "[[ -d ${output} ]]"
+  assertTrue "${LINENO}: Did not create directory at /tmp/ dir" '[[ $output =~ /tmp/ ]]'
 }
 
 invoke_shunit


src/lib/lore.sh
```diff
@@ -415,7 +415,8 @@ function process_patchsets()
   local pids
   local i
 
-  shared_dir_for_parallelism=$(mktemp --directory)
+  shared_dir_for_parallelism=$(mktemp -p /dev/shm --directory)
```
rodrigosiqueira, line 418: Before creating a folder inside `/dev/shm`, ensure it exists and that you have permission to change it. Also, check: https://kworkflow.org/content/codingstyle.html#avoid-short-options

davidbtadokoro: Agree, you should try using `/dev/shm` and fallback to the default directory of `mktemp --directory`.

Also, what you guys think of creating a dedicated function like `create_shared_memory_dir`, to add automated tests? Maybe inside `kwlib.sh`.

tests/unit/lib/kwlib_test.sh
```diff
@@ -965,4 +965,18 @@ function test_show_verbose()
...
+function test_create_shared_memory_dir()
+{
+  local output
+
+  TMP_DIR="$SHUNIT_TMPDIR/test/tmp_dir"
+  mkdir --parents "$TMP_DIR"
+  output=$(create_shared_memory_dir)
+  assertContains 'Unable to create "TMP_DIR" in the specified directory' "$output" "$TMP_DIR"
+
+  rm -r "$TMP_DIR"
```
rodrigosiqueira, line 977: Always check if a variable exists and is valid before the `rm` command.

tests/unit/lib/kwlib_test.sh
```diff
@@ -965,4 +965,18 @@ function test_show_verbose()
...
+  output=$(create_shared_memory_dir)
+  assertContains 'Unable to create TMP_DIR on "/tmp/" directory' "$output" "/tmp/"
```
rodrigosiqueira, line 979: Single quotes around '/tmp/'

rodrigosiqueira (commented):
Keep the commit body message under 75 characters.

src/lib/kwlib.sh
```diff
@@ -126,6 +127,21 @@ function show_verbose()
   [[ "$flag" == 'VERBOSE' ]] && say "$cmd"
 }
 
+# This function create the shared memory directory
+# @TMP_DIR is a global variable that sets the target '/dev/shm' directory
```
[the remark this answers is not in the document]
davidbtadokoro: Although the line starts with `@TMP_DIR` this isn't a argument description :upside_down_face:

davidbtadokoro (approved):
Hey @JGBSouza, and thanks a lot for the continuous updates on this PR! Everything looks legit to me. @rodrigosiqueira, going to wait for your green-light, as you also reviewed this PR.

src/lib/kwlib.sh
```diff
@@ -126,6 +127,21 @@ function show_verbose()
...
+# if @TMP_DIR is not available create the directory on '/tmp/' instead
+# Returns:
+# The path for the new directory
+function create_shared_memory_dir()
+{
+  if [[ -d "$TMP_DIR" ]] && [[ -w "$TMP_DIR" ]]; then
+    printf '%s' "$(mktemp --tmpdir="$TMP_DIR" --directory)"
+    return
+  fi
+  printf '%s' "$(mktemp --directory)"
```
rodrigosiqueira, line 142: Add a new line.

src/lib/kwlib.sh
```diff
@@ -126,6 +127,21 @@ function show_verbose()
...
 
+# This function create the shared memory directory
+# @TMP_DIR is a global variable that sets the target '/dev/shm' directory
+#
+# if @TMP_DIR is not available create the directory on '/tmp/' instead
+# Returns:
+# The path for the new directory
+function create_shared_memory_dir()
+{
+  if [[ -d "$TMP_DIR" ]] && [[ -w "$TMP_DIR" ]]; then
```
rodrigosiqueira, line 138: Why do we need two `[[..]] && [[..]]`? Why not just use `[[ ... && ]]`?

src/lib/kwlib.sh
```diff
@@ -126,6 +127,21 @@ function show_verbose()
   [[ "$flag" == 'VERBOSE' ]] && say "$cmd"
 }
 
+# This function create the shared memory directory
```
rodrigosiqueira, line 130: Technically, you added a fallback behavior that creates the folder in the `/tmp`; you must highlight this in the doc. Also, add an empty line between the doc description and the parameter.

src/lib/kwlib.sh
```diff
@@ -126,6 +127,21 @@ function show_verbose()
...
+# @TMP_DIR is a global variable that sets the target '/dev/shm' directory
+#
+# if @TMP_DIR is not available create the directory on '/tmp/' instead
+# Returns:
```
rodrigosiqueira, line 134: Add an empty comment line before `Return`.

src/lib/kwlib.sh
```diff
@@ -4,6 +4,7 @@ include "${KW_LIB_DIR}/lib/kw_db.sh"
 include "${KW_LIB_DIR}/lib/kw_time_and_date.sh"
 
 ENV_DIR='envs'
+TMP_DIR='/dev/shm'
```
rodrigosiqueira, line 7: This name seems too generic... try something like `KW_SHARED_MEMORY_DEFAULT_DIR`... or something shorter but meaningful. Also, add a small doc about this global variable.

tests/unit/lib/kwlib_test.sh
```diff
@@ -965,4 +965,25 @@ function test_show_verbose()
...
 
+function test_create_shared_memory_dir()
+{
+  local output
+
+  TMP_DIR="${SHUNIT_TMPDIR}/test/tmp_dir"
+  mkdir --parents "$TMP_DIR"
+
+  output=$(create_shared_memory_dir)
+  assertContains "Unable to create TMP_DIR in the specified directory" "$output" "$TMP_DIR"
```
rodrigosiqueira, line 976: Use single-quotes in the assert message.

tests/unit/lib/kwlib_test.sh
```diff
@@ -965,4 +965,25 @@ function test_show_verbose()
   assert_equals_helper 'Expected value of command' "$LINENO" "$cmd" "$output"
 }
 
+function test_create_shared_memory_dir()
```
rodrigosiqueira, line 968: Change this test name to something more meaningful, something like: `test_check_if_create_shared_memory_dir_creates_dir_in_TMP_DIR`.

tests/unit/lib/kwlib_test.sh
rodrigosiqueira: Tbh... I'm not sure if I liked this `prefix` test... it is ok, but not deep enough in terms of validation. How about checking if the `TMP_DIR` folder is clean; next you run the `create_shared_memory_dir`, and you should expect a new folder to become part of the `TMP_DIR`. You can even collect the return of the function, and use it to see if it exists in the assert (e.g., `assertTrue ... [[ -d "$output" ]]`.

tests/unit/lib/kwlib_test.sh
```diff
@@ -965,4 +965,25 @@ function test_show_verbose()
...
+{
+  local output
+
+  TMP_DIR="${SHUNIT_TMPDIR}/test/tmp_dir"
+  mkdir --parents "$TMP_DIR"
```
rodrigosiqueira, line 973: Can we move this thing to the `setUp` phase?

tests/unit/lib/kwlib_test.sh
```diff
@@ -965,4 +965,25 @@ function test_show_verbose()
...
+
+  output=$(create_shared_memory_dir)
+  assertContains "Unable to create TMP_DIR in the specified directory" "$output" "$TMP_DIR"
+
+  is_safe_path_to_remove "$TMP_DIR"
+  if [[ "$?" == 0 ]]; then
+    rm --recursive "$TMP_DIR"
+  else
+    fail "It was not possible to remove TMP_DIR in the specified directory"
+  fi
```
rodrigosiqueira, line 983: Can we move this to the `tearDown` phase?

tests/unit/lib/kwlib_test.sh
```diff
@@ -965,4 +965,25 @@ function test_show_verbose()
...
+
+  output=$(create_shared_memory_dir)
+  assertContains "Unable to create TMP_DIR on /tmp/ directory" "$output" '/tmp/'
```
rodrigosiqueira, line 986: This should be a separated test.

rodrigosiqueira (changes requested):

tests/unit/lib/kwlib_test.sh
JGBSouza: > Use single-quotes in the assert message.

Just to be shure, even in the assert messages from the shunit2 should we use single quotes? because all the assertEquals, even the ones without variable expansion, uses double quotes like in 'test_find_kernel_root()' or in 'test_join_path()'

davidbtadokoro: Hey João. I see how this may cause some confusion. There are some parts of kw that are indeed outdated, so don't take them as reference. The general rule is if there are no expansions, use single quotes.

davidbtadokoro (changes requested):
Hi @JGBSouza, thanks a lot for the update! We are almost at that perfect spot, just left some comments, so please, see if you can address them.

tests/unit/lib/kwlib_test.sh
```diff
@@ -965,4 +981,28 @@ function test_show_verbose()
...
 }
 
+function test_check_if_create_shared_memory_dir_creates_dir_in_KW_SHARED_MEMORY_DEFAULT_DIR()
+{
+  local output
+  local number_of_files
+
+  setupKwSharedMemoryDefaultDir
+
+  number_of_files=$(find "$KW_SHARED_MEMORY_DEFAULT_DIR" -maxdepth 1 -type f | wc -l)
```
rodrigosiqueira, line 991: use `--lines` instead of `-l`.

tests/unit/lib/kwlib_test.sh
```diff
@@ -965,4 +981,28 @@ function test_show_verbose()
...
+  output=$(create_shared_memory_dir)
+  assertTrue "$LINENO: Temporary_dir Was not created" "[[ -d $output ]]"
+  assertTrue "$LINENO: Temporary_dir was not created in the KW_SHARED_MEMORY_DEFAULT_DIR" '[[ $output =~ $KW_SHARED_MEMORY_DEFAULT_DIR ]]'
+
+  teardownSharedMemoryDefaultDir
+}
+
+function test_check_if_create_shared_memory_dir_creates_dir_in_TMP_DIR()
+{
+  output=$(create_shared_memory_dir)
```
rodrigosiqueira, line 1003: Don't forget to declare `output` as `local` in this function scope.

rodrigosiqueira (changes requested):

davidbtadokoro:
Hi @JGBSouza thanks a lot for the many updates you did for this PR, sorry for the delay in this last review, and also thanks @rodrigosiqueira for the thorough reviews. Everything looks great to me, so merged into the unstable branch!

PR #966, opened 2023-11-16
kw: show notice message if kw runs in developer mode

uwla:
When kw runs in developer mode, some env vars are changed to point the cache/data/db/doc/etc/lib/man directories to the repository in which the kw script is at.

This can cause confusion to developers, which may expect some files (such as database files, manual pages, plugin files) to be somewhere else and may think kw has a bug (due to devmode) when it doesn't.

This PR basically shows a notice message to inform the developer that kw is running in developer mode and some envvars have been changed. The message is sent to stderror to prevent failing tests (because it would disrupt the output expected to the commands).

OBS: I made this commit because I myself ran into problems due to the variables being changed without my awareness.

Closes https://github.com/kworkflow/kworkflow/issues/956

### DIFF
--- a/kw
+++ b/kw
@@ -36,11 +36,13 @@ KW_STATE_DIR="${XDG_STATE_HOME:-"${HOME}/.local/state"}/${KWORKFLOW}"
 # ENV dir name
 ENV_DIR='envs'
 
-##BEGIN-DEV-MODE##
+##BEGIN-REPO-MODE##
 KW_BIN="$(readlink -f "${BASH_SOURCE[0]}")"
 KW_BASE_DIR="$(dirname "${KW_BIN}")"
+KW_REPO_MODE='n'
 if [ -f "${KW_BASE_DIR}/src/lib/kwlib.sh" ]; then
   # running from source directory
+  KW_REPO_MODE='y'
   KW_LIB_DIR="${KW_BASE_DIR}/src"
   # KW_DATA_DIR # use default data folder
   KW_DOC_DIR="${KW_BASE_DIR}/documentation"
@@ -51,7 +53,7 @@ if [ -f "${KW_BASE_DIR}/src/lib/kwlib.sh" ]; then
   KW_PLUGINS_DIR="${KW_LIB_DIR}/plugins"
   # KW_CACHE_DIR # use default cache folder
 fi
-##END-DEV-MODE##
+##END-REPO-MODE##
 
 # Export external variables required by kworkflow
 export KWORKFLOW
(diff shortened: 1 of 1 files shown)


kw
```diff
@@ -36,6 +36,18 @@ KW_STATE_DIR="${XDG_STATE_HOME:-"${HOME}/.local/state"}/${KWORKFLOW}"
...
+export KWORKFLOW
+
+#INJECT_CODE_TRACING_SETUP
+
+# This is the one and only time a file will be sourced this way.
+# The include function (sourced from this file) should always be used for file sourcing.
+. "${KW_LIB_DIR}/lib/kw_include.sh" --source-only
+
+# Dev mode has to begin after sourcing lib/kw_include.sh
+# Otherwise, it is not possible to use helpers.
```
davidbtadokoro, line 49: I see that you moved this sourcing to be able to use the `say` function, but moving it to here introduced a regression. When I run kw using the repo executable without having kw installed in my system, kw crashes:

Screen Shot 2023-11-17 at 22 10 18

I think that this is due to the else in the if-else below being executed at the start of the kw file:

```bash
# Kw shell files
if [[ -f '/usr/bin/kw' ]]; then
  KW_LIB_DIR='/usr/share/kw'
else
  KW_LIB_DIR="${KW_LIB_DIR:-"${HOME}/.local/lib"}/${KWORKFLOW}"
fi
```

You can either drop this move and don't use the `say` function (however we lose colors 😢), or source the file directly from the repo without using `include` (but I don't think breaking the rule of sourcing with `include` is wise). You could also use colors directly, but you need to be extra sure to not mess the user terminal session 🙃.

kw
```diff
@@ -36,6 +36,18 @@ KW_STATE_DIR="${XDG_STATE_HOME:-"${HOME}/.local/state"}/${KWORKFLOW}"
...
+
 ##BEGIN-DEV-MODE##
```
davidbtadokoro, line 51: I had an offline conversation with @rodrigosiqueira and he mentioned changing the name DEV-MODE to something else like REPO-MODE, or else (I like REPO-MODE, but feel free to say your piece about other names 👍). So see if you can add a commit previous to this one just to change the name.

kw
```diff
@@ -50,17 +62,27 @@ if [ -f "${KW_BASE_DIR}/src/lib/kwlib.sh" ]; then
...
 
-# Export external variables required by kworkflow
-export KWORKFLOW
+  # include IO utilities
+  include "${KW_LIB_DIR}/lib/kwio.sh"
 
-#INJECT_CODE_TRACING_SETUP
+  # Print a notice to the developers so he knows some envvars have been changed
+  # We redirect to stderror because this output would make tests fails if the developer is testing
+  say "[INFO]: Running kw in developer mode. Envvars set to the following:
```
davidbtadokoro, line 71: Besides changing the reference `developer mode` to the new name, maybe break the line before `Envvars set to (...)`. Also, we could be a little more verbose with and use `kw environment variables set to the following:`.

kw
```diff
@@ -50,17 +62,27 @@ if [ -f "${KW_BASE_DIR}/src/lib/kwlib.sh" ]; then
...
-. "${KW_LIB_DIR}/lib/kw_include.sh" --source-only
+KW_CACHE_DIR =${KW_CACHE_DIR}
+KW_DATA_DIR=${KW_DATA_DIR}
+KW_DB_DIR=${KW_DB_DIR}
+KW_DOC_DIR=${KW_DOC_DIR}
+KW_ETC_DIR=${KW_ETC_DIR}
+KW_LIB_DIR=${KW_LIB_DIR}
+KW_MAN_DIR=${KW_MAN_DIR}
+KW_PLUGINS_DIR=${KW_PLUGINS_DIR}
+KW_SOUND_DIR=${KW_SOUND_DIR}
```
davidbtadokoro, line 81: See if you can break the lines conform the [code style](https://kworkflow.org/content/codingstyle.html#breaking-long-lines-and-strings). To add newlines use something like the following:

```bash
dev_mode_message='[INFO]: Running kw in developer mode.'$'\n'
dev_mode_message+='kw environment variables set to the following:'$'\n'
dev_mode_message+="KW_CACHE_DIR =${KW_CACHE_DIR}"$'\n'
...

printf '%s\n' "$dev_mode_message"
```

kw
```diff
@@ -50,17 +62,27 @@ if [ -f "${KW_BASE_DIR}/src/lib/kwlib.sh" ]; then
...
+" >&2
```
davidbtadokoro, line 82: Could explain to me the reason for using the std-err stream? 😅

davidbtadokoro (changes requested):
Hey @uwla thank you for the proposed change!

Left some comments so see if you can tackle them. Don't hesitate to ping me 😄.

kw
uwla: surely! 

If the developer is running in "repo mode", he/she will probably want to run the tests at some moment; sending the message `you are running kw in developer mode` will modify the expected output of kw in the tests and I thought the tests may fail.

As I said, I thought the tests may fail, but I just ran the tests here with the notice message sent to std-out and all tests passed... That means my guess was wrong. I'll use std-out instead of std-err

kw
uwla: alright, I'll make it

kw
uwla: I won't use the `say` function and will instead use some colors directly

kw
uwla: That is a better name indeed. I'll change it.

uwla:
@davidbtadokoro I think I have made the requested changes

kw
davidbtadokoro: Oh, I get it now! Thanks for the explanation, I confess I agree with your reasoning, but since the tests don't fail using std-out, its better to use it for clarity.

davidbtadokoro:
Hey @uwla, thanks for the update! I made the change of changing the name to REPO-MODE in a previous commit.

Merged into the unstable branch 👍

PR #967, opened 2023-11-17
documentation: conf: Add prefix to man pages

rodrigosiqueira:
While packaging kw for Debian, we noticed that kw man was only partially available due to the lack of a name prefix. This commit introduces the 'kw-' prefix to each man page used by kw, which enables a better way to distribute the man page system-wide.


davidbtadokoro:
Hi @rodrigosiqueira, thanks for this change. Tested and reviewed the code and everything looks legit to me! Merged into the unstable branch 👍

davidbtadokoro:
Hi @rodrigosiqueira, thanks for this change. Tested and reviewed the code and everything looks legit to me! Merged into the unstable branch 👍

Issue #935, opened 2023-10-07
Modernize codestyle


fernandohjunqueira:
**Description:**

The codestyle feature is old and outdated, lets:

- Use the new parser system (main, help, parse...)
- Use a better way to display data
- Update and expand unit tests

jppaulo06:
@andregnl and I will tackle this issue soon.

PR #974, opened 2023-11-22
src: checkpatch_wrapper: Modernize codestyle

Andregnl:
The checkpatch_wrapper script was out of date in terms of code style, so modernize checkpatch_wrapper.sh to follow the current code standard. Changes include:
- Use parse function to parse arguments, so now arguments work in any order
- Rename checkpatch_wrapper.sh to codestyle.sh
- Rename checkpatch_wrapper_test.sh to codestyle_test.sh
- Update man page to include verbose option
- Update autocomplete to include help and verbose options

Issue #935


rodrigosiqueira:
Hi @Andregnl , this PR lgtm. Please, rebase it against the latest unstable branch.

Andregnl:
Ok, will do!

src/codestyle.sh
```diff
@@ -0,0 +1,161 @@
...
+  local checkpatch
+  local cmd_script
+
+  parse_codestyle_options "$@"
+  if [[ "$?" != 0 ]]; then
+    complain "Invalid option: ${options_values['ERROR']}"
+    return 22 # EINVAL
+  fi
+
+  flag=${options_values['TEST_MODE']}
```
davidbtadokoro, line 29: If I understood correctly, you are using this `TEST_MODE` option passed for testing reasons. However, the options accepted in the parse functions (`kw codestyle <option>`) should be dedicated to the user, so adding this option (which the user will not use) doesn't fit well.

My suggestion it to create a function inside `codestyle_main` that is called after the parsing of options and settings of the `flag` variable, and that encapsulate all logic currently done by `codestyle_main`. This created function can then accept a `flag` argument and you can pass the value `TEST_MODE` without having to add it as an option to the parse phase.

**Obs.:** If you look throughout the project, we usually don't test the main function mainly because of this problem of parsing and setting `TEST_MODE` flags. The main function should be a "mediator" and delegate the real work for other functions, which are the actual units to test. That being said, I know you did this way because of the legacy way `checkpatch_wrapper` was implemented :grimacing:

src/codestyle.sh
```diff
@@ -0,0 +1,161 @@
...
+  [[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
+
+  flag=${flag:-'SILENT'}
+  # TODO: Note that codespell file is not specified yet because of the poluted
+  # output. It could be nice if we can add another option just for this sort
+  # of check.
+
+  path="${options_values['PATH']}"
+  path=${path:-'.'}
+  validate_path "$path"
```
davidbtadokoro, line 39: As the check is really simple and lean, I think we can remove `validate_path` and explicitly do the check here.

src/codestyle.sh
```diff
@@ -0,0 +1,161 @@
...
+}
+
+function validate_path()
+{
+  local path="$*"
+  if [[ ! -d "$path" && ! -f "$path" ]]; then
+    return 2 # ENOENT
+  fi
+  return 0
+}
```
davidbtadokoro, line 97: As I commented above, maybe remove this function.

src/codestyle.sh
```diff
@@ -0,0 +1,161 @@
...
+
+# This function gets raw data and based on that fill out the options values to
+# be used in another function.
+#
+# Return:
+# In case of successful return 0, otherwise, return 22.
+function parse_codestyle_options()
+{
+  local long_options='verbose,test_mode,help'
```
davidbtadokoro, line 106: As I commented before, remove `test_mode` as a feature option exposed to the user.

src/codestyle.sh
```diff
@@ -0,0 +1,161 @@
...
+
+  if [[ "$?" != 0 ]]; then
+    options_values['ERROR']="$(kw_parse_get_errors 'kw diff' "$short_options" \
+      "$long_options" "$@")"
+    return 22 # EINVAL
+  fi
+
+  # Default values
+  options_values['VERBOSE']=''
+  options_values['TEST_MODE']=''
```
davidbtadokoro, line 120: Remove

src/codestyle.sh
```diff
@@ -0,0 +1,161 @@
...
+  while [[ "$#" -gt 0 ]]; do
+    case "$1" in
+      --verbose)
+        options_values['VERBOSE']=1
+        shift
+        ;;
+      TEST_MODE)
+        options_values['TEST_MODE']='TEST_MODE'
+        shift
+        ;;
```
davidbtadokoro, line 133: Remove

davidbtadokoro (changes requested):
Hey @Andregnl and @jppaulo06, thanks for this PR! Left some inline comments, so feel free to ping me if any doubts surge.

@rodrigosiqueira I tested it on my end and it seems that it didn't break the feature, but, just to be safe, can you also test it (just the behavior)?

Also, @Andregnl and @jppaulo06, when you update this PR, see if you can add a first commit that just renames the files. I say this because renaming and changing a file is tricky if git does not recognizes that the file was renamed. If you look in the 'Commits' tab of this PR, you will see that `checkpatch_wrapper.sh` was completely removed and that `codestyle.sh` was created, so `codestyle.sh` has only additions and the 'diffs' (the changes you made to the renamed file) get lost. This is bad for both reviewing (as bugs and the like can easily sneak in) and for the git log itself. Does it makes sense?

src/codestyle.sh
jppaulo06: Oh, I understand.  We can surely do it :)

However, I think it's important to mention that some of the other kw features do this same thing when parsing options. Examples are:
- `kw ssh`
- `kw explore`
- `kw diff`
- `kw debug`
- `kw deploy`

So, wouldn't it be more right to change everything instead of just `kw codestyle`? 

Nonetheless, there is one specific line that really doesn't make sense for me now, which we copied from `kw ssh`:
```shell
  local long_options='verbose,test_mode,help'
```
I think it's unnecessary to put 'test_mode' there since the user might never execute `kw codestyle --test_mode`; so we should remove it.

jppaulo06:
> Hey @Andregnl and @jppaulo06, thanks for this PR! Left some inline comments, so feel free to ping me if any doubts surge.
> 
> Also, see if you can add a first commit that just renames the files. I say this because renaming and changing a file is tricky if git does not recognizes that the file was renamed. If you look in the 'Commits' tab of this PR, you will see that `checkpatch_wrapper.sh` was completely removed and that `codestyle.sh` was created, so `codestyle.sh` has only additions and the 'diffs' (the changes you made to the renamed file) get lost. This is bad for both reviewing (as bugs and the like can easily sneak in) and for the git log itself. Does it makes sense?

I'm struggling for a bit now with this. @davidbtadokoro, do you have any suggestion on how to solve this problem in a simple way?

src/codestyle.sh
```diff
@@ -5,41 +5,47 @@
...
+    return 22 # EINVAL
+  fi
+
+  flag=${options_values['TEST_MODE']}
+  [[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
+
   flag=${flag:-'SILENT'}
+  # TODO: Note that codespell file is not specified yet because of the poluted
+  # output. It could be nice if we can add another option just for this sort
+  # of check.
```
rodrigosiqueira, line 35: Please, create an issue for that.

PR #976, opened 2023-11-25
documentation: man: features: Add whatis entry

rodrigosiqueira:
While packaging kw for Debian, we got multiple warnings like this:

 kworkflow: bad-whatis-entry [usr/share/man/man1/kw-build.1.gz]

This commit introduce the whatis for each kw feature.


documentation/man/features/kw-pomodoro.rst
```diff
@@ -4,6 +4,10 @@ kw-pomodoro
 
 .. _pomodoro-doc:
 
+--------------------------------
+UI with lore.kernel.org archives
```
davidbtadokoro, line 8: Maybe 'Create and manage Pomodoro timers'

documentation/man/features/kw-report.rst
```diff
@@ -4,6 +4,10 @@ kw-report
 
 .. _report-doc:
 
+------------------------
+User data report support
```
davidbtadokoro, line 8: Maybe 'Display user reports' or 'Display user data reports'

documentation/man/features/kw-mail.rst
```diff
@@ -4,6 +4,10 @@ kw-mail
 
 .. _mail-doc:
 
+-------
+kw mail
```
davidbtadokoro, line 8: I think it is hard to summarize `kw mail`, but maybe something more specific like 'Send patches through mail'

davidbtadokoro (commented):
Thanks for the PR @rodrigosiqueira! Left some comments that are more about style. See if they have some merit, but overall LGTM, so feel free to merge this (or ping me for it).

davidbtadokoro:
@rodrigosiqueira, actually maybe there is something wrong. Should the 'whatis' should be displayed like in the screenshots below?

Screenshot from 2023-11-27 20-45-21
Screenshot from 2023-11-27 20-45-32

Isn't the description in the `NAME` section the 'whatis' (in the second screenshot is the 'pomodoro style time management' string)?

rodrigosiqueira:
@davidbtadokoro This information came from the config.py; if we do not use this file, we hit the error I mentioned in the commit message. In the Debian package, I just use the rst2man command without referencing the config file.

davidbtadokoro:
Hey @rodrigosiqueira, and thanks for finding a solution that works for both the packaging of kw for Debian and for the local man pages. Merged into the unstable branch :+1:

PR #978, opened 2023-12-01
Remove duplicated config entries

uwla:
Very basic stuff: just remove duplicated entries in config array.

Though trivial, I think it is more relevant than it seems: while I was working on some features for the config module, there were several times while I saw duplicated lines in `config`'s output, leading me to think I had introduced some sort of bug and wasting time trying to find it. Hopefully, other devs won't be fooled by it.


rodrigosiqueira:
Nice change. Merged into the unstable branch.
Thanks

PR #979, opened 2023-12-01
setup.sh: add flag --skip-docs to skip generating man pages

uwla:
With this new flag,the user can install KW without generating man pages, which takes time due to sphinx.  This  drastically  reduce  installation time and is specially useful in cases where the  manual  pages  are  not needed, such as local development and continuous integration.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -9,6 +9,7 @@ SILENT=1
 VERBOSE=0
 FORCE=0
 SKIPCHECKS=0
+SKIPDOCS=0
 ENABLE_TRACING=0
 
 declare -r app_name='kw'
@@ -224,6 +225,7 @@ function usage()
   say "--install   | -i     Install $app_name"
   say "--uninstall | -u     Uninstall $app_name"
   say '--skip-checks        Skip checks (use this when packaging)'
+  say '--skip-docs          Skip creation of man pages (use this when installing)'
   say '--verbose            Explain what is being done'
   say '--force              Never prompt'
   say "--completely-remove  Remove $app_name and all files under its responsibility"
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -570,6 +574,10 @@ for arg; do
     SKIPCHECKS=1
     continue
   fi
+  if [ "$arg" = '--skip-docs' ]; then
```
rodrigosiqueira, line 577: prefer `[[]]`.

rodrigosiqueira (commented):
This option is not working. I tried:

```
./setup.sh --skip-docs
Invalid number of arguments
usage: ./setup.sh option

Where option may be one of the following:
--help      | -h     Display this usage message
--install   | -i     Install kw
--uninstall | -u     Uninstall kw
--skip-checks        Skip checks (use this when packaging)
--skip-docs          Skip generating man pages (use this when installing)
--verbose            Explain what is being done
--force              Never prompt
--completely-remove  Remove kw and all files under its responsibility
--docs               Build kw's documentation as HTML pages into ./build
--enable-tracing     Install kw with tracing enabled (use it with --install)
```

uwla:
> This option is not working. I tried:
> 
> ```
> ./setup.sh --skip-docs
> Invalid number of arguments
> usage: ./setup.sh option
> 
> Where option may be one of the following:
> --help      | -h     Display this usage message
> --install   | -i     Install kw
> --uninstall | -u     Uninstall kw
> --skip-checks        Skip checks (use this when packaging)
> --skip-docs          Skip generating man pages (use this when installing)
> --verbose            Explain what is being done
> --force              Never prompt
> --completely-remove  Remove kw and all files under its responsibility
> --docs               Build kw's documentation as HTML pages into ./build
> --enable-tracing     Install kw with tracing enabled (use it with --install)
> ```

It will only work you add the flag `-i|--install` along with it.

Like one of the following:

```
$ ./setup.sh -i --skip-docs
$ ./setup.sh --install --skip-docs
$ ./setup.sh -i --skip-docs --force
$ ./setup.sh --install --skip-docs --force
```

The flag `--skip-docs` means "don't generate manpages when installing". Therefore, if you are not providing flags `-i` or `--install`, it won't work because it is not installing. Do you agree with this behavior?

setup.sh
```diff
@@ -224,6 +225,7 @@ function usage()
   say "--install   | -i     Install $app_name"
   say "--uninstall | -u     Uninstall $app_name"
   say '--skip-checks        Skip checks (use this when packaging)'
+  say '--skip-docs          Skip generating man pages (use this when installing)'
```
davidbtadokoro, line 228: Maybe 'Skip generation of man pages' instead.

davidbtadokoro (changes requested):
Hey @uwla, and thanks for this change.

Indeed, there are some installations, specially for development and reviewing, where the man pages are not needed and, because they account for the majority of the installation time, for me it makes sense to have this option :smile:.

Overall LGTM. Tested `--install --skip-docs`, `-i --skip-docs`, `--skip-docs --install`, `--skip-docs -i`, and `--install`, and they work as expected (the first four are the same and skip the docs, while the last installs kw just like always).

Left a simple comment about style, so please  address it and the one left by @rodrigosiqueira .

davidbtadokoro (approved):
Hey @uwla, thanks for addressing our comments. Tested again and it look legit.
@rodrigosiqueira, I'm going to wait for your greenlight on this, as this PR adds new behavior and you also reviewed it.

rodrigosiqueira:
@davidbtadokoro you can go ahead and merge it.

davidbtadokoro:
Thanks for the attention guys! Merged into the unstable branch 👍

Issue #700, opened 2022-09-23
Mail in the kw vars looks weird


rodrigosiqueira:
**Description:**

When I use kw vars, I see:

```
kw Mail configuration variables:
  Local Mail config file: Yes 
  kw mail options: 
    Options to be used when sending a patch (send_opts): --annotate --cover-letter --no-chain-reply-to --thread 
  SSH options: 
    SSH user (ssh_user): root 
    SSH port (ssh_port): 22 
    SSH address (ssh_ip): localhost
```

Notice the ssh option in the mail.

    1. cd to a folder with .kw
    2. kw vars

**Expected behavior:**
Only show kw mail options and also show remote, something like:

```
Local Mail config file: Yes 
  kw mail options: 
    Options to be used when sending a patch (send_opts): --annotate --cover-letter --no-chain-reply-to --thread 
kw remote options:
  Default ssh: 
    SSH user (ssh_user): root 
    SSH port (ssh_port): 22 
    SSH address (ssh_ip): localhost
```

**System information:**
 - OS: Debian Sid
 - kw Version: unstable

fernandohjunqueira:
Me and @GuilhermeMoreno started working on a fix for this issue.

rodrigosiqueira:
@fernandohjunqueira let's drop the `vars` feature since `kw config` is the preferred way of checking variables.

PR #981, opened 2023-12-01
src: lib: kw_config_loader: Fix remote kw vars

fernandohjunqueira:
Previously, the remote vars were not being displayed correctly.

In this context, fix the output of remote options inside the kw vars command.

Also fix the remote.config file accordingly.

Closes https://github.com/kworkflow/kworkflow/issues/700

### DIFF
--- a/etc/remote.config
+++ b/etc/remote.config
@@ -1,5 +1,4 @@
 #kw-default=origin
-Host origin
-  Hostname localhost
-  Port 22
-  User root
+hostname=localhost
+remote_port=22
+remote_user=root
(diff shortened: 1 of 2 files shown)


fernandohjunqueira:
Our solution is incorrect. We've realized when it failed the tests.

It is caused by the change of the remote.config file format: it does not support the cases where there are multiple options for SSH.

We'll fix it soon.

@davidbtadokoro or @rodrigosiqueira, do you have any ideas on how to solve it? How can we read the fields from this format?

rodrigosiqueira:
@fernandohjunqueira tbh, I think `kw vars` does not have any value anymore... I believe the `kw config` fully replaces `vars`, I actually recommend you create a new PR that removes `vars` from `kw`.

rodrigosiqueira:
Based on my previous comment, I'll close this PR.

Issue #901, opened 2023-08-20
`kw patch-hub`: Add pagination for 'Registered/Unregistered Mailing Lists' screen


davidbtadokoro:
**Description:**

The 'Registered/Unregistered Mailing Lists' screen has some missing mailing lists, due to the pagination of lore.kernel.org response. We should add this pagination factor to display all the available mailing lists.

**Type:**

Enhancement.

**Why:**

Although busier lists seem to always be present, we should display all available lore lists.

jppaulo06:
@Andregnl and I will tackle this issue

PR #982, opened 2023-12-04
src: lib: lore: Add pagination support to mailing lists

jppaulo06:
The 'Registered/Unregistered Mailing Lists' screen has some missing mailing lists, due to the pagination of lore.kernel.org response.

For example, the mailing list 'lvm-devel' doesn't appear in patch-hub because it is in the second page of lore.kernel.org.

So, add pagination factor to display all the available mailing lists.

Issue: #901


src/lib/lore.sh
```diff
@@ -98,23 +98,36 @@ function retrieve_available_mailing_lists()
   local flag="$1"
```
davidbtadokoro, line 98: Revise this function documentation to briefly explain the pagination and why it is done.

davidbtadokoro (changes requested):
Hi @jppaulo06 and @Andregnl! Thanks a lot for this change. Don't know why, but the fact that some MLs were missing really bothers me 🤣

Left some comments, see if you can address and don't hesitate to ping me. In any case, the code looks legit overall, and the behavior is what we were expecting 👍

davidbtadokoro:
Hi @jppaulo06 and @Andregnl, thanks for the fast update. Everything LGTM, so merged into the unstable 👍 
Thank you guys for the effort, this is a great change!

PR #987, opened 2023-12-08
src: help.sh: show correct version on repo mode

### LINKED ISSUE #986: kw --version mismatch if running in repo mode
**Description:**

When running on repo mode, KW does not show up the correct version because it was simply running `cat` on a file. This works in normal installations because the `kw` installed version is copied to a static file. But when running KW in repomode such information is dynamic, not static.

**How to Reproduce:**

1. Install `kw` on some branch, `unstable` for example:

```
$ git checkout unstable
$ ./setup.sh --install
```

2. Checkout to another branch whose last commit optionally differs, `foo` for example:

```
$ git switch -c foo
$ echo 'temporary silly branch' > a.txt
$ git add a.txt
$ git commit -m "add a.txt" 
```

3. Create a symlink to run `kw` in `repo-mode`:

```
$ ln -s ~/.local/bin/kw ~/.loca/bin/kwdev
```

4. Run `kw` and `kwdev`:

```
$ kw --version          
alpha
Branch: unstable
Commit: 77d6333
$ kdev --version
[INFO]: Running kw using repository executable.

KW environment variables are set to the following:

	KW_CACHE_DIR=<CACHE_DIR>
	KW_DATA_DIR=<DATA_DIR>
        ...
	KW_SOUND_DIR=<SOUND_DIR>

beta
Branch: master
Commit: 1d933ba
```

`repo mode` should show I'm on branch `foo`, not on branch `master`.

It should show this:

```
$ kdev --version
[INFO]: Running kw using repository executable.

KW environment variables are set to the following:

	KW_CACHE_DIR=<CACHE_DIR>
	KW_DATA_DIR=<DATA_DIR>
        ...
	KW_SOUND_DIR=<SOUND_DIR>

beta
Branch: master
Commit: 1d933ba
```

5. Install `kw` again on branch `foo` (our example) to see change in behavior

```
$ ./setup.sh -i
$ kw --version
beta
Branch: master
Commit: 1d933ba
```

But `kwdev --version` still shows the same.

**Expected behavior:**

`kw --version` should work in `repo-mode` the same way it works for normal installation;

**System information:**

 - OS: ArchLinux, Artix, Debian, Fedora (I tested it in these distros using containers)
 - kw Version: unstable and alpha
 - Bash Version: Any

**Additional context:**

Bug detected when trying to optimize integration tests by running kw in repo-mode.

uwla:
When running on repo-mode, KW  did  not  show  up  the  correct  version because it was simply running `cat` on a  file.  This  works  in  normal installations because the `kw` installed version is copied to  a  static file during the installation. In repo-mode such information is  dynamic, not static. In this case, it is necessary to run other commands.

Closes #986

### DIFF
--- a/kw
+++ b/kw
@@ -91,7 +91,12 @@ if [[ "${KW_REPO_MODE}" == 'y' ]]; then
   repo_mode_msg+=$'\t'"KW_PLUGINS_DIR=${KW_PLUGINS_DIR}"$'\n'
   repo_mode_msg+=$'\t'"KW_SOUND_DIR=${KW_SOUND_DIR}"$'\n'$'\n'
   repo_mode_msg+="${SEPARATOR}"
-  say "${repo_mode_msg}"
+
+  # Message is redirected to stderr such that it can be easily ignored if needed.
+  # Otherwise, it could mess up the output expected by  integration  tests  which
+  # are runnning kw in repo-mode and the tests would fail (we may eventually want
+  # such tests because in repo-mode we can execute kw without installing it).
+  say "${repo_mode_msg}" >&2
 fi
 
 function kw()
--- a/setup.sh
+++ b/setup.sh
@@ -4,6 +4,7 @@ KW_LIB_DIR='src'
 include "${KW_LIB_DIR}/lib/kwio.sh"
 include "${KW_LIB_DIR}/lib/kwlib.sh"
 include "${KW_LIB_DIR}/lib/kw_db.sh"
+include "${KW_LIB_DIR}/help.sh"
 
 SILENT=1
 VERBOSE=0
@@ -487,19 +488,7 @@ function safe_remove()
 
 function update_version()
 {
-  local head_hash
-  local branch_name
-  local base_version
-
-  head_hash=$(git rev-parse --short HEAD)
-  branch_name=$(git rev-parse --short --abbrev-ref HEAD)
-  base_version=$(head -n 1 "$libdir/VERSION")
-
-  cat > "$libdir/VERSION" << EOF
-$base_version
-Branch: $branch_name
-Commit: $head_hash
-EOF
+  kworkflow_version_from_repo > "${libdir}/VERSION"
 }
 
 function install_home()
--- a/src/help.sh
+++ b/src/help.sh
@@ -68,9 +68,30 @@ function kworkflow_man()
   exit 2 # ENOENT
 }
 
+function kworkflow_version_from_repo()
+{
+  local head_hash
+  local branch_name
+  local base_version
+  local git_dir
+
+  # get version info from the git repo
+  git_dir=$(realpath "${KW_LIB_DIR}/../.git")
+  head_hash=$(git -C "${git_dir}" rev-parse --short HEAD)
+  branch_name=$(git -C "${git_dir}" rev-parse --short --abbrev-ref HEAD)
+  base_version=$(head --lines 1 "${KW_LIB_DIR}/VERSION")
+
+  printf '%s\nBranch: %s\nCommit: %s\n' "${base_version}" "${branch_name}" "${head_hash}"
+}
+
 function kworkflow_version()
 {
-  local version_path="$KW_LIB_DIR/VERSION"
+  local version_path="${KW_LIB_DIR}/VERSION"
 
-  cat "$version_path"
+  if [[ "${KW_REPO_MODE}" == 'y' ]]; then
+    kworkflow_version_from_repo
+    return
+  fi
+
+  printf '%s' "$(< "$version_path")"
 }
--- a/tests/unit/help_test.sh
+++ b/tests/unit/help_test.sh
@@ -32,6 +32,25 @@ function test_kworkflow_man()
   assertEquals "($LINENO) We expected an error message." "$expect" "$output"
 }
 
+function test_kworkflow_version()
+{
+  local KW_LIB_DIR
+  local output
+  local expected_output
+
+  # the mocked version for KW
+  expected_output=$(printf 'beta\nBranch: bazz\nCommit: ffddee\n')
+
+  # this value mocks a fake location for the version file, which is the file used
+  # by KW to get information about the current version.
+  KW_LIB_DIR="${SHUNIT_TMPDIR}"
+  printf '%s' "${expected_output}" > "${KW_LIB_DIR}/VERSION"
+
+  # check output
+  output=$(kworkflow_version)
+  assert_equals_helper "Got wrong kw version." "$LINE" "${expected_output}" "${output}"
+}
+
 function test_kworkflow_version_in_repomode()
 {
   local branch_name
--- a/tests/unit/help_test.sh
+++ b/tests/unit/help_test.sh
@@ -32,4 +32,57 @@ function test_kworkflow_man()
   assertEquals "($LINENO) We expected an error message." "$expect" "$output"
 }
 
+function test_kworkflow_version_in_repomode()
+{
+  local branch_name
+  local head_hash
+  local version_name
+  local output
+  local original_path
+  local path_to_git_repository
+  local KW_REPO_MODE
+  local KW_LIB_DIR
+
+  # Initialize variables.
+  original_path="$PWD"
+  version_name='omega'
+  branch_name='foo'
+  path_to_git_repository="${SHUNIT_TMPDIR}/git_repo"
+  KW_LIB_DIR="${path_to_git_repository}/src"
+
+  # shellcheck disable=SC2034
+  KW_REPO_MODE='y' # used to trigger repo-mode behavior
+
+  # create the git repository
+  mkdir --parents "${path_to_git_repository}" "${KW_LIB_DIR}"
+
+  # cd into the repo, so we can run git commands there.
+  # This is necessary because our helper `mk_fake_git` assumes the current
+  # working directory is the directory in which we want to run git commands.
+  cd "${path_to_git_repository}" || fail 'Failed to switch to git dir.'
+
+  # print kw's version name
+  printf '%s' "${version_name}" > "${KW_LIB_DIR}/VERSION"
+
+  # Make fake git commits from the created branch.
+  mk_fake_git
+
+  # Create and switch to a new branch
+  git switch --quiet --create "${branch_name}"
+
+  # Get commit hash.
+  head_hash=$(git rev-parse --short HEAD)
+
+  # Get kw version.
+  output=$(kworkflow_version)
+
+  # Compare results.
+  assert_substring_match "Wrong version name" "(${LINENO})" "${version_name}" "${output}"
+  assert_substring_match "Wrong commit" "(${LINENO})" "Commit: ${head_hash}" "${output}"
+  assert_substring_match "Wrong branch" "(${LINENO})" "Branch: ${branch_name}" "${output}"
+
+  # Go back to previous dir.
+  cd "$original_path" || fail 'Failed to switch back to original directory.'
+}
+
 invoke_shunit


davidbtadokoro (changes requested):
Hi @uwla and @jooag. Thanks for another contribution!

Left three comments, so please see if you can address them. Also, add a unit test for `kworkflow_version` in `tests/unit/help_test.sh`. You may need to mock a git repository to test this, so maybe [these two functions on kwlib tests](https://github.com/kworkflow/kworkflow/blob/77d6333a758109e5f6df89a4fb4838a316da0175/tests/unit/lib/kwlib_test.sh#L81) can serve as reference.

uwla:
@davidbtadokoro can you please review it again?

src/help.sh
```diff
@@ -70,7 +70,20 @@ function kworkflow_man()
...
+  local base_version
+  local version_path="${KW_LIB_DIR}/VERSION"
 
-  cat "$version_path"
+  if [[ "${KW_REPO_MODE}" == 'y' ]]; then
+    head_hash=$(git rev-parse --short HEAD)
+    branch_name=$(git rev-parse --short --abbrev-ref HEAD)
+    base_version=$(head -n 1 "${KW_LIB_DIR}/VERSION")
+    printf '%s\nBranch: %s\nCommit: %s\n' "$base_version" "$branch_name" "$head_hash"
+  else
```
rodrigosiqueira, line 83: Add a return inside the `if` part, and drop the `else`.

src/help.sh
```diff
@@ -70,7 +70,20 @@ function kworkflow_man()
...
+  local head_hash
+  local branch_name
+  local base_version
+  local version_path="${KW_LIB_DIR}/VERSION"
 
-  cat "$version_path"
+  if [[ "${KW_REPO_MODE}" == 'y' ]]; then
+    head_hash=$(git rev-parse --short HEAD)
+    branch_name=$(git rev-parse --short --abbrev-ref HEAD)
+    base_version=$(head -n 1 "${KW_LIB_DIR}/VERSION")
```
rodrigosiqueira, line 81: https://kworkflow.org/content/codingstyle.html#avoid-short-options

src/help.sh
```diff
@@ -70,7 +70,20 @@ function kworkflow_man()
...
+    printf '%s\nBranch: %s\nCommit: %s\n' "$base_version" "$branch_name" "$head_hash"
+  else
+    # we use < instead of cat as it is more much performant and also recommended
+    # by the Bash manual:
+    # https://www.gnu.org/software/bash/manual/html_node/Command-Substitution.html
```
rodrigosiqueira, line 86: You don't need this comment; we already have it as part of our code style.

tests/unit/help_test.sh
```diff
@@ -3,6 +3,51 @@
...
+# Execute a git command in our temporary git directory.
+function git_cmd()
+{
+  git -C "${PATH_TO_GIT_REPOSITORY}" "$@"
+}
+
+# We may want supress git's output so it doesn't show up when running tests.
+function git_cmd_silent()
+{
+  git_cmd "$@" > /dev/null 2>&1
```
rodrigosiqueira, line 31: Git already has the `--quiet` option, which is better than using redirect to `/dev/null`.

tests/unit/help_test.sh
```diff
@@ -3,6 +3,51 @@
...
+}
+
+# Execute a git command in our temporary git directory.
+function git_cmd()
+{
+  git -C "${PATH_TO_GIT_REPOSITORY}" "$@"
+}
+
+# We may want supress git's output so it doesn't show up when running tests.
+function git_cmd_silent()
```
rodrigosiqueira, line 29: I think there is no raison d'être for this silent function... You can make it as a parameter in the `git_cmd` and add or not the `--quiet` argument.

tests/unit/help_test.sh
```diff
@@ -3,6 +3,51 @@
...
+{
+  git_cmd "$@" > /dev/null 2>&1
+}
+
+# Copy KW to our temporary git repository and setup git credentials.
+function setupGitRepository()
+{
+  cp -r "${PATH_TO_KW_REPOSITORY}" "${PATH_TO_GIT_REPOSITORY}"
```
rodrigosiqueira, line 37: https://kworkflow.org/content/codingstyle.html#avoid-short-options

tests/unit/help_test.sh
```diff
@@ -3,6 +3,51 @@
...
+  git_cmd_silent config user.name kw
+  git_cmd_silent config user.email kw@kw
```
rodrigosiqueira, line 39: Maybe add the `--local` to avoid default behavior.

rodrigosiqueira: btw, we have `mk_fake_git` in the `utils.sh`, can we leverage it here instead of creating all of these things?

tests/unit/help_test.sh
```diff
@@ -32,4 +77,40 @@ function test_kworkflow_man()
...
+# it does not require installation. This is what this test does.
+function test_kworkflow_version_in_repomode()
+{
+  local branch_name='foo'
+  local head_hash
+  local output
+
+  # Create a new branch (but delete it first if it exists).
+  git_cmd_silent branch -D "${branch_name}"
+  git_cmd_silent switch -c "${branch_name}"
```
rodrigosiqueira, line 91: https://kworkflow.org/content/codingstyle.html#avoid-short-options

tests/unit/help_test.sh
```diff
@@ -32,4 +77,40 @@ function test_kworkflow_man()
...
+
+  # Create a file.
+  printf 'kworkflow\n' > "${PATH_TO_GIT_REPOSITORY}/a.txt"
+
+  # Create a commit.
+  git_cmd_silent add a.txt
+  git_cmd_silent commit -m "add a.txt"
```
rodrigosiqueira, line 98: Again, check `mk_fake_git`

rodrigosiqueira (changes requested):

tests/unit/help_test.sh
uwla: I did not know about `mk_fake_git`. I'll use it then.

uwla:
@rodrigosiqueira I have addressed the requested changes.

kw
```diff
@@ -91,7 +91,11 @@ if [[ "${KW_REPO_MODE}" == 'y' ]]; then
   repo_mode_msg+=$'\t'"KW_PLUGINS_DIR=${KW_PLUGINS_DIR}"$'\n'
   repo_mode_msg+=$'\t'"KW_SOUND_DIR=${KW_SOUND_DIR}"$'\n'$'\n'
   repo_mode_msg+="${SEPARATOR}"
-  say "${repo_mode_msg}"
+
+  # Message is redirected to stderr such that it can be ignored if needed.
+  # Otherwise, it could mess up the output expected by tests which are runnning
+  # kw in repo-mode and the tests would fail.
+  say "${repo_mode_msg}" >&2
```
davidbtadokoro, line 98: From what I suspect this is an issue just for the integration tests, right? If so, shouldn't this come with the integration test in case, or, at least stress it in the comment that this is for the integration tests specifically?

src/help.sh
```diff
@@ -70,7 +70,18 @@ function kworkflow_man()
...
 function kworkflow_version()
 {
-  local version_path="$KW_LIB_DIR/VERSION"
+  local head_hash
+  local branch_name
+  local base_version
+  local version_path="${KW_LIB_DIR}/VERSION"
 
-  cat "$version_path"
+  if [[ "${KW_REPO_MODE}" == 'y' ]]; then
```
davidbtadokoro, line 78: Just as a curiosity, when you are only expanding a variable (not concatenating it with other literals/expansions) you do not need to use the curly braces, i.e., you can use `"$KW_REPO_MODE"`.

src/help.sh
```diff
@@ -70,7 +70,18 @@ function kworkflow_man()
...
+    head_hash=$(git rev-parse --short HEAD)
+    branch_name=$(git rev-parse --short --abbrev-ref HEAD)
+    base_version=$(head --lines 1 "${KW_LIB_DIR}/VERSION")
```
davidbtadokoro, line 81: Take advantage of the value from `version_path` here.

tests/unit/help_test.sh
```diff
@@ -32,4 +32,47 @@ function test_kworkflow_man()
...
+  git switch --quiet --create "${branch_name}"
+
+  # Make fake git commits from the created branch.
+  mk_fake_git
+
+  # Get commit hash.
+  head_hash=$(git rev-parse --short HEAD)
+
+  # Get kw --version using KW executable from the git repo.
+  output=$(./kw --version 2> /dev/null)
```
davidbtadokoro, line 68: By the comment describing this unit test, I see what you are trying to test here, but it seems to me that you are actually doing an integration test. As this is a unit test, I think you should be testing the behavior only of `kworkflow_version`.

The test will probably be a little easier to do as you won't need to copy the whole kw repository, run `./kw --version` directly, can drop the `2> /dev/null`, etc. and would only need to mock the value of `KW_REPO_MODE`, the file `${KW_LIB_DIR}/VERSION` and manipulate a generic git repo (it doesn't need to be a kw one, I think). You can also take the opportunity to test the default behavior of `kworkflow_version`, i.e., the one where `"$KW_REPO_MODE" != 'y'`.

Maybe this is an issue on my end as I have `pre-commit` configured, but, nevertheless, the test should not output these messages (see below).

Screenshot from 2023-12-15 13-24-53

tests/unit/help_test.sh
```diff
@@ -32,4 +32,47 @@ function test_kworkflow_man()
...
+
+  # Compare results.
+  assertEquals "$(grep Commit <<< "${output}")" "Commit: ${head_hash}"
+  assertEquals "$(grep Branch <<< "${output}")" "Branch: ${branch_name}"
```
davidbtadokoro, line 72: See the function `assert_substring_match`, it may help here.

davidbtadokoro (changes requested):
Hey @uwla and @jooag, and thanks for the update! Left some comments, so see if you guys can address them.

src/help.sh
uwla: Yes, but I like curly braces, haha.

kw
uwla: Yes. Eventually, we may want to use KW in repo-mode during integration tests. For example, we may want to run KW without installing it. I'll add such comment.

tests/unit/help_test.sh
```diff
@@ -32,4 +32,70 @@ function test_kworkflow_man()
...
+  expected_output=$(printf 'beta\nBranch: bazz\nCommit: ffddee\n')
+
+  # this value mocks a fake location for the version file, which is the file used
+  # by KW to get information about the current version.
+  KW_LIB_DIR="${SHUNIT_TMPDIR}"
+  printf '%s' "${expected_output}" > "${KW_LIB_DIR}/VERSION"
+
+  # check output
+  output=$(kworkflow_version)
+  assertEquals "($LINENO) Got wrong kw version." "$expected_output" "$output"
```
rodrigosiqueira, line 51: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/help.sh
```diff
@@ -70,7 +70,29 @@ function kworkflow_man()
...
+  local version_path="${KW_LIB_DIR}/VERSION"
+
+  if [[ "${KW_REPO_MODE}" == 'y' ]]; then
+    # in order to run the git commands, we need to switch to the parent  directory
+    # of $KW_LIB_DIR because that is where the git repo is at.
+    cd "${KW_LIB_DIR}/.." || (complain 'Failed to find KW root directory' && return 2)
+
+    # get version info from the git repo
+    head_hash=$(git rev-parse --short HEAD)
+    branch_name=$(git rev-parse --short --abbrev-ref HEAD)
```
rodrigosiqueira, line 85: Instead of `cd`, why not use the `-C` option on the git commands?

src/help.sh
```diff
@@ -70,7 +70,29 @@ function kworkflow_man()
 
 function kworkflow_version()
 {
-  local version_path="$KW_LIB_DIR/VERSION"
+  local head_hash
+  local branch_name
+  local base_version
+  local version_path="${KW_LIB_DIR}/VERSION"
+
+  if [[ "${KW_REPO_MODE}" == 'y' ]]; then
```
rodrigosiqueira, line 78: Maybe all the implementation under this `if`, should be in a separate function.

tests/unit/help_test.sh
```diff
@@ -32,4 +32,70 @@ function test_kworkflow_man()
...
+  version='omega'
+  branch_name='foo'
+  path_to_git_repository="${SHUNIT_TMPDIR}/git_repo"
+  KW_LIB_DIR="${path_to_git_repository}/src"
+
+  # shellcheck disable=SC2034
+  KW_REPO_MODE='y' # used to trigger repo-mode behavior
+
+  # cd into the repo, so we can run git commands there.
+  mkdir -p "${path_to_git_repository}" "${KW_LIB_DIR}"
```
rodrigosiqueira, line 74: https://kworkflow.org/content/codingstyle.html#avoid-short-options

tests/unit/help_test.sh
```diff
@@ -32,4 +32,70 @@ function test_kworkflow_man()
...
+  git switch --quiet --create "${branch_name}"
+
+  # Get commit hash.
+  head_hash=$(git rev-parse --short HEAD)
+
+  # Get kw version.
+  output=$(kworkflow_version)
+
+  # Compare results.
+  assert_substring_match "Wrong version" "($LINENO)" "${version}" "$output"
```
rodrigosiqueira, line 93: https://kworkflow.org/content/codingstyle.html#string-concatenation

rodrigosiqueira (changes requested):

src/help.sh
uwla: Because the `mk_fake_git` helper in `unit/utilsh` assumes the current directory is a git directory, and runs commands in it, such as `git init`.

uwla:
@rodrigosiqueira I have addressed the requested changes.

I would also like to point out that I'm waiting for this PR to be merged onto unstable branch such that I can add an integration test for kw version in repo mode. Currently, PR #953 has an integration test for kw version, but not for kw version in repo mode.

setup.sh
```diff
@@ -4,6 +4,7 @@ KW_LIB_DIR='src'
 include "${KW_LIB_DIR}/lib/kwio.sh"
 include "${KW_LIB_DIR}/lib/kwlib.sh"
 include "${KW_LIB_DIR}/lib/kw_db.sh"
+include "${KW_LIB_DIR}/lib/help.sh"
```
davidbtadokoro, line 7: The correct path is `${KW_LIB_DIR}/help.sh` instead of `${KW_LIB_DIR}/lib/help.sh`. This causes the file to not be included, which breaks `kw --version` for the installed case (see screenshot below).

Screenshot from 2023-12-23 17-19-11

src/help.sh
```diff
@@ -68,9 +68,30 @@ function kworkflow_man()
...
-  local version_path="$KW_LIB_DIR/VERSION"
+  local version_path="${KW_LIB_DIR}/VERSION"
+
+  if [[ "${KW_REPO_MODE}" == 'y' ]]; then
+    kworkflow_version_from_repo
+    return
+  fi
 
-  cat "$version_path"
+  cat "${version_path}"
```
davidbtadokoro, line 96: Just to keep the code style and avoid using `cat` explicitly, we can make this line as:

```bash
printf '%s' "$(< "$version_path")"
```

As a curiosity, you do not need the curly braces `{}` when the string and the expanded variable are one in the same.

tests/unit/help_test.sh
```diff
@@ -32,4 +32,74 @@ function test_kworkflow_man()
...
+  # shellcheck disable=SC2034
+  KW_REPO_MODE='y' # used to trigger repo-mode behavior
+
+  # create the git repository
+  mkdir --parents "${path_to_git_repository}" "${KW_LIB_DIR}"
+
+  # cd into the repo, so we can run git commands there.
+  # This is necessary because our helper `mk_fake_git` assumes the current
+  # working directory is the directory in which we want to run git commands.
+  cd "${path_to_git_repository}" || fail 'Failed to switch to git dir.'
```
davidbtadokoro, line 79: Before this line, I think it is cleaner to store the original path as `original_path="$PWD"` (don't forget to declare it as local) and use this value to go back to the starting dir at the end of the test.

tests/unit/help_test.sh
```diff
@@ -32,4 +32,74 @@ function test_kworkflow_man()
...
+  # Get kw version.
+  output=$(kworkflow_version)
+
+  # Compare results.
+  assert_substring_match "Wrong version name" "(${LINENO})" "${version_name}" "${output}"
+  assert_substring_match "Wrong commit" "(${LINENO})" "Commit: ${head_hash}" "${output}"
+  assert_substring_match "Wrong branch" "(${LINENO})" "Branch: ${branch_name}" "${output}"
+
+  # Go back to previous dir.
+  cd - > /dev/null || fail 'Failed to switch back to original directory.'
```
davidbtadokoro, line 102: Use the value of `original_path` here instead of `-`. I know it is the same thing, but lets keep the code (even test code) as readable as possible :grimacing:

tests/unit/help_test.sh
```diff
@@ -32,4 +32,74 @@ function test_kworkflow_man()
...
+  expected_output=$(printf 'beta\nBranch: bazz\nCommit: ffddee\n')
+
+  # this value mocks a fake location for the version file, which is the file used
+  # by KW to get information about the current version.
+  KW_LIB_DIR="${SHUNIT_TMPDIR}"
+  printf '%s' "${expected_output}" > "${KW_LIB_DIR}/VERSION"
+
+  # check output
+  output=$(kworkflow_version)
+  assertEquals "(${LINENO}) Got wrong kw version." "${expected_output}" "${output}"
```
davidbtadokoro, line 51: Prefer the use of `assert_equals_helper` over `assertEquals` when comparing complete strings, as the failing output is much better for the latter. As an illustration, the first screenshot below is the output of this assertion using `assertEquals`, and the second is the output using `assert_equals_helper`.

Screenshot from 2023-12-23 17-31-31
Screenshot from 2023-12-23 17-32-26

davidbtadokoro (changes requested):
Hey @uwla and @jooag, thanks for updating this PR. Left some comments that are simple to address (we are almost at the perfect point).

Also, I'm not sure if the second commit should be separated from the first (at least part of it). IMHO, the test `test_kworkflow_version_in_repomode` should be included in the first commit as it tests behavior introduced by the first commit. Don't have a solid opinion about the test `test_kworkflow_version`, though, but I think squashing this PR into a single commit (the first) is better than having a commit for implementation and a commit for the tests of the implementation. @rodrigosiqueira, what are your thoughts on this?

rodrigosiqueira:
Hi @uwla do you have plans to keep working on this PR?

uwla:
Hi @rodrigosiqueira. I have addressed the latest requests from David.
@davidbtadokoro, can you review it again?

src/help.sh
```diff
@@ -68,9 +68,30 @@ function kworkflow_man()
   exit 2 # ENOENT
 }
 
+function kworkflow_version_from_repo()
```
rodrigosiqueira, line 71: I added documentation to this function.

src/help.sh
```diff
@@ -68,9 +68,30 @@ function kworkflow_man()
...
+  # get version info from the git repo
+  git_dir=$(realpath "${KW_LIB_DIR}/../.git")
+  head_hash=$(git -C "${git_dir}" rev-parse --short HEAD)
+  branch_name=$(git -C "${git_dir}" rev-parse --short --abbrev-ref HEAD)
+  base_version=$(head --lines 1 "${KW_LIB_DIR}/VERSION")
+
+  printf '%s\nBranch: %s\nCommit: %s\n' "${base_version}" "${branch_name}" "${head_hash}"
+}
+
 function kworkflow_version()
```
rodrigosiqueira, line 87: I added documentation to this function.

src/help.sh
```diff
@@ -68,9 +68,30 @@ function kworkflow_man()
...
-  local version_path="$KW_LIB_DIR/VERSION"
+  local version_path="${KW_LIB_DIR}/VERSION"
+
+  if [[ "${KW_REPO_MODE}" == 'y' ]]; then
+    kworkflow_version_from_repo
+    return
+  fi
 
-  cat "$version_path"
+  printf '%s' "$(< "$version_path")"
```
rodrigosiqueira, line 96: I added `\n` after `%s`.

tests/unit/help_test.sh
```diff
@@ -32,4 +32,57 @@ function test_kworkflow_man()
...
+  version_name='omega'
+  branch_name='foo'
+  path_to_git_repository="${SHUNIT_TMPDIR}/git_repo"
+  KW_LIB_DIR="${path_to_git_repository}/src"
+
+  # shellcheck disable=SC2034
+  KW_REPO_MODE='y' # used to trigger repo-mode behavior
+
+  # create the git repository
+  mkdir --parents "${path_to_git_repository}" "${KW_LIB_DIR}"
```
rodrigosiqueira, line 76: You don't need these two paths, just `KW_LIB_DIR`; I dropped `path_to_git_repository.`

tests/unit/help_test.sh
```diff
@@ -32,4 +32,57 @@ function test_kworkflow_man()
...
+  # Get commit hash.
+  head_hash=$(git rev-parse --short HEAD)
+
+  # Get kw version.
+  output=$(kworkflow_version)
+
+  # Compare results.
+  assert_substring_match "Wrong version name" "(${LINENO})" "${version_name}" "${output}"
+  assert_substring_match "Wrong commit" "(${LINENO})" "Commit: ${head_hash}" "${output}"
+  assert_substring_match "Wrong branch" "(${LINENO})" "Branch: ${branch_name}" "${output}"
```
rodrigosiqueira, line 101: I replaced those double quotes in the message with a single quote.

tests/unit/help_test.sh
```diff
@@ -32,6 +32,25 @@ function test_kworkflow_man()
...
+  expected_output=$(printf 'beta\nBranch: bazz\nCommit: ffddee\n')
+
+  # this value mocks a fake location for the version file, which is the file used
+  # by KW to get information about the current version.
+  KW_LIB_DIR="${SHUNIT_TMPDIR}"
+  printf '%s' "${expected_output}" > "${KW_LIB_DIR}/VERSION"
+
+  # check output
+  output=$(kworkflow_version)
+  assert_equals_helper "Got wrong kw version." "$LINE" "${expected_output}" "${output}"
```
rodrigosiqueira, line 51: It should be LINENO, not LINE. Also, you should use single quotes. I changed all of those things.

PR #988, opened 2023-12-08
Add option to install KW systemwide

### LINKED ISSUE #985: add option to install kw system wide
**Description:**

Add option to install kw system wide, copying files to `/usr/share`.

`kw` already performs a check for system-wide installation:

```bash
# If kw gets installed in the system instead of the user's home, the main
# binary will be available at /usr/bin. Let's use it to update the global
# paths.
KW_SYSTEM_WIDE_INSTALLATION='/usr/bin/kw'

# ...

# Set system-wide path
if [[ -f "$KW_SYSTEM_WIDE_INSTALLATION" ]]; then
  KW_LIB_DIR='/usr/share/kw'
  KW_SOUND_DIR='/usr/share/sounds/kw'
  KW_DOC_DIR='/usr/share/doc/kw/html/'
  KW_SOUND_DIR='/usr/share/sounds/kw'
  KW_ETC_DIR='/etc/kw'
else
  KW_LIB_DIR="${KW_LIB_DIR:-"${HOME}/.local/lib"}/${KWORKFLOW}"
fi
```

But `setup.sh` only installs `kw` at the user's home directory.
It does not have option to install system wide.

**Type:**

Functionality enhancement.

**Why:**

System wide installation allows several users to run `kw`.

System wide installation does not require modifying `PATH`, such as in `.bashrc`.

It can be useful when testing because some testing environments (such as docker or podman containers) do NOT have `$HOME/.local/bin` in their `$PATH` and do NOT read `.bashrc` files by default, which requires manually changing `PATH`.

uwla:
Add `--system` flag in `setup.sh` for systemwide installation.

This enhancement brings the following benefits:

1. System wide installation allows several users to run kw.
2. System wide installation does not require modifying PATH, such as in .bashrc.
3. It can be useful when testing because some testing environments (such as docker or podman containers) do NOT have $HOME/.local/bin in their $PATH and do NOT read .bashrc files by default, which requires manually changing PATH.

Closes #985

### DIFF
--- a/kw
+++ b/kw
@@ -23,8 +23,8 @@ KW_ETC_DIR="${XDG_CONFIG_HOME:-"${HOME}/.config"}/${KWORKFLOW}"
 if [[ -f "$KW_SYSTEM_WIDE_INSTALLATION" ]]; then
   KW_LIB_DIR='/usr/share/kw'
   KW_SOUND_DIR='/usr/share/sounds/kw'
-  KW_DOC_DIR='/usr/share/doc/kw/html/'
-  KW_SOUND_DIR='/usr/share/sounds/kw'
+  KW_DOC_DIR='/usr/share/doc/kw/html'
+  KW_MAN_DIR='/usr/share/man/man1'
   KW_ETC_DIR='/etc/kw'
 else
   KW_LIB_DIR="${KW_LIB_DIR:-"${HOME}/.local/lib"}/${KWORKFLOW}"
--- a/setup.sh
+++ b/setup.sh
@@ -11,6 +11,7 @@ VERBOSE=0
 FORCE=0
 SKIPCHECKS=0
 SKIPDOCS=0
+SYSTEMWIDE=0
 ENABLE_TRACING=0
 
 declare -r app_name='kw'
@@ -19,37 +20,34 @@ declare -r app_name='kw'
 ## Following are the install paths
 ##
 # Paths used during the installation process
-declare -r kwbinpath="$HOME/.local/bin/$app_name"
-declare -r binpath="$HOME/.local/bin"
-declare -r libdir="$HOME/.local/lib/$app_name"
-declare -r sharedir="${XDG_DATA_HOME:-"$HOME/.local/share"}/$app_name"
-declare -r docdir="$sharedir/doc"
-declare -r mandir="$sharedir/man"
-declare -r sounddir="$sharedir/sound"
-declare -r databasedir="$sharedir/database"
-declare -r datadir="${XDG_DATA_HOME:-"$HOME/.local/share"}/$app_name"
-declare -r etcdir="${XDG_CONFIG_HOME:-"$HOME/.config"}/$app_name"
-declare -r cachedir="${XDG_CACHE_HOME:-"$HOME/.cache/$app_name"}"
-declare -r tracingdir="${datadir}/tracing"
-declare -r dot_configs_dir="${datadir}/configs"
+declare BINDIR="${HOME}/.local/bin"
+declare LIBDIR="${HOME}/.local/lib/${app_name}"
+declare KWBINPATH="${BINDIR}/${app_name}"
+declare ETCDIR="${XDG_CONFIG_HOME:-"${HOME}/.config"}/${app_name}"
+declare SHAREDIR="${XDG_DATA_HOME:-"${HOME}/.local/share"}/${app_name}"
+declare DOCDIR="${SHAREDIR}/doc"
+declare MANDIR="${SHAREDIR}/man"
+declare SOUNDDIR="${SHAREDIR}/sound"
+declare DATABASEDIR="${SHAREDIR}/database"
+declare -r DATADIR="${XDG_DATA_HOME:-"${HOME}/.local/share"}/${app_name}"
+declare -r CACHEDIR="${XDG_CACHE_HOME:-"${HOME}/.cache/"}/${app_name}"
+declare -r TRACINGDIR="${DATADIR}/tracing"
+declare -r DOTCONFIGDIR="${DATADIR}/configs"
 
 ##
 ## Source code references
 ##
-declare -r SRCDIR='src'
-declare -r MAN='documentation/man/'
-declare -r CONFIG_DIR='etc/'
-declare -r KW_CACHE_DIR="$cachedir"
-declare -r TRACING_CODE_EXCERPTS_DIR='tracing/code_excerpts'
-
-declare -r SOUNDS='sounds'
-declare -r DATABASE='database'
 declare -r BASH_AUTOCOMPLETE='bash_autocomplete'
-declare -r DOCUMENTATION='documentation'
-
+declare -r CONFIG_DIR='etc/'
 declare -r CONFIGS_PATH='configs'
-
+declare -r DATABASE='database'
 declare -r DOCS_VIRTUAL_ENV='docs_virtual_env'
+declare -r DOCUMENTATION='documentation'
+declare -r KW_CACHE_DIR="$CACHEDIR"
+declare -r MAN='documentation/man/'
+declare -r SOUNDS='sounds'
+declare -r SRCDIR='src'
+declare -r TRACING_CODE_EXCERPTS_DIR='tracing/code_excerpts'
 
 function check_dependencies()
 {
@@ -167,9 +165,14 @@ function update_path()
 {
   local shellrc=${1:-'.bashrc'}
 
+  # there is no need to update the path if the installation is system wide
+  if [[ "${SYSTEMWIDE}" == 1 ]]; then
+    return
+  fi
+
   IFS=':' read -ra ALL_PATHS <<< "$PATH"
   for path in "${ALL_PATHS[@]}"; do
-    [[ "$path" -ef "$binpath" ]] && return
+    [[ "$path" -ef "$BINDIR" ]] && return
   done
 
   safe_append "PATH=${HOME}/.local/bin:\$PATH # kw" "${HOME}/${shellrc}"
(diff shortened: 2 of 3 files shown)


setup.sh
```diff
@@ -18,16 +19,16 @@ declare -r app_name='kw'
...
-declare -r binpath="$HOME/.local/bin"
-declare -r libdir="$HOME/.local/lib/$app_name"
-declare -r sharedir="${XDG_DATA_HOME:-"$HOME/.local/share"}/$app_name"
-declare -r docdir="$sharedir/doc"
-declare -r mandir="$sharedir/man"
-declare -r sounddir="$sharedir/sound"
-declare -r databasedir="$sharedir/database"
+declare kwbinpath="$HOME/.local/bin/$app_name"
+declare binpath="$HOME/.local/bin"
+declare libdir="$HOME/.local/lib/$app_name"
```
rodrigosiqueira, line 24: https://kworkflow.org/content/codingstyle.html#string-concatenation

kw
```diff
@@ -22,9 +22,10 @@ KW_ETC_DIR="${XDG_CONFIG_HOME:-"${HOME}/.config"}/${KWORKFLOW}"
 # Set system-wide path
 if [[ -f "$KW_SYSTEM_WIDE_INSTALLATION" ]]; then
   KW_LIB_DIR='/usr/share/kw'
+  KW_DB_DIR="/usr/share/kw/database"
   KW_SOUND_DIR='/usr/share/sounds/kw'
   KW_DOC_DIR='/usr/share/doc/kw/html/'
-  KW_SOUND_DIR='/usr/share/sounds/kw'
+  KW_MAN_DIR='/usr/share/man/en/kw'
```
rodrigosiqueira, line 28: Are you sure this path is correct?

kw
```diff
@@ -22,9 +22,10 @@ KW_ETC_DIR="${XDG_CONFIG_HOME:-"${HOME}/.config"}/${KWORKFLOW}"
 # Set system-wide path
 if [[ -f "$KW_SYSTEM_WIDE_INSTALLATION" ]]; then
   KW_LIB_DIR='/usr/share/kw'
+  KW_DB_DIR="/usr/share/kw/database"
```
rodrigosiqueira, line 25: We don't want kw database in a protected directory; it is better to be at the user's home.

setup.sh
```diff
@@ -18,16 +19,16 @@ declare -r app_name='kw'
...
 # Paths used during the installation process
-declare -r kwbinpath="$HOME/.local/bin/$app_name"
-declare -r binpath="$HOME/.local/bin"
-declare -r libdir="$HOME/.local/lib/$app_name"
-declare -r sharedir="${XDG_DATA_HOME:-"$HOME/.local/share"}/$app_name"
-declare -r docdir="$sharedir/doc"
-declare -r mandir="$sharedir/man"
-declare -r sounddir="$sharedir/sound"
-declare -r databasedir="$sharedir/database"
+declare kwbinpath="$HOME/.local/bin/$app_name"
```
rodrigosiqueira, line 22: Why remove `-r`? It looks unsafe...

setup.sh
```diff
@@ -226,6 +232,7 @@ function usage()
   say "--uninstall | -u     Uninstall $app_name"
   say '--skip-checks        Skip checks (use this when packaging)'
   say '--skip-docs          Skip creation of man pages (use this when installing)'
+  say '--system    | -s     Install kw system wide (use this when installing)'
```
rodrigosiqueira, line 235: You also need to manage the uninstall in this scenario.

rodrigosiqueira (changes requested):

kw
```diff
@@ -22,9 +22,10 @@ KW_ETC_DIR="${XDG_CONFIG_HOME:-"${HOME}/.config"}/${KWORKFLOW}"
...
   KW_SOUND_DIR='/usr/share/sounds/kw'
   KW_DOC_DIR='/usr/share/doc/kw/html/'
-  KW_SOUND_DIR='/usr/share/sounds/kw'
```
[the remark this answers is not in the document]
uwla: Look that there is already `KW_SOUND_DIR` in line 26. Therefore, line 27 was unnecessary duplicated code.

kw
uwla: More or less sure. Debian keeps man pages at `/usr/share/` under their language abbreviation, such as: jp, ch, es, pt. That is why I put it in `/usr/share/man/en/kw` instead of `/usr/share/man/kw`.

kw
uwla: Ok.

setup.sh
uwla: Because I need to change the value of the variable if it is system-wide installation. It will be at `/usr/bin/kw`. It can't be readonly, therefore.

setup.sh
uwla: Yes, we managed uninstall too. I left some comments about it in the code:

```bash
# if system wide, change the installation directories.
# This approach works both for installing and uninstalling KW.
if [[ "${SYSTEMWIDE}" == 1 ]]; then
  use_system_wide_installation_directories
fi
```

uwla:
@rodrigosiqueira I have addressed the comments and here is how we can test the PR:

Install kw in a podman container to test system-wide installation:

```
$ kw_path=/home/user/code/kworkflow # change it accordingly
$ podman run -v "${kw_path}":/kw -e PATH='/usr/bin' -w /kw --rm -it docker.io/library/archlinux bash
# ./setup.sh --install --system --force
```

The `$` prefix denotes running shell in host machine as a normal user.
The `#` prefix denotes running shell in container machine as root user.

Check the files were installed:

```
# find /usr ~/ -type f | grep /kw
...
```

Test commands:

```
# kw --version
# kw config --show mail
# kw patchhub
# kw mail --help
```

Uninstall:

```
# ./setup.sh --completely-remove --force --system
```

Check the files were uninstalled:

```
# find /usr ~/ -type f | grep /kw
```


By the way: the above commands are kind of a sketch for a `install_kw_system_wide` integration test.

kw
rodrigosiqueira: Tbh, I'm a little bit confused about the man path.

I checked this path in my Debian and Arch Linux systems, and none have the subfolder `en` (my system is set to English). I'm under the impression that the default English documentation for a command publicly available for the system should be in the subfolder `man1`. However, I'm not 100% sure... let's try to be sure about this.

setup.sh
```diff
@@ -18,16 +19,16 @@ declare -r app_name='kw'
...
-declare -r sounddir="$sharedir/sound"
-declare -r databasedir="$sharedir/database"
+declare kwbinpath="${HOME}/.local/bin/$app_name"
+declare binpath="${HOME}/.local/bin"
+declare libdir="${HOME}/.local/lib/$app_name"
+declare sharedir="${XDG_DATA_HOME:-"$HOME/.local/share"}/$app_name"
+declare docdir="${sharedir}/doc"
+declare mandir="${sharedir}/man"
+declare sounddir="${sharedir}/sound"
+declare databasedir="${sharedir}/database"
```
rodrigosiqueira, line 29: Consider making another commit just to use the upper case for all of these constant variables since it is more aligned with the overall code style.

setup.sh
```diff
@@ -166,6 +167,11 @@ function update_path()
 {
   local shellrc=${1:-'.bashrc'}
 
+  # there is no need to update the path if the installation is system wide
+  if [[ "${SYSTEMWIDE}" == 1 ]]; then
```
rodrigosiqueira, line 171: I think you can just use `[[ -n "${SYSTEMWIDE}" ]]`.

setup.sh
```diff
@@ -10,6 +10,7 @@ VERBOSE=0
 FORCE=0
 SKIPCHECKS=0
 SKIPDOCS=0
+SYSTEMWIDE=0
```
rodrigosiqueira, line 13: How about `SYSTEM_WIDE=''

setup.sh
```diff
@@ -445,7 +453,11 @@ function synchronize_files()
     mkdir --parents "$tracingdir"
   fi
 
-  say "$app_name installed into $HOME"
+  if [[ "$SYSTEMWIDE" == 1 ]]; then
```
rodrigosiqueira, line 456: How about `[[ -n "$SYSTEM_WIDE" ]]`.

setup.sh
```diff
@@ -502,15 +514,36 @@ Commit: $head_hash
...
+  # modify the default installation paths
+  libdir="${base_path}/kw"
+  docdir="${base_path}/doc/kw/html"
+  mandir="${base_path}/man/en/kw"
+  sounddir="${base_path}/sounds/kw"
+
+  # The exception is the ETC directory and the BIN path. System wide config
+  # files are usually located at /etc and not at /usr while binary files are
+  # usually at /usr/bin
+  binpath="/usr/bin"
```
rodrigosiqueira, line 531: single quotes.

setup.sh
```diff
@@ -502,15 +514,36 @@ Commit: $head_hash
...
+function install_kw()
 {
   # Check Dependencies
   if [[ "$SKIPCHECKS" == 0 ]]; then
     say 'Checking dependencies ...'
     check_dependencies
   fi
-  # Move old folder structure to new one
-  legacy_folders
+  if [[ "$SYSTEMWIDE" == 0 ]]; then
```
rodrigosiqueira, line 543: Same as my previous comments.

src/help.sh
```diff
@@ -54,13 +54,13 @@ function kworkflow_man()
     feature="kw-${feature}"
   fi
 
-  if [[ -f "$KW_SYSTEM_WIDE_INSTALLATION" ]]; then
-    cmd_manager "$flag" "man ${feature}"
+  if [[ -r "$doc/$feature.1" ]]; then
+    cmd_manager "$flag" "man -l $doc/$feature.1"
```
rodrigosiqueira, line 58: String concatenation.

rodrigosiqueira (changes requested):

setup.sh
uwla: ok

uwla:
@rodrigosiqueira , I have done some research, and I learned the following:

- Man searches for pages located in a directory defined by `MANPATH`. It usually searchs in `/usr/share/man` and `/usr/local/share/man`.
- Man pages are divided into sections (usually, from 1 to 9). If the man pages for a program has multiple sections, they would be at `${MANPATH}/man1/<programname>.1`, `${MANPATH}/man2/<programname>.2`, and so on. So, there is a suffix for the section.
- If the man page is not in English, it will be in a subdirectory denoted by the language acronym

uwla:
I closed it by mistake and I just realized it now....

uwla:
@rodrigosiqueira if you agree in putting kworkflow's manpages at /usr/share/man/man1/ then I can improve the PR very quickly.

uwla:
I pushed a commit but forgot to address one change. I'll do it later today.

uwla:
@rodrigosiqueira I think we have it now

uwla:
@rodrigosiqueira, when you have time, let's finish it.

rodrigosiqueira:
Hi @uwla ,
I'm sorry I never got to this important change. I'll revive it soon, and I'll keep your credits for future changes.

PR #989, opened 2023-12-08
src: ui: patch_hub: patchset_details_and_actions: Add bookmarking info

JGBSouza:
Bookmarking patches didn't show any information about where the patch set will be downloaded.

To solve this problem, add a checkbox message when the patchset is bookmarked or removed so the user can be sure where to find it or that it was removed. Solve the last task from #880

### DIFF
--- a/src/ui/patch_hub/patchset_details_and_actions.sh
+++ b/src/ui/patch_hub/patchset_details_and_actions.sh
@@ -122,8 +122,8 @@ function handle_download_action()
       if [[ "$ret" != 0 ]]; then
         create_message_box 'Error' 'Could not download patchset:'$'\n'"${_patchset['patchset_title']}"$'\n'"[error message] ${output}"
       else
-        message_box='Downloaded patchset:'$'\n'"- ${_patchset['patchset_title']}"$'\n'$'\n'
-        message_box+='Filepath:'$'\n'"$output"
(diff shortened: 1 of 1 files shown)


src/ui/patch_hub/patchset_details_and_actions.sh
```diff
@@ -186,4 +190,7 @@ function handle_remove_bookmark_action()
   if [[ "$?" != 0 ]]; then
     create_message_box 'Error' 'Could not unbookmark patchset'$'\n'"- ${_patchset['patchset_title']}"
   fi
+
+  message_box='Deleted patchset:'$'\n'"- ${_patchset['patchset_title']}"$'\n'$'\n'
```
davidbtadokoro, line 194: Same nitpick, maybe say 'Removed bookmark from patchset: (...)' (or something better, but stress that it is being unbookmarked).

davidbtadokoro (changes requested):
Hey @JGBSouza, and thanks a lot for your change, it is simple but it will have impact for the User Experience. Left two simple comments about style/language, so please see if you can address them. Outside of that, tested and it looks really nice!

davidbtadokoro:
Hey @JGBSouza, and thanks for the update. Made some small adaptations to your changes, but, overall, looks great to me, so merged into the unstable branch :+1:

Issue #995, opened 2023-12-14
`kw remote`: Remove dashes in --list


jppaulo06:
**Description:**

In `kw remote`, it's a bit strange that the two dashes ('--') is necessary to `kw remote --list` command. 
It's possible to run `kw remote add`, `kw remote remove`, `kw remote rename`, but not `kw remote list`. 

**Type:**

Other

**Why:**

- `kw remote` should be intuitive.
- It's faster to type `kw remote list`.

rodrigosiqueira:
Just to give a little bit of background on why we have this inconsistency.

This command was heavily inspired by `git remote`, and as a result, I tried to copy its interface. Notice that `--list` is not part or `git remote`, and for this reason, I decided to add `--list`.

In hindsight, I think I should try to be more consistent with `kw` in general by adding `--` in all of the remote commands. The price that we pay adds a little bit of extra overhead for the developer to learn this command.

@matheustavares since you worked with git, do you have any comment about the `git remote` interface?
@davidbtadokoro any thoughts?

matheustavares:
> This command was heavily inspired by `git remote`, and as a result, I tried to copy its interface. Notice that `--list` is not part or `git remote`, and for this reason, I decided to add `--list`.

Got it. One thing to notice is that `git remote` with no args will behave as "--list", while `kw remote` with no args is an user error. (That's probably not a big deal, though).

> In hindsight, I think I should try to be more consistent with `kw` in general by adding `--` in all of the remote commands. The price that we pay adds a little bit of extra overhead for the developer to learn this command.
> 
> @matheustavares since you worked with git, do you have any comment about the `git remote` interface? 

Nothing specific about git, I think... But I agree that making the remote options consistent in kw is a nice improvement. Something that occurred to me while reading the man page is that all "--" options with the exception of "--set-default" can be used with any of the remote subcomands. So I think it would make sense to either turn both "--list" and "--set-default" into subcomands, or turn all subcommands into "--" options, as you suggested.

PR #996, opened 2023-12-14
src: kw_remote.sh: Remove dashes from `kw remote --list` command

jppaulo06:
The dashes inside the command `kw remote --list` are confusing since some of the others `kw remote` options don't require them. For example, there aren't dashes in `kw remote add`, `kw remote remove` and `kw remote rename`.

So, change `kw remote --list` command to `kw remote list` to make it more clear and intuitive.

Solves: #995


rodrigosiqueira:
tbh, I prefer to go in the opposite direction, i.e., adding `--` to the other parameters.

rodrigosiqueira:
@jppaulo06 Will you keep working on this PR?

PR #999, opened 2023-12-15
src: add shellbang to all bash scripts

### LINKED ISSUE #998: Add shell bang to shellscripts at src/
**Description:**

Bash scripts at the root of kw and at the `tests` directory have the shellbang `#!/bin/bash`, but those at `src` directory do not have. This makes the source code inconsistent and causes "warnings" in IDEs that have LSP for Bash.

**Type:**

Refactor

**Why:**

1. Make code more consistent.
2. Better development experience for KW developers running IDEs such as LunarVim or Vscodium.

uwla:
The scripts at src/ were missing shellbang, while  the  scripts  at  the root of kw (kw, setup.sh, run_tests.sh) and at tests/ do have shellbang. So, we add shellbang to files at src/ to make code more consistent.

Closes #998.

A quick notice. I used the following bash script to automate adding the shellbang:

```bash
tmp_file='/tmp/a.txt'
find src/ -type f -name '*.sh' | while read fname; do
  firstline=$(head --lines 1 "${fname}")
  if [ "${firstline}" != '#!/bin/bash' ]; then
    (printf '#!/bin/bash\n' ; cat "${fname}" ) > "${tmp_file}"
    mv "${tmp_file}" "${fname}"
  fi
done
```

### DIFF
--- a/src/backup.sh
+++ b/src/backup.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # This file deals with backing up and restoring kw data files stored under the
 # the KW_DATA_DIR directory. These files are configs, pomodoro reports and
 # kw usage statistics.
--- a/src/bash_autocomplete.sh
+++ b/src/bash_autocomplete.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 function _kw_autocomplete()
 {
   declare -A kw_options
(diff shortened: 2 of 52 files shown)


rodrigosiqueira (changes requested):
Iirc, this was a decision we made many moons ago.

The code under the `src/` directory works as a library and is not meant to be executed alone. We expect that other scripts include it, and those scripts should have the bash shebang. For example, `kw` file is the script that must be executed; the same for each test.

PR #1001, opened 2023-12-20
documentation: dependencies: add device-module related dependencies

### LINKED ISSUE #1000: fresh install of KW in Debian or Fedora may fail to run kw device --local due to missing dependencies
**Description:**

KW device runs a few commands to detect hardware information, but some of those commands rely on libraries which are not listed in the dependency list.

**How to Reproduce:**

1. Run a docker or podman container using Debian or Fedora
2. Install git
3. Clone kw in the container
4. Install KW
5. Run `kw device --local` and check some errors say the commands were not found
6. Install the missing dependencies: `<pkg> install -y pciutils procps`
where `<pkg>` is `apt` for Debian and `dnf` for Fedora.
8. Run `kw device --local` again and check it does not show up the previous errors about the missing commands.

**Expected behavior:**

KW should  install any library it depends on in order to run properly.

**System information:**

 - OS: Debian or Fedora
 - kw Version: *
 - Bash Version: *

uwla:
The device module depends on `lscpi` and `ps`, which are  available  via the packages `pciutils` and `procps`. Normally these  dependencies  will already be installed by default, but we cannot assume they are,  so  add them to the dependency lists.

Closes #1000

### DIFF
--- a/documentation/dependencies/arch.dependencies
+++ b/documentation/dependencies/arch.dependencies
@@ -26,3 +26,5 @@ curl
 perl-xml-xpath
 coreutils
 b4
+procps-ng
+pciutils
--- a/documentation/dependencies/debian.dependencies
+++ b/documentation/dependencies/debian.dependencies
@@ -28,3 +28,5 @@ curl
 libxml-xpath-perl
 coreutils
 b4
+procps
+pciutils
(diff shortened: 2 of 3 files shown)


rodrigosiqueira:
Merged! Thanks

Issue #1010, opened 2024-01-11
`setup.sh` always require to install `procps` in fedora


jppaulo06:
**Description:**

Everytime I reinstall kw unstable, setup requests to install the package procps:

```
$ ./setup.sh --install
Checking dependencies ...
The following packages are required: procps
May we install them? [y/N]:
```
However, it is always already installed in my system, then nothing is done:
```
The following packages are required: procps
May we install them? [y/N]: y
[sudo] senha para jppauloi:                                                               (password for sudo use)
Fedora 39 - x86_64 - Updates
O pacote procps-ng-4.0.3-5.fc39.x86_64 já está instalado.          (The package is already installed)
Dependências resolvidas.                                                                  (Dependencies resolved)
Nada para fazer.                                                                                   (Nothing to do)
Concluído!                                                                                             (Done!)
```

**How to Reproduce:**

In a fedora system, inside a kw directory, execute `./setup.sh --install`.

**Expected behavior:**

It is expected to not received a request to install an already installed package.

**System information:**

 - OS: Linux fedora 6.6.9
 - kw Version: 
 ```
 alpha
Branch: unstable
Commit: 1764e42
```
 - Bash Version: 5.2.21

**Additional context:**

I think the problem might be with the package name.
If I do a check with `rpm -q procps`, it says that the package does not exist in my system. However, it works well if I write `rpm -q procps-ng`.
So, the simplest solution probably is to change `procps` to `procps-ng` in fedora dependencies.

PR #1011, opened 2024-01-11
documentation: dependencies: fedora: Fix procps dependency in fedora

jppaulo06:
In fedora distro, when installing kw with setup.sh, it is always requested to install the package procps, even when it's already installed.

The problem is that the command `rpm -q procps` seems to specifically not recognize that this package is already installed. However, with `rpm -q procps-ng` (which is the same exact package) it works fine.

So, change the dependency to 'procps-ng' instead of 'procps' to fix this problem.

Fixes: #1010

### DIFF
--- a/documentation/dependencies/fedora.dependencies
+++ b/documentation/dependencies/fedora.dependencies
@@ -24,5 +24,5 @@ curl
 perl-XML-XPath
 coreutils
 b4
-procps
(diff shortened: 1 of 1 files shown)


rodrigosiqueira:
Thanks a lot for your PR. Change already merged into the unstable branch.

Issue #1008, opened 2024-01-10
`kw config` does not support values with space


jppaulo06:
**Description:**

Currently, when a user attempts to add a value with spaces to a configuration using kw config — even with single or double quotes — kw will only capture the first word as the value.

**How to Reproduce:**

For example, you could try to change the pomodoro sound:

```bash
kw config notification.sound_alert_command 'paplay /home/<user>/.local/share/kw/sound/berimbau.wav'
```

But this will turn into

```notification.sound_alert_command=paplay```

**Expected behavior:**

It is expected that values with space are supported.

That is, the command
```bash
kw config notification.sound_alert_command 'paplay /home/<user>/.local/share/kw/sound/berimbau.wav'
```

should result in the configuration

```notification.sound_alert_command=paplay /home/<user>/.local/share/kw/sound/berimbau.wav```

**System information:**

 - OS: Linux fedora 6.6.9
 - kw Version:
```
beta
Branch: master
Commit: af4d2e3
```
 - Bash Version: 5.2.21

PR #1012, opened 2024-01-11
src: config: Fix values with space separation in `kw config`

jppaulo06:
When manually configuring kw with `kw config`, it does not accept configurations with space separation, even with single ou double quotes put.

For example, the command
`kw config notification.sound_alert_command 'paplay berimbau.wav'`

will result in the wrong configuration
`notification.sound_alert_command=paplay`

Because of that, change `kw config` to receive the value as all characters after the first space following configuration name, instead of just the first word.

Fixes: #1008


rodrigosiqueira:
That's a very nice fix! Thanks a lot for your contribution.

Change already merged into the unstable branch.

PR #1018, opened 2024-01-12
documentation: content: tests: add instructions for rootless containers

uwla:
Provide instructions for setting up podman to run in rootless mode (indicating the official tutorial). This will be helpful for those who are installing podman for the first time and don't know how to configure it to run rootless containers.

Also, fix a codeblock in `tests.rst` which was not  correctly  formatted and was displayed as inline instead of multiline.

If necessary, I could split the fix to another PR because it is unrelated to the main purpose of this PR. But, since it is a small fix (two lines added), I think it was reasonable to include it here to avoid an extra PR.


rodrigosiqueira:
Merged into the unstable branch, thanks!

PR #1019, opened 2024-01-12
run_tests.sh: add --clear-cache flag to clear all test-related cache

### LINKED ISSUE #1017: Container images are not removed in `setup.sh --completely-remove`
**Description:**

After completely removing kw, images from integration tests are not removed.

**How to Reproduce:**

1. With unstable kw installed, execute in kw source tree
```bash
./run_tests --integration
```
2. After it finishes, you can see that the images are installed runnning
```bash
podman image ls
```
3. Then, completely remove kw with
```bash
./setup.sh --completely-remove
```
4. Run this command again to confirm that the images are still there:
 ```bash
podman image ls
```
You should see something like
```
$ podman image ls                                                                                                                  [19:43:36]
REPOSITORY                   TAG         IMAGE ID      CREATED       SIZE
localhost/kw-fedora          latest      eef06db6345a  6 hours ago   1.16 GB
localhost/kw-debian          latest      40227dd4c41a  6 hours ago   2.39 GB
localhost/kw-archlinux       latest      e912e0d8d8dd  6 hours ago   1.24 GB
docker.io/library/debian     latest      a6916e41aa87  20 hours ago  121 MB
docker.io/library/archlinux  latest      69f38d8f6347  10 days ago   446 MB
docker.io/library/fedora     latest      8404925a71fd  6 weeks ago   183 MB
```

**Expected behavior:**

It is expected that the images are gone after completely removing kw from system.

**System information:**

 - OS: Linux fedora 6.6.9
 - kw Version: 
``` 
alpha
Branch: unstable
Commit: 1764e42
```
 - Bash Version 5.2.21

### PR DESCRIPTION
**Pull request by uwla**
Currently, this new option clears cached container images.

Fixes #1017

### DIFF
--- a/documentation/content/tests.rst
+++ b/documentation/content/tests.rst
@@ -1,5 +1,5 @@
 ===============
-  About Tests
+About Tests
 ===============
 
 .. _tests:
@@ -25,7 +25,7 @@ List all available test files::
 
 Or run individual tests with::
 
-  ./run_tests.sh test TESTFILE1 ...
+  ./run_tests.sh test TEST_FILE_1 ...
 
 To limit the scope of the tests, pass the flag `--unit` or `--integration` as
 the first argument to any of the examples above. So, the syntax is::
@@ -33,7 +33,7 @@ the first argument to any of the examples above. So, the syntax is::
   ./run_tests.sh [scope] [command] [args]
 
 Where `[scope]` can be `--unit` or `--integration`. The placeholder  `[command]`
-can be either `list`, `test` or simply omited in order to run  all  tests.  Here
+can be either `list`, `test` or simply omitted in order to run  all  tests. Here
 are some examples:
 
 .. code-block:: bash
@@ -41,12 +41,15 @@ are some examples:
   ./run_tests.sh --unit                       # run all unit tests
   ./run_tests.sh --unit list                  # list all unit tests
   ./run_tests.sh --unit test device           # test device unit test
+  ./run_tests.sh --unit clear-cache           # clear unit tests cache
   ./run_tests.sh --integration                # run all integration tests
   ./run_tests.sh --integration list           # list all integration tests
   ./run_tests.sh --integration test device    # test device integration test
+  ./run_tests.sh --integration clear-cache    # clear integration tests cache
   ./run_tests.sh                              # run all tests
   ./run_tests.sh list                         # list all tests
   ./run_tests.sh test device                  # run all device tests
+  ./run_tests.sh clear-cache                  # clear all cache
 
 The integration tests can take over 10 minutes to run in the first time  because
 podman is building the container images to be used in the tests, which  requires
@@ -56,7 +59,7 @@ seconds each time.
 
 Then, the local kw repo is copied to the containers and installed  again,  which
 takes very few seconds. For optimization purposes,  the  containers  are  reused
-accross tests. If you add a new commit or checkout to another branch, such  that
+across  tests. If you add a new commit or checkout to another branch, such  that
 HEAD points to another commit, the containers  will  be  destroyed  and  created
 again in order to install the current local version of kw.
 
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -18,6 +18,7 @@ function show_help()
     '         Limit tests to unit tests' \
     '' \
     'COMMANDS' \
+    '  clear-cache - clears tests cache' \
     '  help - displays this help message' \
     '  list - lists all test files under tests/' \
     '  test - runs the given test'
@@ -121,12 +122,23 @@ function run_tests()
   # amount of time for the integration tests.
   if [[ "${integration_tests_setup}" == 1 ]]; then
     printf '\n' # add new line after the last "OK"
-    say 'Tearing down environment used in integration tests...'
-    teardown_container_environment
+    say 'Tearing down containers used in integration tests...'
+    teardown_containers
     printf '\n'
   fi
 
   report_results "$total" "$success" "$notfound" "$fail" "$test_failure_list"
+}
+
+function clear_unit_tests_cache()
+{
+  say 'Unit tests: nothing cached to clear.'
+}
+
+function clear_integration_tests_cache()
+{
+  say 'Integration tests: cleaning cache...'
+  teardown_container_environment "$@"
 }
 
 declare -a TESTS
@@ -139,12 +151,16 @@ function set_tests()
   done
 }
 
+declare TESTS_UNIT=1
+declare TESTS_INTEGRATION=1
 declare TESTS_DIR=./tests
 if [[ "$1" == '--unit' || "$1" == '-u' ]]; then
   TESTS_DIR=./tests/unit
+  TESTS_INTEGRATION=0
   shift
 elif [[ "$1" == '--integration' || "$1" == '-i' ]]; then
   TESTS_DIR=./tests/integration
+  TESTS_UNIT=0
   shift
 fi
 
@@ -178,6 +194,10 @@ elif [[ "$1" == 'test' ]]; then
   files_list=$(find "$TESTS_DIR" | grep --perl-regexp "${regex}" | grep --extended-regexp --invert-match 'samples/.*|/shunit2/')
   set_tests $files_list
   LANGUAGE=en_US.UTF_8 run_tests
+elif [[ "$1" == 'clear-cache' ]]; then
+  shift
+  [[ "${TESTS_UNIT}" == 1 ]] && clear_unit_tests_cache "$@"
+  [[ "${TESTS_INTEGRATION}" == 1 ]] && clear_integration_tests_cache "$@"
 else
   show_help
 fi
--- a/tests/integration/utils.sh
+++ b/tests/integration/utils.sh
@@ -2,6 +2,7 @@
 
 include './src/lib/kwio.sh'
 
+declare -gr CONTAINER_BASE_IMAGE='docker.io/library'
 declare -g CONTAINER_DIR # Has the container files to build the container images
 declare -g SAMPLES_DIR   # Has sample files used accross the integration tests
 declare -g KWROOT_DIR    # Local kw dir to be copied to and installed in the containers
@@ -64,14 +65,14 @@ function setup_container_environment()
 
     # Only build the image if it does not exist. That's because trying to build
     # the podman image takes a second or two even if it exists and is cached.
-    podman image exists "${container_img}"
+    image_exists "${container_img}"
     if [[ "$?" -ne 0 ]]; then
       current_step=$((current_step + 1))
       say "[${current_step}/${total_steps}] Building container image for ${distro}. This might take a while..."
 
       build_distro_image "$distro"
       if [[ "$?" -ne 0 ]]; then
-        complain "Failed to setup container environment for distro ${distro}"
+        complain "Failed to setup container for distro ${distro}"
 
         current_step=$((current_step + 1))
         say "[${current_step}/${total_steps}] Skip creating ${distro} container."
@@ -129,7 +130,7 @@ function setup_container_environment()
 }
 
 # Destroy all containers used in the tests
-function teardown_container_environment()
+function teardown_containers()
 {
   local distro
   local i=0
@@ -144,7 +145,7 @@ function teardown_container_environment()
 
 # Destroy a single container
 #
-# @container: Name or ID of the container
+# @container    Name or ID of the container
 function teardown_single_container()
 {
   local container="$1"
@@ -157,6 +158,177 @@ function teardown_single_container()
   fi
 }
 
+# Completely remove the container environment: the containers and the images
+#
+# @flag   Optional. Currently, only accepts '-f' or '--force' to force removal of cached images.
+function teardown_container_environment()
+{
+  local flag="$1"
+  local container_name
+  local distro
+  local img
+  local img_layer_list
+  local dangling_img
+  local dangling_img_layer_list
+
+  for distro in "${DISTROS[@]}"; do
+    container_name="kw-${distro}"
+    podman container exists "${container_name}"
+    if [[ "$?" -eq 0 ]]; then
+      say "Removing podman container kw-${distro}."
+      teardown_single_container "${container_name}"
+    fi
+
+    img="kw-${distro}"
+    image_exists "${img}"
+    if [[ "$?" -eq 0 ]]; then
+      say "Removing podman ${img} image."
+
+      # Get this image layers, so we can safely remove its children dangling images.
+      img_layer_list=$(image_inspect -f '{{.RootFS.Layers}}' "${img}" | tr -d '[]')
+
+      # Remove possible dangling images
+      while read -r dangling_img; do
+        # We check if the image exist  because  it  could  happen  that  in  the
+        # previous iteration the father dangling image was deleted, causing  the
+        # child dangling image to be deleted as well.
+        image_exists "${dangling_img}"
+        if [[ "$?" -ne 0 ]]; then
+          continue
+        fi
+
+        # We check if the dangling image have layers which are exact prefixes of
+        # the current image we are trying to delete.
+        dangling_img_layer_list=$(image_inspect -f '{{.RootFS.Layers}}' "${dangling_img}" | tr -d '[]')
+        grep --perl-regexp "^${dangling_img_layer_list}" <<< "${img_layer_list}" > /dev/null
+
+        # Delete the dangling image if that is the case.
+        if [[ "$?" -eq 0 ]]; then
+          # Dangling images are always force-removed.
+          image_rm --force "${dangling_img}"
+        fi
+      done <<< "$(image_ls --all --quiet --filter 'intermediate=true')"
+
+      # Remove main image.
+      image_rm --force "${img}"
+      if [[ "$?" -ne 0 ]]; then
+        complain "Failed to remove ${img} image."
+      fi
+    fi
+
+    # Remove the base image for the distro.
+    img="${CONTAINER_BASE_IMAGE}/${distro}"
+    image_exists "${img}"
+    if [[ "$?" -eq 0 ]]; then
+      say "Removing podman ${img} image."
+
+      if [[ -n "$flag" ]]; then
+        image_rm "$flag" "$img"
+      else
+        image_rm "$img"
+      fi
+    fi
+  done
+}
+
+# Check if the given image exists.
+#
+# @image    The image name or id.
+function image_exists()
+{
+  podman image exists "$1"
+}
+
+# Remove the given images.
+#
+# @flags    Optional flags to be passed to podman.
+#           If it is '-f' or '--force', then it should be the ONLY flag.
+# @images   Image names or ids.
+function image_rm()
+{
+  local force_remove=0
+  local img
+
+  if [[ "$1" == "-f" || "$1" == "--force" ]]; then
+    force_remove=1
+    shift
+  fi
+
+  if [[ "$#" -lt 1 ]]; then
+    complain "(${LINENO}): no image provided to be removed."
+    return 1
+  fi
+
+  # shellcheck disable=SC2068
+  podman image rm $@ > /dev/null 2>&1
+
+  # command succeed.
+  if [[ "$?" -eq 0 ]]; then
+    return 0
+  fi
+
+  # comman failed and we should not force-attempt it.
+  if [[ "$force_remove" == 0 ]]; then
+    complain "(${LINENO}): kw failed to execute \`podman image rm ${*}\`. Consider using --force flag."
+    return 1
+  fi
+
+  # We try it again, this time forcing the removal of the images.
+  # WARNING: If the first argument is '-f', we suppose all the other ones are the image names.
+  while [[ "$#" -gt 0 ]]; do
+    img="$1"
+    shift
+
+    # Make sure the image exists.
+    podman image exists "$img"
+    if [[ "$?" -ne 0 ]]; then
+      complain "(${LINENO}) provided image '${img}' does not exist."
+      continue
+    fi
+
+    # Force remove all containers that depend on that image, one by one.
+    (podman container ls --all --quiet --filter ancestor="$img" |
+      xargs -n1 podman container rm --force --time 0) > /dev/null 2>&1
+
+    # Force remove the image now.
+    podman image rm --force "$img" > /dev/null 2>&1
+  done
+}
+
+# List the given images.
+#
+# @options  Options to be passed to podman.
+# @images   Image names or ids.
+function image_ls()
+{
+  if [[ "$#" -le 1 ]]; then
+    complain "(${LINENO}): no image provided to be listed."
+    return 1
+  fi
+
+  # shellcheck disable=SC2068
+  podman image ls $@
+
+  if [[ "$?" -ne 0 ]]; then
+    complain "(${LINENO}): kw failed to execute \`podman image ls ${*}\`."
+  fi
+}
+
+# Inspect the given images.
+#
+# @options  Options to be passed to podman.
+# @images   Image names or ids.
+function image_inspect()
+{
+  # shellcheck disable=SC2068
+  podman image inspect $@
+
+  if [[ "$?" -ne 0 ]]; then
+    complain "(${LINENO}): kw failed to execute \`podman image inspect ${*}\`."
+  fi
+}
+
+# run a container
 function container_run()
 {
   # shellcheck disable=SC2068
@@ -210,6 +382,10 @@ function container_copy()
   fi
 }
 
+# Inspect the given containers.
+#
+# @options  Options to be passed to podman.
+# @images   Container names or ids.
 function container_inspect()
 {
   # shellcheck disable=SC2068

### CODE REVIEW

File: run_tests.sh [hunk 1]
```diff
@@ -178,6 +183,9 @@ elif [[ "$1" == 'test' ]]; then
   files_list=$(find "$TESTS_DIR" | grep --perl-regexp "${regex}" | grep --extended-regexp --invert-match 'samples/.*|/shunit2/')
   set_tests $files_list
   LANGUAGE=en_US.UTF_8 run_tests
+elif [[ "$1" == "--clear-cache" ]]; then
+  say "Clearing tests cache"
```
rodrigosiqueira, line 187: Use single quotes; also, add this option in the help output.

rodrigosiqueira (changes requested):

run_tests.sh
```diff
@@ -178,6 +183,9 @@ elif [[ "$1" == 'test' ]]; then
   files_list=$(find "$TESTS_DIR" | grep --perl-regexp "${regex}" | grep --extended-regexp --invert-match 'samples/.*|/shunit2/')
   set_tests $files_list
   LANGUAGE=en_US.UTF_8 run_tests
+elif [[ "$1" == "--clear-cache" ]]; then
```
davidbtadokoro, line 186: By the way the conditionals are, the following cases are acceptable
```bash
./run_tests.sh --unit --clear-cache
./run_tests.sh --integration --clear-cache
./run_tests.sh --clear-cache
```
Which one are you aiming to?

IMHO, I think it makes sense to tie the option to `--integration` like in the the second case and maybe have
```bash
./run_tests.sh --integration clear-cache
```
just to keep the pattern of set by the `test`, `list`, and `help` options that come after `--unit`/`--integration`.

davidbtadokoro (changes requested):
Hey @uwla, and thank you so much this change.

Tested it and it seems to do what it proposes to. In any case, I came upon the errors from podman shown below when removing some images. I think that this may be a problem on my end, but can you double check the behavior, just to be sure?

Screenshot from 2024-01-17 15-19-35

tests/integration/utils.sh
```diff
@@ -167,6 +167,43 @@ teardown_container_environment()
...
+    podman image exists "${img}"
+    if [[ "$?" -eq 0 ]]; then
+      say "Removing podman ${img} image."
+      podman image rm "${img}"
+      if [[ "$?" -ne 0 ]]; then
+        complain "Failed to remove ${img} image."
+      fi
+    fi
+
+    img="docker.io/library/${distro}"
```
rodrigosiqueira, line 195: I think we should make this docker prefix path a global read-only variable because, in the future, if we want to use something different, we can just change the global variable.

rodrigosiqueira (commented):
I can confirm that I had a similar issue as the one reported by @davidbtadokoro .

Screenshot from 2024-01-18 17-45-15

tests/integration/utils.sh
uwla: True, indeed

run_tests.sh
uwla: ok

run_tests.sh
```diff
@@ -122,13 +122,18 @@ function run_tests()
   if [[ "${integration_tests_setup}" == 1 ]]; then
     printf '\n' # add new line after the last "OK"
     say 'Tearing down environment used in integration tests...'
```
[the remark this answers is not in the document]
uwla: Ok

run_tests.sh
uwla: That is a good idea: to have cached cleared on a scope basis.

uwla:
> Hey @uwla, and thank you so much this change.
> 
> Tested it and it seems to do what it proposes to. In any case, I came upon the errors from podman shown below when removing some images. I think that this may be a problem on my end, but can you double check the behavior, just to be sure?
> 
> Screenshot from 2024-01-17 15-19-35

As the podman's output points out, the kw images are being used by other images. I had experienced this behavior and deleted intermediary images before. The problem is this: it seems Podman generates duplicated entries of the built images; you can verify that by `podman image ls -a` and see the images seem to be equal.

One way we can address that is to force remove the images.

davidbtadokoro:
> One way we can address that is to force remove the images.

This may be a way to do this. Adding the flag `--force` to the command
```bash
podman image rm --force "$img"
```
seems to do the trick.

However, we need to be sure that doing this won't mess any other podman image besides the ones related to kw's integration tests :grimacing:

uwla:
> However, we need to be **really** sure that doing this won't mess any other podman image besides the ones related to kw's integration tests 😬

I have two suggestions:

1. Interactvely ask the user if KW should proceed and force the removal of the images. This is the easiest way, but I suspected the other "duplicated" podman images would remain (this is problem of duplicating entries of image layers is related to podman, not KW)
2. Figure out a way to inspect which other images are dependent on the one we are trying to remove, and if they are just duplicated entries, we remove them first.

uwla:
I cleared all podman images and container in my system, then ran the integration tests, which built the images locally. Here is the output:

```
[kworkflow][ unstable][$⇡] ❯ podman image ls                             
REPOSITORY                   TAG         IMAGE ID      CREATED        SIZE
localhost/kw-fedora          latest      dba3438ea2ed  2 minutes ago  1.17 GB
localhost/kw-debian          latest      668dde0f82f7  5 minutes ago  2.39 GB
localhost/kw-archlinux       latest      beab320280c0  7 minutes ago  1.34 GB
docker.io/library/archlinux  latest      69f38d8f6347  3 weeks ago    446 MB
docker.io/library/debian     latest      2a033a8c6371  5 weeks ago    121 MB
docker.io/library/fedora     latest      8404925a71fd  8 weeks ago    183 MB
[kworkflow][ unstable][$⇡] ❯ podman image ls -a                        
REPOSITORY                   TAG         IMAGE ID      CREATED        SIZE
localhost/kw-fedora          latest      dba3438ea2ed  2 minutes ago  1.17 GB
<none>                       <none>      e76b911a33d6  3 minutes ago  509 MB
<none>                       <none>      b916531e6fee  3 minutes ago  509 MB
localhost/kw-debian          latest      668dde0f82f7  5 minutes ago  2.39 GB
<none>                       <none>      6353772123c4  7 minutes ago  267 MB
<none>                       <none>      c9c5cc748609  7 minutes ago  267 MB
localhost/kw-archlinux       latest      beab320280c0  8 minutes ago  1.34 GB
<none>                       <none>      bc960f2cb82f  9 minutes ago  755 MB
<none>                       <none>      83b04e8e5557  9 minutes ago  755 MB
docker.io/library/archlinux  latest      69f38d8f6347  3 weeks ago    446 MB
docker.io/library/debian     latest      2a033a8c6371  5 weeks ago    121 MB
docker.io/library/fedora     latest      8404925a71fd  8 weeks ago    183 MB
```

The images marked by `<none>` are byproducts of the building process and can be removed without any loss. Really, they can be removed and the other images will work fine. So, we could do:

```
$ podman image ls -a | grep '<none>' | awk '{ print $3 }' | xargs podman image rm -f
```

However, I cannot ensure removing images marked by `<none>` won't remove images which the user actually wants to keep.

uwla:
I also noticed all these `temporary` images have the `-tmp` prefix in their names:

```
❯ podman image ls -aq | xargs podman inspect -f "{{.NamesHistory}}"
[localhost/kw-fedora:latest]
[docker.io/library/c1b1a7ee84e9430ab72ca01da8d8589c21cf2495a6144e5218dda596990c8b73-tmp:latest]
[docker.io/library/51bda72fc5b773d2aec762bf4cea2b1532c4fd6ee59575bed7c202230d30cce1-tmp:latest]
[localhost/kw-debian:latest]
[docker.io/library/0f4612409cf1443e4b462185c3942187a3753692ff6263bfa68b2acef9d59cb3-tmp:latest]
[docker.io/library/5b22c6a784bbbe871b20f772cc184815c5e5f73506bf09f53d55dfd783fd0400-tmp:latest]
[localhost/kw-archlinux:latest]
[docker.io/library/f378a37dd967a15f6bc8b67bf8a0ab117c6e4875581cfbc3fa6b6903d13790c9-tmp:latest]
[docker.io/library/7bfa6850b870c614a20211ba34ed081338e53d3e5d92f47caa8d4565d18eaa96-tmp:latest]
[docker.io/library/archlinux:latest]
[docker.io/library/debian:latest]
[docker.io/library/fedora:latest]
```

We could assume if a image is marked by `-tmp`, it is a temporary image, byproduct of the building process, and can be discard without any problems. I believe this is the way to go.

If I find a better solution, I'll post it here.

uwla:
Ok, I found out a better solution but this will require some extra string parsing.

We can see the podman images, including the temporary images, share layers:

```
 podman image ls -a | grep -v "IMAGE ID" | awk '{ print $3}' | xargs podman inspect -f "{{.RootFS.Layers}}" | sort
[sha256:53f86715cdbaf45ddf227b5582e881e3ad0d5eb056eda3b0094a66f7cf0c0f1f]
[sha256:53f86715cdbaf45ddf227b5582e881e3ad0d5eb056eda3b0094a66f7cf0c0f1f sha256:a0f125ca6de8ea112b7a6fe7c99afe5b0cac482a666b8b1a0b637d18902358b1]
[sha256:53f86715cdbaf45ddf227b5582e881e3ad0d5eb056eda3b0094a66f7cf0c0f1f sha256:a0f125ca6de8ea112b7a6fe7c99afe5b0cac482a666b8b1a0b637d18902358b1 sha256:3155923fa362a69998dd8afadf32efe9571ede98920a38c2f659bee46b0f7fec]
[sha256:53f86715cdbaf45ddf227b5582e881e3ad0d5eb056eda3b0094a66f7cf0c0f1f sha256:a0f125ca6de8ea112b7a6fe7c99afe5b0cac482a666b8b1a0b637d18902358b1 sha256:3155923fa362a69998dd8afadf32efe9571ede98920a38c2f659bee46b0f7fec sha256:72c62a91e48f8188ff0acb936ae9ebabcf1477e26ea4220860936016ae591a1d]
[sha256:6de0fb71036567b0339788eaa924723efbd5be22f9683b8aeb3b3c23d4e41a9d sha256:d822bba4f032dfd2df5b57457840d3ef708ec0f592b3bc29cc2cba6cf271de23]
[sha256:6de0fb71036567b0339788eaa924723efbd5be22f9683b8aeb3b3c23d4e41a9d sha256:d822bba4f032dfd2df5b57457840d3ef708ec0f592b3bc29cc2cba6cf271de23 sha256:a0030b74034672418b6f47e34735be3c31e1591a87b601afaf695a79e279610b]
[sha256:6de0fb71036567b0339788eaa924723efbd5be22f9683b8aeb3b3c23d4e41a9d sha256:d822bba4f032dfd2df5b57457840d3ef708ec0f592b3bc29cc2cba6cf271de23 sha256:a0030b74034672418b6f47e34735be3c31e1591a87b601afaf695a79e279610b sha256:3155923fa362a69998dd8afadf32efe9571ede98920a38c2f659bee46b0f7fec]
[sha256:6de0fb71036567b0339788eaa924723efbd5be22f9683b8aeb3b3c23d4e41a9d sha256:d822bba4f032dfd2df5b57457840d3ef708ec0f592b3bc29cc2cba6cf271de23 sha256:a0030b74034672418b6f47e34735be3c31e1591a87b601afaf695a79e279610b sha256:3155923fa362a69998dd8afadf32efe9571ede98920a38c2f659bee46b0f7fec sha256:02a0ac4c6157eb3f57f23838293c1668f4205fbbe2be38716f99313885f708de]
[sha256:ae134c61b154341a1dd932bd88cb44e805837508284e5d60ead8e94519eb339f]
[sha256:ae134c61b154341a1dd932bd88cb44e805837508284e5d60ead8e94519eb339f sha256:c2ff07d5b87862d11b7e954cdb1a261fcc3b8f53f83b9b6b11a8ce74988c0f04]
[sha256:ae134c61b154341a1dd932bd88cb44e805837508284e5d60ead8e94519eb339f sha256:c2ff07d5b87862d11b7e954cdb1a261fcc3b8f53f83b9b6b11a8ce74988c0f04 sha256:3155923fa362a69998dd8afadf32efe9571ede98920a38c2f659bee46b0f7fec]
[sha256:ae134c61b154341a1dd932bd88cb44e805837508284e5d60ead8e94519eb339f sha256:c2ff07d5b87862d11b7e954cdb1a261fcc3b8f53f83b9b6b11a8ce74988c0f04 sha256:3155923fa362a69998dd8afadf32efe9571ede98920a38c2f659bee46b0f7fec sha256:91dd6e2a58e3d7ed997362d41b005ff03080eca25da14226252595e4b1a3c1e3]
```

We can then delete all podman images whose layers are a prefix of the KW images.

```
tmp image:  abc def ghi
kw archlinux image: abc def ghi jkl mno
other image: abc def ghi poq
```

In the example above, `tmp` is a prefix of `kw archlinux image`, but `other image` is not because `poq` layer is not present in `kw archlinux image`.

Using this heuristic, we can be confident only the byproduct images will be removed along with kw images.

Please, give me our opinions @davidbtadokoro or @rodrigosiqueira.

uwla:
Ok, I'll implement the solution soon and update the PR

uwla:
@davidbtadokoro please review it again

tests/integration/utils.sh
```diff
@@ -167,6 +167,106 @@ teardown_container_environment()
...
+      # remove possible dangling images
+      image_ls --all --quiet --filter 'intermediate=true' | while read -r tmp_img; do
+
+        # we check if the dangling image have layers which are exact prefixes of
+        # the current image we are trying to delete.
+        tmp_img_layers=$(image_inspect -f '{{.RootFS.Layers}}' "${tmp_img}" | tr -d '[]')
+        grep --perl-regexp "^${tmp_img_layers}" <<< "${img_layers}" > /dev/null
+
+        # delete the dangling image if that is the case.
+        if [[ "$?" -eq 0 ]]; then
```
davidbtadokoro, line 211: A cleaner way to do this would be
```bash
var=$(printf '%s' "$img_layers" | grep --perl-regexp "^${tmp_img_layers}")
if [[ -n "$var" ]]; then
 ...
```
Of course, change the name `var` to a more adequate one.

I say that this is cleaner, because we don't need to redirect the output to `/dev/null`, reading it from left to right (beginning to end) makes more sense, and, most importantly, it follows our coding style.

davidbtadokoro (changes requested):
Hey @uwla, and thanks for the update fixing the issue me and @rodrigosiqueira reported!

Talking about the changes requested, one relates to the commit message, and the others are mainly about coding style and improving readability.

In terms of solving the reported issue, your update really did the trick, as it deletes the temporary dangling images created by the integration tests, and only those! However, I got some strange errors that don't seem to affect the outcome. See the screenshots below:

Screenshot from 2024-01-31 00-00-07
Screenshot from 2024-01-31 00-01-04
Screenshot from 2024-01-31 00-01-15

uwla:
@davidbtadokoro I made the requested changes

run_tests.sh
```diff
@@ -178,6 +194,9 @@ elif [[ "$1" == 'test' ]]; then
   files_list=$(find "$TESTS_DIR" | grep --perl-regexp "${regex}" | grep --extended-regexp --invert-match 'samples/.*|/shunit2/')
   set_tests $files_list
   LANGUAGE=en_US.UTF_8 run_tests
+elif [[ "$1" == "clear-cache" ]]; then
```
davidbtadokoro, line 197: Use single-quotes instead of double-quotes when there is no expansion.

davidbtadokoro (commented):
Hi @uwla, and many many thanks for the update! Everything looks legit to me, from both the coding style and behavior standpoints :+1:. It also fixed those error messages about failing to inspect. There is a small nitpicky comment about single quotes that I left just for documentation, but I can fix this myself, no problems. Nevertheless, great contribution and great effort all around :smiley: 

@rodrigosiqueira, as you also reviewed this, I'll wait for your green light to merge this PR.

rodrigosiqueira:
Hi @uwla , it looks like we have a conflict in this PR. Could you update it against the latest code from the unstable branch?

uwla:
yes, I'll

davidbtadokoro:
Hi @uwla, any update here?

davidbtadokoro:
Thanks for resolving the conflicts @uwla. @rodrigosiqueira, I'll wait for your greenlight to merge it.

tests/integration/utils.sh
```diff
@@ -157,6 +158,123 @@ function teardown_single_container()
...
+        complain "Failed to remove ${img} image."
+      fi
+    fi
+  done
+}
+
+function image_rm()
+{
+  # shellcheck disable=SC2068
+  podman image rm $@ > /dev/null
```
rodrigosiqueira, line 250: In my case, I had a failure that looked like this:

```
Removing podman docker.io/library/debian image.
Error: image used by c23446c400585541e620124d8202d5ed41dac46b5a7b0ee9b92907ae8d2406f3: image is in use by a container: consider listing external containers and force-removing image
(253): Failed to remove the images.
```

In the output, it suggests listing external containers and force-removing. Could you add these steps as part of the error message? I mean, add the podman command to list external containers and the command that forces the removal.

rodrigosiqueira, line 250: Just in case, check if `$@` has at least one parameter.

tests/integration/utils.sh
```diff
@@ -157,6 +158,123 @@ function teardown_single_container()
...
+
+  if [[ "$?" -ne 0 ]]; then
+    complain "(${LINENO}): Failed to remove the images."
+  fi
+}
+
+function image_ls()
+{
+  # shellcheck disable=SC2068
+  podman image ls $@
```
rodrigosiqueira, line 260: Just in case, check if `$@` has at least one parameter.

tests/integration/utils.sh
```diff
@@ -157,6 +158,123 @@ function teardown_single_container()
...
+
+  if [[ "$?" -ne 0 ]]; then
+    complain "(${LINENO}): Failed to list the images."
+  fi
+}
+
+function image_inspect()
+{
+  # shellcheck disable=SC2068
+  podman image inspect $@
```
rodrigosiqueira, line 270: Just in case, check if `$@` has at least one parameter.

tests/integration/utils.sh
```diff
@@ -157,6 +158,123 @@ function teardown_single_container()
...
+  fi
+}
+
+function image_ls()
+{
+  # shellcheck disable=SC2068
+  podman image ls $@
+
+  if [[ "$?" -ne 0 ]]; then
+    complain "(${LINENO}): Failed to list the images."
```
rodrigosiqueira, line 263: To make debugging easier in case of failure, add the command you tried to execute as part of the error message. Something like: `kw failed to execute the command: podman image ls SOMETHING SOMETHING`

tests/integration/utils.sh
```diff
@@ -157,6 +158,123 @@ function teardown_single_container()
...
+  fi
+}
+
+function image_inspect()
+{
+  # shellcheck disable=SC2068
+  podman image inspect $@
+
+  if [[ "$?" -ne 0 ]]; then
+    complain "(${LINENO}): Failed to inspect the images."
```
rodrigosiqueira, line 273: To make debugging easier in case of failure, add the command you tried to execute as part of the error message. Something like: `kw failed to execute the command: podman inspect ls SOMETHING SOMETHING`

tests/integration/utils.sh
```diff
@@ -157,6 +158,123 @@ function teardown_single_container()
...
+      say "Removing podman ${img} image."
+      image_rm "${flags}" "${img}"
+      if [[ "$?" -ne 0 ]]; then
+        complain "Failed to remove ${img} image."
+      fi
+    fi
+  done
+}
+
+function image_rm()
```
rodrigosiqueira, line 247: Add function documentation on top, in particular, explaining that "$@" is a set of flags that will be passed to Podman. Do the same for all the other functions.

tests/integration/utils.sh
```diff
@@ -157,6 +158,123 @@ function teardown_single_container()
...
+        # the current image we are trying to delete.
+        tmp_img_layers=$(image_inspect -f '{{.RootFS.Layers}}' "${tmp_img}" | tr -d '[]')
+        tmp_output=$(grep --perl-regexp "^${tmp_img_layers}" <<< "${img_layers}")
+
+        # Delete the dangling image if that is the case.
+        if [[ -n "$tmp_output" ]]; then
+          # Dangling images are always force-removed.
+          image_rm --force "${tmp_img}"
+        fi
+      done <<< "$(image_ls --all --quiet --filter 'intermediate=true')"
```
rodrigosiqueira, line 225: I'm wondering if we can fail in this part... maybe we can run this command before this while, save the output in a variable/array, and use it in this while loop if everything is fine. This can make the code more reliable and easier to debug in case something happens.

tests/integration/utils.sh
```diff
@@ -157,6 +158,123 @@ function teardown_single_container()
...
+function teardown_container_environment()
+{
+  local flags="$1"
+  local container_name
+  local distro
+  local img
+  local img_layers
+  local tmp_img
+  local tmp_img_layers
+  local tmp_output
```
rodrigosiqueira, line 187: This variable looks very generic, and it is hard to guess what it does from its name. Fwiu, you want to get part of the layer image name, right? If so, rename this variable to something that expresses that idea.

tests/integration/utils.sh
```diff
@@ -157,6 +158,123 @@ function teardown_single_container()
...
+
+# Completely remove the container environment: the containers and the images
+function teardown_container_environment()
+{
+  local flags="$1"
+  local container_name
+  local distro
+  local img
+  local img_layers
+  local tmp_img
```
rodrigosiqueira, line 185: This name looks a little bit generic; maybe use something like podman_image_list or something like that.

tests/integration/utils.sh
```diff
@@ -157,6 +158,123 @@ function teardown_single_container()
...
+  local tmp_img_layers
```
rodrigosiqueira, line 186: Same here... this variable name is a little bit generic.

run_tests.sh
```diff
@@ -121,14 +122,25 @@ function run_tests()
...
+    teardown_containers
     printf '\n'
   fi
 
   report_results "$total" "$success" "$notfound" "$fail" "$test_failure_list"
 }
 
+function clear_unit_tests_cache()
+{
+  say 'Clearing unit tests cache...'
```
rodrigosiqueira, line 135: If there is nothing to clean, just say that; otherwise, you can give the wrong idea to the user. Maybe say, `Skip unit test clean` or `Clear unit test is not necessary` or anything like that.

run_tests.sh
```diff
@@ -18,6 +18,7 @@ function show_help()
     '         Limit tests to unit tests' \
     '' \
     'COMMANDS' \
+    '  clear-cache - clears tests cache' \
```
rodrigosiqueira, line 21: Don't forget to update documentation/content/tests.rst with this new option.

rodrigosiqueira (changes requested):
Hi @uwla , thanks a lot for the update, and sorry for my delay in reviewing it. You can find some of my comments in this PR.

run_tests.sh
uwla: ok

tests/integration/utils.sh
uwla: ok

uwla:
Everything should be working fine by now.

@rodrigosiqueira and @davidbtadokoro,  I tested it and worked on my side, but I also ask you to test it locally, because we do not have automated integration tests yet.

davidbtadokoro:
Nice @uwla ! I'll take a look at it : )

run_tests.sh
davidbtadokoro: Make it single-quotes here.

documentation/content/tests.rst
```diff
@@ -41,23 +41,26 @@ are some examples:
...
 seconds each time.
 
-Then, the local kw repo is copied to the containers and installed  again,  which
-takes very few seconds. For optimization purposes,  the  containers  are  reused
-accross tests. If you add a new commit or checkout to another branch, such  that
-HEAD points to another commit, the containers  will  be  destroyed  and  created
+Then, the local kw repo is copied to the containers and installed again, which
+takes very few seconds. For optimization purposes, the containers are reused
+accross tests. If you add a new commit or checkout to another branch, such that
+HEAD points to another commit, the containers will be destroyed and created
```
davidbtadokoro, line 63: I think your formatter changed these two paragraphs. Revert them to the previous state.

tests/integration/utils.sh
```diff
@@ -144,7 +145,8 @@ function teardown_container_environment()
 
 # Destroy a single container
 #
-# @container: Name or ID of the container
+# @flags        Optional flags, such as -f to force remove it.
```
davidbtadokoro, line 148: There is no `flags` argument, so remove it from the function doc.

tests/integration/utils.sh
```diff
@@ -157,6 +159,177 @@ function teardown_single_container()
...
+
+    img="kw-${distro}"
+    image_exists "${img}"
+    if [[ "$?" -eq 0 ]]; then
+      say "Removing podman ${img} image."
+
+      # Get this image layers, so we can safely remove its children dangling images.
+      img_layer_list=$(image_inspect -f '{{.RootFS.Layers}}' "${img}" | tr -d '[]')
+
+      # Gemove possible dangling images
```
davidbtadokoro, line 191: Fix `Gemove` typo.

tests/integration/utils.sh
```diff
@@ -157,6 +159,177 @@ function teardown_single_container()
...
+      while read -r dangling_img; do
+        # We check if the image exist  because  it  could  happen  that  in  the
+        # previous iteration the father dangling iamge was deleted, causing  the
```
davidbtadokoro, line 194: Fix `iamge` typo.

tests/integration/utils.sh
```diff
@@ -157,6 +159,177 @@ function teardown_single_container()
...
+      xargs -n1 podman container rm --force --time 0) > /dev/null 2>&1
+
+    # Force remove the image now.
+    podman image rm --force "$img" > /dev/null 2>&1
+  done
+}
+
+# List the given images.
+#
+# @flags    Optional flags to be passed to podman.
```
davidbtadokoro, line 301: FWIU, this `flags` argument are podman options that will be passed through `$@` (along with the `images`) to the podman command. Instead of separating the documentation of these two arguments, I think it is better to describe them as one (all arguments passed to the function), and explicitly mentioning that it is composed of (first) podman options and (second) image names/IDs.

tests/integration/utils.sh
```diff
@@ -157,6 +159,177 @@ function teardown_single_container()
...
+# @images   Image names or ids.
+function image_ls()
+{
+  if [[ $# -le 1 ]]; then
```
davidbtadokoro, line 305: Use double-quotes around `$#`.

tests/integration/utils.sh
```diff
@@ -157,6 +159,177 @@ function teardown_single_container()
...
+  podman image ls $@
+
+  if [[ "$?" -ne 0 ]]; then
+    complain "(${LINENO}): kw failed to execute \`podman image ls ${*}\`."
+  fi
+}
+
+# Inspect the given images.
+#
+# @flags    Optional flags to be passed to podman.
```
davidbtadokoro, line 320: Same comment as the one I made on the `image_ls` function.

davidbtadokoro (changes requested):
Hi @uwla, and thanks for the update! We are really close to the perfect point!

Left just some simple comments this time. Also, a comment that didn't fit the inline review: remove the final dot in the commit message subject.

uwla:
@davidbtadokoro I've made the requested changes. Hopefully it passes all checks now

rodrigosiqueira:
Hi @uwla , thanks a lot for your update. I decided to merge this change into the unstable branch.
Thanks a lot.

Issue #1016, opened 2024-01-11
Error when preparing container environment


jppaulo06:
**Description:**

When executing integration tests, I keep getting these errors:
```
Preparing environment for integration tests...
[1/6] Using cached container image for archlinux.
[2/6] Creating archlinux container.
./tests/integration/utils.sh: line 129: fail: command not found
[3/6] Using cached container image for debian.
[4/6] Creating debian container.
./tests/integration/utils.sh: line 129: fail: command not found
[5/6] Using cached container image for fedora.
[6/6] Creating fedora container.
./tests/integration/utils.sh: line 129: fail: command not found
```

The tests continue to execute normally, with the containers created (see additional context).

Note that there is this command called `fail`, which is either a program that I don't have or a function that does not exist. Maybe this is also a bug.

**How to Reproduce:**

[Steps to reproduce the behavior:
1. Which commands were invoked
2. Where the error appears]

**Expected behavior:**

Test environment should be prepared without errors.

**System information:**

 - OS: Linux fedora 6.6.9
 - kw Version: 
``` 
alpha
Branch: unstable
Commit: 1764e42
```
 - Bash Version 5.2.21

**Additional context:**

I've opened some other issues due to additional problems I'm experiencing with integration tests on the unstable branch. While they may be related, I've decided to address them separately to minimize confusion as much as possible.

rodrigosiqueira:
@uwla

rodrigosiqueira:
@jppaulo06, it looks like you are missing some dependencies. Could you check which command fails in the below line:

`./tests/integration/utils.sh: line 129: fail: command not found`

I suppose it is `container_exec`.

jppaulo06:
> @jppaulo06, it looks like you are missing some dependencies. Could you check which command fails in the below line:
> 
> `./tests/integration/utils.sh: line 129: fail: command not found`
> 
> I suppose it is `container_exec`.

@rodrigosiqueira thanks for the suggestion :)

However, I searched the package for a while but didn't find it in fedora yet :(

But I did understand why I'm getting at this "fail" line, which, I think, is the main error:

Since I'm using fedora, SELinux mode is configured to "enforcing" as default. Apparently, this is causing permissions problems to the mounted volumes inside the containers.

The fix is easy; I will open a PR in a moment.

PR #1020, opened 2024-01-15
tests: integration: Fix permission denied inside containers due to SELinux

jppaulo06:
SELinux is causing permission troubles to podman containers when the current mode is set to "enforcing". It does not allow to execute setup.sh inside the container, for example.

This problem is related with how podman is built, being, currently, a quite common podman issue. Yet, it has an easy fix: add :z when mounting the volume.

For more references, read [podman troubleshooting page](https://github.com/containers/podman/blob/main/troubleshooting.md#2-cant-use-volume-mount-get-permission-denied).

Helps: #1016

### DIFF
--- a/tests/integration/utils.sh
+++ b/tests/integration/utils.sh
@@ -111,7 +111,7 @@ function setup_container_environment()
     # processes) and get the output of each of them separately.
     container_run \
       --workdir "${working_directory}" \
-      --volume "${KWROOT_DIR}":"${working_directory}" \
+      --volume "${KWROOT_DIR}":"${working_directory}:Z" \
       --env PATH='/root/.local/bin:/usr/bin' \
       --name "${container_name}" \
       --detach \


davidbtadokoro (commented):
Hey @jppaulo06, nice change!

I couldn't reproduce the issue reported in #1016 (I was using a fedora VM, don't know if it was because of this). However, by your explanation and the reference from the podman troubleshoot, I agree conceptually with your change and it doesn't seems to add regressions (i.e., my integration tests are working like before).

As a side note, I was [reading further](https://unix.stackexchange.com/questions/651198/podman-volume-mounts-when-to-use-the-z-or-z-suffix) on the use of `:z` instead of `:Z`, and it seems that `:z` is the correct one as the mounted volume pointed by `KWROOT_DIR` will be shared among the multiple containers created in the tests. Just as a sanity check, could you confirm that we can't use `Z:`? Because, if we can, we should opt for it as it would make the testing environments even more isolated (I think :sweat_smile:).

rodrigosiqueira:
Hi @jppaulo06, any update here?

jppaulo06:
Hey, @davidbtadokoro
Everything is OK with the `:Z` instead of `:z` :)

I've just got a Fedora 39 cloud image for testing purposes in a VM. The same problem happened :/

Maybe your VM has some config changes? In Redhat based systems, as Fedora, SELinux mode is, by default, set to `enforcing`, but maybe something happened...

Check if your SELinux mode is set to `enforcing` with the command `sestatus`. If you get `Current mode: permissive`, run `sudo setenforce enforcing` to change it. 

Then, try the integration tests again.

This configuration is not permanent. However, you might want to go back with `sudo setenforce permissive`.

davidbtadokoro:
Hi @jppaulo06, and thanks for the reply!

> Maybe your VM has some config changes?

You are right, just downloaded the same setup on a VM and could reproduce the bug :sweat_smile: 

> Hey, @davidbtadokoro Everything is OK with the `:Z` instead of `:z` :)

Nice! It has been some time since I looked their differences, but I remembered that, according to the reference you commented, `Z:` and `z:` were the same, but the big one made the containers more isolated from each one.

If my assertion is right, `Z:` works the same as `z:` (from the integration tests perspective), and there is no significant performance loss, can you update the PR switching the options?

jppaulo06:
> You are right, just configured the same setup on a VM and could reproduce the bug (not that I was saying that the bug didn't exist first 😅).
Nice! Just had to check if the problem wasn't solely happening in my system (which is kind of a mess) hahah

> If my assertion is right, `Z:` works the same as `z:` (from the integration tests perspective), and there is no significant performance loss, can you update the PR switching the options?
You are right. Ok, I will update it!

rodrigosiqueira:
Hi @jppaulo06 , I noticed that you replaced the option. Thanks a lot.

PR #1023, opened 2024-01-16
src: kw_remote: Add dashes to remote commands

jppaulo06:
This change was asked by @rodrigosiqueira in [this closed pull request](https://github.com/kworkflow/kworkflow/pull/996).

`kw remote` commands lack a clear pattern related to double dashes (--). Notably, `kw remote add`, `kw remote remove` and `kw remote rename` do not include dashes, but `kw remote --list` does.

So, to make it more clear, execute the following changes:
- `kw remote add` to `kw remote --add`
- `kw remote remove` to `kw remote --remove`
- `kw remote rename` to `kw remote --rename`

### DIFF
--- a/documentation/man/features/kw-remote.rst
+++ b/documentation/man/features/kw-remote.rst
@@ -7,9 +7,9 @@ kw-remote - Manage set of tracked test machines
 SYNOPSIS
 ========
 | *kw remote* [-v | \--verbose]
-| *kw remote* [--global] add <name> <user>@<remote>[:<port>]
-| *kw remote* [--global] remove <name>
-| *kw remote* [--global] rename <old-name> <new-name>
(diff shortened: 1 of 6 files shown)


davidbtadokoro (approved):
Thanks for tackling the proposed change of PR #996. I agree with going in the direction of adding double dashes for these options as it fits with the rest of the kw features. Tested the change and it doesn't introduce any regressions, nor it seems to alter any behavior.

@rodrigosiqueira, I'm going to wait for your green light as, although you said in the closed PR that you agreed with this change, I feel like this is an important UI change to not have your opinion on it :grimacing:

rodrigosiqueira:
Hi, @jppaulo06, thanks a lot for this new version. Change merged into the unstable branch.

Issue #824, opened 2023-03-21
`kw b --help` shows unrelated warning


magalilemes:
**Description:**

When running `kw b --help` outside a kernel tree, apart from the manual page, users are also prompted with an unrelated warning.
```
$ kw b --help
Please use kw init to update your config files
```
**How to Reproduce:**

Run `kw b --help` outside a kernel directory.

**Expected behavior:**

Only the man page of kw build should show up.

**System information:**

 - OS: Ubuntu 22.04.2 LTS 
 - kw Version: alpha, Branch: unstable, Commit: edd0f16
 - Bash Version: GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)

GuilhermeMoreno:
After reviewing the issue, it seems that the offenders are those lines in `build.sh`
```
# Every time build.sh is loaded its proper configuration has to be loaded as well
load_build_config
load_notification_config
```
That is, when `kw` includes `build.sh` it automatically calls these function, without checking whether the user wants to run the build or see the man pages. 

Maybe it would be useful to add a parameter in the `include` function to indicate whether the file is being added as module or main file.

PR #1029, opened 2024-01-25
src: build: fix unrelated warning in kw b --help

Andregnl:
When including the file build.sh, all the configurations were loaded under all conditions, so simply running kw b --help would show unrelated warning if used in a folder with no config files. So the following changes are made:
- Change "include" function to pass arguments to included file
- Change "build.sh" to interrupt loading the config if "--help" flag was detected

Fix suggestion for issue #824

### DIFF
--- a/kw
+++ b/kw
@@ -119,7 +119,7 @@ function kw()
       ;;
     build | b)
       (
-        include "${KW_LIB_DIR}/build.sh"
+        include "${KW_LIB_DIR}/build.sh" "$@"
 
         build_kernel_main '' "$@"
         local ret="$?"
--- a/src/build.sh
+++ b/src/build.sh
@@ -441,6 +441,23 @@ function build_help()
     '  build (--verbose) - Show a detailed output'
 }
 
-# Every time build.sh is loaded its proper configuration has to be loaded as well
+function check_include_interrupt()
+{
+  while [[ "$#" -gt 0 ]]; do
+    case "$1" in
+      --help | -h)
+        build_help "$1"
+        exit
+        ;;
+      *)
+        shift
+        ;;
+    esac
+  done
+}
+
+# Check if loading config is necessary
+check_include_interrupt "$@"
+
 load_build_config
 load_notification_config
--- a/src/lib/kw_include.sh
+++ b/src/lib/kw_include.sh
@@ -30,5 +30,6 @@ function include()
     declare -gA KW_INCLUDED_PATHS=(["$fullpath"]=1)
   fi
 
-  . "$filepath" --source-only
+  shift
+  . "$filepath" --source-only "$@"
 }


davidbtadokoro (commented):
Hi @Andregnl, and many thanks for tackling this issue.

I think we've discussed this issue and concluded that it wasn't much of a good first issue, as the problem, although simple, didn't seem to have a straightforward fix. I see that your change fixes the bug, but it comes with the cost of altering the behavior of `include`, a function used all around the project. I don't think this trade-off is worth this workaround :disappointed: 

Don't get me wrong, I see that you've taken the utmost care to make it safe and backward compatible, but, IMHO, I think that the need for this tricky workaround in the first place is the fact that `build.sh` is still outdated in some parts, and `load_build_config` is an example of this.

On this note, I think we will need to update this part of the code sooner or later, but I just came up with a more straightforward workaround to address the issue, i.e., having the mentioned error don't appear when running `kw build --help` **outside** a kernel tree. See the diff below, in which `load_build_config` just returns, in case the command is running outside a kernel tree:

```
diff --git a/src/build.sh b/src/build.sh
index 78e074c..919e656 100644
--- a/src/build.sh
+++ b/src/build.sh
@@ -283,6 +283,8 @@ function load_build_config()
 
   parse_configuration "${XDG_CONFIG_HOME:-"$HOME/.config"}/$KWORKFLOW/$BUILD_CONFIG_FILENAME" build_config
 
+  ! is_kernel_root "$PWD" && return
+
   if [[ -f "$PWD/$KW_DIR/$BUILD_CONFIG_FILENAME" ]]; then
     parse_configuration "$PWD/$KW_DIR/$BUILD_CONFIG_FILENAME" build_config
   else
```
Feel free to use this approach to solve the bug, but the requirement is that you really revise it, verify if this is a correct fix (I've kinda just sketched it :grimacing:), and add a test for `load_build_config` in this scenario :smiley_cat:.

Andregnl:
Good afternoon @davidbtadokoro, thank you for the response, and sorry for taking this much time to answer.

I think this solution may be more elegant and straightforward, but what about the case where the command "kw b --help" is used in a kernel directory with no kw config files? The warning would still show for the help command. Would this behavior be expected for this command?

davidbtadokoro:
> [...] sorry for taking this much time to answer.

No worries!!!
 
> I think this solution may be more elegant and straightforward, but what about the case where the command "kw b --help" is used in a kernel directory with no kw config files? The warning would still show for the help command. Would this behavior be expected for this command?

I agree with your point of view. Nevertheless, that is why I said that the real reason behind the issue is this outdated part of `build.sh`. FWIU, the intent of this part of `load_build_config` was to inform legacy-users of kw to run `kw init` (which I assume was just introduced into the project) to create local configs. Notice that showing the warning inside a kernel tree is not incorrect, and the issue pointed to the fact that the user was suggested to run `kw init` even outside a kernel tree, which doesn't makes sense.

With all that being said, I agree that showing this message inside a kernel tree in the special case of `kw build --help` still doesn't makes a lot of sense, but, as stated, the "correct" fix for this entails a bigger overhaul of `build.sh` :grimacing:

Andregnl:
Hi @davidbtadokoro, If the function is going to be removed anyway I don't see any reason for this PR to continue.

davidbtadokoro:
> Hi @davidbtadokoro, If the function is going to be removed anyway, I don't see any reason for this PR to continue. Thanks for letting me know.

Sure, I understand. Thanks for the fast reply :smiley:

PR #1021, opened 2024-01-15
Improve readability integrations utils

jppaulo06:
This pull request introduces minor changes to the integration tests code with the primary goal of improving code readability. The modifications do not alter the behavior or fix errors of the tests but focus on enhancing the clarity and maintainability of the code.

Key improvements include:
- Refactoring certain sections of the code to adhere to already used coding conventions.
- Adjusting variable names for better descriptive clarity.
- Reordering lines or sections to improve the logical flow of the code.

In addition to these changes, the PR slightly modifies the messages outputted during integration tests.

PR #1030, opened 2024-01-30
tests: refactor run_tests and integration tests

uwla:
- move global variable declarations to top of the file
- replace `if/elif/else` blocks with `case` statement
- extract long blocks of code to their own functions
- encapsulate podman commands in library functions

This refactor does not "refactor" everything, but just a few things.

It does not:
- rename variables for more clarity
- improve comments
- reorder lines
- improve general readbility

This refactoring is NOT a duplicate of #1021 because its scope is different: it changes things that PR does not change, and it does not changes things that PR changes.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -5,6 +5,11 @@ include './tests/unit/utils.sh'
 include './tests/integration/utils.sh'
 include './src/lib/kwio.sh'
 
+declare -a TESTS
+declare TESTS_DIR
+declare TESTS_UNIT=1
+declare TESTS_INTEGRATION=1
+
 function show_help()
 {
   printf '%s\n' "Usage: $0 [--flags] [help] [list] [test <tfile1> ...]" \
@@ -141,7 +146,6 @@ function clear_integration_tests_cache()
   teardown_container_environment "$@"
 }
 
-declare -a TESTS
 function set_tests()
 {
   local file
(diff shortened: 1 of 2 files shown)


rodrigosiqueira:
Hi @uwla , overall, your change lgtm. The only thing I disagree with is the replacement of `fail` with `complain` in the tests. `fail` is part of the shunit framework, and I think it makes more send in the test part.

uwla:
Done.

rodrigosiqueira:
Hi @uwla , it looks like this PR has conflicts. Could you rebase it against the latest unstable branch?
Thanks

uwla:
sure

davidbtadokoro:
Hey @uwla, any update here?

davidbtadokoro:
Thanks for also resolving the conflicts here @uwla. I took the opportunity to skim through the change and I have two observations:

1. Maybe we can split this commit into two: one for the refactors in `run_tests.sh` and the other for the ones in `tests/integration/utils.sh`.
2. I think this PR and #1019 conflict because both define `teardown_container_environment`. However I don't think it will be a difficult conflict to resolve and we should worry about it when the first is merged into the unstable.

rodrigosiqueira:
@uwla, from David's comment, maybe remove the container function that you added here that is already in PR #1019 .

uwla:
ok, I'll do it in a few days

uwla:
I'll go back to this PR after #1019 has been merged, because it introduces a new command and modifies utils used by integration tests. This PR needs to be build on top of that, in order for the code to be coeherent after merged.

davidbtadokoro:
> I'll go back to this PR after #1019 has been merged, because it introduces a new command and modifies utils used by integration tests. This PR needs to be build on top of that, in order for the code to be coeherent after merged.

For sure, it makes sense.

uwla:
@rodrigosiqueira I believe it should be good by now.

davidbtadokoro:
Hi @uwla, and sorry for the delay on our end. The PR is legit IMHO, but you didn't split the commit into two like I asked in my previous comment. Because you are refactoring two files in isolated contexts, as one refactor does not depend on the other, it is best practice to make them more atomic (notice that in the commit message you say `refactor run_tests.sh, tests/integration/utils.sh`, which means that this commit is doing two things at the same time).

Nevertheless, as we took too much time for feedback, I did the split myself. Don't worry, didn't altered the code, neither your credits or the authorship. Besides the split, I added more context to the commit messages. Anyway, thanks a lot! Merged into the unstable branch :+1:

Issue #965, opened 2023-11-16
test_create_async_loading_screen_notification_rely_on_some_default_options fails locally


uwla:
**Description:**

A recent new test is not passing locally.

**How to Reproduce:**

1. Clone kworkflow
2. Switch to unstable
3. Run ` ./run_tests.sh test tests/lib/dialog_ui_test.sh`

**Expected behavior:**

The test `test_create_async_loading_screen_notification_rely_on_some_default_options` should not fail even if it runs on another Linux system
OR
we should know what extra dependencies we have to install in order to develop KW and be able to run the tests consistenly accross machines.

**Screenshots:**

```
./run_tests.sh test tests/lib/dialog_ui_test.sh
Running test [dialog_ui_test]
=========================================================
egrep: warning: egrep is obsolescent; using grep -E
test_ui_setup_no_layout
test_ui_setup_with_layout
test_create_menu_options_rely_on_some_default_options
test_create_menu_options_use_all_options
test_create_simple_checklist_rely_on_some_default_options
test_create_simple_checklist_use_all_options
test_create_loading_screen_notification_rely_on_some_default_options
test_create_loading_screen_notification_use_all_options
test_create_async_loading_screen_notification_rely_on_some_default_options
ASSERT: line 180: Expected async loading screen with some default options
  Expected Result: dialog --colors --infobox $'loading latest patchsets from amd-gfx ⣾⣽⣻' '8' '60'dialog --colors --infobox $'loading latest patchsets from amd-gfx ⣾⣽⣻' '8' '60'dialog --colors --infobox $'loading latest patchsets from amd-gfx ⢿⡿⣟' '8' '60'
    Actual Result: dialog --colors --infobox $'loading latest patchsets from amd-gfx ⣾' '8' '60'dialog --colors --infobox $'loading latest patchsets from amd-gfx ⣾' '8' '60'dialog --colors --infobox $'loading latest patchsets from amd-gfx ⣽' '8' '60'
test_create_async_loading_screen_notification_use_all_options
ASSERT: line 205: Expected async loading screen with some default options
  Expected Result: dialog --colors --infobox $'loading latest patchsets from amd-gfx ⣾⣽⣻' '1234' '2718281'dialog --colors --infobox $'loading latest patchsets from amd-gfx ⣾⣽⣻' '1234' '2718281'dialog --colors --infobox $'loading latest patchsets from amd-gfx ⢿⡿⣟' '1234' '2718281'
[...]

PR #1031, opened 2024-01-31
src: lib: dialog_ui: Remove `LC_CTYPE` override for async loading screen spinner

davidbtadokoro:
The function `create_async_loading_screen_notification` creates a Dialog loading screen with a spinner to signal users that a delayed action is happening underneath. However, users reported environments that had problems with the characters used for the spinner while using `kw patch-hub` and running the test suite for `src/lib/dialog_ui.sh`.

The problem stems from the environment variable `LC_CTYPE` (used to determine character handling rules) being overridden to account for using the spinner characters that are Unicode (multibyte). The logic behind this is sound, as the encoding aspect should be considered because these spinner characters have a length of 3 bytes. Nevertheless, not handling this matter fixes the problem for the environments mentioned above and doesn't introduce regressions for the environments that didn't face the issues.

In this context, remove the override of `LC_CTYPE` and consider the spinner characters of length 1.

Closes: #965

### DIFF
--- a/src/lib/dialog_ui.sh
+++ b/src/lib/dialog_ui.sh
@@ -250,12 +250,10 @@ function create_loading_screen_notification()
 function spin_frame()
 {
   local frame_offset="$1"
-  local LC_CTYPE=C
   local spin='⣾⣽⣻⢿⡿⣟⣯⣷'
-  local char_width=3
 
   frame_offset=$(((frame_offset) % ${#spin}))
-  printf "%s" "${spin:$frame_offset:$char_width}"
+  printf "%s" "${spin:$frame_offset:1}"
 }
 
 # Create simple async loading screen notification for delayed actions.
@@ -293,7 +291,7 @@ function create_async_loading_screen_notification()
     # We should not use --clear because this flushes the infobox
     cmd+='dialog --colors'
     spin="$(spin_frame $frame_offset)"
-    frame_offset=$((frame_offset + 3))
+    frame_offset=$((frame_offset + 1))
 
     # Add Infobox screen
     cmd+=" --infobox $'${loading_message} ${spin}'"


davidbtadokoro:
@Andregnl, can you weigh in on the matter? Am I interpreting the intention behind the override of `LC_CTYPE` correctly? If my assumptions are correct, can you test this PR to see if it doesn't break anything on your end?

Andregnl:
Hi @davidbtadokoro, I tested this change on my end and everything seems to be in order. 

I looked a little more into the LC_CTYPE flag, and apparently it is used in scripts to set how a sequence of bytes of text is interpreted in different systems, but apparently mixing LC_CTYPE with other language settings may have unexpected results. So if the spin character is the same in other locales, the removal of the flag shouldn't be an issue. 

We may face a problem with how text is handled in different locales, in that case, the LC_ALL flag may be a better option since it overrides the other language settings for the script, but I am not sure if it is good practice to use it.

davidbtadokoro:
Hey @Andregnl, and thank you so much for the reply!

> So if the spin character is the same in other locales, the removal of the flag shouldn't be an issue.
> 
> We may face a problem with how text is handled in different locales, in that case, the LC_ALL flag may be a better option since it overrides the other language settings for the script, but I am not sure if it is good practice to use it.

That is the thing, indeed the best course of action is to handle this locale matter, as we shouldn't assume anything about the setup of users. However, overriding these locales (especially `LC_ALL`) may also have unpredictable outcomes...

I'm leaning towards assuring that ignoring the matter solves the problems for now, and look to handle it in case we face issues in the future.

@rodrigosiqueira, what is your take on all of this (recommend reading the PR description and @Andregnl, response for context).

rodrigosiqueira:
tbh, it looks good enough to me. If we have issues in the future, we will try to hand it.

Thanks

PR #1032, opened 2024-02-07
src: build: Remove redundant override of `load_build_config`

davidbtadokoro:
In kw, the `src/lib/kw_config_loader.sh` module takes care of configuration precedence for all the configurations, i.e., it composes configurations from the most preferred scope to the least one. Although the function `load_build_config` defined inside
`src/lib/kw_config_loader.sh` takes care of this behavior for the configurations of `kw build`, this function is also defined in `src/build.sh`, doing virtually the same as the former definition. This double definition causes the one from build to override the one in the config loader (config loader is sourced first).

Three problems arise from this situation:
 - Duplicated/redundant code.
 - The definition from build has a bug that outputs a warning to the user that doesn't makes sense.
 - It breaks the project's pattern of delegating any and all configuration loading to the `src/lib/kw_config_loader.sh` module.

Probably, `load_build_config` from build was created before the inception of the config loader and we forgot to remove it after adding the module. That may be the reason why the code is duplicated and has a bug that doesn't exist in the config loader version.

To address the three problems described above, remove the definition of `load_build_config` from `src/build.sh`.

Fixes: #824

Thanks-to: Rodrigo Siqueira \<Rodrigo.Siqueira@amd.com\>
Thanks-to: André Gustavo Nakagomi Lopez \<andregnl@usp.br\>

### DIFF
--- a/src/build.sh
+++ b/src/build.sh
@@ -247,50 +247,6 @@ function full_cleanup()
   cmd_manager "$flag" "$cmd"
 }
 
-# This function loads the kw build configuration files into memory, populating
-# the $build_config hashtable. The files are parsed in a specific order,
-# allowing higher level setting definitions to overwrite lower level ones.
-function load_build_config()
-{
-  if [[ -v build_config && "$OVERRIDE_BUILD_CONFIG" != 1 ]]; then
-    unset OVERRIDE_BUILD_CONFIG
-    return
-  fi
-
-  local -a config_dirs
-  local config_dirs_size
-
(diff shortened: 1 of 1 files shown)


davidbtadokoro:
@rodrigosiqueira and @Andregnl, I added a `Thanks-to` tag with you guys. Feel free to ask the change/removal of the tag.

rodrigosiqueira:
Thanks a lot for this PR!

PR #1033, opened 2024-02-10
Fix env switch bug with remote

rodrigosiqueira:


davidbtadokoro:
Hey @rodrigosiqueira, and thank you so much for the PR! I just adapted the calls to `assertTrue` to `assert_equals_helper`, and, apart from that LGTM, so merged into the unstable branch :+1:

PR #1034, opened 2024-02-11
Expand kw env list to show the path info

rodrigosiqueira:
Recently, we got some user feedback that they wish to know the env path to enable them to use other kernel custom commands. This commit addresses this issue by improving the kw env --list to also show the env path.

### DIFF
--- a/src/kw_env.sh
+++ b/src/kw_env.sh
@@ -293,6 +293,12 @@ function destroy_env()
   success "The \"${env_name}\" environment has been destroyed."
 }
 
+function no_env_message()
+{
+  say 'Kw did not find any environment. You can create a new one with the --create option.'
+  say 'See kw env --help'
+}
+
 # This function searches for any folder inside the .kw directory and considers
 # it an env.
 #
@@ -302,8 +308,7 @@ function list_env_available_envs()
 {
   local local_kw_configs="${PWD}/.kw"
   local current_env
-  local output
-  local ret
+  declare -a all_envs
 
   if [[ ! -d "$local_kw_configs" ]]; then
     complain 'It looks like that you did not setup kw in this repository.'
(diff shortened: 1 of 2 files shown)


src/kw_env.sh
```diff
@@ -303,30 +314,37 @@ function list_env_available_envs()
...
-  local ret
+  declare -a all_envs
 
   if [[ ! -d "$local_kw_configs" ]]; then
     complain 'It looks like that you did not setup kw in this repository.'
     complain 'For the first setup, take a look at: kw init --help'
     exit 22 # EINVAL
   fi
 
-  output=$(ls "${local_kw_configs}/${ENV_DIR}" 2> /dev/null)
```
davidbtadokoro, line 314: With this deletion, we can remove the local declaration of `output`.

src/kw_env.sh
```diff
@@ -303,30 +314,37 @@ function list_env_available_envs()
...
+    no_env_message
     return 0
   fi
 
   if [[ -f "${local_kw_configs}/${ENV_CURRENT_FILE}" ]]; then
     current_env=$(< "${local_kw_configs}/${ENV_CURRENT_FILE}")
-    say "Current env:"
-    printf ' -> %s\n' "$current_env"
+    say 'Current env:'
+    printf ' -> %s: %s\n\n' "$current_env" "${KW_CACHE_DIR}/${ENV_DIR}/${current_env}"
```
davidbtadokoro, line 339: When testing, the path `"${KW_CACHE_DIR}/${ENV_DIR}/${current_env}"`, points only to the directory where the environment `.config` is stored. FWIU, the env comprises the kernel `.config` plus the kw configs (stored at `.kw/envs`).

I get that the kw configs location may be trivial, as `kw env` doesn't have local scopes (i.e., the kw configs listed will always be inside `${PWD}/.kw/envs`), but is the kernel `.config` file sufficient? If so, ignore this comment.

Screenshot from 2024-02-12 20-24-09

src/kw_env.sh
```diff
@@ -303,30 +314,37 @@ function list_env_available_envs()
...
   fi
 
-  say 'All kw environments set for your local folder:'
-  printf '%s\n' "$output"
+  warning 'Other kw environments:'
+  # For the below loop, we want to split the array
+  # shellcheck disable=SC2068
+  for env in ${all_envs[@]}; do
+    printf ' * %s: %s\n' "$env" "${KW_CACHE_DIR}/${ENV_DIR}/${env}"
```
davidbtadokoro, line 346: Same question as the previous comment.

davidbtadokoro (changes requested):
Hi @rodrigosiqueira, and thanks for the change! I tested the PR, and it delivers what it proposes :+1: .

I left a few comments, so address them if you can.

src/kw_env.sh
```diff
@@ -33,6 +33,11 @@ function env_main()
     return "$?"
   fi
 
+  if [[ -n "${options_values['INFO']}" ]]; then
+    env_info
+    return "$?"
+  fi
+
```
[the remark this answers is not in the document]
rodrigosiqueira: This is a leftover from my first approach; I'll drop it.

src/kw_env.sh
rodrigosiqueira: From your update, you got the idea, but just to give you one context, I usually want to run the below command for running the unit test:

```
./tools/testing/kunit/kunit.py run --arch=x86_64 --build_dir=/home/siqueira/.cache/kw/envs/X86_64_KUNIT  --kunitconfig=drivers/gpu/drm/amd/display/test/kunit
```

However, I constantly forget the cache path. This improvement in the list feature will help me and others with that.

davidbtadokoro: Thanks for the context. It makes even clearer now :+1:

davidbtadokoro:
Hi @rodrigosiqueira, and thanks so much for the update! FWIU this is a great change for the users. Everything looks legit to me, so merged into the unstable branch :+1:

PR #1035, opened 2024-02-13
src: _kw: Revise `kw remote` Zsh completions

davidbtadokoro:
The Zsh completions for the `kw remote` feature are broken, because:

  1. The `--list`, `--add`, `--rename`, and `remove` aren't mutually exclusive, as they should.
  2. The `--add` and `--rename` options require 2 arguments, but the completions assume they don't need any and suggest other options.
  3. The `--remove` option requires 1 argument, but the completions assume it doesn't need any and suggest other options.
  4. The `--set-default|-s` option followed by `--add` shouldn't accept arguments, yet it does.

In this sense, fix problems 1 to 3 by declaring `--add`, `--rename`, and `--remove` as completions to `_arguments` instead of `_describe`, and fix 4 by altering the completion of `--set-default|-s`, if `--add` has been completed.

### DIFF
--- a/src/_kw
+++ b/src/_kw
@@ -438,20 +438,26 @@ _kw_pomodoro()
 
 _kw_remote()
 {
-  local -a remote_commands=(
-    '--add:add a named remote to kw management'
-    '--remove:remove an existing remote from kw management'
-    '--rename:rename an existing remote'
+  local -a set_default=()
+
+  set_default=(
+    '(-s --set-default --add --list --remove --rename)'{-s-,--set-default=}'[set existing remote as default]:remote-name: '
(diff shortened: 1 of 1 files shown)


davidbtadokoro:
@aquilamacedo, as you use Zsh, could you test this to see if the `kw remote` completions are fixed with this PR?

aquilamacedo:
@davidbtadokoro, everything is working fine now. Great fix; it worked well on my zsh, thx :)

davidbtadokoro:
@aquilamacedo, thank you so much for testing this PR. Change merged into the unstable branch.

PR #1037, opened 2024-02-14
src: lib: lore: Make query filter for Lore request more robust

davidbtadokoro:
When fetching patches from Lore, patch-hub composes a query URL that contain base filters along with additional filters. The base filters basically aims to filter out any message from the given mailing list that is not a patch. However, this query filter simply filters out messages that have the string 'Re:' in the subject.

In this context, make this query filter more robust by enforcing that the subject mustn't contain the string 're:', while also containing the strings 'patch' or 'rfc' (all case insensitive). The query filter is also constructed as an isolated logical expression to not conflict with potential added filters.

Note: Took the opportunity to anonymize a test information.

### DIFF
--- a/src/lib/lore.sh
+++ b/src/lib/lore.sh
@@ -361,11 +361,8 @@ function compose_lore_query_url_with_verification()
     return 22 # EINVAL
   fi
 
-  # TODO: We need to use the query prefix 's:Re:' to filter out replies and match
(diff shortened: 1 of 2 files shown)


rodrigosiqueira:
Very nice improvement! This will make the search way more robust :)

PR #1038, opened 2024-02-14
src: lib: lore: Fix empty message ID bug in `get_patchset_bookmark_st…

davidbtadokoro:
…atus`

The function `get_patchset_bookmark_status` is used by patch-hub UI to determine if given patchset is already in the Lore bookmarked database. However, because the function uses `grep` to check if the patchset message ID is present in the file that represents the bookmark database, an empty message ID returns 1, which means true to being in the database.

To fix this, check if the message ID passed as argument is empty and return 22 EINVAL in case it is.

Note: Took the opportunity to rename the argument name from `patchset_url` to `message_id`, as it is more precise.


rodrigosiqueira:
Another great PR :)

Issue #1028, opened 2024-01-24
`patch-hub`: Error in fetching patchsets after returning from 'Details and Actions' screen


davidbtadokoro:
**Description:**

In `patch-hub`, when fetching the latest patchsets either from a specific mailing list or from a string search, if any patchset is selected (which brings the 'Details and ACtions' screen), then return to the listing of patchsets, the next fetch will fail with an error `Couldn't fetch patchsets from <list> list.`

**How to Reproduce:**

1. Run patch-hub with `kw patch-hub`
2. Select any mailing list in the 'Registered Mailing Lists` menu
3. When the list is loaded, select any patchset with the 'OK' button, then return to the listing with the 'Return' button
4. In the listing screen, use the 'Next' button until you get the error mentioned.

**Expected behavior:**

It is fundamental for users to be able to go back and forth in the listing without the fetching breaking.

**System information:**

 - OS: Debian Bookworm
 - kw Version: `alpha Branch: unstable Commit: 999f5d3`
 - Bash Version 5.2.15

**Additional context:**

The problem is most likely due to the implementation of the 'Search String in Lore' feature, in which we allowed additional filters to be passed when fetching latest patchsets. This parameter is passed through the variable `screen_sequence['SHOW_SCREEN_PARAMETER']` (see [here](https://github.com/kworkflow/kworkflow/blob/999f5d346fb1ed2c266d8a5e628fd85ec66e553a/src/ui/patch_hub/search_string_in_lore.sh#L47)). This variable is also used to pass the patchset data to be displayed in the 'Details and Actions' screen (see [here](https://github.com/kworkflow/kworkflow/blob/999f5d346fb1ed2c266d8a5e628fd85ec66e553a/src/ui/patch_hub/latest_patchsets_from_mailing_list.sh#L51)). So, in the 'Details and Actions' screen, the value of `screen_sequence['SHOW_SCREEN_PARAMETER']=<patchset_data>`, and because this value isn't reset when we return from this screen (see [here](https://github.com/kworkflow/kworkflow/blob/999f5d346fb1ed2c266d8a5e628fd85ec66e553a/src/ui/patch_hub/patchset_details_and_actions.sh#L53)), patch-hub model creates an invalid query URL to Lore assuming `screen_sequence['SHOW_SCREEN_PARAMETER']` contains additional filters.

PR #1039, opened 2024-02-14
src: ui: patch_hub: patchset_details_and_actions: Fix returning from 'Details and Actions' bug

davidbtadokoro:
[What]

With the implementation of 'Search String in Lore', the function `show_latest_patchsets_from_mailing_list` accepts an argument for additional filters to be passed in the Lore request. This argument is passed via the `screen_sequence` data structure used to pass information between the screens that compose patch-hub UI. However, because in the screen 'Details and Actions' the
`screen_sequence['SHOW_SCREEN_PARAMETER']` contains data about a patchset (not an additional filter) and because this value isn't reset when returning from this screen, `show_latest_patchsets_from_mailing_list` makes an invalid request that causes the fetch to fail, when it shouldn't.

In this context, make the screen 'Details and Actions' reset the value of `screen_sequence['SHOW_SCREEN_PARAMETER']` when returning and adapt other parts to not also break the 'Search String in Lore' feature.

[Why]

Undoubtedly, users will go back and forth between the listing of latest patchsets and a patchset details and actions, so this bug is a major flaw in patch-hub.

[How]

The fix goes by simply resetting the value of
`screen_sequence['SHOW_SCREEN_PARAMETER']` in the function `show_patchset_details_and_actions` when the user hits the button 'Return'. However, because 'Search String in Lore' also uses this value to work, just this reset breaks this feature. So, besides the reset, we add a global variable to patch-hub UI declared in `patch_hub_core.sh` to pass additional filters instead of using
`screen_sequence['SHOW_SCREEN_PARAMETER']` for this.

Closes: #1028

### DIFF
--- a/src/ui/patch_hub/latest_patchsets_from_mailing_list.sh
+++ b/src/ui/patch_hub/latest_patchsets_from_mailing_list.sh
@@ -7,7 +7,6 @@ declare -ga formatted_patchsets_list
 # These patchsets are ordered by their recieved time in the lore.kernel.org servers.
 function show_latest_patchsets_from_mailing_list()
 {
-  local additional_filters="$1"
   local starting_index
   local ending_index
   local box_title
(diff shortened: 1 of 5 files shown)


rodrigosiqueira:
Very nice change!

Issue #1040, opened 2024-02-15
Integration tests error in container_exec with pipes inside


aquilamacedo:
**Description:**

The current implementation of the container_exec function encounters a critical issue when executing commands within a container involving pipes (|). Specifically, Podman interprets the pipe as a command for its internal use, resulting in unexpected behavior and hindering the proper execution of commands within the container, consequently leading to errors.

![image](https://github.com/kworkflow/kworkflow/assets/43859386/c2647a92-c9c3-4575-b03b-f0c3e99318f2)

**How to Reproduce:**

1. Run a command with a pipe inside a container, mirroring the implementation of the container_exec function. For instance:

```bash
# Where `ls | kw --version` is the command intended to be executed inside the container.
podman container exec kw-archlinux ls | kw --version
```

2. Observe the ensuing error or unexpected behavior caused by the misinterpretation of the pipe character.

**Expected Behavior:**
The `container_exec` function should handle command-line arguments safely, avoiding potential word splitting issues when executing commands within a specified container. Any errors during container command execution should be appropriately reported.

**System information:**

 - OS: Debian GNU/Linux 12 (bookworm)
 - kw Version: alpha, Branch: unstable, Commit: aa867af

 - zsh Version zsh 5.9 (x86_64-debian-linux-gnu)

aquilamacedo:
The theme's name is "Oxocarbon" by NVChad. ngl, like this theme too :)

PR #1041, opened 2024-02-15
tests: integration: utils: fix container_exec for enhanced command execution

aquilamacedo:
This adjustment, including proper handling of options, container names, and the command to execute with special consideration for the `/bin/sh -c` option, promotes a more reliable and secure execution process. The commit aims to enhance the overall functionality and robustness of the container_exec function when dealing with various command scenarios.

Closes: #1040


tests/integration/utils.sh
```diff
@@ -168,10 +167,24 @@ function container_run()
   fi
 }
 
+# Execute a command within a container.
+#
+# @container_options   Additional options for the 'podman container exec' command.
```
davidbtadokoro, line 172: Maybe rename this to something that makes it clearer that the options are specific to `podman container exec`.

tests/integration/utils.sh
```diff
@@ -168,10 +167,24 @@ function container_run()
...
+# @container_name      The name or ID of the target container.
+# @container_command   The command to be executed within the container.
 function container_exec()
 {
-  # shellcheck disable=SC2068
-  podman container exec $@ 2> /dev/null
+  local container_options="$1"
+  local container_name="$2"
+  local container_command="$3"
```
davidbtadokoro, line 179: I think that if we reorder the arguments to

1. `container_name`
2. `container_command`
3. `container_options`

as `container_options` is optional and will be less used, the callers won't have to pass `''` when not using the `podman container exec` options, making the signature of the function cleaner (?)

I see this as a style decision, so feel free to stick to the way it is if you think it is better :grimacing:

davidbtadokoro (changes requested):
Hey @aquilamacedo, and thanks for this PR! Conceptually, the logic is sound and makes `container_exec` more robust, and I just left some style comments. Also, I reran the integration tests suite, which doesn't seem to add any regressions.

However, I couldn't reproduce issue #1040. I revived a container from the tests and ran the command you suggested, but I wasn't met with any errors. Am I doing something wrong?

Screenshot from 2024-02-16 12-36-54

**Note:** This doesn't invalidate your change.

aquilamacedo:
> Hey @aquilamacedo, and thanks for this PR! Conceptually, the logic is sound and makes `container_exec` more robust, and I just left some style comments. Also, I reran the integration tests suite, which doesn't seem to add any regressions.
> 
> However, I couldn't reproduce issue #1040. I revived a container from the tests and ran the command you suggested, but I wasn't met with any errors. Am I doing something wrong?
> 
> Screenshot from 2024-02-16 12-36-54
> 
> **Note:** This doesn't invalidate your change.
Hi @davidbtadokoro, first of all, thanks for your review.

So, in this case, an unexpected output occurs because the ls && kw --version command is supposed to run inside the container and display the results of both ls and kw --version. However, if you observe closely, the kw --version command is being executed on the host machine rather than within the container. This is problematic as the pipe operation in this case is affecting the Podman command.

I made this modification to address potential issues when a kw command requires interaction. This format of command, such as printf "y\n" | kw --self-update -u, can be particularly useful in scenarios where a kw command necessitates a response. Therefore, this adjustment is valuable for ensuring proper command execution within the container and enhancing compatibility with interactive kw commands in the future.

tests/integration/utils.sh
aquilamacedo: I just did this :)

davidbtadokoro:
> Hi @davidbtadokoro, first of all, thanks for your review.
> 
> So, in this case, an unexpected output occurs because the `ls | kw --version` command is supposed to run inside the container and display the results of both `ls` and `kw --version`. However, if you observe closely, the `kw --version` command is being executed on the host machine rather than within the container. This is problematic as the pipe operation in this case is affecting the Podman command.
> 
> I made this modification to address potential issues when a kw command requires interaction. This format of command, such as `printf "y\n" | kw --self-update -u`, can be particularly useful in scenarios where a kw command necessitates a response. Therefore, this adjustment is valuable for ensuring proper command execution within the container and enhancing compatibility with interactive kw commands in the future.

Oh, I see it now! You are right, sorry for the mess :sweat_smile: 
Going to review it again (and probably merge it) now!

davidbtadokoro:
Hi @aquilamacedo, and thanks again for the update!

Really sorry for the delay. I was validating your PR through the wrong way, because I was creating a container from a kw image and simply running `podman container exec kw-cont /bin/sh -c 'ls | kw --version'`, at which point I was met with a `/bin/sh: line 1: kw: command not found` 😅

Gibberish aside, great change! Merged into the unstable branch 👍

Issue #1026, opened 2024-01-22
Remove `kw vars`


davidbtadokoro:
**Description:**

With the implementation of `kw config --show`, `kw vars` is deprecated.

**Type:**

Refactor

PR #1042, opened 2024-02-17
remove kw vars

MarceloSpessoto:
This PR deletes the `kw vars` functionality as requested by #1026.
Commit `f7898dc` removes kw vars itself and its documentation. It also adjusts some pieces of documentation to recommend `kw config --show` instead of `kw vars`.
Commit `6d8ee59` removes show_variables_main function and its tests, since the function was only used by `kw vars`.


src/_kw
```diff
@@ -406,8 +405,7 @@ _kw_man()
   _values 'kw commands' \
     'backup' 'build' 'codestyle' 'config' 'debug' 'deploy' 'device' 'diff' 'drm' 'env' \
     'explore' 'init' 'mail' 'maintainers' 'pomodoro' 'remote' 'report' 'self-update' 'ssh' \
-    'patch-hub' 'vars'
-}
```
davidbtadokoro, line 410: I think you deleted this curly bracket by mistake 😅. Doing this breaks the Zsh completions, so please put them back.

src/lib/kw_config_loader.sh
```diff
@@ -56,103 +56,6 @@ declare -gA lore_config_local
 # Default target option from kworkflow.config
 declare -gA deploy_target_opt=(['local']=2 ['remote']=3)
```
davidbtadokoro, line 58: Don't forget to also remove the function `vars_help` defined in this file.

davidbtadokoro (changes requested):
Hi @MarceloSpessoto, and thank you so much for the change! kw vars has been deprecated for a time now, so nice change 👍 

Left two in-line comments so please see if you can address them. Also, two important comments that don't fit the in-line review:

- I think you can squash both commits into one. I interpret that the change is the removal of the `kw vars` as a whole, and the way you splitted it as (1) remove docs plus the user interface to the feature, then (2) remove its inner implementation, can be more atomic.
- Revise the commit message. The commit message subject should follow the pattern `path: to: file: Do foo in bar`, and the body should the describe the motivation to the change (why) and the change (what). For reference see [here](https://kworkflow.org/content/howtocontribute.html#commit-messages).

**Obs.:** as this is a big change in number of files involved, it may be difficult to see a clear `path: to: file`. In these cases, use the main file that represents the change. Maybe the `kw` executable at the root of the repo, as the removal in it "represents" the removal of the feature (?)

MarceloSpessoto:
Hey @davidbtadokoro , I've made the requested changes!

davidbtadokoro:
Hey @MarceloSpessoto, and thank you so much for the update! Everything looks legit, so merged into the unstable branch 👍

Issue #1015, opened 2024-01-11
Integration tests don't pass due to system language


jppaulo06:
**Description:**

Running integration tests on unstable branch, not all the tests pass for me.
It's seems like it's because of how my system prints float numbers.
These are the outputs of the failing tests:

```
test_device_archlinux
ASSERT:(55): kw device failed for archlinux expected:<Chassis:
  Type: Desktop
CPU:
  Model: AMD
  Max frequency (MHz): 4208.2031
  Min frequency (MHz): 2200.0000
RAM:
  Total RAM: 16G
Storage devices:
Operating System:
Motherboard:
  Vendor: ASUSTeK COMPUTER INC.
  Name: PRIME B550M-A AC
GPU:
  Model: ASUSTeK Computer Inc. Navi 22 [Radeon RX 6700/6700 XT/6750 XT / 6800M/6850M XT]
  Provider: Advanced Micro Devices, Inc. > but was:<Chassis:
  Type: Desktop
CPU:
  Model: AMD
  Max frequency (MHz): 4208,2031
  Min frequency (MHz): 2200,0000
RAM:
  Total RAM: 16G
Storage devices:
Operating System:
Motherboard:
  Vendor: ASUSTeK COMPUTER INC.
  Name: PRIME B550M-A AC
GPU:
  Model: ASUSTeK Computer Inc. Device 05d7
  Provider: Advanced Micro Devices, Inc. >
test_device_debian
ASSERT:(55): kw device failed for debian expected:<Chassis:
  Type: Desktop
CPU:
  Model: AMD
  Max frequency (MHz): 4208.2031
  Min frequency (MHz): 2200.0000
RAM:
  Total RAM: 16G
Storage devices:
Operating System:
Motherboard:
  Vendor: ASUSTeK COMPUTER INC.
  Name: PRIME B550M-A AC
GPU:
  Model: ASUSTeK Computer Inc. Device 05d7
  Provider: Advanced Micro Devices, Inc. > but was:<Chassis:
  Type: Desktop
CPU:
  Model: AMD
  Max frequency (MHz): 4208,2031
  Min frequency (MHz): 2200,0000
RAM:
  Total RAM: 16G
Storage devices:
Operating System:
Motherboard:
  Vendor: ASUSTeK COMPUTER INC.
  Name: PRIME B550M-A AC
GPU:
  Model: ASUSTeK Computer Inc. Device 05d7
  Provider: Advanced Micro Devices, Inc. >
[...]

PR #1044, opened 2024-02-20
create integration_tests.yml

aquilamacedo:
This commit creates a workflow in GitHub Actions for integration tests

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -1,0 +1,36 @@
+name: Integration tests
+on:
+  [push, pull_request]
+
+jobs:
+  initialize-repository:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+
+  integration-tests:
+    needs: initialize-repository
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+
+      - name: Update the system
+        run: sudo apt update -y
+
+      - name: Install dependencies
+        run: |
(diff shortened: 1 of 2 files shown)


davidbtadokoro (changes requested):
Hi @aquilamacedo, and thank you so much for implementing this important workflow in the CI!

I reviewed the change and left two simple comments. The workflow executed correctly in GitHub Actions, so I think this validates the PR.

Two points that I think are worth addressing:

- In the commit message subject, I think the prefix `.github: workflows: integration_tests:` fits (run `git log .github`, for reference);
- I see that the workflow failed because of a known issue in the `kw device` integration test. So, see if you can address this issue in another commit, or comment out the failing tests in this commit, without forgetting to describe this in the commit message body and creating a issue for it (issue #1015 is related but is a different problem). Either approach is fine by me 😬

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 3
```
rodrigosiqueira, line 16: I don't remember why do we need `fetch-depth: 3`. Why not 1? Am I missing something?

rodrigosiqueira: @aquilamacedo any comment here?

aquilamacedo: Upon closer examination of this option, I now think that a fetch depth of 1 would be more appropriate. There is no necessity to delve into the historical data for these particular operations.

rodrigosiqueira: @aquilamacedo please update your PR.

rodrigosiqueira:
Hi @aquilamacedo , thanks a lot for this change. I just merged it in the unstable branch.

PR #1045, opened 2024-02-24
src: remote: Improve ssh handling in case of failure

rodrigosiqueira:
As part of the kernel workflow, it is common for a developer to have multiple hard disks with different distros for checking the kernel behavior in those scenarios. When the developer swaps the disk but the IP is still the same for the target machine, it is expected to hit the remote host identification issue. This commit alleviates this problem by detecting and taking action based on this type of failure.

### DIFF
--- a/src/lib/remote.sh
+++ b/src/lib/remote.sh
@@ -15,6 +15,24 @@ REMOTE_KW_DEPLOY='/root/kw_deploy'
 
 declare -gA remote_parameters
 
+# This function checks if connecting to the host machine via ssh is possible.
+# It is helpful to invoke this function before performing any meaningful
+# operation that depends on the ssh connection.
+#
+# @flag: How to display a command, the default value is
+#   "HIGHLIGHT_CMD". For more options see `src/lib/kwlib.sh` function `cmd_manager`
+# @remote: IP or domain name.
+# @port: TCP Port. Default value is 22.
+# @user: User in the host machine. Default value is "root"
+# @remote_file: Path to the remote file
+# @remote_file_host: Hostname in the ssh file
+#
+# Returns:
+# In case of success, this function returns 0, otherwise, it can return:
+# 2   - It did not find the remote_file
+# 101 - Network is not reachable
+# 125 - The operation was canceled
+# 255 - Unknown error
 function is_ssh_connection_configured()
 {
   local flag=${1:-'SILENT'}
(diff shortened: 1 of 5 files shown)


src/lib/remote.sh
```diff
@@ -40,6 +42,132 @@ function is_ssh_connection_configured()
...
+
+  warning 'kw was not able to ssh into:'
+  warning " Host: ${remote_parameters['REMOTE_IP']}"
+  warning " Port: ${remote_parameters['REMOTE_PORT']}"
+  warning " User: ${remote_parameters['REMOTE_USER']}"
+  warning 'Due to a remote host identification issue.'
+  if [[ $(ask_yN 'Do you want kw to remove the current host from the know_hosts file?') =~ '0' ]]; then
+    say ' No problem, the operation have been canceled.'
+    say ' You can try to manually check the kw issue by using the below command:'
+    say "  ${command}"
```
davidbtadokoro, line 120: Currently, the value of `command` passed as an argument to `remove_key_from_known_hosts` is the `ssh_cmd` value from `is_ssh_connection_configured`, which means that the message suggests the user should run the command with the `-q` flag, which outputs nothing. To avoid this, either pass `drop_quiet_from_ssh_command` instead of `sshd_cmd` or remove the `-q` flag inside this function.

src/lib/remote.sh
```diff
@@ -40,6 +42,132 @@ function is_ssh_connection_configured()
...
+    return 125 # ECANCELED
+  fi
+
+  remove_key_cmd+=" '${HOME}/.ssh/known_hosts' -R '${remote_parameters['REMOTE_IP']}'"
```
davidbtadokoro, line 124: Most importantly, you should also consider the remote port value in case the port is not the default 22. So make this line like this:
```bash
remove_key_cmd+=" '${HOME}/.ssh/known_hosts' -R '[${remote_parameters['REMOTE_IP']}]:${remote_parameters['REMOTE_PORT']}'"
```

If we don't consider the port, we may face the following error:

Screenshot from 2024-02-29 19-42-17

**Obs.:** Tested with this "fix", and it solves the issue.

davidbtadokoro (changes requested):
Hey, @rodrigosiqueira, and thank you so much for this change! It solves what it seems to be an annoying problem :grimacing: 

I left some comments, but most of them are about coding style. However, pay close attention to the comment with the screenshot about adding the port value when removing the host in `known_hosts`.

Aside from that, I tested the PR, and it does what it proposes to do flawlessly :+1:

src/lib/remote.sh
```diff
@@ -40,6 +42,132 @@ function is_ssh_connection_configured()
...
+    ((line_number--))
+    trimmed_remote_file=$(printf '%s' "$trimmed_remote_file" | head --lines "$line_number")
+  fi
+
+  # Here, we are matching lines with Hostname/Port/User occurrences that happen between the target
+  # host line and the next host line (if it exists) in the remote config file and only have leading
+  # spaces (excludes commented lines). We then cut the resulting string to get only the value of
+  # the Hostname/Port/User line.
+  line=$(printf '%s' "$trimmed_remote_file" | grep --max-count=1 '^[[:blank:]]*Hostname')
+  remote=$(printf '%s' "$line" | rev | cut --delimiter=' ' -f1 | rev)
```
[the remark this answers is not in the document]
davidbtadokoro: I think you are right, as I guess that rev optimizes the reversion by just printing in the inverse order :+1:

davidbtadokoro:
Hi @rodrigosiqueira, and thank you for the great update! Made some really minimal adjustments, but it looks legit to me, so merged into the unstable branch :+1: . Also, sorry for the delay.

PR #1055, opened 2024-03-07
tests: integration: self_update_test: add self-update test

aquilamacedo:
Introduce integration tests for self-update with
the --unstable option and regular self-update.

### DIFF
--- a/tests/integration/self_update_test.sh
+++ b/tests/integration/self_update_test.sh
@@ -1,0 +1,247 @@
+#!/bin/bash
+
+include './src/lib/kwio.sh'
+include './src/lib/kwlib.sh'
+include './tests/unit/utils.sh'
+include './tests/integration/utils.sh'
+
+function oneTimeSetUp()
+{
+  local kw_repo_url='https://github.com/kworkflow/kworkflow'
+  local ret
+  declare -ar KW_BRANCHES=('master' 'unstable')
+  declare -g HOST_TEMP_DIR
+  declare -g HOST_KW_CLONED_PATH
+
+  HOST_TEMP_DIR=$(mktemp --directory)
+  HOST_KW_CLONED_PATH="${HOST_TEMP_DIR}/kworkflow"
+
+  # Clone the repository once on the host
+  git clone --quiet "$kw_repo_url" "$HOST_KW_CLONED_PATH"
+  ret="$?"
+  if [[ "$ret" -ne 0 ]]; then
+    complain "Failed to clone ${kw_repo_url}"
+    if [[ -n "$HOST_KW_CLONED_PATH" ]]; then
+      if is_safe_path_to_remove "$HOST_TEMP_DIR"; then
+        rm --recursive --force "$HOST_TEMP_DIR"
+      else
+        complain "Unsafe path: ${HOST_TEMP_DIR} - Not removing"
+      fi
+    fi
+    exit "$ret"
+  fi
+
+  # The VERBOSE variable is set and exported in the run_tests.sh script based
+  # on the command-line options provided by the user. It controls the verbosity
+  # of the output during the test runs.
+  setup_container_environment "$VERBOSE" 'self-update'
+}
+
+function oneTimeTearDown()
+{
+  # Check if the path is safe to remove
+  if is_safe_path_to_remove "$HOST_TEMP_DIR"; then
+    rm --recursive --force "$HOST_TEMP_DIR"
+  fi
+}
+
+# Function to set up an outdated kworkflow environment in a container
+#
+# @container: The name of the container.
+# @kw_cloned_container_path: Temporary directory inside the container where kworkflow is cloned.
+# @kw_branch: The branch of kworkflow to be used.
+# @commit_point: The specific commit to be checked out.
+function setup_outdated_kworkflow_environment()
+{
+  local container="$1"
+  local kw_cloned_container_path="$2"
+  local kw_branch="$3"
+  local commit_point="$4"
+  local ret
+
+  # Check if container and kw_cloned_container_path are non-empty
+  if [[ -z "$container" ]]; then
+    complain 'Container name is not provided'
+    return 1
+  fi
+
+  if [[ -z "$kw_cloned_container_path" ]]; then
+    complain 'kworkflow cloned directory path inside the container is not provided'
+    return 1
+  fi
+
+  # Copy the repository into the container
+  container_copy "$container" "$HOST_KW_CLONED_PATH" "$kw_cloned_container_path"
+  ret="$?"
+  if [[ "$ret" -ne 0 ]]; then
+    complain "Failed to copy kworkflow to container $container"
+    return "$ret"
+  fi
+
+  container_exec "$container" "git -C ${kw_cloned_container_path} checkout remotes/origin/${kw_branch}"
+  ret="$?"
+  if [[ "$ret" -ne 0 ]]; then
+    complain 'Git checkout operation failed'
+    complain "git -C ${kw_cloned_container_path} checkout remotes/origin/${kw_branch}"
+    return "$ret"
+  fi
+
+  container_exec "$container" "git -C ${kw_cloned_container_path} checkout ${commit_point}"
+  ret="$?"
+  if [[ "$ret" -ne 0 ]]; then
+    complain 'Git checkout to specific commit failed'
+    complain "git -C ${kw_cloned_container_path} checkout ${commit_point}"
+    return "$ret"
+  fi
+
+  container_exec "$container" "cd ${kw_cloned_container_path} && ./setup.sh --install --force --skip-checks > /dev/null 2>&1"
+  ret="$?"
+  if [[ "$ret" -ne 0 ]]; then
+    complain 'kw setup failed'
+    complain './setup.sh --install --force --skip-checks --skip-docs'
+    return "$ret"
+  fi
+}
+
+# This function gets the commit hash and base version of the specified
+# branch, constructs the expected output, updates kworkflow based on
+# the branch, and then checks the version.
+#
+# @container: The name of the container.
+# @kw_cloned_container_path: Temporary directory inside the container where kworkflow is cloned
+# @kw_branch: The branch of kworkflow to be used.
+#
+# Return:
+# Return 0: On success. Prints a formatted string that includes the branch name,
+# the commit hash of the branch, and the base version of the branch.
+function generate_formatted_branch_info()
+{
+  local container="$1"
+  local kw_cloned_container_path="$2"
+  local kw_branch="$3"
+  local head_commit_hash
+  local base_version
+  local ret
+
+  head_commit_hash=$(container_exec "$container" "git -C ${kw_cloned_container_path} rev-parse --short origin/${kw_branch}")
+  ret="$?"
+  if [[ "$ret" -ne 0 ]]; then
+    complain 'Git checkout to specific commit failed'
+    complain "git -C ${kw_cloned_container_path} checkout ${commit_point}"
+    return "$ret"
+  fi
+
+  base_version=$(container_exec "$container" "git -C ${kw_cloned_container_path} show remotes/origin/${kw_branch}:./src/VERSION | head --lines 1")
+  ret="$?"
+  if [[ "$ret" -ne 0 ]]; then
+    complain 'Git checkout to specific commit failed'
+    complain "git -C ${kw_cloned_container_path} checkout ${commit_point}"
+    return "$ret"
+  fi
+
+  printf '%s\nBranch: %s\nCommit: %s\n' "$base_version" "$kw_branch" "$head_commit_hash"
+}
+
+# This function tests the self-update functionality of kworkflow for the master
+# branch.
+function test_self_update_from_master_branch()
+{
+  local kw_cloned_container_path
+  local container
+  local distro
+  local expected_output
+  local output
+
+  for distro in "${DISTROS[@]}"; do
+    container="kw-${distro}"
+    kw_cloned_container_path="$(generate_temporary_directory_in_container "$container")/kworkflow"
+
+    container_copy "$container" "$HOST_KW_CLONED_PATH" "$kw_cloned_container_path"
+
+    expected_output=$(generate_formatted_branch_info "$container" "$kw_cloned_container_path" 'master')
+
+    # Specific update command for master branch
+    container_exec "$container" 'yes | bash ./kw self-update > /dev/null'
+
+    output=$(container_exec "$container" 'kw --version')
+    assert_equals_helper "kw self-update master failed for ${distro}" "$LINENO" "$expected_output" "$output"
+  done
+}
+
+# This function tests the self-update functionality of kworkflow for the
+# unstable branch.
+function test_self_update_from_unstable_branch()
+{
+  local kw_cloned_container_path
+  local container
+  local distro
+  local expected_output
+  local output
+
+  for distro in "${DISTROS[@]}"; do
+    container="kw-${distro}"
+    kw_cloned_container_path="$(generate_temporary_directory_in_container "$container")/kworkflow"
+
+    container_copy "$container" "$HOST_KW_CLONED_PATH" "$kw_cloned_container_path"
+
+    expected_output=$(generate_formatted_branch_info "$container" "$kw_cloned_container_path" 'unstable')
+
+    # Specific update command for unstable branch
+    container_exec "$container" "yes | bash ./kw self-update --unstable > /dev/null"
+
+    output=$(container_exec "$container" 'kw --version')
+    assert_equals_helper "kw self-update unstable failed for ${distro}" "$LINENO" "$expected_output" "$output"
+  done
+}
+
+# This function tests the self-update functionality of kworkflow for
+# the master branch with an old version of kworkflow.
+function test_self_update_with_kw_old_version_of_master_branch()
+{
+  local kw_commit_point='HEAD~4'
+  local kw_cloned_container_path
+  local container
+  local distro
+
+  for distro in "${DISTROS[@]}"; do
+    container="kw-${distro}"
+    kw_cloned_container_path="$(generate_temporary_directory_in_container "$container")/kworkflow"
+
+    # Configuring the kworkflow to test an old version of the master branch
+    setup_outdated_kworkflow_environment "$container" "$kw_cloned_container_path" 'master' "$kw_commit_point"
+
+    expected_output=$(generate_formatted_branch_info "$container" "$kw_cloned_container_path" 'master')
+    container_exec "$container" 'yes | bash ./kw self-update > /dev/null'
+
+    output=$(container_exec "$container" 'kw --version')
+    assert_equals_helper "kw self-update master failed for ${distro}" "$LINENO" "$expected_output" "$output"
+  done
+}
+
+# This function tests the self-update functionality of kworkflow for
+# the unstable branch with an old version of kworkflow.
+function test_self_update_with_kw_old_version_of_unstable_branch()
+{
+  local kw_commit_point='HEAD~3'
+  local kw_cloned_container_path
+  local container
+  local distro
+
+  for distro in "${DISTROS[@]}"; do
+    container="kw-${distro}"
+    kw_cloned_container_path="$(generate_temporary_directory_in_container "$container")/kworkflow"
+
+    # Configuring the kworkflow to test an old version of the unstable branch
+    setup_outdated_kworkflow_environment "$container" "$kw_cloned_container_path" 'unstable' "$kw_commit_point"
+
+    expected_output=$(generate_formatted_branch_info "$container" "$kw_cloned_container_path" 'unstable')
+
+    # Specific update command for master branch
+    container_exec "$container" "yes | bash ./kw self-update --unstable > /dev/null"
+
+    output=$(container_exec "$container" 'kw --version')
+    assert_equals_helper "kw self-update unstable failed for ${distro}" "$LINENO" "$expected_output" "$output"
+  done
+}
+
+invoke_shunit
--- a/tests/integration/utils.sh
+++ b/tests/integration/utils.sh
@@ -130,7 +130,7 @@ function setup_container_environment()
 
     # Container images already have kw installed. Install it again, overwriting
     # the installation.
-    container_exec "${container_name}" './setup.sh --install --force --skip-checks --skip-docs > /dev/null 2>&1'
+    container_exec "${container_name}" './setup.sh --install --force --skip-docs > /dev/null 2>&1'
 
     if [[ "$?" -ne 0 ]]; then
       fail "(${LINENO}): Failed to install kw in the container ${container_name}"
--- a/tests/integration/utils.sh
+++ b/tests/integration/utils.sh
@@ -453,3 +453,19 @@ function container_inspect()
     fail "(${LINENO}): Failed to inspect the container."
   fi
 }
+
+# This function generates a temporary directory inside a specified container.
+#
+# @container  The name of the container.
+function generate_temporary_directory_in_container()
+{
+  local container="$1"
+  local ret
+
+  container_exec "$container" 'mktemp --directory'
+  ret="$?"
+  if [ "$ret" -ne 0 ]; then
+    fail 'Failed to create a temporary directory inside the container'
+    return "$ret"
+  fi
+}
--- a/documentation/dependencies/arch.dependencies
+++ b/documentation/dependencies/arch.dependencies
@@ -29,3 +29,4 @@ b4
 procps-ng
 pciutils
 libnotify
+python-sphinx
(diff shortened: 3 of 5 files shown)


aquilamacedo:
TODO:

- [ ] Hide the output (it's printing the entire self-update process).

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,61 @@
...
+function test_self_update_master_helper()
+{
+  self_update_test_helper 'master'
+}
+
+function test_self_update_unstable_helper()
+{
+  self_update_test_helper 'unstable'
+}
+
```
rodrigosiqueira, line 60: One test that I want to see:
1. Manually install kw in the container. Check if everything is all right.
2. Run an update to master (do the same for the unstable) and check if everything looks all right.

Do a similar test, but in this case, intentionally install an old version of the kw (maybe three commits old, you can force it via git) and ensure the update works as expected.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,61 @@
...
+  head_commit_hash=$(git rev-parse --short origin/"${kw_branch}")
+
+  # Get the base version from the specified branch
+  base_version=$(git show "${kw_branch}":./src/VERSION | head --lines 1)
+
+  # Format the expected output
+  expected_output=$(printf '%s\nBranch: %s\nCommit: %s' "$base_version" "$kw_branch" "$head_commit_hash")
+
+  for distro in "${DISTROS[@]}"; do
+    say "Running tests on ${distro} environment..."
```
davidbtadokoro, line 30: I think this is unnecessary as it is outputted whether the test fails or succeeds. IMHO, the distro info is only relevant if the test fails (which is displayed by the assertion).

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,61 @@
...
+    container="kw-${distro}"
+
+    case "$update_type" in
+      'master')
+        container_exec "${container}" 'yes | kw self-update'
```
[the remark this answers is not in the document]
davidbtadokoro: Also, redirect the output of this command to `/dev/null` so as not to pollute the test suite output.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,61 @@
...
+        ;;
+      'unstable')
+        container_exec "${container}" 'yes | kw self-update --unstable'
```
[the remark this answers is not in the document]
davidbtadokoro: Also, redirect the output of this command to `/dev/null` so as not to pollute the test suite output.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,61 @@
...
+        ;;
+    esac
+
+    output=$(container_exec "${container}" 'kw --version')
+
+    assertEquals "(${LINENO}): kw version failed for ${distro}" "$expected_output" "$output"
+  done
+}
+
+function test_self_update_master_helper()
```
davidbtadokoro, line 51: Remove the suffix `_helper` here and in the next test function.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,61 @@
...
+  local kw_branch="${update_type}"
+  local base_version
+  local distro
+  local container
+  local output
+  local expected_output
+  local head_commit_hash
+
+  # Get the short commit hash of the branch
+  head_commit_hash=$(git rev-parse --short origin/"${kw_branch}")
```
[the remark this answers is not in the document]
davidbtadokoro: Also, this git command is run in the host repository (i.e., the developer's actual kw repo), in which we can't assume too much. For example, my kw fork is labeled `davidbtadokoro-fork` and the kw upstream repo `upstream`, so when I run the tests, I encounter the error `fatal: Needed a single revision` and the `Commit` field of the expected output gets blank in turn (see screenshot below).

Screenshot from 2024-03-10 19-18-31

**Note:** The test output differs from this PR version because I cleaned the extra output and used `assert_equals_helper` for better visualization.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,61 @@
...
+
+  # Get the base version from the specified branch
+  base_version=$(git show "${kw_branch}":./src/VERSION | head --lines 1)
```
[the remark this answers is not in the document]
davidbtadokoro: Like in the git command above, we assume some things about the dev's local kw repo. For example, I didn't have the `master` branch update with the upstream, and `src/VERSION` still contains `alpha` as the base version in this local branch. This causes the expected base version to be wrong (see screenshot below).

Screenshot from 2024-03-10 19-23-35

**Note:** As in the previous screenshot, the output differs because I cleaned the extra output and used `assert_equals_helper`. Also, I "fixed" the problem pointed in the screenshot to make sure that the test failed strictly due to the expected base version being wrong.

davidbtadokoro (changes requested):
Hi @aquilamacedo, and thank you so much for this change!

I left some comments, so please see if you can address them (ping me if you have any questions). The most critical comments are about the verbose output that pollutes the test suite and the weak assumptions that the git commands make (the ones with screenshots).

**UPDATE:** I see now that you made a task to hide the output. Nevertheless, don't forget to do it!

aquilamacedo:
Hi @davidbtadokoro. Could you check if everything is in order?

I addressed all your suggestions, thx :)

Now I'm working on @rodrigosiqueira  suggestions

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,66 @@
...
+  local tmp_kw_dir
+  local repo_url='https://github.com/kworkflow/kworkflow'
+  local base_version
+  local distro
+  local container
+  local output
+  local expected_output
+  local head_commit_hash
+
+  tmp_kw_dir=$(mktemp --directory)
```
davidbtadokoro, line 21: I'm kind of confused: shouldn't this command be run inside the container? I think the temporary directory is being created on the host machine and that the test doesn't fail because the path returned will be something like `/tmp/tmp.XXXXXX`, which is a valid path for the container. Confirm if this is the case and, if so, run this command inside each container with `container_exec`.

davidbtadokoro:
Hey, @aquilamacedo, and thank you for tackling my comments :smile: 

I ran the tests on my local machine, and it works. I think you solved the problem of the environment assumptions I mentioned. I just had a simple comment about creating the temp dir inside the host rather than inside the containers, so please address it.

Apart from that, everything LGTM!

tests/integration/self_update_test.sh
aquilamacedo: Hi @davidbtadokoro.

I hadn't noticed that, it's been fixed now. Thank you :-)

rodrigosiqueira:
@aquilamacedo, any updates here?

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,141 @@
+#!/bin/bash
+
+include './src/lib/kwio.sh'
+include './tests/unit/utils.sh'
+include './tests/integration/utils.sh'
+
+declare -g expected_output
```
davidbtadokoro, line 7: Do we need this global variable? FWIU, `expected_output` is only used inside `self_update_and_check_version`, so can't we use a local variable for this?

tests/integration/utils.sh
```diff
@@ -219,3 +219,14 @@ function container_inspect()
...
 }
+
+function completely_remove_kworkflow_from_container()
+{
+  local container="$1"
+
+  container_exec "$container" 'yes | ./setup.sh --completely-remove > /dev/null'
+
+  if [ "$?" -ne 0 ]; then
+    fail "Failed to completely remove kworkflow"
```
davidbtadokoro, line 230: Use single quotes here

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,141 @@
...
+      container_exec "$container" 'yes | kw self-update --unstable > /dev/null'
+      ;;
+    *)
+      complain "Unsupported update for branch '${kw_branch}'"
+      ;;
+  esac
+
+  output=$(container_exec "$container" "kw --version")
+
+  assert_equals_helper "kw self-update ${kw_branch} failed for ${distro}" "$LINENO" "$expected_output" "$output"
```
davidbtadokoro, line 89: I really like the modular approach you've taken here, as the setup and the self-update vary across the cases in just some parameters. But can't we move the assertions (this line and the line where we get `output`) to the callers, i.e., the functions that represent the test cases? I say this because the way the test for the feature is implemented, we lose the information `LINENO` (all will indicate fail in line 89) and a more fine-grained error message.

Sorry for being too demanding, but as test cases grow, these two pieces of information will matter when debugging failing tests.

davidbtadokoro (changes requested):
Hi @aquilamacedo, and thanks again for the update (self-update :thinking:)

This PR is legitimate, and I just left some comments that I think will be simple for you to tackle. The test cases are aligned with what @rodrigosiqueira asked, and they are running well, at least on my end and in the CI.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,141 @@
...
+function setup_kworkflow_environment()
+{
+  local container="$1"
+  local kw_tmp_dir="$2"
+  local kw_branch="$3"
+  local commit_point="$4"
+
+  container_exec "$container" "git clone ${kw_repo_url} ${kw_tmp_dir}"
+  if [[ "$?" -ne 0 ]]; then
+    complain 'Git clone operation failed'
```
rodrigosiqueira, line 26: Add another `complain` highlighting the failed command, i.e., `git clone ${kw_repo_url} ${kw_tmp_dir}.
Do it for all of these types of checks.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,141 @@
...
+      ;;
+    'unstable')
+      container_exec "$container" 'yes | kw self-update --unstable > /dev/null'
+      ;;
+    *)
+      complain "Unsupported update for branch '${kw_branch}'"
+      ;;
+  esac
+
+  output=$(container_exec "$container" "kw --version")
```
rodrigosiqueira, line 87: Use single quotes for the kw --version

tests/integration/self_update_test.sh
rodrigosiqueira: I agree

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,141 @@
...
+    kw_tmp_dir=$(container_exec "$container" 'mktemp --directory')
+
+    # Remove the kworkflow from container
+    completely_remove_kworkflow_from_container "$container"
+
+    # Configuring the kworkflow to test an old version of unstable branch
+    setup_kworkflow_environment "$container" "$kw_tmp_dir" 'unstable' 'HEAD~4'
+
+    for kw_branch in "${kw_branches[@]}"; do
+      self_update_and_check_version "$container" "$kw_tmp_dir" "$kw_branch" "$distro"
```
rodrigosiqueira, line 136: It will be more readable if the assert is here.

rodrigosiqueira (changes requested):

tests/integration/self_update_test.sh
aquilamacedo: I agree too, in this regard of pointing to the error line it would be more difficult to debug, i've already fixed this

rodrigosiqueira:
@aquilamacedo , could you check the CI failure?

aquilamacedo:
> @aquilamacedo , could you check the CI failure?

Yeah, I'm looking into that rn

aquilamacedo:
Since this PR was dependent on another one that has already been accepted, I have just updated it :)

rodrigosiqueira (changes requested):

aquilamacedo:
I made all the changes we discussed in the call, could you check them? @rodrigosiqueira @davidbtadokoro

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,213 @@
...
+include './tests/integration/utils.sh'
+
+declare -a KW_BRANCHES=('master' 'unstable')
+declare -g HOST_KW_CLONED_PATH
+
+function oneTimeSetUp()
+{
+  local kw_repo_url='https://github.com/kworkflow/kworkflow'
+
+  HOST_KW_CLONED_PATH="$(mktemp --directory)/kworkflow"
```
davidbtadokoro, line 15: I think this is a nitpick, but maybe we want to create another readonly global variable to store the path returned by `mktemp --directory`, so we can remove this temporary dir in `oneTimeTearDown` or below, in case of fail. At the moment, we are only removing the `kworkflow` repo inside the temp dir (again, a nitpick, as this temp dir wont occupy practically any space...)

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,213 @@
...
+  setup_container_environment "$VERBOSE" 'self-update'
+}
+
+# Function to set up an outdated kworkflow environment in a container
+#
+# @container: The name of the container.
+# @kw_cloned_container_path: Temporary directory inside the container where kworkflow is cloned
+# @kw_branch: The branch of kworkflow to be used.
+# @commit_point: The specific commit to be checked out.
+function setup_outdated_kworkflow_environment()
```
davidbtadokoro, line 44: Just to be sure, the idea is to:

1) in each `_old_version_` test case, install an arbitrary outdated upstream version of kw, using the cloned repo
2) in each `_old_version_` test case, get the expected version from `kw --version` using the the cloned repo
3) in each `_old_version_` test case, run the current version of `kw self-update` (from the local repo actually being tested.

If so, I feel like it is a great approach to circumvent some of the difficulties we mentioned in the offline conversation.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,213 @@
...
+  base_version=$(container_exec "$container" "git -C ${kw_cloned_container_path} show remotes/origin/${kw_branch}:./src/VERSION | head --lines 1")
+
+  expected_output=$(printf '%s\nBranch: %s\nCommit: %s' "$base_version" "$kw_branch" "$head_commit_hash")
+
+  printf '%s\n' "$expected_output"
+}
+
+# This function tests the self-update functionality of kworkflow for the master
+# branch.
+function test_self_update_master()
```
davidbtadokoro, line 101: Maybe I am missing something, but in this test case, and the below one, the kw version installed before the `container_exec "$container" 'yes | bash ./kw self-update > /dev/null'` line, is the version of the current repository that is being tested?

davidbtadokoro (changes requested):
Hi @aquilamacedo, thanks a lot for the great update and sorry for the **massive** delay (on all PRs).

This PR is looking really great and refined. I left really small comments (some are just questions). Tested it locally and it works really nice.

The only "serious" issue is the execution time that it is really slow. I suppose there is no trivial way to solve this due to the fact that we need to clone instances of the kw repo to isolate the test git commands and don't mess with the current local repo. The `kw self-update` feature itself clones a repo, so I think there isn't much where to run...

Overall, wonderful work!

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,213 @@
+#!/bin/bash
+
+include './src/lib/kwio.sh'
+include './src/lib/kwlib.sh'
+include './tests/unit/utils.sh'
+include './tests/integration/utils.sh'
+
+declare -a KW_BRANCHES=('master' 'unstable')
+declare -g HOST_KW_CLONED_PATH
```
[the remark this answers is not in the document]
aquilamacedo: `KW_BRANCHES` can be read-only with -ar. However, `HOST_KW_CLONED_PATH` needs to be assigned a value later, so it can't be read-only (-gr).

Also, combining declaration and assignment triggers a shellcheck `SC2155` error.

tests/integration/self_update_test.sh
aquilamacedo: Nice catch, I added this change :)

tests/integration/self_update_test.sh
aquilamacedo: Exactly, that's the idea.

tests/integration/self_update_test.sh
aquilamacedo: Yes, before calling these functions, the version of `kw` matches the current version modified by the developer on the host.

aquilamacedo:
> Hi @aquilamacedo, thanks a lot for the great update and sorry for the **massive** delay (on all PRs).
> 
> This PR is looking really great and refined. I left really small comments (some are just questions). Tested it locally and it works really nice.
> 
> The only "serious" issue is the execution time that it is really slow. I suppose there is no trivial way to solve this due to the fact that we need to clone instances of the kw repo to isolate the test git commands and don't mess with the current local repo. The `kw self-update` feature itself clones a repo, so I think there isn't much where to run...
> 
> Overall, wonderful work!

Thx for the review, David :)

I believe I have addressed all your comments. Could you please check?

Regarding the execution time, it is indeed a bit slow. I believe it is taking a long time because of the installations and clones of the kw repository that `self-update` does, as you said

rodrigosiqueira:
Hi @aquilamacedo ,
When I tried to run it in my local, I saw this error:

```bash
./run_tests.sh -v --integration test self_update_test 
Running test [integration/self_update_test]
Preparing environment for kw self-update integration tests...
[1/6] Using cached container image for archlinux.
[2/6] Creating archlinux container.
[3/6] Using cached container image for debian.
[4/6] Creating debian container.
[5/6] Using cached container image for fedora.
[6/6] Creating fedora container.
test_self_update_master
test_self_update_unstable
test_self_update_with_kw_old_version_master
ASSERT:(416): Failed to execute the command in the container.
kw setup failed
./setup.sh --install --force --skip-checks --skip-docs
test_self_update_with_kw_old_version_unstable

Ran 4 tests.

FAILED (failures=1)

Tearing down containers used in integration tests...
[1/3] Removing archlinux container.
[2/3] Removing debian container.
[3/3] Removing fedora container.
```

Any idea of what could be the issue?

@davidbtadokoro , could you also try on your side?

rodrigosiqueira:
@davidbtadokoro did you see my previous comment?

aquilamacedo:
Hi @rodrigosiqueira.

Try to clean your environment first:

 `
./run_tests.sh --integration clear-cache
`

Then, execute the self-update test with this command:

`
  ./run_tests.sh -v --integration test self_update_test 
 `

davidbtadokoro:
> @davidbtadokoro did you see my previous comment?

Hi, really sorry for the delay. I've tried on my end and couldn't reproduce your errors. I tested on a new environment where I hadn't ever run the integration tests, so (based on Aquila's response above) maybe this is why...

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,214 @@
...
+}
+
+# This function gets the commit hash and base version of the specified
+# branch, constructs the expected output, updates kworkflow based on
+# the branch, and then checks the version.
+#
+# @container: The name of the container.
+# @kw_cloned_container_path: Temporary directory inside the container where kworkflow is cloned
+# @kw_branch: The branch of kworkflow to be used.
+function get_expected_output()
```
rodrigosiqueira, line 83: This name is very generic; use a name that better describes what the expected output is.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,214 @@
...
+  fi
+}
+
+# This function gets the commit hash and base version of the specified
+# branch, constructs the expected output, updates kworkflow based on
+# the branch, and then checks the version.
+#
+# @container: The name of the container.
+# @kw_cloned_container_path: Temporary directory inside the container where kworkflow is cloned
+# @kw_branch: The branch of kworkflow to be used.
```
rodrigosiqueira, line 82: Document the return part of this function.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,214 @@
...
+  local head_commit_hash
+  local base_version
+  local expected_output
+
+  head_commit_hash=$(container_exec "$container" "git -C ${kw_cloned_container_path} rev-parse --short origin/${kw_branch}")
+  base_version=$(container_exec "$container" "git -C ${kw_cloned_container_path} show remotes/origin/${kw_branch}:./src/VERSION | head --lines 1")
+
+  expected_output=$(printf '%s\nBranch: %s\nCommit: %s' "$base_version" "$kw_branch" "$head_commit_hash")
+
+  printf '%s\n' "$expected_output"
```
rodrigosiqueira, line 97: I don't see a good reason to save the output in a variable and use it again in the `printf`. Why not use the printf directly?

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,214 @@
...
+}
+
+# This function tests the self-update functionality of kworkflow for the master
+# branch.
+function test_self_update_master()
```
rodrigosiqueira, line 102: maybe use `test_self_update_from_master_branch`.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,214 @@
...
+{
+  local kw_branch='master'
```
rodrigosiqueira, line 104: Since this function is dedicated to the master, why do you need this variable? You can use this constant direcly. Or if you still prefer to use the variable, at least use a name that makes clear that you are using the master branch.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,214 @@
...
+  for distro in "${DISTROS[@]}"; do
+    container="kw-${distro}"
+    kw_cloned_container_path="$(generate_temporary_directory_in_container "$container")/kworkflow"
+
+    container_copy "$container" "$HOST_KW_CLONED_PATH" "$kw_cloned_container_path"
+
+    expected_output=$(get_expected_output "$container" "$kw_cloned_container_path" "$kw_branch")
+
+    # Specific update command for master branch
+    container_exec "$container" 'yes | bash ./kw self-update > /dev/null'
```
rodrigosiqueira, line 120: Why you are using `.kw`? Fwiu, you should use the installed version, i.e., `kw self-update`.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,214 @@
...
+    output=$(container_exec "$container" 'kw --version')
+    assert_equals_helper "kw self-update ${kw_branch} failed for ${distro}" "$LINENO" "$expected_output" "$output"
+  done
+}
+
+# This function tests the self-update functionality of kworkflow for the
+# unstable branch.
+function test_self_update_unstable()
+{
+  local kw_branch='unstable'
```
rodrigosiqueira, line 131: My comments from the previous tests also apply to this test.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,214 @@
...
+
+    # Specific update command for master branch
+    container_exec "$container" "yes | bash ./kw self-update --${kw_branch} > /dev/null"
+
+    output=$(container_exec "$container" 'kw --version')
+    assert_equals_helper "kw self-update ${kw_branch} failed for ${distro}" "$LINENO" "$expected_output" "$output"
+  done
+}
+
+function oneTimeTearDown()
```
rodrigosiqueira, line 206: This should be after the `oneTimeStartUp()` function.

tests/integration/self_update_test.sh
```diff
@@ -0,0 +1,214 @@
...
+{
+  # Check if the path is safe to remove
+  if is_safe_path_to_remove "$HOST_TEMP_DIR"; then
+    rm -rf "$HOST_TEMP_DIR"
```
rodrigosiqueira, line 210: avoid short parameters.

tests/integration/utils.sh
```diff
@@ -448,3 +448,17 @@ function container_inspect()
...
+
+# This function generates a temporary directory inside a specified container.
+#
+# @container  The name of the container.
+function generate_temporary_directory_in_container()
+{
+  local container="$1"
+
+  container_exec "$container" 'mktemp --directory'
+
```
rodrigosiqueira, line 460: remove this extra line.

tests/integration/utils.sh
```diff
@@ -448,3 +448,17 @@ function container_inspect()
...
+  if [ "$?" -ne 0 ]; then
+    fail 'Failed to create a temporary directory inside the container'
```
rodrigosiqueira, line 462: If it fails, you must return an error code and check this function return in the test.

rodrigosiqueira (changes requested):

tests/integration/self_update_test.sh
aquilamacedo: I use `bash ./kw self-update` because it's the best approach. It is more reliable than using the global version, as running `kw self-update` could **change** the installed version, making it different from the one used in the integration test.

aquilamacedo:
changes made, could you give it a final review? @rodrigosiqueira

rodrigosiqueira:
PR merged into the unstable branch. Thanks

Issue #911, opened 2023-09-08
`kw patch-hub`: Optimize performance of patch fetch


davidbtadokoro:
**Description:**

`kw patch-hub` fetch times are really slow. They are also really inconsistent as some seem to take 20 seconds up to more than 1 minute.

**Type**

Optimization.

**Why**

The time it takes for these actions is not acceptable.

Issue #900, opened 2023-08-20
`kw patch-hub`: Fix invisible patches because of message-ID pattern


davidbtadokoro:
**Description:**

`kw patch-hub` processes patches individually and, for patchsets (patch series), elects one to represent the set, be it either the cover letter (patch 0/N) or the first patch in the set (patch 1/N). However, the way that the feature defines a "representant" through its message-ID pattern causes false negatives, i.e., some patches that should be considered "representants" aren't.

**Type:**

Enhancement, but also fixing correctness.

**Why:**

Wrongly filtering patches is really bad and, although there seems to be no false positives, this behaviour should not be present in the feature.

PR #1056, opened 2024-03-07
`kw patch-hub`: Overhaul listing of patchsets

davidbtadokoro:
There are two major problems with the listing of patchsets in the `kw patch-hub` feature:

1. Inconsistent performance that at best revolves around 20+ seconds #911 
2. Incorrectly determining representative patches #900 

This PR aims to fix both. In the case of the first problem, we could consistently maintain, at worst, times below 20 seconds (didn't verify with all lists).

### DIFF
--- a/src/lib/lore.sh
+++ b/src/lib/lore.sh
@@ -542,6 +542,30 @@ function process_individual_patches()
   done <<< "$pre_processed_patches"
 }
 
+# This function makes the lore request to get the raw contents of a lore
+# message.
+#
+# @message_id: Message-ID of desired lore message.
+# @flag: Flag to control function output.
+#
+# Returns:
+# - Output: Response from requisiton to raw lore message.
+function get_raw_lore_message()
+{
+  local message_id="$1"
+  local flag="${2:-SILENT}"
+  local raw_message_url
+
+  if [[ "$message_id" =~ /$ ]]; then
+    raw_message_url=$(replace_http_by_https "${message_id}raw")
+  else
+    raw_message_url=$(replace_http_by_https "${message_id}/raw")
+  fi
+
+  # TODO: Add cache functionality
+  cmd_manager "$flag" "curl --silent '${raw_message_url}'"
+}
+
 # This function processes representative patches (i.e. the first message in a
 # patchset) from a list of processed individual patches. The patches determined
 # as representatives are stored (in the same order as the argument
@@ -556,11 +580,13 @@ function process_individual_patches()
 function process_representative_patches()
 {
   local -n _individual_patches_array="$1"
+  local is_representative_patch
   local message_id
   local in_reply_to_message_id
   local -a patch_metadata
   local -a in_reply_to_metadata
-  local is_representative_patch
+  local in_reply_to_title
+  local in_reply_to_patch_metadata
 
   for patch in "${_individual_patches_array[@]}"; do
     is_representative_patch=''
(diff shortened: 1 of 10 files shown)


src/lib/lore.sh
```diff
@@ -502,6 +502,30 @@ function process_individual_patches()
   done <<< "$pre_processed_patches"
 }
 
+# This function makes a network requisition to get the raw contents of a lore
```
rodrigosiqueira, line 549: What do you think of using `makes the lore request to get`?

src/lib/lore.sh
```diff
@@ -532,29 +558,36 @@ function process_representative_patches()
...
+      IFS=',' read -ra patch_metadata <<< "${individual_patches_metadata["$message_id"]}"
 
-        # Assume that, if 'In-Reply-To' is not patch number 0 from the same
-        # version, number 1 is the representative
-        if [[ "${patch_metadata[0]}" != "${in_reply_to_metadata[0]}" || "${in_reply_to_metadata[1]}" != 0 ]]; then
-          is_representative_patch=1
+      if [[ -n "${individual_patches_metadata["$in_reply_to_message_id"]}" ]]; then
+        IFS=',' read -ra in_reply_to_metadata <<< "${individual_patches_metadata["$in_reply_to_message_id"]}"
+      else
+        in_reply_to_title=$(get_raw_lore_message "$message_id" | grep --perl-regexp '^Subject:' | sed 's/^Subject: //')
```
rodrigosiqueira: Is it possible that the subject information should be two lines long with or without \n? If so, this grep/sed will work?

tests/unit/lib/lore_test.sh
```diff
@@ -889,6 +901,104 @@ function test_process_representative_patches_duplicated_patches()
...
+  assert_equals_helper 'Wrong size of `processed_representative_patches` hashtable' "$LINENO" 1 "${#processed_representative_patches[@]}"
+  assert_equals_helper 'Wrong value of `REPRESENTATIVE_PATCHES_PROCESSED`' "$LINENO" 1 "$REPRESENTATIVE_PATCHES_PROCESSED"
+  assert_equals_helper 'Wrong representative patch 0' "$LINENO" "$patch1" "${representative_patches[0]}"
+  assert_equals_helper 'Representative patch 0 should be marked in hashtable' "$LINENO" 1 "${processed_representative_patches['id1']}"
+
+  # Case 5: In reply isn't patch from same patchset, but is cover letter
+  # shellcheck disable=SC2317
+  function get_raw_lore_message()
+  {
+    printf 'Subject: [PATCH 0/3] Some title'
```
rodrigosiqueira, line 1041: I believe it is worth to test:
1. Very large subjects.
2. Very small subjects.
3. Multi-line subject (if it makes sense)
4. Subjects that have Æ.

src/lib/lore.sh
```diff
@@ -570,6 +570,64 @@ function get_raw_lore_message()
...
+  local shared_dir_for_parallelism="$3"
+  local is_representative_patch
+  local message_id
+  local in_reply_to_message_id
+  local -a patch_metadata
+  local -a in_reply_to_metadata
+  local in_reply_to_title
+  local in_reply_to_patch_tag
+
+  is_representative_patch=''
```
rodrigosiqueira, line 586: Could you move this line after the declaration section?

rodrigosiqueira (changes requested):
Hi @davidbtadokoro ,
Overall, this change looks fantastic! Great job.
I just added some minor comments.

davidbtadokoro:
> Hi @davidbtadokoro , Overall, this change looks fantastic! Great job. I just added some minor comments.

Hey @rodrigosiqueira, and thank you so much for the review! I will work on the update ASAP :+1:

src/lib/lore.sh
```diff
@@ -160,57 +160,147 @@ function is_introduction_patch()
...
 {
-  local url="$1"
-  local curl_cmd='curl --insecure --silent --fail --silent --head'
-  local raw_curl_output
-  local url_status_code
+  local message_title="$1"
+  local patch_tag
 
-  [[ -z "$url" ]] && return 22 # EINVAL
+  if [[ "$message_title" =~ \[[^\]]*(RFC|Rfc|rfc|PATCH|Patch|patch)[^\[]*\] ]]; then
```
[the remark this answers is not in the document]
davidbtadokoro: For sure, thanks for the catch :+1: Going to also add tests for this case.

src/lib/lore.sh
```diff
@@ -160,57 +160,147 @@ function is_introduction_patch()
...
+    patch_tag="${BASH_REMATCH[0]}"
+  fi
```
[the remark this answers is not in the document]
davidbtadokoro: In case there is no match, I implicitly determined that the default tag is an empty string. The rest of the `get_<patch_attribute>` handle an empty patch tag as "undefined"; for this situation, the version, number in series, and the total will be `X`. Do you agree with this approach?

src/lib/lore.sh
davidbtadokoro: Sure :+1:

davidbtadokoro (commented):
Hey, @rodrigosiqueira, and thank you so much for the review!

I addressed most of your comments. I mark these as resolved (but most have a comment, if you want to see them) to not pollute the ones still open. I left some open because I have to do some digging before addressing them, so this update isn't ready to be reviewed again (I'll ping you when it is).

src/lib/lore.sh
davidbtadokoro: Hey @rodrigosiqueira, I experimented a little more with some lists, and I am confident that this heuristic is valid. We really are only fetching the first line of `Subject:`, i.e., with multi-line subjects in the raw patch, the `grep`  command only gets its first line.

This is OK, because we assume that the patch tag will surely be in the first line (which I "confirmed" in my "experiments"), and, considering that we are already handling an edge case (we only fetch the raw patch, because we fell in the case of a patch number 1 that has an in reply to a message that wasn't processed yet), I think we are safe.

Anyway, give me your thoughts on this.

davidbtadokoro:
Hey @rodrigosiqueira, I did the digging, and the PR is good for review now (this last force-push is just a rebase to the lastest unstable update).

If possible, give me your thoughts on the conversations that I left open.

tests/unit/lib/lore_test.sh
davidbtadokoro: After thinking further,  I don't think we need to test those cases in this context, but we surely must handle subjects that have Æ in processing patches (which is a more critical context). I will create an issue for that.

src/lib/lore.sh
```diff
@@ -160,57 +160,187 @@ function is_introduction_patch()
...
+}
+
+# This function removes a patch metadata substring from a message title.
+#
+# For example, considering `@message_title` equal to
+#   '[additional tag][RFC/PATCH v23 12/57] some/subsys: Do bar',
+# and `@patch_metadata` equal to
+#   '[RFC/PATCH v23 12/57]',
+# the outputted stripped title will be
+#   '[additional tag] some/subsys: Do bar'.
```
rodrigosiqueira, line 306: Very nice example :)

src/lib/lore.sh
```diff
@@ -440,126 +374,194 @@ function compose_lore_query_url_with_verification()
...
     fi
 
-    # Based on the way that we build our xpath expression, we can rely on this sequence:
-    # Name, Email, Title, Link
-    # Since we have a dedicated function to extract title metadata, we want to
-    # save the title in a separated variable for later processing.
     case "$count" in
-      0) # NAME
-        processed_patchset="$(process_name "$line")${SEPARATOR_CHAR}"
+      0) # Author's name
```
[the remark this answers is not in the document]
rodrigosiqueira: Don't worry about my suggestion in this case.

src/lib/lore.sh
rodrigosiqueira: I think we are fine!

rodrigosiqueira:
Hi @davidbtadokoro , thanks a lot for all your efforts in optimizing this! After thinking a lot about the current status, I decided to merge it now since it is reliable and good enough.

Great job!

Issue #1052, opened 2024-02-28
Add context to kw explore results


rodrigosiqueira:
**Description:**

Expand `kw explore` to show some extra context after matching a string, and I believe that we have a perfect tool for that: `batcat`. After the normal exploration matches a file and its lines, we can show X lines above and after the match and show it to the user with the below command:

```
batcat -P  --line-range <START>:<END> <FILE>
```

e.g.:

```sh
batcat -P  --line-range 30:+10 Makefile
```

**Type:**

Feature improvement

**Why:**

Sometimes, when searching for a code, it has multiple matches in the code, and it helps to see the context. Maybe we can have something like:

```sh
kw e -H --show-context dc_stream_state
```

iamtvk:
I want to work on this, I am new to this codebase, so any input is much appreciated.

sahil-sagwekar2652:
@rodrigosiqueira I would like to know your thoughts on this PR https://github.com/kworkflow/kworkflow/pull/1058

PR #1058, opened 2024-03-18
Add `--show-context` to kw explore

sahil-sagwekar2652:
The issue proposed to use the batcat command but this commit uses the built-in -C flags in git grep and GNU grep instead, to avoid adding more dependencies to kw.

This commit also updates the docs as well as the tests for kw-explore.

Is the choice of using `-C` appropriate? Or would it be confusing because there is also the `-c` which exists. Please suggest your thoughts on this.

Closes: #1052

### DIFF
--- a/documentation/man/features/kw-explore.rst
+++ b/documentation/man/features/kw-explore.rst
@@ -7,7 +7,7 @@ kw-explore - Explore folder
 SYNOPSIS
 ========
 *kw* (*e* | *explore*) [(-l | \--log) | (-g | \--grep) | (-a | \--all) | \--verbose]
-                       [(-c | \--only-source) | (-H | \--only-header)] <expr>
+                       [(-c | \--only-source) | (-H | \--only-header)] [(-C[<num>] | \--show-context[=<num>])] <expr>
                        [-p] [<dir> | <file>]
 
 DESCRIPTION
@@ -45,5 +45,23 @@ OPTIONS
 -H | \--only-header:
   With this option, it is possible to show only the results from the header.
 
+-C[<num>] | \--show-context[=<num>]:
+  Show <num> lines of additional context above and below the matched line.
+  If <num> is not specified, the default value of 3 will be used.
+
 \--verbose:
   Verbose mode allows the user to see the commands executed under the hood.
+
+EXAMPLES
+========
+To show matched line with context using long-form flag::
+
+  kw explore --show-context=5 search_string
+
+To show matched line with context using short flag::
+
+  kw explore -C5 search_string
+
+Search through all tracked and untracked files, default value of 3 will be used for context::
+
+  kw explore -C --all search_string
--- a/src/_kw
+++ b/src/_kw
@@ -282,6 +282,7 @@ _kw_explore()
     '(-a --all -l --log -g --grep)'{-a,--all}'[search for all matches in files under or not git management]' \
     '(-c --only-source -H --only-header)'{-c,--only-source}'[show only results from the source]' \
     '(-H --only-header -c --only-source)'{-H,--only-header}'[show only results from the header]' \
+    '(-C --show-context)'{-C-,--show-context=-}'[set the context value]' \
     '1: : ' \
     '2: :_files'
 }
--- a/src/explore.sh
+++ b/src/explore.sh
@@ -15,6 +15,7 @@ function explore_main()
   local flag
   local search
   local path
+  local context
   local ret
 
   if [[ "$1" =~ -h|--help ]]; then
@@ -31,6 +32,7 @@ function explore_main()
   flag="${options_values['TEST_MODE']:-'SILENT'}"
   search="${options_values['SEARCH']}"
   path="${options_values['PATH']:-'.'}"
+  context="${options_values['CONTEXT']:-0}"
 
   [[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
 
@@ -48,19 +50,19 @@ function explore_main()
 
   if [[ "${options_values['TYPE']}" -eq 2 ]]; then
     # Use GNU GREP
-    explore_files_gnu_grep "$search" "$path" "$flag"
+    explore_files_gnu_grep "$search" "$path" "$context" "$flag"
     return
   fi
 
   if [[ "${options_values['TYPE']}" -eq 3 ]]; then
     # Search in directories controlled or not by git
-    explore_all_files_git "$search" "$path" "$flag"
+    explore_all_files_git "$search" "$path" "$context" "$flag"
     return
   fi
 
   if [[ -z "${options_values['TYPE']}" ]]; then
     # Search in files under git control
-    explore_files_under_git "$search" "$path" "$flag"
+    explore_files_under_git "$search" "$path" "$context" "$flag"
     return
   fi
 }
@@ -75,8 +77,8 @@ function explore_main()
 # This function also set options_values
 function parse_explore_options()
 {
-  local long_options='log,grep,all,only-header,only-source,exactly,verbose'
-  local short_options='l,g,a,H,c'
+  local long_options='log,grep,all,only-header,only-source,exactly,verbose,show-context::'
+  local short_options='l,g,a,H,c,C::'
   local options
 
   if [[ "$#" -eq 0 ]]; then
@@ -97,6 +99,7 @@ function parse_explore_options()
   options_values['TYPE']=''
   options_values['SCOPE']=''
   options_values['EXACTLY']=''
+  options_values['CONTEXT']=''
   options_values['VERBOSE']=''
 
   eval "set -- $options"
@@ -160,6 +163,24 @@ function parse_explore_options()
         options_values['VERBOSE']=1
         shift
         ;;
+      --show-context | -C)
+        if [[ -n "${options_values['CONTEXT']}" ]]; then
+          options_values['ERROR']='Invalid arguments: Multiple context values!'
+          return 22 # EINVAL
+        fi
+
+        if [[ ! "$2" ]]; then
+          options_values['CONTEXT']=3
+        elif [[ ! "$2" =~ ^[0-9]+$ ]]; then
+          options_values['ERROR']='Context value must be a non-negative integer!'
+          return 22 # EINVAL
+        else
+          options_values['CONTEXT']="$2"
+          shift
+        fi
+
+        shift
+        ;;
       --) # End of options, beginning of arguments
         shift
         ;;
@@ -196,7 +217,7 @@ function explore_git_log()
 
   flag=${flag:-'SILENT'}
 
-  cmd_manager "$flag" "git log --grep='$search_string' $path"
+  cmd_manager "$flag" "git log --grep='${search_string}' ${path}"
 }
 
 # This function searches string in files under git control.
@@ -209,11 +230,12 @@ function explore_files_under_git()
 {
   local regex="$1"
   local path="$2"
-  local flag="$3"
+  local context="$3"
+  local flag="$4"
 
   flag=${flag:-'SILENT'}
 
-  cmd_manager "$flag" "git grep -e '$regex' -nI $path"
+  cmd_manager "$flag" "git grep --context ${context} -e '${regex}' --line-number -I ${path}"
 }
 
 # This function uses git grep tool to search string in files under or not git
@@ -228,11 +250,12 @@ function explore_all_files_git()
 {
   local regex="$1"
   local path="$2"
-  local flag="$3"
+  local context="$3"
+  local flag="$4"
 
   flag=${flag:-'SILENT'}
 
-  cmd_manager "$flag" "git grep --no-index -e '$regex' -nI $path"
+  cmd_manager "$flag" "git grep --no-index --context ${context} -e '${regex}' --line-number -I ${path}"
 }
 
 # This function allows the use of gnu grep utility to manages the search for
@@ -246,11 +269,12 @@ function explore_files_gnu_grep()
 {
   local regex="$1"
   local path="$2"
-  local flag="$3"
+  local context="$3"
+  local flag="$4"
 
   flag=${flag:-'SILENT'}
 
-  cmd_manager "$flag" "grep --color -nrI $path -e '$regex'"
+  cmd_manager "$flag" "grep --color --line-number --recursive -I ${path} --context ${context} -e '${regex}'"
 }
 
 function explore_help()
@@ -267,5 +291,6 @@ function explore_help()
     '  explore,e --all,-a <string> - Search for all <string> match under or not of git management' \
     '  explore,e --only-source,-c <string> - Search for all <string> in source files' \
     '  explore,e --only-header,-H <string> - Search for all <string> in header files' \
+    '  explore,e --show-context[=<num>],-C[<num>] <string> - Print <num> lines of output context (default: 3)' \
     '  explore,e --verbose - Show a detailed output'
 }
--- a/tests/unit/explore_test.sh
+++ b/tests/unit/explore_test.sh
@@ -67,8 +67,13 @@ function test_explore_files_under_git_repo()
   assertEquals "($LINENO)" "$MSG_OUT" "$output"
 
   output=$(explore_main 'GNU grep' '.' 'TEST_MODE')
-  expected_result="git grep -e 'GNU grep' -nI ."
+  expected_result="git grep --context 0 -e 'GNU grep' --line-number -I ."
   assertEquals "($LINENO)" "$expected_result" "$output"
+
+  # Test for non zero context value
+  output=$(explore_main --show-context=5 'GNU grep' '.' 'TEST_MODE')
+  expected_result="git grep --context 5 -e 'GNU grep' --line-number -I ."
+  assertEquals "(${LINENO})" "$expected_result" "$output"
 
   # Test if search only in files under git control
   cp "$current_path/tests/unit/samples/grep_check.c" ./
@@ -82,13 +87,13 @@ function test_explore_files_under_git_repo()
 
   # Test only-source and only-header
   MSG_OUT='3'
-  output=$(explore_main 'camelCase' | wc -l)
+  output=$(explore_main 'camelCase' | wc --lines)
   assertEquals "($LINENO)" "$MSG_OUT" "$output"
   MSG_OUT='2'
-  output=$(explore_main -c 'camelCase' | wc -l)
+  output=$(explore_main -c 'camelCase' | wc --lines)
   assertEquals "($LINENO)" "$MSG_OUT" "$output"
   MSG_OUT='1'
-  output=$(explore_main -H 'camelCase' | wc -l)
+  output=$(explore_main -H 'camelCase' | wc --lines)
   assertEquals "($LINENO)" "$MSG_OUT" "$output"
 
   cd "$current_path" || {
@@ -133,7 +138,11 @@ function test_explore_grep()
   assertEquals "($LINENO)" '.git' "$output"
 
   output=$(explore_main --grep 'GNU grep' '.' 'TEST_MODE')
-  expected_result="grep --color -nrI . -e 'GNU grep'"
+  expected_result="grep --color --line-number --recursive -I . --context 0 -e 'GNU grep'"
+  assertEquals "(${LINENO})" "$expected_result" "$output"
+
+  output=$(explore_main --grep --show-context=5 'GNU grep' '.' 'TEST_MODE')
+  expected_result="grep --color --line-number --recursive -I . --context 5 -e 'GNU grep'"
   assertEquals "($LINENO)" "$expected_result" "$output"
 
   cd "$current_path" || {
@@ -153,7 +162,11 @@ function test_explore_git()
   }
 
   output=$(explore_main --all 'GNU grep' '.' 'TEST_MODE')
-  expected_result="git grep --no-index -e 'GNU grep' -nI ."
+  expected_result="git grep --no-index --context 0 -e 'GNU grep' --line-number -I ."
+  assertEquals "(${LINENO})" "$expected_result" "$output"
+
+  output=$(explore_main --all --show-context=5 'GNU grep' '.' 'TEST_MODE')
+  expected_result="git grep --no-index --context 5 -e 'GNU grep' --line-number -I ."
   assertEquals "($LINENO)" "$expected_result" "$output"
 
   # Test if the search ignores files in .git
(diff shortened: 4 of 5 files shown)


davidbtadokoro (changes requested):
Hi @sahil-sagwekar2652, and thank you so much for your first change!

As starting feedback, it seems you get the gist of the contribution practices and processes (which aren't trivial), so great job!

I left some review comments, and I recommend you visualize them by going into the `Commits` tab and seeing my comments with context (kinda meta considering the purpose of your PR :rofl: ). Most are about coding style but are as important as any other, so see if you can tackle them. Feel free to ping me if you have doubts.

Also, here is a comment about the commit message that doesn't fit the inline review. Think that the issue you are alluding to won't be accessible when the change gets merged into the project, so make your commit message auto-contained, i.e., describe what, how, and why you are doing it (see [here](https://kworkflow.org/content/howtocontribute.html#commit-messages) for reference).

Apart from all of this, I tested the change in the three modalities (git grep, git grep with all files, and GNU grep), and it works as intended! @rodrigosiqueira, when you have the opportunity, check to see if this PR introduced the behavior you described in the issue.

src/explore.sh
```diff
@@ -48,19 +49,19 @@ function explore_main()
 
   if [[ "${options_values['TYPE']}" -eq 2 ]]; then
     # Use GNU GREP
-    explore_files_gnu_grep "$search" "$path" "$flag"
+    explore_files_gnu_grep "$search" "$path" "$flag" "$context"
```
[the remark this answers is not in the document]
sahil-sagwekar2652: I did think of this, but then I thought that `flag` being number 3 would maintain some consistency in all the functions, including `explore_git_log`.

I agree with your opinion too. Do let me know your final thoughts, I'm happy to go either way.

src/explore.sh
```diff
@@ -210,10 +227,11 @@ function explore_files_under_git()
   local regex="$1"
   local path="$2"
   local flag="$3"
+  local context="$4"
 
   flag=${flag:-'SILENT'}
 
-  cmd_manager "$flag" "git grep -e '$regex' -nI $path"
+  cmd_manager "$flag" "git grep -C $context -e '$regex' -nI $path"
```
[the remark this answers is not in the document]
sahil-sagwekar2652: Yes, I actually wanted to confirm this. I understand now.
Similarly, if I understand correctly, I don't think we are following [this rule](https://kworkflow.org/content/codingstyle.html#avoid-short-options) of the coding style which suggests using long options instead of short ones, wherever available. 

According to it the command should look like - `git grep --context ${context} -e '${regex}' --line-number -I ${path}`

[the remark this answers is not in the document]
sahil-sagwekar2652: Yes, I actually wanted to confirm this. I understand now.
Similarly, if I understand correctly, I don't think we are following [this rule](https://kworkflow.org/content/codingstyle.html#avoid-short-options) of the coding style which suggests using long options instead of short ones, wherever available. 

According to it the command should look like - `git grep --context ${context} -e '${regex}' --line-number -I ${path}`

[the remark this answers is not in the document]
davidbtadokoro: Nice catch! You are correct, this rule should also be enforced, so take the opportunity to do this here (and in the other two functions).

Also, just as a reminder, you probably will find many other places in `src/explore.sh` that don't follow the coding style, but **don't change them** in this PR, as they would be outside the scope of the change.

davidbtadokoro:
Hi @sahil-sagwekar2652, I see that you've requested a review, and I imagine you did it because of the comment I just answered. Keep in mind that we maintainers usually take a couple of days or more to look into new PRs or updates.

I'll seize the opportunity to add another comment I remembered I need to add to the review. As you added a new option to the `explore` feature, you must add Bash and Zsh completions. For reference, see [here](https://kworkflow.org/content/project_structure.html#adding-a-new-feature), but you can just analyze the `src/bash_autocomplete.sh` and `src/_kw` files to understand how the completions work (don't forget to test **both** completions, i.e., if you only use Bash or Zsh, install the other shell, install this PR version of kw and validate that the completion is the correct one).

sahil-sagwekar2652:
Hello @davidbtadokoro, I mistakenly asked for the review, forgetting that this is an ongoing review. I apologise for any inconvenience. I have modified the code according to your suggestions.

I want to take this opportunity to reach out to you for the special ID required to submit a GSoC application.
Apologies for this digression I've been trying to contact you by email for the past couple of weeks.

Thanks in advance!

davidbtadokoro:
> Hello @davidbtadokoro, I mistakenly asked for the review, forgetting that this is an ongoing review. I apologise for any inconvenience. I have modified the code according to your suggestions.

No worries! Sorry if I sounded annoyed, that really wasn't the case :smile:! I will review this new version ASAP. 

> 
> I want to take this opportunity to reach out to you for the special ID required to submit a GSoC application. Apologies for this digression I've been trying to contact you by email for the past couple of weeks.

Oops! I'm sorry if your message slipped through me. Funny enough, I can't find these messages :sweat:. I will send you an email with some information. 

> 
> Thanks in advance!

davidbtadokoro:
Hi @sahil-sagwekar2652, and thank you once again for your update. From what I saw, you tackled all my comments.

However, you've updated the PR by adding commits on top of the original one, so you should squash all three commits into one. In case this doesn't make sense, think that there are two important premisses when dealing with PRs:

1. There is no problem in rewriting the commit history as a PR isn't merged code into the upstream yet (this **is** a great problem if we are talking about pushed code that is being used);
2. Every commit should be atomic in its sense. Each commit's "actual" code should be accompanied by related tests and documentation. Also, each commit should not fail in the CI jobs.

In your update, the second commit fixes the "problems" of the first by addressing my comments, and the third commit fixes the problems of the second (number 2 fails in the unit tests). So, as we have a chance to "rewrite" the first commit, we can refine the change into a single self-contained commit. Overall, when updating a PR, you should use tools like `git rebase --interactive` and `git commit --amend`, and if you are unfamiliar with these concepts, take a look at this [post](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) in the Git doc.

In any case, this PR is really promising, and besides my verbose comments, you are doing great!!

sahil-sagwekar2652:
Hi @davidbtadokoro, thanks for the review. I have squashed the commits into a single commit.
Thanks again for all the help!

tests/unit/explore_test.sh
```diff
@@ -67,7 +67,12 @@ function test_explore_files_under_git_repo()
   assertEquals "($LINENO)" "$MSG_OUT" "$output"
 
   output=$(explore_main 'GNU grep' '.' 'TEST_MODE')
-  expected_result="git grep -e 'GNU grep' -nI ."
+  expected_result="git grep -C 0 -e 'GNU grep' --line-number -I ."
+  assertEquals "($LINENO)" "$expected_result" "$output"
+
+  # Test for non zero context value
+  output=$(explore_main --show-context 5 'GNU grep' '.' 'TEST_MODE')
+  expected_result="git grep -C 5 -e 'GNU grep' --line-number -I ."
```
rodrigosiqueira, line 75: This type of test is okay, but I would prefer something that uses the feature. For example, it would be better to search for something in one of the sample files, capture the output, and check if it worked as expected.

rodrigosiqueira (changes requested):
Hi @sahil-sagwekar2652 ,
Very nice PR! You can see some of my comments in the PR.

src/explore.sh
```diff
@@ -160,6 +163,21 @@ function parse_explore_options()
         options_values['VERBOSE']=1
         shift
         ;;
+      --show-context | -C)
+        if [[ -n "${options_values['CONTEXT']}" ]]; then
+          options_values['ERROR']="Invalid arguments: Multiple context values!"
```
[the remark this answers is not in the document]
sahil-sagwekar2652: There are other instances like line `150` where double quotes are used. Should I change them too?

documentation/man/features/kw-explore.rst
```diff
@@ -45,5 +45,8 @@ OPTIONS
 -H | \--only-header:
   With this option, it is possible to show only the results from the header.
 
+-C | \--show-context <num>:
+  Show <num> lines of additional context above and below the matched line.
```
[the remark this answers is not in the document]
sahil-sagwekar2652: Hi @rodrigosiqueira, 

Thank you for the review and suggestions.
I do like the idea of defining a default value, but it brings up some inconsistencies.
From what I understand (I might be wrong), the `::` notation to define optional arguments doesn't work for short options. Additionally, I read on StackOverflow that it works in `getopts.c` but not in the bash version of getopts.

Using the `::` notation for `C` and `show-context` and changing the code inside the case, we get the following viable combinations:

1. `kw e -C string` and `kw e --show-context string` - uses the default context value, i.e., 3.
2. `kw e --show-context=5 string` - shows 5 lines of context.
3. `kw e -C=3 string` will not work.

sahil-sagwekar2652:
Hello @rodrigosiqueira, I will squash this into one commit after the final approval.

documentation/man/features/kw-explore.rst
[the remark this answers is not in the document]
rodrigosiqueira: I'm ok with `-C3`.

src/explore.sh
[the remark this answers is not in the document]
sahil-sagwekar2652: Hi @rodrigosiqueira , Double quotes are also being used throughout the switch case block to set the `options_values['ERROR']`
I think it would be a good idea to use single quotes in other cases too.

documentation/man/features/kw-explore.rst
```diff
@@ -45,5 +45,9 @@ OPTIONS
 -H | \--only-header:
   With this option, it is possible to show only the results from the header.
 
+-C<num> | \--show-context=<num>:
+  Show <num> lines of additional context above and below the matched line.
+  If <num> is not specified, the default value of 3 will be used.
```
rodrigosiqueira, line 50: How about creating an example section and adding one or two examples of this feature?

src/_kw
```diff
@@ -405,7 +406,7 @@ _kw_man()
   _values 'kw commands' \
     'backup' 'build' 'codestyle' 'config' 'debug' 'deploy' 'device' 'diff' 'drm' 'env' \
     'explore' 'init' 'mail' 'maintainers' 'pomodoro' 'remote' 'report' 'self-update' 'ssh' \
-    'patch-hub' 
+    'patch-hub'
```
rodrigosiqueira, line 409: Is this change part of this PR, or is this unrelated?

src/explore.sh
```diff
@@ -158,6 +161,22 @@ function parse_explore_options()
...
+          options_values['ERROR']='Invalid arguments: Multiple context values!'
+          return 22 # EINVAL
+        fi
+
+        # Check if the next argument exists and is an integer
+        if [[ ! "$2" =~ ^[0-9]+$ ]]; then
+          options_values['CONTEXT']=3
+        else
+          options_values['CONTEXT']="$2"
+          shift
```
rodrigosiqueira, line 177: Don't you need to use `shift 2` since it can have a second item?

tests/unit/explore_test.sh
```diff
@@ -67,9 +67,18 @@ function test_explore_files_under_git_repo()
...
+  expected_result="git grep --context 0 -e 'GNU grep' --line-number -I ."
   assertEquals "($LINENO)" "$expected_result" "$output"
 
+  # Test for non zero context value
+  output=$(explore_main --show-context=5 'GNU grep' '.' 'TEST_MODE')
+  expected_result="git grep --context 5 -e 'GNU grep' --line-number -I ."
+  assertEquals "($LINENO)" "$expected_result" "$output"
+
+  MSG_OUT='7'
+  output=$(explore_main --show-context=3 'avoid' codestyle_error.c | wc -l)
```
rodrigosiqueira, line 79: Use `--lines` instead of `-l`.

tests/unit/explore_test.sh
```diff
@@ -67,9 +67,18 @@ function test_explore_files_under_git_repo()
...
+  assertEquals "($LINENO)" "$MSG_OUT" "$output"
```
rodrigosiqueira, line 80: Counting the number of lines is good, but you could also add some tests that check for the string before and after the match. You can leverage some of the functions in the `test/unit/utils.sh`; for example, `assert_substring_match` could be handy for this type of test, but we have other options.

tests/unit/explore_test.sh
```diff
@@ -67,9 +67,18 @@ function test_explore_files_under_git_repo()
   assertEquals "($LINENO)" "$MSG_OUT" "$output"
 
   output=$(explore_main 'GNU grep' '.' 'TEST_MODE')
-  expected_result="git grep -e 'GNU grep' -nI ."
+  expected_result="git grep --context 0 -e 'GNU grep' --line-number -I ."
   assertEquals "($LINENO)" "$expected_result" "$output"
 
+  # Test for non zero context value
```
rodrigosiqueira, line 73: btw, the explore test has some bad test patterns, such as adding many checks in the same function. Since you are adding a new feature, I recommend you create specific functions for your tests. For example, you could create a function named `test_explore_with_git_and_context`. Ideally, we want one test function per use case since this keeps things easy to debug and update.

tests/unit/explore_test.sh
```diff
@@ -236,6 +257,20 @@ function test_parse_explore_options()
   assertEquals "($LINENO)" '0' "$ret"
   assertEquals "($LINENO)" 'HEADER' "${options_values['SCOPE']}"
 
+  unset options_values
+  declare -gA options_values
+  parse_explore_options -C
+  ret="$?"
+  assertEquals "($LINENO)" '0' "$ret"
```
rodrigosiqueira, line 264: Use `(${LINENO})` in all of the assertions, and you don't need quotes around numbers.

rodrigosiqueira (changes requested):

src/_kw
sahil-sagwekar2652: No, this shouldn't be here.

sahil-sagwekar2652:
Hello, @rodrigosiqueira 
I'll send a commit ASAP, was caught up in college work.

src/explore.sh
sahil-sagwekar2652: There is a `shift` statement outside the else block. It adds up to have the same effect.

tests/unit/explore_test.sh
sahil-sagwekar2652: I think this change would be out of place for this PR. Would it be a good idea to change this in another commit?

sahil-sagwekar2652:
Hello @rodrigosiqueira ,
I have added a new test function and also added some examples in the documentation.
I have not tackled [this](https://github.com/kworkflow/kworkflow/pull/1058#discussion_r1550670863) comment of yours in this commit.

sahil-sagwekar2652:
Hello @rodrigosiqueira ,
I have added a new test function and also added some examples in the documentation.
I have not tackled [this](https://github.com/kworkflow/kworkflow/pull/1058#discussion_r1550670863) comment of yours in this commit.

tests/unit/explore_test.sh
davidbtadokoro: I think Siqueira was alluding to changing all occurrences in the additions of this PR as we are concatenating the parenthesis with the value of `LINENO`. The pattern in this file is misleading, so don't use it as basis :smile:

src/_kw
```diff
@@ -282,6 +282,7 @@ _kw_explore()
     '(-a --all -l --log -g --grep)'{-a,--all}'[search for all matches in files under or not git management]' \
     '(-c --only-source -H --only-header)'{-c,--only-source}'[show only results from the source]' \
     '(-H --only-header -c --only-source)'{-H,--only-header}'[show only results from the header]' \
+    '(-C --show-context)'{-C,--show-context}'[set the context value]' \
```
davidbtadokoro, line 285: The short option accepts arguments that come directly after, like in `-C5`, and the long option accepts arguments after the equal sign, like in `--show-context=5`. However, the Zsh completions leave a blank space after completing `-C` and `--show-context,` both of which will result in a parser error and a bad experience for the user. To fix this, change the line to

```bash
'(-C --show-context)'{-C-,--show-context=-}'[set the context value]' \
```

with the minus after `-C` and an equal sign and minus after `--show-context` inside the curly braces.

You probably shouldn't know this if you never played with Zsh completions (which aren't intuitive), so don't worry.

davidbtadokoro (changes requested):
First, thanks again for the update and dedication to this change (it looks neat).

I've reviewed the comments of @rodrigosiqueira, and it seems that you've tackled all of them besides one, which I commented on. I've resolved the conversations that were clearly resolved but left the ones that can be interpretative open for Siqueira to check if they follow. I also left a comment which is more about how the native completions system for Zsh is wild and obscure :sweat_smile:

sahil-sagwekar2652:
Hello @davidbtadokoro ,
I have changed `($LINENO)` to `(${LINENO})` only in context of my changes. We could change the other occurences in a separate PR. And I've also modified the `_kw` file according to your suggestion, thanks again.

davidbtadokoro:
Hi @sahil-sagwekar2652, and thank you so much for the update! @rodrigosiqueira, I know I've already said this in this PR, but now I think everything looks legit on my end, so will wait for your green light. Tested it again in default, grep and all mode, and both options (short and long) work great (and the Zsh completions as well).

tests/unit/explore_test.sh
```diff
@@ -172,6 +185,47 @@ function test_explore_git()
   }
 }
 
+function test_explore_context()
+{
+  local -r current_path="$PWD"
+  local EXPECTED_CONTEXT='3'
+  local EXPECTED_MATCH='avoid'
+  local MSG_OUT='7'
```
rodrigosiqueira, line 193: Why are you using upper case here? Lowercase makes more sense here.

tests/unit/explore_test.sh
```diff
@@ -172,6 +185,47 @@ function test_explore_git()
...
+  }
+
+  # Check the number of output lines
+  output=$(explore_main --show-context=3 'avoid' codestyle_error.c | wc --lines)
+  assertEquals "(${LINENO})" "$MSG_OUT" "$output"
+
+  output=$(explore_main --show-context="$EXPECTED_CONTEXT" "$EXPECTED_MATCH" codestyle_error.c)
+
+  # Check if the expected match and context lines are present in the output
+  assert_substring_match "Expected match not found!" ${LINENO} "${EXPECTED_MATCH}" "$output"
```
rodrigosiqueira, line 207: Use single quotes in the string and add double quotes around `${LINENO}`.

tests/unit/explore_test.sh
```diff
@@ -172,6 +185,47 @@ function test_explore_git()
...
+
+  # Check context lines below the match
+  for ((i = 1; i <= EXPECTED_CONTEXT; i++)); do
+    CONTEXT_LINE=$((i + 4)) # Assuming match is on line 4
+    CONTEXT_LINE_CONTENT=$(echo "$output" | head -n $CONTEXT_LINE | tail -n 1)
```
rodrigosiqueira, line 212: use `printf` instead of `echo`.
https://kworkflow.org/content/codingstyle.html#use-printf-instead-of-echo

Also use --lines= instead of -n

and add double quotes around CONTEXT_LINE.

tests/unit/explore_test.sh
```diff
@@ -172,6 +185,47 @@ function test_explore_git()
...
+    assert_line_match "Context line $CONTEXT_LINE below match" "$CONTEXT_LINE_CONTENT" "$output"
```
rodrigosiqueira, line 213: add `{}` around `CONTEXT_LINE`.

tests/unit/explore_test.sh
```diff
@@ -172,6 +185,47 @@ function test_explore_git()
...
+  done
+
+  # Check context lines above the match
+  for ((i = 1; i <= EXPECTED_CONTEXT; i++)); do
+    CONTEXT_LINE=$((4 - i)) # Assuming match is on line 4
+    CONTEXT_LINE_CONTENT=$(echo "$output" | head -n $CONTEXT_LINE | tail -n 1)
+    assert_line_match "Context line $CONTEXT_LINE above match" "$CONTEXT_LINE_CONTENT" "$output"
+  done
```
rodrigosiqueira, line 221: Check my previous comments since they also apply here.

tests/unit/explore_test.sh
```diff
@@ -236,6 +290,20 @@ function test_parse_explore_options()
...
+  unset options_values
+  declare -gA options_values
+  parse_explore_options -C
+  ret="$?"
+  assertEquals "(${LINENO})" '0' "$ret"
+  assertEquals "(${LINENO})" '3' "${options_values['CONTEXT']}"
+
+  unset options_values
+  declare -gA options_values
+  parse_explore_options --show-context=5
```
rodrigosiqueira, line 302: Could you also add a test with an invalid option?

src/explore.sh
```diff
@@ -267,5 +289,6 @@ function explore_help()
     '  explore,e --all,-a <string> - Search for all <string> match under or not of git management' \
     '  explore,e --only-source,-c <string> - Search for all <string> in source files' \
     '  explore,e --only-header,-H <string> - Search for all <string> in header files' \
+    '  explore,e --show-context=<num>,-C<num> - Print <num> lines of output context (default: 3)' \
```
rodrigosiqueira, line 292: Since num is optional, add [] around it (e.g., --show-context[=<num>]

documentation/man/features/kw-explore.rst
```diff
@@ -45,5 +45,23 @@ OPTIONS
 -H | \--only-header:
   With this option, it is possible to show only the results from the header.
 
+-C<num> | \--show-context=<num>:
```
rodrigosiqueira, line 48: Since num is optional, add [] around it, e.g., --show-context[=<num>].

rodrigosiqueira (changes requested):

sahil-sagwekar2652:
Hello @rodrigosiqueira , I have tried to address all your comments, except the one for using `--lines` instead of `-n`,
I thought using `-n` with the `head` and `tail` commands is a well-known convention which is why I went with the short-hand flag.

There are many more instances of using the short flag with `head` and `tail` within the file, if we are going to be updating it to `--lines` should we do it for all occurrences? Please let me know your thoughts on this. Thanks.

rodrigosiqueira:
Hi @sahil-sagwekar2652 ,
First of all, I'm sorry for my huge delay in reviewing this change. Anyway, it looks great! I already merged it into the unstable branch.

Thanks a lot.

PR #1059, opened 2024-03-19
Add the Handbook index on documentation

dudszzz:
There is some information not covered by the current documentation, mainly those related to the project workflow that are not clear to new contributors.

add the new Handbook index in index.rst file.

### DIFF
--- a/documentation/index.rst
+++ b/documentation/index.rst
@@ -26,6 +26,16 @@ check:
    content/developmentworkflow
    content/tracingandprofiling
 
+Kw Handbook
+===========
+
+Device and tips for new contributors.
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Handbook:
+
+
 Kw code doc
 ===========
 


documentation/index.rst
```diff
@@ -26,6 +26,16 @@ check:
    content/developmentworkflow
    content/tracingandprofiling
 
+Kw Handbook
```
davidbtadokoro, line 29: We should make it clear to readers that this handbook is directed to contributors, so make it something like `Kw Contributor's Handbook` (feel free to propose other possibilities). We need to nail the name here as this may be decisive on converting newcomers as contributors.

documentation/index.rst
```diff
@@ -26,6 +26,16 @@ check:
...
+===========
+
+Device and tips for new contributors.
```
davidbtadokoro, line 32: On the same note as above, we need to "sell" to the reader that all that new contributors need to know is included in this section. In this sense, reformulate the sentence to vehiculate this information.

davidbtadokoro (changes requested):
Hey @dudszzz, and thank you so much for your first contribution!

In terms of the project's practices and processes, you nailed it!

Just a minor comment on your commit message (it looks great overall, btw): in the last sentence, add something like `In this sense, add (...)` or `Taking this into account, add (...)`. Also, update the message with the idea that the handbook will be destined for contributors.

Issue #666, opened 2022-08-06
Enable kw build to build a range of commit


rodrigosiqueira:
**Description:**

When kernel developers prepare a patchset, they have to check if every commit compiles without any issue. They usually use something like this:

```
git rebase -i --exec "yes \"\" | make prepare ; make -j16 modules M=drivers/gpu/drm/amd/amdgpu"
```

**API draft:**

```
kw build --from-sha SHA [--to-sha SHA]
```

PR #1060, opened 2024-03-20
add from-sha option for kw build

MarceloSpessoto:
implement kw build --from-sha option requested in issue #666. By default, it builds every commit from requested hash to actual commit (changing target commit is something to be added in the future with to-sha option).


src/build.sh
```diff
@@ -362,6 +382,11 @@ function parse_build_options()
         options_values['LOG_PATH']="$2"
         shift 2
         ;;
+      --from-sha)
+        options_values['FROM_SHA_ARG']="$2"
```
davidbtadokoro, line 386: We should check the `$2` argument to see if it is a valid past commit.

src/build.sh
```diff
@@ -362,6 +382,11 @@ function parse_build_options()
...
+        echo "${options_values['FROM_SHA_ARG']}"
```
davidbtadokoro, line 387: Remove this line.

src/build.sh
```diff
@@ -394,7 +419,8 @@ function build_help()
     '  build (-c | --clean) - Clean option integrated into env' \
     '  build (-f | --full-cleanup) - Reset the kernel tree to its default option integrated into env' \
     '  build (--cflags) - Customize kernel compilation with specific flags' \
-    '  build (--verbose) - Show a detailed output'
+    '  build (--verbose) - Show a detailed output' \
+    '  build (--from-sha) - Build all commits from given commit to actual commit'
```
davidbtadokoro, line 423: Don't forget to indicate the argument of the option.

davidbtadokoro (changes requested):
Hi @MarceloSpessoto, and thanks for the change! It seems like it can provide for many workflows :+1: 

I tested the change, and it does what it proposes. However, the biggest problem I noticed is that we aren't building the kernel in "using kw" for each commit, just directly running `make`. I left some other comments, so please ping me if you have any doubts.

Also, as you are adding a new option to the feature, you need to implement its Bash and Zsh completions. Don't forget to add test cases for this option as well!

@rodrigosiqueira, as you've opened the issue related to this change, give your feedback when you can.

src/build.sh
```diff
@@ -247,9 +254,21 @@ function full_cleanup()
   cmd_manager "$flag" "$cmd"
 }
 
+function from_sha()
```
rodrigosiqueira, line 257: Don't forget to add documentation. Also, I think we can find a better name for this function... maybe something like `build_from_sha`

src/build.sh
```diff
@@ -247,9 +254,21 @@ function full_cleanup()
...
+{
+  local flag="$1"
+  local env_path="$2"
+  local sha="$3"
+  local cmd
+
+  flag=${flag:-'SILENT'}
+  cmd="git rebase ${sha} --exec 'yes \"\" | make prepare${env_path}; make${env_path}'"
+  cmd_manager "$flag" "$cmd"
+}
```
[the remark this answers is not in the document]
rodrigosiqueira: > If I am not mistaken, we should find a way to make the --from-sha option only automate the process of the user directly running git rebase -i, stopping in every commit, and running kw build

I'm not sure if I got what you mean here... As a user of this feature, I want to use: `kw b --from-sha <SHA>`, and I only want to stop in the middle of the rebase to take any action if it fails to build for some reason.

> @rodrigosiqueira, do you have any thoughts on this?

I liked the second approach since it will bring all of the kw build benefits, such as the parallel build.

src/build.sh
```diff
@@ -247,9 +254,21 @@ function full_cleanup()
   cmd_manager "$flag" "$cmd"
 }
 
+function from_sha()
+{
+  local flag="$1"
+  local env_path="$2"
+  local sha="$3"
```
rodrigosiqueira, line 261: You should not trust that SHA is valid. You must check it first.

src/build.sh
```diff
@@ -247,9 +254,21 @@ function full_cleanup()
...
+  local cmd
+
+  flag=${flag:-'SILENT'}
+  cmd="git rebase ${sha} --exec 'yes \"\" | make prepare${env_path}; make${env_path}'"
```
rodrigosiqueira, line 265: You can't start to run git rebase directly; you first need to check if the repository is in a good state to run it.

rodrigosiqueira (changes requested):

flake.nix
```diff
@@ -0,0 +1,65 @@
+{
```
rodrigosiqueira, line 1: @MarceloSpessoto , I think you added this commit by accident, right?

MarceloSpessoto: yes, going to remove it

MarceloSpessoto:
I've made most of the requested changes.  I am soon going to add the zsh completion and the repository and sha validation check. 
Do you want the test cases to be added in this PR or should we open another issue/PR for that?

documentation/man/features/kw-build.rst
```diff
@@ -102,6 +103,10 @@ OPTIONS
     | **sv** or **vs** enables both.
     | **n** (or any other option) disables notifications (this is the default).
 
+\--from-sha:
+  Build every commit from <SHA> to actual commit. Useful for testing if all patches in
```
rodrigosiqueira, line 107: Maybe: [..] from <SHA> to the branch head.

documentation/man/features/kw-build.rst
```diff
@@ -102,6 +103,10 @@ OPTIONS
     | **sv** or **vs** enables both.
     | **n** (or any other option) disables notifications (this is the default).
 
+\--from-sha:
```
rodrigosiqueira, line 106: How about expanding the build tutorial by adding this option?
https://kworkflow.org/tutorials/buildlinux.html

src/build.sh
```diff
@@ -247,9 +254,20 @@ function full_cleanup()
   cmd_manager "$flag" "$cmd"
 }
 
+function build_from_sha()
```
rodrigosiqueira, line 257: Add function documentation.

src/build.sh
```diff
@@ -247,9 +254,20 @@ function full_cleanup()
...
+{
+  local flag="$1"
+  local sha="$2"
+  local cmd
+
+  flag=${flag:-'SILENT'}
+  cmd="git rebase ${sha} --exec 'kw build'"
```
rodrigosiqueira, line 264: Before using `sha` validate whether it is valid.

rodrigosiqueira (changes requested):
Hi @MarceloSpessoto ,
This PR looks great! I tested it, and it is working as expected. Could you please address my comments and also add unit tests?

Thanks

MarceloSpessoto:
@davidbtadokoro @rodrigosiqueira 
I've implemented most of the requested changes:
+ Documented the use of this option in the kw tutorials page
+ Wrote the function documentation. I wrote it using the template from `build_kernel_main()`, but noticed many other functions are not properly documented. Should a tracking issue be open to record the function documentation of each main file?
+ Modified the text descriptions as requested.
+ Made verifications relative to the use of the given SHA. The verification of being used in a valid repo was already present for `kw build` by default, so I didn't touch it. Others, such as being a real commit and an ancestor of the branch head were implemented. Since `--from-sha` relies on a `git rebase`, be alert that a possible bug here may mess with the git tree. I think it is safe for use, but your review of it will be very important.
+ While fixing the commit, I found out about the lack of proper handling of rebases in progress. I've added a check of rebase in progress before trying the compilation (warning the user to abort the rebase before proceeding, if there is any). I've placed a mechanism to abort the rebase if `kw build` fails to compile one of the commits (which freezes the rebase of `--from-sha`).
+ When a verification fails, it returns error 22. I am not sure if this is the proper way of handling these.

I'll wait for your next review and, if there is no major issue, I will proceed to finish the unit tests and zsh autocomplete.

src/build.sh
```diff
@@ -124,6 +126,35 @@ function build_kernel_main()
     return "$?"
   fi
 
+  if [[ -n "$from_sha_arg" ]]; then
+    # Check if there is a rebase in process.
+    printf '%s\n' "Verifying if there is a rebase in process..."
+    if [[ -e .git/rebase-merge/done ]]; then
+      printf '%s\n' "ERROR: Abort the repository rebase before continuing with build from sha (use 'git rebase --abort')!"
+      return 22
```
rodrigosiqueira, line 134: Usually, in the kw, we add a comment after the return error that shows the error. For example, return 22 # EINVAL

src/build.sh
```diff
@@ -124,6 +126,35 @@ function build_kernel_main()
...
+    fi
+
+    # Check if given SHA represents real commit
+    eval "git cat-file -e ${from_sha_arg}^{commit} 2> /dev/null"
```
rodrigosiqueira, line 138: Why not use `cmd_manager`?

src/build.sh
```diff
@@ -124,6 +126,35 @@ function build_kernel_main()
...
+    if [[ "$?" != 0 ]]; then
+      printf '%s\n' "ERROR: The given SHA doesn't represent a real commit in this repository."
+      return 22
+    fi
+
+    # Check if given SHA is in working tree.
+    local sha_base
+    local merge_base
```
rodrigosiqueira, line 146: local variables should be at the beginning of the function.

src/build.sh
```diff
@@ -247,9 +278,34 @@ function full_cleanup()
...
+  local sha="$2"
+  local cmd
+
+  flag=${flag:-'SILENT'}
+  cmd="git rebase ${sha} --exec 'kw build'"
+  cmd_manager "$flag" "$cmd"
+
+  if [[ "$?" != 0 ]]; then
+    echo "kw build failed during the compilation of a patch! Aborting rebase..."
+    git rebase --abort
```
rodrigosiqueira, line 301: https://kworkflow.org/content/codingstyle.html#use-printf-instead-of-echo

If it fails in some specific commit, this is useful since the developer can try to fix the issue. Instead of just abort, ask the user if they want to abort.

rodrigosiqueira (changes requested):

src/build.sh
```diff
@@ -124,6 +126,35 @@ function build_kernel_main()
     return "$?"
   fi
 
+  if [[ -n "$from_sha_arg" ]]; then
+    # Check if there is a rebase in process.
+    printf '%s\n' "Verifying if there is a rebase in process..."
+    if [[ -e .git/rebase-merge/done ]]; then
```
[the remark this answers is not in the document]
matheustavares: Hmm, git prompt does this by checking for the presence of `.git/rebase-merge` as well: https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh#L452

You may want to take a look at that function as it also checks for apply/cherry-pick/bisect, etc., which could also get in the way of the feature you are working on.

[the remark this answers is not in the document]
MarceloSpessoto: @matheustavares , thanks for the suggestions, added verifications for merge, bisect and apply!

src/build.sh
MarceloSpessoto: I've added a warning message reporting the failed rebase.

MarceloSpessoto:
Hey @rodrigosiqueira and @davidbtadokoro, I've implemented the requested changes.

MarceloSpessoto:
The current `--from-sha` implementation ignores other `kw build` options, such as `--cpu-scaling`. Do you want me to add support for that? If you do, which options would you want to work along with `--from-sha`?

src/_kw
davidbtadokoro, line 1: FWIU, `--from-sha` is mutually exclusive with every other option, except `--verbose`, i.e., if any other option is already on the prompt then `--from-sha` shouldn't be suggested, and vice-versa. If I am right, you should add `--from-sha` inside the parenthesis of every option except verbose

src/build.sh
```diff
@@ -124,6 +128,41 @@ function build_kernel_main()
     return "$?"
   fi
 
+  if [[ -n "$from_sha_arg" ]]; then
+    # Check if there is a rebase in process.
+    if [[ -d .git/rebase-merge ]]; then
+      printf '%s\n' 'ERROR: Abort the repository rebase before continuing with build from sha (use "git rebase --abort")!'
+      return 22 #EINVAL
```
davidbtadokoro, line 135: Add a space between `#` and `EINVAL`. Do it for every addition you've introduced like this

src/build.sh
```diff
@@ -124,6 +128,41 @@ function build_kernel_main()
...
+    elif [[ -f .git/MERGE_HEAD ]]; then
+      printf '%s\n' 'ERROR: Abort the repository merge before continuing with build from sha (use "git rebase --abort")!'
+      return 22 #EINVAL
+    elif [[ -f .git/BISECT_LOG ]]; then
+      printf '%s\n' 'ERROR: Stop the repository bisect before continuing with build from sha (use "git bisect reset")!'
+      return 22 #EINVAL
+    elif [[ -d .git/rebase-apply ]]; then
+      printf '%s\n' 'ERROR: Abort the repository patch apply before continuing with build from sha (use "git am --abort")!'
+      return 22 #EINVAL
```
davidbtadokoro, line 144: IMHO, for all the checks above, we get to an error not because the argument is invalid (the meaning of 22 EINVAL), but because we shouldn't try to start a new `git rebase` in any of the cases above. That being said, I don't know an errno that would fit perfectly, but I think that 125 ECANCELED works better.

@rodrigosiqueira, what are your thoughts on this?

src/build.sh
```diff
@@ -124,6 +128,41 @@ function build_kernel_main()
...
+    fi
+
+    # Check if given SHA represents real commit
+    cmd_manager "SILENT" "git cat-file -e ${from_sha_arg}^{commit} 2> /dev/null"
+    if [[ "$?" != 0 ]]; then
+      printf '%s\n' 'ERROR: The given SHA does not represent a valid commit sha.'
```
davidbtadokoro, line 150: Maybe we can output the value of `from_sha_arg` in this print to assert the user about the value being used.

src/build.sh
```diff
@@ -124,6 +128,41 @@ function build_kernel_main()
...
+      return 22 #EINVAL
+    fi
+
+    # Check if given SHA is in working tree.
+    sha_base=$(git rev-parse --verify "$from_sha_arg")
+    merge_base=$(git merge-base "$from_sha_arg" HEAD)
+    if [[ "$sha_base" != "$merge_base" ]]; then
+      printf '%s\n' 'ERROR: Given SHA is invalid. Check if it is an ancestor of the branch head.'
```
davidbtadokoro, line 158: Maybe we can output the value of `from_sha_arg` in this print to assert the user about the value being used.

davidbtadokoro (changes requested):
Hey @MarceloSpessoto, and thank you so much for the constant updates!

I reviewed the code and found mostly just nitpicky comments about typos and other minor stuff. Pay close attention to my comment on `_kw` to nail the correct Zsh completion behavior.

The most useful review I have is that we should make clear to the user that the commit `<SHA>` won't be built, in the sense that, if we have a git log like `A -> B -> C <- HEAD`, and ran `kw build --from-sha A`, only commits `B`, and `C` will be built.

Besides that, tested both invalid and valid cases and the PR looks really legit :+1:

src/build.sh
rodrigosiqueira: @MarceloSpessoto

src/build.sh
rodrigosiqueira: I agree.

Actually, I'm also thinking about this code. How about creating a new file under `src/lib/` named `git.sh`? We should add checks like this in a library since many parts of kw rely on git, and we constantly reinvent the wheel in different files. @MarceloSpessoto, you don't need to do it here, but it would be nice if you opene an issue to address this and create a PR moving this code to this new lib file.

src/build.sh
rodrigosiqueira: I agree with David, and also use `complain`.

rodrigosiqueira (changes requested):

tests/unit/build_test.sh
```diff
@@ -861,6 +861,121 @@ function test_kernel_build_cpu_scaling_llvm_warning_sava_log_to()
...
+  local sha="HEAD^"
+
+  mk_fake_git
+  mkdir .git/rebase-merge
+
+  output=$(build_kernel_main 'TEST_MODE' --from-sha ${sha})
+  declare -a expected_result=(
+    'ERROR: Abort the repository rebase before continuing with build from sha (use "git rebase --abort")!'
+  )
+  compare_command_sequence '' "($LINENO)" 'expected_result' "$output"
```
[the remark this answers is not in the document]
MarceloSpessoto: I don't understand where I should apply a string concatenation fix

rodrigosiqueira:
Thanks a lot for all your updates. I merged this change into the unstable branch.

PR #1053, opened 2024-02-29
Kw mail group

JGBSouza:
Add basic resolutions for #657 and #550. The code still need improvments but a initial review would be nice

Issue #659, opened 2022-08-06
Save cover-letter


rodrigosiqueira:
**Description:**

It is so frustrating when you write a giant cover letter, and we have a failure to send... we have to write everything again if we did not save it.

**Why:**

Avoid developer's frustration with the public mailing lists.

**Targeted users:**

* Developers that uses cover letter.

**API draft:**

Maybe we can integrate this feature with:
https://github.com/kworkflow/kworkflow/issues/658

However, we could do something like "*.swp" files like vim. In other words, before we send the file, we save the cover letter in a temporary file. If we fail to send the patch, we keep this file, and the next time that the developer wants to use kw mail, we ask the user if they want to reuse the cover letter. If everything is fine in sending, we can remove this temporary file.

JGBSouza:
Working on it!

Issue #658, opened 2022-08-06
Mail Dashboard


rodrigosiqueira:
**Description/Why:**

As a kernel developer, I constantly send patches. Some of them are like sidequest work; after completing the task, I tend to forget to follow up with the community about the patch. Additionally, sometimes I want to see the patches that I sent.

[Explain why do we want this feature]

* Developers that use public mailing list

**API draft:**

For this feature, we probably want:
* Registers patches that the developer sent
* Provide a simple way to list recent patches
* Provide some metadata mechanism that allows developers to update the status of some specific patch manually

Follows the first draft for this feature:

* Every patch sent, should be registered in the kw database. Next, we should add it as part of kw report. Something like this:
```bash
...
# Patches
 * [PATCH V2] something/la: this is something
 * [PATCH] something/la: this is something
```

* List recent patches

```bash
kw mail --dashboard
ID Date                              Status             Title
1   2022/04/23:15:00                                 [PATCH V2] something/la: this is something
2   2022/06/23:15:00                                 [PATCH] something/la: this is something
...
```
We should be able to use a range
```
```bash
kw mail --dashboard --from "YYYY/MM/DD"
ID Date                              Status             Title
1   2022/04/23:15:00                                [PATCH V2] something/la: this is something
2   2022/06/23:15:00                                [PATCH] something/la: this is something
...
```

* Change patch status

It would be nice if we can say if a patch was merged or not, and use it for generate statistic data later on. Maybe we can have something like:
```
```bash
kw mail --dashboard --patch-id 1 merged
kw mail --dashboard
ID Date                              Status             Title
1   2022/04/23:15:00         Merged            [PATCH V2] something/la: this is something
...

P.s.: Maybe we can create a new feature under kw, instead of use `mail`, e.g., kw patches or kw patch-track.

@melissawen @kwy95 I would like to hear your thoughts about this feature idea.

JGBSouza:
Working on it!

PR #1061, opened 2024-03-21
Update database handling functions

JGBSouza:
This changes will be usefull to the other issues i'm working on (#1053/#657 and #658, #659)

### DIFF
--- a/src/lib/kw_db.sh
+++ b/src/lib/kw_db.sh
@@ -221,7 +221,7 @@ function select_from()
     complain 'Empty table.'
     return 22 # EINVAL
   fi
- 
+
   if [[ -n "$_condition_array" ]]; then
     where_clause="$(generate_where_clause "$_condition_array")"
   fi
@@ -231,6 +231,61 @@ function select_from()
   if [[ -n "${order_by}" ]]; then
     query="${query::-2} ORDER BY ${order_by} ;"
   fi
+
+  cmd="sqlite3 -init "${KW_DB_DIR}/pre_cmd.sql" -cmd \"${pre_cmd}\" \"${db_path}\" -batch \"$query\""
+  cmd_manager "$flag" "$cmd"
+}
+
+# This function update the set of values in the table of given database
+# with the given conditions.
+#
+# @table:     Table to select info from
+# @set:   An array reference of condition pairs that will be
+# updated in the db
+# @pre_cmd:   Pre command to execute
+# @_condition_array: An array reference of condition pairs
+# specifing the data that will be updated
+# @flag:      Flag to control function output
+# @db:        Name of the database file
+# @db_folder: Path to the folder that contains @db
+#
+# Return:
+# 2 if db doesn't exist; 22 if table is empty
+# 0 if succesful; non-zero otherwise
+function update_into()
+{
+  local table="$1"
+  local _updates_array="$2"
+  local pre_cmd="$3"
+  local _condition_array="$4"
+  local flag=${5:-'SILENT'}
+  local db="${6:-"$DB_NAME"}"
+  local db_folder="${7:-"$KW_DATA_DIR"}"
+  local where_clause=''
+  local db_path
+  local query
+
+  db_path="$(join_path "$db_folder" "$db")"
+
+  if [[ ! -f "$db_path" ]]; then
+    complain 'Database does not exist'
+    return 2
+  fi
+
+  if [[ -z "$table" ]]; then
+    complain 'Empty table.'
+    return 22 # EINVAL
+  fi
+
+  if [[ -z "$_condition_array" || -z "$_updates_array" ]]; then
+    complain 'Empty condition or updates array.'
+    return 22 #EINVAL
+  fi
+
+  where_clause="$(generate_where_clause "$_condition_array")"
+  set_clause="$(generate_set_clause "$_updates_array")"
+
+  query="UPDATE ${table} SET ${set_clause} ${where_clause} ;"
 
   cmd="sqlite3 -init "${KW_DB_DIR}/pre_cmd.sql" -cmd \"${pre_cmd}\" \"${db_path}\" -batch \"$query\""
   cmd_manager "$flag" "$cmd"
@@ -252,7 +307,7 @@ function generate_where_clause()
   local attribute
   local where_clause="WHERE "
   local value
-  
+
   for clause in "${!condition_array_ref[@]}"; do
     attribute="$(cut --delimiter=',' --fields=1 <<< "$clause")"
     value="${condition_array_ref["${clause}"]}"
@@ -266,6 +321,30 @@ function generate_where_clause()
   done
 
   printf '%s' "${where_clause::-5}" # Remove trailing ' AND '
+}
+
+# This function receives a condition_array and then generate
+# the infos that will be used by the SET clause to update
+# the data fields whe want.
+#
+# @condition_array_ref: The condition array containing the conditions
+#
+# Returns:
+# A string containing the generated clause
+function generate_set_clause()
+{
+  local -n condition_array_ref="$1"
+  local attribute
+  local set_clause
+  local value
+
+  for attribute in "${!condition_array_ref[@]}"; do
+    value="${condition_array_ref["${attribute}"]}"
+    set_clause+="${attribute} = '${value}'"
+    set_clause+=', '
+  done
+
+  printf '%s' "${set_clause::-2}" # Remove trailing ', '
 }
 
 # This function takes arguments and assembles them into the correct format to
--- a/tests/unit/kw_db_test.sh
+++ b/tests/unit/kw_db_test.sh
@@ -497,6 +497,85 @@ function test_remove_from()
   assert_equals_helper 'Wrong output' "$LINENO" "$expected" "$output"
 }
 
+function test_update_into()
+{
+  local columnns
+  local values
+  declare -A condition_array
+  declare -A updates_array
+  local output
+  local expected
+  local ret
+
+  # invalid operations
+  output=$(update_into 'table' 'updates_array' '' '' '' 'wrong/path/invalid_db.db')
+  ret="$?"
+  expected='Database does not exist'
+  assert_equals_helper 'Invalid db, error expected' "$LINENO" 2 "$ret"
+  assert_equals_helper 'Wrong error message' "$LINENO" "$expected" "$output"
+
+  output=$(update_into '' 'updates_array' '' '')
+  ret="$?"
+  expected='Empty table.'
+  assert_equals_helper 'Empty table, error expected' "$LINENO" 22 "$ret"
+  assert_equals_helper 'Wrong error message' "$LINENO" "$expected" "$output"
+
+  output=$(update_into table 'update_array' '' '')
+  ret="$?"
+  expected='Empty condition or updates array.'
+  assert_equals_helper 'Empty condition array expected' "$LINENO" 22 "$ret"
+  assert_equals_helper 'Wrong error message' "$LINENO" "$expected" "$output"
+
+  columns=$(concatenate_with_commas 'name' 'attribute1' 'attribute2' 'rank')
+
+  values=$(format_values_db 4 \
+    'name19' 'att1' 'att2' '10' \
+    'name20' 'att1' 'att2' '11' \
+    'name21' 'att1' 'att2' '2' \
+    'name22' 'att1' 'att2' '2')
+  insert_into 'fake_table' "$columns" "$values"
+
+  # update one row using one unique attribute
+  condition_array=(['name']='name19')
+  updates_array=(['attribute1']='att1.1')
+  update_into 'fake_table' 'updates_array' '' 'condition_array'
+  ret="$?"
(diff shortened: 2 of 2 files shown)


src/lib/kw_db.sh
```diff
@@ -232,6 +229,75 @@ function select_from()
...
+  fi
+
+  where_clause="$(generate_where_clause '_condition_array')"
+  query="SELECT ${columns} FROM ${table} WHERE ${where_clause} ;"
+
+  cmd="sqlite3 -init "${KW_DB_DIR}/pre_cmd.sql" \"${db_path}\" -batch \"$query\""
+  cmd_manager "$flag" "$cmd"
+}
+
+function generate_where_clause()
```
davidbtadokoro, line 286: Add test cases and documentation for this function.

src/lib/kw_db.sh
```diff
@@ -232,6 +229,75 @@ function select_from()
...
+{
+  local -n c_array="$1"
```
davidbtadokoro, line 288: I assume you are using this name because the use of `condition_array` or `_condition_array` probably causes problems (Bash, sometimes you are complicated :cry:). However, find a more descriptive name, like `conditions_array_ref` (surely there is something better than this :rofl:).

src/lib/kw_db.sh
```diff
@@ -232,6 +229,75 @@ function select_from()
...
+  local clause
+  local relational_op='='
+  local database_att=''
```
davidbtadokoro, line 291: Maybe just the name `attribute` is fine.

src/lib/kw_db.sh
```diff
@@ -232,6 +229,75 @@ function select_from()
...
+  local where_clause=''
+
+  for clause in "${!c_array[@]}"; do
+    database_att="$(cut -d',' -f1 <<< "$clause")"
+    att_value="${c_array["${clause}"]}"
```
davidbtadokoro, line 296: Maybe just `value` is good. Also, declare this variable as local.

src/lib/kw_db.sh
```diff
@@ -232,6 +229,75 @@ function select_from()
...
+
+    if [[ $clause == *,* ]]; then
```
davidbtadokoro, line 298: Use double-quotes around `$clause`.

src/lib/kw_db.sh
```diff
@@ -232,6 +229,75 @@ function select_from()
...
+function generate_where_clause()
+{
+  local -n c_array="$1"
+  local clause
+  local relational_op='='
+  local database_att=''
+  local where_clause=''
+
+  for clause in "${!c_array[@]}"; do
+    database_att="$(cut -d',' -f1 <<< "$clause")"
```
davidbtadokoro, line 295: Use the long option `--delimiter`, like described [here](https://kworkflow.org/content/codingstyle.html#avoid-short-options).

src/lib/kw_db.sh
```diff
@@ -232,6 +229,75 @@ function select_from()
...
+    att_value="${c_array["${clause}"]}"
+
+    if [[ $clause == *,* ]]; then
+      relational_op=$(cut -d',' -f2 <<< "$clause")
```
davidbtadokoro, line 299: Use the long option here too.

src/lib/kw_db.sh
```diff
@@ -232,6 +229,75 @@ function select_from()
...
+# @pre_cmd:   Pre command to execute
+# @_condition_array: An array reference of condition pairs
+# @flag:      Flag to control function output
+# @db:        Name of the database file
+# @db_folder: Path to the folder that contains @db
+#
+# Return:
+# 2 if db doesn't exist; 22 if table is empty
+# 0 if succesful; non-zero otherwise
+function select_from_where()
```
davidbtadokoro, line 246: I see much duplication with the function `select_from`. Can't we just expand the behavior of `select_from` to allow the addition of where conditions?

src/lib/kw_db.sh
```diff
@@ -232,6 +229,75 @@ function select_from()
...
+
+  for clause in "${!c_array[@]}"; do
+    database_att="$(cut -d',' -f1 <<< "$clause")"
+    att_value="${c_array["${clause}"]}"
+
+    if [[ $clause == *,* ]]; then
+      relational_op=$(cut -d',' -f2 <<< "$clause")
+    fi
+
+    where_clause+="$database_att${relational_op}'${att_value}'"
```
davidbtadokoro: Use curly braces around `database_att` to correctly  concatenate strings.

src/lib/kw_db.sh
```diff
@@ -174,14 +174,10 @@ function remove_from()
     return 22 # EINVAL
```
davidbtadokoro, line 174: In this commit message, you missed a colon in `src: lib: kw_db Update database handling functions ` between `kw_db` and `Update`.

src/lib/kw_db.sh
davidbtadokoro, line 298: Also, use regex instead of this syntax. So, make it something like
```bash
    if [[ "$clause" =~ <regexp> ]]; then
```

src/lib/kw_db.sh
```diff
@@ -275,6 +275,14 @@ function select_from_where()
   cmd_manager "$flag" "$cmd"
```
davidbtadokoro, line 275: As each commit should be atomic, the "code that tests the code" (tests) and documentation should be included in the same commit. So, you should squash this commit with the other one.

davidbtadokoro (changes requested):
Hi @JGBSouza, and thank you for this change! It is a great catch, as making more elaborate select queries and removals should be supported in our database lib.

Left some comments, so please see if you can address them. I would like to stress the comment about squashing both commits into one, as the code, the code that test the code, and its documentation should be into the same commit for it to be atomic.

davidbtadokoro:
Hi @JGBSouza, and thank you so much for the update! I gave it a quick look and it seems you've tackled all my comments. Going to give a thorough review later in the day :+1:

tests/unit/kw_db_test.sh
```diff
@@ -226,37 +226,95 @@ function test_select_from()
...
+  ret="$?"
+  expected=$(sqlite3 "$KW_DATA_DIR/kw.db" -batch 'SELECT "start_date","start_time","description" FROM "pomodoro" WHERE start_time = '2021-11-18' ;')
+  assert_equals_helper 'No error expected' "$LINENO" "$ret" 0
+  assert_equals_helper 'Wrong output' "$LINENO" "$output" "$expected"
+
+  output=$(select_from 'pomodoro' "$entries" '' 'condition_array')
+  ret="$?"
+  expected=$(sqlite3 "$KW_DATA_DIR/kw.db" -batch 'SELECT "start_date","start_time","description" FROM "pomodoro" WHERE start_time = '2021-11-18' ;')
+  assert_equals_helper 'No error expected' "$LINENO" "$ret" 0
+  assert_equals_helper 'Wrong output' "$LINENO" "$output" "$expected"
```
davidbtadokoro, line 270: These three cases are the same, so remove two of them.

tests/unit/kw_db_test.sh
```diff
@@ -226,37 +226,95 @@ function test_select_from()
...
+  expected=$(sqlite3 "$KW_DATA_DIR/kw.db" -batch 'SELECT "start_date","start_time","description" FROM "pomodoro" WHERE start_time > '2021-11-18' ;')
+  assert_equals_helper 'No error expected' "$LINENO" "$ret" 0
+  assert_equals_helper 'Wrong output' "$LINENO" "$output" "$expected"
+
+  condition_array=(['start_time,<=']='2021-11-18')
+  output=$(select_from 'pomodoro' "$entries" '' 'condition_array')
+  ret="$?"
+  expected=$(sqlite3 "$KW_DATA_DIR/kw.db" -batch 'SELECT "start_date","start_time","description" FROM "pomodoro" WHERE start_time <= '2021-11-18' ;')
+  assert_equals_helper 'No error expected' "$LINENO" "$ret" 0
+  assert_equals_helper 'Wrong output' "$LINENO" "$output" "$expected"
```
davidbtadokoro, line 305: This is also a duplicated case. Remove this

tests/unit/kw_db_test.sh
```diff
@@ -226,37 +226,95 @@ function test_select_from()
...
+
+  condition_array=(['start_time,<>']='2021-11-18')
+  output=$(select_from 'pomodoro' "$entries" '' 'condition_array')
+  ret="$?"
+  expected=$(sqlite3 "$KW_DATA_DIR/kw.db" -batch 'SELECT "start_date","start_time","description" FROM "pomodoro" WHERE start_time <> '2021-11-18' ;')
   assert_equals_helper 'No error expected' "$LINENO" "$ret" 0
   assert_equals_helper 'Wrong output' "$LINENO" "$output" "$expected"
 }
```
davidbtadokoro, line 320: Also, for all the `assert_equals_helper` calls of this function (`test_select_from`), invert the order of `output/ret` with `expected/<expected-ret-code>`, as number 3 of `assert_equals_helper` is the expected and number 4 is the actual.

src/lib/kw_db.sh
```diff
@@ -223,15 +222,132 @@ function select_from()
...
+# @_condition_array: An array reference of condition pairs
+# specifing the data that will be updated
+# @flag:      Flag to control function output
+# @db:        Name of the database file
+# @db_folder: Path to the folder that contains @db
+#
+# Return:
+# 2 if db doesn't exist; 22 if table is empty
+# 0 if succesful; non-zero otherwise
+function update_into()
```
davidbtadokoro, line 256: Make a dedicated commit for the implementation of this function (along with its doc and tests). It is indeed a necessary DB operation, but, for organizational reasons, we should separate it into an isolated commit.

src/lib/kw_db.sh
```diff
@@ -203,13 +201,14 @@ function select_from()
   local table="$1"
   local columns="${2:-"*"}"
   local pre_cmd="$3"
-  local order_by="$4"
-  local flag=${5:-'SILENT'}
-  local db="${6:-$DB_NAME}"
-  local db_folder="${7:-$KW_DATA_DIR}"
+  local -n _condition_array="$4"
```
davidbtadokoro, line 204: About the comment of making `select_from` accept an empty `@_condition_array` value, this line will cause problems if the argument is empty, as Bash will complain:
```bash
local: `': not a valid identifier
```

I managed to alter the function to accept an empty value (without changing the behavior of when the value isn't empty) with the below:

```
diff --git a/src/lib/kw_db.sh b/src/lib/kw_db.sh
index 80169c4..3039752 100644
--- a/src/lib/kw_db.sh
+++ b/src/lib/kw_db.sh
@@ -201,7 +201,7 @@ function select_from()
   local table="$1"
   local columns="${2:-"*"}"
   local pre_cmd="$3"
-  local -n _condition_array="$4"
+  local _condition_array="$4"
   local order_by=${5:-''}
   local flag=${6:-'SILENT'}
   local db="${7:-"$DB_NAME"}"
@@ -224,8 +224,8 @@ function select_from()
 
   query="SELECT ${columns} FROM ${table} ;"
 
-  if [[ -n "${!_condition_array[*]}" ]]; then
-    where_clause="$(generate_where_clause '_condition_array')"
+  if [[ -n "$_condition_array" ]]; then
+    where_clause="$(generate_where_clause "$_condition_array")"
     query="${query::-2} WHERE ${where_clause} ;"
   fi
```

 As in `select_from` we don't do any associative array manipulations (only `generate_where_clause`, FWIU) we can only check to see if `_condition_array` is empty or not and treat it as a normal Bash variable (not a `local -n` reference), which won't complain about it being empty. However, you **must** pass an empty value, in case there is no `WHERE` clause (otherwise, this will cause an error).

I leave to you to decide if you want to use my approach, but confirm if this doesn't invalidates the non-empty cases.

tests/unit/kw_db_test.sh
```diff
@@ -226,37 +226,95 @@ function test_select_from()
...
   expected='Empty table.'
   assert_equals_helper 'Empty table, error expected' "$LINENO" "$ret" 22
   assert_equals_helper 'Expected error msg' "$LINENO" "$output" "$expected"
 
   # valid
-  output=$(select_from 'tags' 'tag')
+  output=$(select_from 'pomodoro' "$entries" '' 'condition_array')
   ret="$?"
-  expected=$(sqlite3 "$KW_DATA_DIR/kw.db" -batch 'SELECT tag FROM tags;')
+  expected=$(sqlite3 "$KW_DATA_DIR/kw.db" -batch 'SELECT * FROM "pomodoro" ;')
```
davidbtadokoro, line 247: Same comment about keeping calls where there is no `WHERE` clause unchanged.

davidbtadokoro (changes requested):
Hi @JGBSouza, and thank you so much for the new update. We are really close to merging, IMHO.

It may seem like a lot of comments, but there are two main points:

1. Make `select_from` accept an empty `@_condition_array` argument, and keep calls to it unchanged for when there is no `WHERE` clause;
2. Make the introduction of `update_into` (and `generate_set_clause`) on a dedicated commit.

There are some other minor comments, so please see if you can address them.

PR #1076, opened 2024-04-01
Update `kw remote` documentation

sahil-sagwekar2652:
The -s being an optional argument the long form flag can read the argument with the "=" but the short flag cannot read  "=" or space.
This commit fixes the documentation to reflect the above points.

`-s<name` works but `-s=<name>` doesn't.


davidbtadokoro:
Hi @sahil-sagwekar2652, and thanks for the catch and fix! Merged into the unstable branch :+1:

PR #1081, opened 2024-04-02
src: report: Fix the default error messages

sahil-sagwekar2652:
If the kw report is called without any arguments or with --all flag and the statistics and Pomodoro data are not present then ideally we should get two separate errors. It looks like this currently - 

image

Because the `options_values['ERROR']` value is getting overwritten in the 

I'm opening this PR for discussion and subsequent improvement, there could be a better way to implement this.

image

### DIFF
--- a/src/report.sh
+++ b/src/report.sh
@@ -287,9 +287,8 @@ function show_report()
     done
     printf '\n'
   elif [[ -n "${options_values['STATISTICS']}" ]]; then
-    warning "${options_values['ERROR']}"
+    warning "kw doesn't have any statistics of the target period: ${target_period}"
   fi
-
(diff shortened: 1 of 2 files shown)


davidbtadokoro:
Hi, @sahil-sagwekar2652, and thanks for another fix! TBH, I noticed this some time ago and was lazy about doing it :sweat_smile: 

Looks legit, but I've taken the opportunity to also delete the settings of `options_values['ERROR']` for these errors in the `process_and_format_...` functions, as we won't need them. In any case, merged into the unstable branch :+1:

sahil-sagwekar2652:
Thanks for updating the tests too @davidbtadokoro

PR #1084, opened 2024-04-18
kw: fix unneeded variable redefinition

OJarrisonn:
The variable `KW_SOUND_DIR` was being setted to
`'/usr/share/sounds/kw'` twice


rodrigosiqueira:
Nice catch! Change merged into the unstable branch.
Thanks

PR #1085, opened 2024-04-18
Added --repeat flag to pomodoro

th-duvanel:
Added the --repeat function in the `kw pomodoro` command.

If the user haven't executed any pomodoro command yet, the program shows an error.
The command stays in a temporary file, in the /tmp. If the computer is restarted, all the
/tmp files are lost, so the last command used with pomodoro.

### DIFF
--- a/src/pomodoro.sh
+++ b/src/pomodoro.sh
@@ -234,6 +234,43 @@ function is_valid_argument()
   return 0
 }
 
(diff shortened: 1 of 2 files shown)


th-duvanel:
shfmt -i=2 -ln=bash -fn -ci -sr -w filename

Issue #1089, opened 2024-04-21
replace "#!/bin/bash" shebangs with "#!/usr/bin/env bash"


MarceloSpessoto:
**Description:**

kworkflow, being a bash-written project, requires the use of shebangs to define the interpreter environment where kw is run. Nowadays, the scripts contain the shebang "#!/bin/bash", which works normally for most of the distros. 
There are, however, some exceptions, such as NixOS, where there is no /bin/bash (bash is located in /nix/store). Using workflow on these types of distros may not be smooth, since the user is required to run kw commands preceded by `bash` or manually modify the code.
A solution to fix the problem and not affect the experience on other distros is to replace the shebangs with "#!/usr/bin/env bash".

**Type:**

Others

**Why:**

Make kworkflow experience portable in a wider range of Linux distributions.

PR #1090, opened 2024-04-21
kw: replace every shebang with an updated version

MarceloSpessoto:
This PR makes the requested changes in issue #1089, by modifying every shebang in kw source code.
It also adds the corresponding coding guideline in the code style section from kw documentation.

### DIFF
--- a/database/migrate_legacy_data_20220101.sh
+++ b/database/migrate_legacy_data_20220101.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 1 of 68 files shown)


davidbtadokoro:
Tested on Arch (Manjaro) across most of the features, and it works normal!

I just think you could explain further on the commit message what you've described in #1089. When this change gets merged, the information detailed in the issue won't carry to the commit message, so it is important to give that context. Also, add the tag `Closes: #1089`.

rodrigosiqueira:
Thanks a lot! Change merged into the unstable branch.

PR #1092, opened 2024-04-21
Fix reverse argument order of assert_equals_helper in docs and test cases

### LINKED ISSUE #1091: Order of arguments in assert_equals_helper is the oposite of what is decalred
Both in docs and in its usage in evey test file of kworkflow, the `assert_equals_helper` arguments are defined to be ` "($LINENO)" "$output" "$expected_output"`.
But in reality, the function interprets it as ` "($LINENO)" "$expected_output" "$output"`

I stumbled upon this while testing a different change, and getting very confused by the output of the test I was adding.

**How to Reproduce:**

[Steps to reproduce the behavior:
1. Change a expected value in any test
2. Run `./run_tests.sh test <file>`
3. Check what gets printed out

**Expected behavior:**

```bash
./run_tests.sh test mail                               
...
test_mail_parser
ASSERT: line 324: Set to flag
  Expected Result: intentional-error@demo.com
    Actual Result: some@mail.com
```

**Actual behavior:**

```bash
./run_tests.sh test mail                               
...
ASSERT: line 324: Set to flag
  Expected Result: some@mail.com
    Actual Result: intentional-error@demo.com
```

auyer:
The order of arguments in `assert_equals_helper` is declared in docs and used in one way, but defined in another way

Definition in `codingstyle.rst`
```rst
   assert_equals_helper "$error_message" "($LINENO)" "$output" "$expected_output"

or::

   assertEquals "($LINENO)" "$output" "$expected_output"

```

I introduced a fake error to demonstrate this in the `mail_test.sh` file.

```bash
parse_mail_options '--to=some@mail.com'
expected='intentional-error@demo.com'
assert_equals_helper 'Set to flag' "$LINENO" "${options_values['TO']}" "$expected"
```

Expected Result:
```bash
./run_tests.sh test mail                               
...
test_mail_parser
ASSERT: line 324: Set to flag
  Expected Result: intentional-error@demo.com
    Actual Result: some@mail.com
```

Result before my patch:
```bash
./run_tests.sh test mail                               
...
ASSERT: line 324: Set to flag
  Expected Result: some@mail.com
    Actual Result: intentional-error@demo.com
```

I fixed the docs, and I am fixing the usage in most files.

Fixes #1091

### DIFF
--- a/documentation/content/codingstyle.rst
+++ b/documentation/content/codingstyle.rst
@@ -244,11 +244,11 @@ a developer. We also encourage using the ``assert_equals_helper`` helper
 function, which provides a wrapper capable of spitting a useful error message
 in case the assertion fails. Ideally, one should do either::
 
-   assert_equals_helper "$error_message" "($LINENO)" "$output" "$expected_output"
+   assert_equals_helper "$error_message" "($LINENO)" "$expected_output" "$output"
 
 or::
 
-   assertEquals "($LINENO)" "$output" "$expected_output"
+   assertEquals "($LINENO)" "$expected_output" "$output"
 
 Help functions
 --------------
--- a/tests/unit/config_test.sh
+++ b/tests/unit/config_test.sh
@@ -108,11 +108,11 @@ function test_set_config_value_changing_default_value()
 
   set_config_value 'use_llvm' 'lala' "${KW_CONFIG_BASE_PATH}/build.config"
   output=$(grep 'use_llvm' "${KW_CONFIG_BASE_PATH}/build.config")
-  assert_equals_helper 'Change llvm' "($LINENO)" "$output" 'use_llvm=lala'
+  assert_equals_helper 'Change llvm' "($LINENO)" 'use_llvm=lala' "$output"
 
   set_config_value 'menu_config' 'menuconfig' "${KW_CONFIG_BASE_PATH}/build.config"
   output=$(grep 'menu_config' "${KW_CONFIG_BASE_PATH}/build.config")
-  assert_equals_helper 'Change llvm' "($LINENO)" "$output" 'menu_config=menuconfig'
+  assert_equals_helper 'Change llvm' "($LINENO)" 'menu_config=menuconfig' "$output"
 }
 
 function test_set_config_value_with_dot_in_the_value()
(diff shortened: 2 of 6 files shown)


davidbtadokoro:
Hi @auyer, and sorry for the delay. Nice catch and proactiveness in solving the matter. However, I think that we should use a different approach to solve this. See the comment that I left in the issue for the explanation.

auyer:
I had the impression that the order ` "$output" "$expected_output"` was used everywhere. 
But I just found examples using the opposite, like you mentioned.

 will check what order is used more, and modify the others. Ok ?

davidbtadokoro:
> I had the impression that the order ` "$output" "$expected_output"` was used everywhere. But I just found examples using the opposite, like you mentioned.
> 
> will check what order is used more, and modify the others. Ok ?

I can't assure you that `expected` at 3 and `output` at 4 is the most used, but I think it is best just to update the wrong usages (i.e., where expected is 4 and output is 3, invert). My interpretation is that the use isn't standardized, so we should enforce the correct, not the most used, which, in my opinion, is `expected` at 3 and `output` at 4 (it makes more sense to me).

Also, if you want to tackle this issue, I think is best to either open one PR with one commit for each test file correction, or one PR for each file correction (I think this is best, as we can do it more incrementally and agile).

auyer:
I did not fix every file, but I got most of them.
There are two files I tried to fix, but `shellcheck` started complaining about unrelated issues I wont fix now.

rodrigosiqueira:
Hi @auyer ,
This is a very nice clean-up!
I think all of these commits could be squashed into a single one with a good commit message.

auyer:
Hi @rodrigosiqueira. I just squashed them all, as you requested.

davidbtadokoro (changes requested):
Hi @auyer, and thanks again for the change!

@rodrigosiqueira, actually, I was the one that prompted auyer to split the changes into dedicated commits for each test file :sweat_smile:. I thought it would make it more organized, but as these are clean replaces, I get why it is better to squash them (sorry @auyer :smiling_face_with_tear:).

In any case, the PR looks legit to me. The only are about the commit message. Change the subject to follow the pattern `target: file: Do something`. As the fixing the wrong usage of `assert_equals_helper` test suites is the primarily goal, I think it makes sense for the subject to be something like `tests: unit: Fix usage of <...>`.

auyer:
@davidbtadokoro  Your subject suggestion wold better fit a 1commit/file situation. Nevertheless, I adapted the subject to match your request.

davidbtadokoro:
> @davidbtadokoro Your subject suggestion would better fit a 1commit/file situation. Nevertheless, I adapted the subject to match your request.

I get that as you are also changing the documentation, there doesn't seem to be a clear subject to encapsulate every specific change. But mind that it isn't rare (I would say it is fairly common) to have a commit that directly or collaterally changes more than one file, so you should find a broader subject that conveys the overarching context in these cases without using subheads and such. Personally I would prefer a subject that focus on the fixing of the test suites and a body that mentions the change in the documentation as collateral, but this may be a matter of style.

In any case, I think the subhead in the subject isn't great and you can avoid using it, like

```
tests: unit: fix reverse argument order of assert_equals_helper in docs and test cases
```

instead of

```
tests: unit: fix docs and usage of assert_equals_helper - argument order is reversed
```

Notice that you are maybe trying to explain too much in the subject, which you can use the body for that.

(Obs.: This may seem like nitpick territory, but commit messages are almost as important as the changes itself).

davidbtadokoro:
Hi @auyer, and thanks again for the (many) updates! Everything looked legit, and with the adjustment on the commit message, I feel like we are good. Merged into the unstable branch :+1:

Issue #670, opened 2022-08-11
Enable sending files and directories through `kw mail`


kwy95:
**Description:**

The `git send-email` program allows the users to send the contents of a file or a directory as the patches, our implementation should support this behavior as well. Currently the biggest problem lies with the automatic inclusion of the reference to `@^` when no revision ranges are passed.

JGBSouza:
Working on it!

PR #1093, opened 2024-04-25
Enable kw mail file and directory sending

JGBSouza:
solve issue #670

### DIFF
--- a/src/mail.sh
+++ b/src/mail.sh
@@ -342,6 +342,10 @@ function find_commit_references()
     parsed=''
   done <<< "$(git rev-parse -- $args 2> /dev/null)"
 
+  if [[ -f "$i" || -d "$i" ]]; then
+    return 0
+  fi
+
   if [[ -n "$commit_range" ]]; then
     printf '%s' "$(str_strip "$commit_range")"
     return 0
--- a/tests/unit/mail_test.sh
+++ b/tests/unit/mail_test.sh
@@ -10,6 +10,9 @@ function oneTimeSetUp()
   declare -gr FAKE_KERNEL="$FAKE_GIT/fake_kernel/"
   declare -ga test_config_opts=('test0' 'test1' 'test2' 'user.name' 'sendemail.smtpuser')
 
+  declare -gr FAKE_PATCHES_DIR='fake_patches'
+  declare -gr FAKE_PATCH_FILE="${FAKE_PATCHES_DIR}/fake_patch_file"
+
   export KW_ETC_DIR="$SHUNIT_TMPDIR/etc/"
   export KW_CACHE_DIR="$SHUNIT_TMPDIR/cache/"
 
@@ -29,6 +32,9 @@ function oneTimeSetUp()
   }
 
   mk_fake_git
+
+  mkdir "./${FAKE_PATCHES_DIR}"
+  touch "./${FAKE_PATCH_FILE}"
 
   cd "$ORIGINAL_DIR" || {
     ret="$?"
@@ -494,6 +500,16 @@ function test_mail_send()
   expected='git send-email --to="mail@test.com" extra_args --other_arg -13 -v2'
   assert_equals_helper 'Testing no options option' "$LINENO" "$expected" "$output"
 
+  parse_mail_options "$FAKE_PATCH_FILE" '--to=mail@test.com'
+  output=$(mail_send 'TEST_MODE')
+  expected="git send-email --to=\"mail@test.com\" $FAKE_PATCH_FILE"
+  assert_equals_helper 'Testing send with a file' "$LINENO" "$expected" "$output"
+
+  parse_mail_options "$FAKE_PATCHES_DIR" '--to=mail@test.com'
+  output=$(mail_send 'TEST_MODE')
+  expected="git send-email --to=\"mail@test.com\" $FAKE_PATCHES_DIR"
+  assert_equals_helper 'Testing send with a directory' "$LINENO" "$expected" "$output"
+
   parse_mail_options '--to=mail@test.com'
 
   parse_configuration "$KW_MAIL_CONFIG_SAMPLE" mail_config
@@ -513,6 +529,7 @@ function test_mail_send()
     fail "($LINENO): Failed to move back to original dir"
     exit "$ret"
   }
+
 }
 
 function test_get_configs()


src/mail.sh
```diff
@@ -186,8 +188,21 @@ function pre_generate_patches()
   mkdir -p "$patch_cache"
 
   cmd_manager 'SILENT' "git format-patch --quiet --output-directory $patch_cache $version $commit_range"
+}
```
rodrigosiqueira, line 198: It looks like you add an extra space here.

src/mail.sh
```diff
@@ -99,6 +99,7 @@ function mail_send()
   local kernel_root
   local patch_count=0
   local cmd='git send-email'
+  declare -g patch_cache="${KW_CACHE_DIR}/patches"
```
rodrigosiqueira, line 103: Tbh, I don't see the value of making this variable global and in this function. If you think we can reuse it in multiple areas and we do not change it, declare it at the beginning of the file using upper case and the read-only option. Making this variable global inside this function is confusing.

tests/unit/mail_test.sh
```diff
@@ -9,7 +9,9 @@ function oneTimeSetUp()
   declare -gr FAKE_GIT="$SHUNIT_TMPDIR/fake_git/"
   declare -gr FAKE_KERNEL="$FAKE_GIT/fake_kernel/"
   declare -ga test_config_opts=('test0' 'test1' 'test2' 'user.name' 'sendemail.smtpuser')
-
+  declare -gr FAKE_PATCHES_DIR="fake_patches"
```
rodrigosiqueira, line 12: single quotes

tests/unit/mail_test.sh
```diff
@@ -30,6 +32,14 @@ function oneTimeSetUp()
 
   mk_fake_git
 
+  mkdir "./$FAKE_PATCHES_DIR"
```
rodrigosiqueira, line 36: https://kworkflow.org/content/codingstyle.html#string-concatenation

tests/unit/mail_test.sh
```diff
@@ -30,6 +32,14 @@ function oneTimeSetUp()
...
+  mkdir "./$FIRST_TEST_PATCHES_DIR"
+  cp -f "${ORIGINAL_DIR}/tests/unit/samples/test.patch" "${FIRST_TEST_PATCHES_DIR}/first.patch"
```
rodrigosiqueira, line 39: https://kworkflow.org/content/codingstyle.html#avoid-short-options

tests/unit/mail_test.sh
```diff
@@ -204,6 +214,32 @@ function test_validate_email_list()
...
+function test_count_patches()
+{
+  cd "$FAKE_GIT" || {
+    ret="$?"
+    fail "($LINENO): Failed to move to fake git dir"
+    exit "$ret"
+  }
+
+  local output
+  local expected
```
rodrigosiqueira, line 228: move this to the beginning of this function.

tests/unit/mail_test.sh
```diff
@@ -204,6 +214,32 @@ function test_validate_email_list()
   assert_equals_helper 'Expected a success' "$LINENO" "$ret" 0
 }
 
+function test_count_patches()
+{
+  cd "$FAKE_GIT" || {
+    ret="$?"
```
rodrigosiqueira, line 222: declare `ret` as local variable.

tests/unit/mail_test.sh
```diff
@@ -204,6 +214,32 @@ function test_validate_email_list()
...
+  local output
+  local expected
+
+  output="$(count_patches "$FIRST_TEST_PATCHES_DIR")"
+  expected=2
+  assert_equals_helper 'Expected two patches' "$LINENO" "$output" "$expected"
+
+  output="$(count_patches "$SECOND_TEST_PATCHES_DIR")"
+  expected=1
+  assert_equals_helper 'Expected two patches' "$LINENO" "$output" "$expected"
```
rodrigosiqueira, line 236: You should add other tests that validate:
1. There is not patch.
2. The target file is not a patch.

src/mail.sh
```diff
@@ -115,8 +116,14 @@ function mail_send()
     cmd+=" --cc=\"$cc_recipients\""
   fi
 
-  pre_generate_patches "$commit_range" "$version"
-  patch_count="$(count_patches "$patch_cache")"
+  if [[ -z "$patch_source" ]]; then
+    pre_generate_patches "$commit_range" "$version"
+    patch_count="$(count_patches "$patch_cache")"
```
rodrigosiqueira, line 121: Did you validate if your change does not break the original behavior? I mean, the one using the commit SHA.

src/mail.sh
```diff
@@ -115,8 +116,14 @@ function mail_send()
...
+  elif [[ -d "$patch_source" ]]; then
+    patch_count="$(count_patches "$patch_source")"
+  elif [[ $(is_a_patch "$patch_source") ]]; then
+    patch_count=1
+  fi
```
rodrigosiqueira, line 126: Add a comment inside each if that highlight the case the if condition is handling.

src/mail.sh
```diff
@@ -135,7 +142,7 @@ function mail_send()
   [[ -n "$private" ]] && cmd+=" $private"
   [[ -n "$rfc" ]] && cmd+=" $rfc"
   [[ -n "$extra_opts" ]] && cmd+=" $extra_opts"
-
+  [[ -n "$patch_source" ]] && cmd+=" $patch_source"
```
rodrigosiqueira, line 145: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/mail.sh
```diff
@@ -1119,6 +1126,8 @@ function parse_mail_options()
         for i in "$@"; do
           if [[ "${i}" =~ " " ]]; then
             pass_option_to_send_email+=" ${i@Q}"
+          elif [[ -f "$i" || -d "$i" ]]; then
+            options_values['PATCHES_SOURCE']="$i"
```
rodrigosiqueira, line 1130: The interface is not clear to me. You must update the man page by adding how this new option looks like.

rodrigosiqueira (changes requested):

davidbtadokoro:
Hi @JGBSouza, and thank you so much for your change!

I spent maybe an hour+ reviewing your PR and thinking if it was the best approach and testing it, but it suddenly hit me that maybe we don't need to do all the handling of the number of patches, and the processing of files/directories which can represent a patchset.

Below, are my takes on the matter, which you can consider my "review" (@rodrigosiqueira, if possible, also give me your thoughts on the matter).

# My interpretation of the problem:

Look at what Rubens described in the issue:

> Currently the biggest problem lies with the automatic inclusion of the reference to @^ when no revision ranges are passed.

This is in reference to [this part](https://github.com/kworkflow/kworkflow/blob/af4d2e3100bd33c8dd07c8dba2c42cd7af020c9b/src/mail.sh#L1130) of kw mail, which means that, if we don't pass any commit reference (either a range, or a number of commits) to `--send`, the value of `options_values['COMMIT_RANGE']` would be `@^`, which, FWIU, is the same as `HEAD^`, meaning the last commit. What he is alluding to is the following:

Take this repo that I have in my machine with this git log:

2024-04-28-125222_509x189_scrot

I create the following patch files that represent the patchset with

2024-04-28-125343_987x279_scrot

I can then use kw mail to send the patchset (almost) correctly with the current unstable `kw mail` implementation (notice that `kw mail --send -- *.patch` would work the same)

2024-04-28-125635_744x353_scrot

The problem is that the last commit (represented by file `/tmp/4kZTrnWu7j/0001-Add-Sofware.patch`) is also included in the sending of the patch.

# How `--send` really works:

It is really important to notice that `--send` actually doesn't receive a argument, per se. Look at the [case](https://github.com/kworkflow/kworkflow/blob/af4d2e3100bd33c8dd07c8dba2c42cd7af020c9b/src/mail.sh#L1000) in the parser and see that `--send` is treated as a standalone option for `kw mail`, so values like `-3` or `abcd..e1d3` are treated as additional arguments (which fall in [this case](https://github.com/kworkflow/kworkflow/blob/af4d2e3100bd33c8dd07c8dba2c42cd7af020c9b/src/mail.sh#L1097) of the parser).

This means that commit ranges are directly processed by `git send-email`, which is good as delegating this to git reduces the complexity and problems to us.

# My opinion on the best approach:

Pay attention that `kw mail` currently supports the passing of files/directories with patches like Rubens described, but it sends the last commit together, which is wrong.

I think the best approach would be to modify the function `find_commit_references` (see [this line](https://github.com/kworkflow/kworkflow/blob/af4d2e3100bd33c8dd07c8dba2c42cd7af020c9b/src/mail.sh#L1113) for reference) to not return 22 EINVAL when files or directories are passed as arguments.

In other words, IMHO, the best approach is to expand the current behavior to just **also** support the passing of files and directories. If I am correct, with this approach the change in the codebase will be minimal and we would also benefit from the complexity to handle the processing of files and directories being totally delegated to `git send-email`.

davidbtadokoro:
# UPDATE

@JGBSouza, I've managed to implement the necessary support for files and directories with the following

```
diff --git a/src/mail.sh b/src/mail.sh
index f4a6981..d5ca143 100644
--- a/src/mail.sh
+++ b/src/mail.sh
@@ -347,6 +347,10 @@ function find_commit_references()
     return 0
   fi
 
+  for arg in $args; do
+    [[ -f "$arg" || -d "$arg" ]] && return 0
+  done
+
   return 22 # EINVAL
 }
```

As always, code and approach suggestions from my end are to be taken with a grain of salt. In other words, you need to thoroughly and actively think on my arguments and don't take anything I say for granted. If you feel like my approach is best, heavily validate, refine and test it!

src/mail.sh
JGBSouza: I also think that this is the case of making the variable global in the beginnig of the file, but for some reason, "$KW_CACHE_DIR" is an empty var in the begining of the code and i didn't understand where it's value is updated inside this code

src/mail.sh
```diff
@@ -342,6 +342,10 @@ function find_commit_references()
     parsed=''
   done <<< "$(git rev-parse -- $args 2> /dev/null)"
 
+  if [[ -f "$i" || -d "$i" ]]; then
+    return 0
+  fi
```
davidbtadokoro, line 347: I am a little confused, where does `i` come from? I don't see it declared anywhere in this function neither see any relation with `args`, which is the variable that stores the files/directory arguments. It seems to work, as `kw mail --send` and `kw mail --send *.patch` work as intended (i.e., the first sends the last commit as a patch, and the second sends the patch files as patches **without** also including the last commit), but it isn't clear how. Please, comment your reasoning behind this, in case I am mistaken about it being wrong.

There are also two points:

1. As `kw mail` "enhances" `git send-email` we have an opportunity to check if the arguments are valid patch files. So I think we could do some kind of loop to check if each argument is a patch file (in case they are files) or a directory containing patch files (in case it is a directory).
2. I think the check to see if the arguments are patch files/directories should come after the below check (`if [[ -n "$commit_range" ]]; then`), as this last one is related with the while loop above (`while read -r arg; do`). I am saying this, because this function now finds commit references **or** patch files/directories, so it should check those separately.

src/mail.sh
davidbtadokoro, line 1: 1. You should probably rename `find_commit_references` as now it checks for commit references **or** patch files/dirs, Find a name that transmits both, in a more succinct way than `find_commit_references_or_patch_files_dirs`, if possible.
2. You should modify the function documentation to reflect this change in behavior.

tests/unit/mail_test.sh
```diff
@@ -10,6 +10,9 @@ function oneTimeSetUp()
   declare -gr FAKE_KERNEL="$FAKE_GIT/fake_kernel/"
   declare -ga test_config_opts=('test0' 'test1' 'test2' 'user.name' 'sendemail.smtpuser')
 
+  declare -gr FAKE_PATCHES_DIR='fake_patches'
+  declare -gr FAKE_PATCH_FILE="${FAKE_PATCHES_DIR}/fake_patch_file"
```
davidbtadokoro, line 14: I know that because of the `cd` statement below, we will eventually be at `FAKE_GIT`, and it will be safe to do what you did:

```bash
mkdir "./${FAKE_PATCHES_DIR}"
touch "./${FAKE_PATCH_FILE}"
```

However, I think the cleanest way is to define explicitly as `FAKE_PATCHES_DIR=${FAKE_GIT}/fake_patches` and make the statements below simply:

```bash
  mkdir "$FAKE_PATCHES_DIR"
  touch "$FAKE_PATCH_FILE"
```

this way we don't depend on relative paths which can induce errors that will be hard to debug.

tests/unit/mail_test.sh
```diff
@@ -30,6 +33,9 @@ function oneTimeSetUp()
 
   mk_fake_git
 
+  mkdir "./${FAKE_PATCHES_DIR}"
+  touch "./${FAKE_PATCH_FILE}"
```
davidbtadokoro, line 37: See the comment above to change these lines accordingly

tests/unit/mail_test.sh
```diff
@@ -513,6 +529,7 @@ function test_mail_send()
     fail "($LINENO): Failed to move back to original dir"
     exit "$ret"
   }
+
```
davidbtadokoro, line 532: You don't need to add this line, remove it

tests/unit/mail_test.sh
```diff
@@ -494,6 +500,16 @@ function test_mail_send()
   expected='git send-email --to="mail@test.com" extra_args --other_arg -13 -v2'
   assert_equals_helper 'Testing no options option' "$LINENO" "$expected" "$output"
 
+  parse_mail_options "$FAKE_PATCH_FILE" '--to=mail@test.com'
+  output=$(mail_send 'TEST_MODE')
+  expected="git send-email --to=\"mail@test.com\" $FAKE_PATCH_FILE"
```
davidbtadokoro, line 505: String concatenation in `FAKE_PATCH_FILE`

tests/unit/mail_test.sh
```diff
@@ -494,6 +500,16 @@ function test_mail_send()
...
+  assert_equals_helper 'Testing send with a file' "$LINENO" "$expected" "$output"
+
+  parse_mail_options "$FAKE_PATCHES_DIR" '--to=mail@test.com'
+  output=$(mail_send 'TEST_MODE')
+  expected="git send-email --to=\"mail@test.com\" $FAKE_PATCHES_DIR"
```
davidbtadokoro, line 510: String concatenation in `FAKE_PATCHES_DIR`

tests/unit/mail_test.sh
```diff
@@ -494,6 +500,16 @@ function test_mail_send()
   expected='git send-email --to="mail@test.com" extra_args --other_arg -13 -v2'
   assert_equals_helper 'Testing no options option' "$LINENO" "$expected" "$output"
 
+  parse_mail_options "$FAKE_PATCH_FILE" '--to=mail@test.com'
```
davidbtadokoro, line 503: Also, I would like to see more real life cases:

1. More than one patch file without cover letter
2. More than one patch file with cover letter
3. Use of patch files with prefix `000` (as in `0003` or `0010`) and suffix `.patch` to relate to what `git format-patches` outputs

src/mail.sh
davidbtadokoro, line 1: Add an example in `kw mail` manpage to exemplify the passing of patch files/directories as arguments.

davidbtadokoro (changes requested):
Hi @JGBSouza, and thank you for the update!

I really think this approach is a cleaner and direct way to implement the desired behavior, without the risk of adding any regression.

I left important comments so please tackle everyone of them (ping in case of **any** doubt), but mind that this PR is looking legit. It works as intended, so now we should focus on refining it.

**Update:** Forgot to mention that you didn't add a `Signed-off-by` line in the commit message, so please do it.

rodrigosiqueira:
@JGBSouza any updates here?

rodrigosiqueira:
PR closed due to the lack of updates.

Issue #344, opened 2021-06-24
add regex support for kw d --uninstall


melissawen:
To easy uninstall multiples kernel with similar names.
For example: same customised name with different numeric version (5.*-CUSTOM_NAME)

BrizaMel:
Me and Lorenzo are working on it!

PR #1099, opened 2024-04-30
src: plugins: kernel_install: utils: add regex support for kw d --uninstall

lorenzoberts:
Currently, the command kw d --uninstall just supports a comma-separated kernel list, but for some cases it's better to provide a regex pattern that will match the desired kernels to be uninstalled.
   
Closes: #344

### DIFF
--- a/documentation/man/features/kw-deploy.rst
+++ b/documentation/man/features/kw-deploy.rst
@@ -85,9 +85,10 @@ OPTIONS
 -s, \--ls-line:
   List available kernels separated by comma.
 
--u <kernel-name>[,...], \--uninstall <kernel-name>[,...]:
+-u (<kernel-name> | regex:<pattern>)[,...], \--uninstall (<kernel-name> | regex:<pattern>)[,...]:
   Remove a single kernel or multiple kernels; for removing
-  multiple kernels it is necessary to separate them with comma.
+  multiple kernels it is necessary to separate them with comma. A regex pattern
+  can also be passed as input, prefixed with 'regex:'.
 
 -f, \--force:
   Remove kernels even if they were not installed by kw (only valid with
@@ -153,3 +154,21 @@ any other kw user. If you want to install a custom kernel from this package,
 you can use::
 
   kw deploy --from-package 5.19.0-THIS-IS-AN-EXAMPLE+.kw.tar
+
+Using kw deploy --uninstall:
+
+Uninstall argument can be a full kernel name.
+
+  kw deploy --uninstall kernel1
+
+Or a comma-separated list of full names.
+
+  kw deploy --uninstall kernel1,kernel2,kernel3
+
+Or a regular expression prefixed with regex: .
+
+  kw deploy --uninstall regex:kernel*
+
+Or a list comma-separated list of full names and regular expressions.
+
+  kw deploy --uninstall regex:kernel[1-3],kernel4,regex:kernel[5-6]
--- a/src/plugins/kernel_install/utils.sh
+++ b/src/plugins/kernel_install/utils.sh
@@ -551,32 +551,88 @@ function do_uninstall()
   fi
 }
 
+# Element existence check. Checks if element belongs to given array.
+#
+# @target_element: element being checked
+# @array: target array
+#
+# Return:
+# returns 0 if element is in array and 1 otherwise
+function is_in_array()
+{
+  local target_element="$1"
+  local -n _array="$2"
+
+  for element in "${_array[@]}"; do
+    if [[ "$element" == "$target_element" ]]; then
+      return 0
+    fi
+  done
+  return 1
+}
+
+# Return a list of unique available kernels names
+#
+# @prefix: set a base prefix for searching for kernels.
+# @all_kernels: list all available kernels, not only the ones installed by kw
+#
+# Return:
+# returns array containing available kernels
+function process_installed_kernels()
+{
+  local all_kernels="$1"
+  local prefix="$2"
+  local -n _processed_installed_kernels="$3"
+  local kernels
+
+  kernels=$(list_installed_kernels 'SILENT' 1 "$all_kernels" "$prefix")
+  IFS=, read -r -a available_kernels <<< "$kernels"
+  mapfile -t _processed_installed_kernels <<< "$(printf "%s\n" "${available_kernels[@]}" | sort --unique)"
+}
+
 function kernel_uninstall()
 {
   local reboot="$1"
   local target="$2"
-  local kernel_list_string="$3"
+  local kernel_list_string_or_regex="$3"
   local flag="$4"
   local force="$5"
   local prefix="$6"
   local update_grub=0
+  local -a deduped_all_kernels
+  local -a deduped_kw_managed_kernels
+  local prefix_for_regex='regex:'
+  declare -A kernel_names
 
-  cmd_manager "$flag" "sudo mkdir -p $REMOTE_KW_DEPLOY"
-  cmd_manager "$flag" "sudo touch '$INSTALLED_KERNELS_PATH'"
-
-  kernel_list_string=$(printf '%s' "$kernel_list_string" | tr --delete ' ')
-
-  if [[ -z "$kernel_list_string" ]]; then
+  if [[ -z "$kernel_list_string_or_regex" ]]; then
     printf '%s\n' 'Invalid argument'
     exit 22 #EINVAL
   fi
 
-  IFS=', ' read -r -a kernel_names <<< "$kernel_list_string"
-  for kernel in "${kernel_names[@]}"; do
-    cmd="sudo grep -q '$kernel' '$INSTALLED_KERNELS_PATH'"
-    cmd_manager "$flag" "$cmd"
+  cmd_manager "$flag" "sudo mkdir -p $REMOTE_KW_DEPLOY"
+  cmd_manager "$flag" "sudo touch '$INSTALLED_KERNELS_PATH'"
+
+  process_installed_kernels 1 "$prefix" 'deduped_all_kernels'
+  process_installed_kernels '' "$prefix" 'deduped_kw_managed_kernels'
+
+  IFS=', ' read -r -a kernel_names_array <<< "$kernel_list_string_or_regex"
+
+  for input_string in "${kernel_names_array[@]}"; do
+    for installed_kernel in "${deduped_all_kernels[@]}"; do
+      if [[ "$input_string" =~ ^$prefix_for_regex ]]; then
+        input_regex=^${input_string#"$prefix_for_regex"}$
+        [[ $installed_kernel =~ $input_regex ]] && kernel_names["$installed_kernel"]=1
+      else
+        [[ "$installed_kernel" == "$input_string" ]] && kernel_names["$installed_kernel"]=1
+      fi
+    done
+  done
+
+  for kernel in "${!kernel_names[@]}"; do
+    # is_in_array "$kernel" "deduped_kw_managed_kernels[@]"
+    is_in_array "$kernel" 'deduped_kw_managed_kernels'
     if [[ "$?" != 0 && -z "$force" ]]; then
-      printf '%s\n' "$kernel not managed by kw. Use --force/-f to uninstall anyway."
+      printf '%s\n' "${kernel} not managed by kw. Use --force/-f to uninstall anyway."
       continue # EINVAL
     fi
 
--- a/tests/unit/plugins/kernel_install/utils_test.sh
+++ b/tests/unit/plugins/kernel_install/utils_test.sh
@@ -52,7 +52,7 @@ function setUp()
   # Creating fake installed kernels
   touch "$INSTALLED_KERNELS_PATH"
   printf '5.5.0-rc2-VKMS+' >> "$INSTALLED_KERNELS_PATH"
-  printf '5.6.0-rc2-AMDGPU+' >> "$INSTALLED_KERNELS_PATH"
+  printf '\n%s' '5.6.0-rc2-AMDGPU+' >> "$INSTALLED_KERNELS_PATH"
 
   # Replace KW_DEPLOY_TMP_FILE
   test_tmp_file="$SHUNIT_TMPDIR/tmp/kw"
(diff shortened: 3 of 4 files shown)


rodrigosiqueira:
Hi @lorenzoberts ,

In terms of design, I would prefer to avoid a new option. I was more inclined to use the same interface but also accept regex. For example:

```
kw d -u=".*kernel-name$,full-kernel-name,.*kernel"
```

src/plugins/kernel_install/utils.sh
```diff
@@ -554,23 +554,39 @@ function kernel_uninstall()
...
+    all_kernels=$(list_installed_kernels 'SILENT' 1 1 "$prefix")
+    IFS=, read -r -a available_kernels <<< "$all_kernels"
+    mapfile -t deduped_kernel_names < <(printf "%s\n" "${available_kernels[@]}" | sort -u)
+    kernel_names=()
+    for kernel in "${deduped_kernel_names[@]}"; do
+      # matching kernels to be uninstalled with the given regex
+      if [[ $kernel =~ $regex_pattern ]]; then
+        kernel_names+=("$kernel")
+      fi
+    done
```
davidbtadokoro, line 578: Hi, @lorenzoberts and @BrizaMel, and thank you guys so much for the change! You've really nailed in this first PR, in terms of following the project rules and practices :+1:

The change seems legit from pure reviewing (didn't test it yet), but I think that there is an easier and more robust approach to implement this. I think that we can leverage the fact that any execution of `kw deploy --uninstall` passes through this function (`kernel_uninstall`) to just alter the processing of the variable `kernel_list_string` (that contains the arguments for `--uninstall`) to **also** support regex arguments.

My idea was to generalize this excerpt of code to handle both types of arguments (full names and regexs) indistinguishably. For example, suppose `kernel_list_string='kernel1,kernel[2-4]'`, and the kernels installed on the target are `kernel1`, `kernel2`, and `kernel3`. When we get to the assertion

```bash
[[ "$kernel" =~ $full_name_or_regex ]]
```

`kernel1` will match with `kernel1`, and `kernel2` and `kernel3` will match with `kernel[2-4]`.

There is probably some complexity that I am overlooking (we would need to think about the boundaries of the regex, how to handle kernel names of the type `<some_kernel>+`, etc.), but IMHO I think that we can cleanly implement the proposed while keeping the feature interface untouched with this.

Feel free to ping me for any doubts related or not with this :smiley:

tests/unit/plugins/kernel_install/utils_test.sh
```diff
@@ -206,6 +206,136 @@ function test_reboot_machine()
...
+    'run_bootloader_update_mock'
+  )
+
+  for cmd in "${cmd_last_part[@]}"; do
+    cmd_sequence["$index"]="$cmd"
+    ((index++))
+  done
+
+  # Check
+  output=$(kernel_uninstall 0 'local' '5.*VKMS' 'TEST_MODE' '' "$SHUNIT_TMPDIR/" 1)
```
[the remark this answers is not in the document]
lorenzoberts: We addressed this issue in 19ac4bd3ee19d5105c48cedd9699fff72af2c56a. Thanks, @melissawen!

src/plugins/kernel_install/utils.sh
lorenzoberts: Hi @davidbtadokoro and @rodrigosiqueira, thanks for the appointment! We believe that this solution may complicate the command's usability since users would need to escape all special characters. In our opinion, using a flag would be the best option as it already allows multiple patterns to be passed. In the case of this example, the input regex would be (kernel1|kernel[2-4]). Additionally, the implementation cost of adding a prefix to each element of the list may not be reflected in the impact on users. What's your opinion on this?

davidbtadokoro: Hi @lorenzoberts and @BrizaMel, and thank you for the update!

> Additionally, the implementation cost of adding a prefix to each element of the list may not be reflected in the impact on users. What's your opinion on this?

I'm not sure if you are saying that it would be a good approach or not, but, IMHO, I think it would have the best trade-off, as the change would be more self-contained, and we wouldn't need a new option.

Below, is a rough draft that I managed to implement which adds the regex support by just modifying `kernel_uninstall`, with the restriction that a regex arguments should be prefixed with `regex:`. (Note that I hardcoded the value`installed_kernels`, and added an `exit` statement to "lazy-test" this draft :sweat_smile:)

```bash
@@ -559,6 +559,10 @@ function kernel_uninstall()
   local force="$5"
   local prefix="$6"
   local update_grub=0
+  declare -A kernel_names
+  local installed_kernels
+  local kernel_name
+  local kernel_regex
 
   cmd_manager "$flag" "sudo mkdir -p $REMOTE_KW_DEPLOY"
   cmd_manager "$flag" "sudo touch '$INSTALLED_KERNELS_PATH'"
@@ -570,7 +574,29 @@ function kernel_uninstall()
     exit 22 #EINVAL
   fi
 
-  IFS=', ' read -r -a kernel_names <<< "$kernel_list_string"
+  installed_kernels='kernel1,kernel2,kernel3,kernel4'
+  # installed_kernels=$(list_installed_kernels 'SILENT' 1 1 "$prefix")
+  IFS=, read -r -a installed_kernels_array <<< "$installed_kernels"
+  IFS=', ' read -r -a kernel_name_or_regex_array <<< "$kernel_list_string"
+  for installed_kernel in "${installed_kernels_array[@]}"; do
+    for kernel_name_or_regex in "${kernel_name_or_regex_array[@]}"; do
+      # In case the element starts with `regex:` we use regex to match
+      if [[ "$kernel_name_or_regex" =~ ^regex: ]]; then
+        kernel_regex=$(printf '%s' "$kernel_name_or_regex" | cut --delimiter=':' --fields=2)
+        [[ "$installed_kernel" =~ ^$kernel_regex$ ]] && kernel_names["$installed_kernel"]=1
+      else
+        kernel_name="$kernel_name_or_regex"
+        [[ "$installed_kernel" == "$kernel_name" ]] && kernel_names["$installed_kernel"]=1
+      fi
+    done
+  done
+
+  for kernel_name in "${!kernel_names[@]}"; do
+    printf 'Removing: %s\n' "$kernel_name"
+  done
+
+  exit
+
   for kernel in "${kernel_names[@]}"; do
     cmd="sudo grep -q '$kernel' '$INSTALLED_KERNELS_PATH'"
     cmd_manager "$flag" "$cmd"
```

Below, is a screenshot of an example of execution in which the user passes one full name argument `kernel` and one regex argument `regex:kernel[2-4]`.

2024-05-12-002055_867x501_scrot

Notice that the function correctly interprets `regex:kernel[2-4]` as a regex argument and matches it with `kernel2`, `kernel3`, and `kernel4`, while not matching the full name `kernel` with any installed kernel.

lorenzoberts: Hi @davidbtadokoro, the commit 0eebc3aafc8bb5cd4d40ad1832910cde32754397 was based on your draft. We adapted the tests we had done, and now everything seems to be fine, including the fix for the issue mentioned by @melissawen. Let us know what you think. 
ps: after all the remaining revision iterations and before merging we probably should squash this three commits into one.

rodrigosiqueira:
Hi @lorenzoberts , I think we don't need the first commit, right? Could you drop it? Keep only the necessary commits in the PR.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,6 +549,18 @@ function do_uninstall()
   fi
 }
 
+function is_in_array()
```
rodrigosiqueira, line 553: Please add documentation.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,6 +549,18 @@ function do_uninstall()
...
+{
+  target_element=$1
```
rodrigosiqueira, line 555: Use `""` around $1. Also set it to local.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,6 +549,18 @@ function do_uninstall()
...
+  array=("${!2}")
```
rodrigosiqueira, line 556: Please, declare the local array together with target_element and add one new line after it.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,6 +549,18 @@ function do_uninstall()
...
+  for element in "${array[@]}"; do
+    if [ "$element" = "$target_element" ]; then
```
rodrigosiqueira, line 558: Use `[[ ]]` and `==`.

src/plugins/kernel_install/utils.sh
```diff
@@ -564,35 +575,48 @@ function kernel_uninstall()
   cmd_manager "$flag" "sudo mkdir -p $REMOTE_KW_DEPLOY"
   cmd_manager "$flag" "sudo touch '$INSTALLED_KERNELS_PATH'"
 
+  all_kernels=$(list_installed_kernels 'SILENT' 1 1 "$prefix")
+  IFS=, read -r -a available_kernels <<< "$all_kernels"
+  mapfile -t deduped_all_kernels < <(printf "%s\n" "${available_kernels[@]}" | sort -u)
```
rodrigosiqueira, line 583: https://kworkflow.org/content/codingstyle.html#avoid-short-options

src/plugins/kernel_install/utils.sh
```diff
@@ -564,35 +575,48 @@ function kernel_uninstall()
...
+
+  kw_managed_kernels=$(list_installed_kernels 'SILENT' 1 '' "$prefix")
+  IFS=, read -r -a available_kernels <<< "$kw_managed_kernels"
+  mapfile -t deduped_kw_managed_kernels < <(printf "%s\n" "${available_kernels[@]}" | sort -u)
```
rodrigosiqueira, line 587: This code is very similar to the code above it; create a function for that to avoid code duplication.

src/plugins/kernel_install/utils.sh
```diff
@@ -564,35 +575,48 @@ function kernel_uninstall()
...
     kernel_names=()
-    for kernel in "${deduped_kernel_names[@]}"; do
+    for kernel in "${deduped_all_kernels[@]}"; do
       # matching kernels to be uninstalled with the given regex
       if [[ $kernel =~ $regex_pattern ]]; then
         kernel_names+=("$kernel")
       fi
     done
+    if [ ${#kernel_names[@]} -eq 0 ]; then
+      printf '%s\n' "Regex $regex_pattern didn't match any kernels."
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/plugins/kernel_install/utils.sh
```diff
@@ -564,35 +575,48 @@ function kernel_uninstall()
...
+      exit 1
```
rodrigosiqueira: EINVAL looks more appropriate here.

src/plugins/kernel_install/utils.sh
```diff
@@ -564,35 +575,48 @@ function kernel_uninstall()
...
-    cmd="sudo grep -q '$kernel' '$INSTALLED_KERNELS_PATH'"
-    cmd_manager "$flag" "$cmd"
-    if [[ "$?" != 0 && -z "$force" ]]; then
-      printf '%s\n' "$kernel not managed by kw. Use --force/-f to uninstall anyway."
-      continue # EINVAL
+    is_in_array "$kernel" "deduped_all_kernels[@]"
+    if [[ "$?" != 0 ]]; then
+      is_in_array "$kernel" "deduped_kw_managed_kernels[@]"
+      if [[ "$?" != 1 && -z "$force" ]]; then
+        printf '%s\n' "$kernel not managed by kw. Use --force/-f to uninstall anyway."
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/plugins/kernel_install/utils.sh
```diff
@@ -564,35 +575,48 @@ function kernel_uninstall()
...
+        continue
+      fi
+    else
+      printf '%s\n' "$kernel not found."
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#string-concatenation

tests/unit/plugins/kernel_install/utils_test.sh
```diff
@@ -206,13 +206,28 @@ function test_reboot_machine()
...
 }
 
+function test_kernel_uninstall_regex_didnt_match()
+{
+  local kernel_name='unknown'
+  # Composing command sequence list
+  local -a cmd_sequence=(
+    "sudo mkdir -p $REMOTE_KW_DEPLOY"
+    "sudo touch '$INSTALLED_KERNELS_PATH'"
+    "Regex ^$kernel_name$ didn't match any kernels."
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#string-concatenation

tests/unit/plugins/kernel_install/utils_test.sh
```diff
@@ -380,19 +391,68 @@ function test_kernel_force_uninstall_unmanaged()
...
-    "Can't find $modules_lib_path"
-    "sudo sed -i '/xpto/d' '$INSTALLED_KERNELS_PATH'"
+    "Removing: $boot_path"
+    "sudo -E rm $boot_path"
+    "Removing: $mkinitcpio_d_path"
+    "sudo -E rm $mkinitcpio_d_path"
+    "Removing: $initramfs_tools_var_path"
+    "sudo -E rm $initramfs_tools_var_path"
+    "Removing: $modules_lib_path"
+    "sudo -E rm -rf $modules_lib_path"
```
rodrigosiqueira, line 396: https://kworkflow.org/content/codingstyle.html#string-concatenation

documentation/man/features/kw-deploy.rst
```diff
@@ -84,9 +84,10 @@ OPTIONS
 -s, \--ls-line:
   List available kernels separated by comma.
 
--u <kernel-name>[,...], \--uninstall <kernel-name>[,...]:
+-u (<kernel-name> | regex:<pattern>)[,...], \--uninstall (<kernel-name> | regex:<pattern>)[,...]:
   Remove a single kernel or multiple kernels; for removing
-  multiple kernels it is necessary to separate them with comma.
+  multiple kernels it is necessary to separate them with comma. A regex pattern
+  can also be passed as input, prefixed with 'regex:'.
```
rodrigosiqueira, line 90: Add some examples in the EXAMPLE section. You can also consider expanding the tutorial part to describe this new regex option.

rodrigosiqueira (changes requested):

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
   fi
 }
 
+# Element existence check
+# Checks if element belongs to given array
```
davidbtadokoro, line 554: I think this can be a single sentence.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
+#
+# @target_element element being checked
+# @array target array
```
davidbtadokoro, line 557: Use colon : between the argument name and the description, as in `# @target_element: element being checked`.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
+# returns 1 if element is in array and 0 otherwise
```
davidbtadokoro, line 558: Add a blank line between the arguments and the return documentation. Also, make the returns section like the below:

```bash
# Return:
# Returns X in case Y....
```

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
+function is_in_array()
+{
+  local target_element="$1"
+  local array=("${!2}")
```
davidbtadokoro, line 562: When dealing with indexed/associative arrays in Bash that you need a function to consult/populate/manipulate them, prefer the use of the form `local -n _reference` to create references to arrays declared outside the scope of these functions. In practical terms, you can make this line
```bash
local -n _array="$2"
```
and call the function below like `is_in_array "$kernel" 'kw_managed_installed_kernels'`.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
+
+  for element in "${array[@]}"; do
+    if [[ "$element" == "$target_element" ]]; then
+      return 1
```
davidbtadokoro, line 566: By convention, 0 denotes true and not 0 denotes false. As the predicate is `is_in_array` invert the return values.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
+    fi
+  done
+  return 0
+}
+
+# Return available kernels
```
davidbtadokoro, line 572: Refine this description to incorporate the idea of a unique list.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
+#
+# @prefix Set a base prefix for searching for kernels.
+# @all_kernels List all available kernels, not only the ones installed by kw
+# returns array containing available kernels
```
davidbtadokoro, line 576: Apply the same mentioned in the comment above about the documentation format of the args and returns of functions.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
+function process_installed_kernels()
+{
+  local all_kernels="$1"
+  local prefix="$2"
+  local kernels
+
```
davidbtadokoro, line 582: Just like in the function `is_in_array` prefer the use o functions that manipulate array references using `local -n` over functions that "globally" do this. If I am not mistaken, notice that you are creating an array `deduped_kernels` that "bypasses" the function API, in the sense that the caller must know that the function processed will be in this specific array.

My request is to define another argument `local -n _processed_installed_kernels="$3"` that will be declared by the caller, and it will passed it as reference to this function to store the processed installed kernels.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
+  kernels=$(list_installed_kernels 'SILENT' 1 "$all_kernels" "$prefix")
+  IFS=, read -r -a available_kernels <<< "$kernels"
+  mapfile -t deduped_kernels < <(printf "%s\n" "${available_kernels[@]}" | sort --unique)
```
davidbtadokoro, line 585: Following the comment above, here you will use `mapfile -t _processed_installed_kernels ...`.

Also, IMHO, we can refine the right part of this line to be cleaner by making it
```bash
mapfile -t _processed_installed_kernels <<< "$(printf "%s\n" "${available_kernels[@]}" | sort --unique)"
```

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
   local target="$2"
-  local kernel_list_string="$3"
+  local kernel_list_string_or_regex="$3"
   local flag="$4"
   local force="$5"
   local prefix="$6"
   local update_grub=0
+  declare -A kernel_names
+
+  prefix_for_regex="regex:"
```
davidbtadokoro, line 599: Declare this as a local value and use single-quotes with `local prefix_for_regex='regex:'`

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
 
   cmd_manager "$flag" "sudo mkdir -p $REMOTE_KW_DEPLOY"
   cmd_manager "$flag" "sudo touch '$INSTALLED_KERNELS_PATH'"
 
-  kernel_list_string=$(printf '%s' "$kernel_list_string" | tr --delete ' ')
+  process_installed_kernels 1 "$prefix"
+  deduped_all_kernels=("${deduped_kernels[@]}")
+
+  process_installed_kernels '' "$prefix"
+  deduped_kw_managed_kernels=("${deduped_kernels[@]}")
```
davidbtadokoro, line 608: Following my comment about the use of an array reference in the `process_installed_kernels` function, here you will call it as `process_installed_kernels 'deduped_all_kernels' 1 "$prefix"`, and analogously in the second call to the function. Don't forget to declare `deduped_all_kernels` and `deduped_kw_managed_kernels` at the top of the function with `local -a deduped_all_kernels`.

Notice how, in this way, it is clearer to the caller the function API and effects.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
 
-  if [[ -z "$kernel_list_string" ]]; then
+  if [[ -z "$kernel_list_string_or_regex" ]]; then
     printf '%s\n' 'Invalid argument'
     exit 22 #EINVAL
   fi
```
davidbtadokoro, line 613: Move this check to before the processing, as it will be meaningless if it fails anyway.

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
+        input_regex=^${input_string#"$prefix_for_regex"}$
+        [[ $installed_kernel =~ $input_regex ]] && kernel_names["$installed_kernel"]=1
+      else
+        [[ "$installed_kernel" == "$input_string" ]] && kernel_names["$installed_kernel"]=1
+      fi
+    done
+  done
+
+  for kernel in "${!kernel_names[@]}"; do
+    is_in_array "$kernel" "deduped_kw_managed_kernels[@]"
```
davidbtadokoro, line 629: By using an array reference in the implementation of `is_in_array`, you will need to alter this line to be
```bash
is_in_array "$kernel" 'deduped_kw_managed_kernels'
```

src/plugins/kernel_install/utils.sh
```diff
@@ -550,32 +550,85 @@ function do_uninstall()
...
+    if [[ "$?" != 1 && -z "$force" ]]; then
```
davidbtadokoro, line 630: Notice that by changing the returns of is_in_array, you will need to check for != 0

davidbtadokoro (changes requested):
Hi @lorenzoberts and @BrizaMel, and thank again for continuing evolving this PR. `kw deploy` is one of the core features of kw and this change will definitely have a great impact.

I left a bunch of comments, but they pertain more to the refinement of the PR instead of the correctness of it (IMHO). It may seem like a lot, but I tried detailing as much as possible my requests. In any case, ping me if you need.

Besides that, it is looking really good :+1:

davidbtadokoro:
Hi @lorenzoberts and @BrizaMel, and thank you for the constant improvement of the PR. There were some things that I felt needed adjustments, but I did them myself to compensate the delay since your last update. I did tested it on an actual with some kernels and the implementation works as intended :+1: 

Issue #69, opened 2019-02-01
Improve "kw m --authors" for multiline MODULE_AUTHOR statements


matheustavares:
When we run a `kw m --authors`, only authors found in a single line MODULE_AUTHOR statement are captured and printed. In the future, it would be nice to look for multiline MODULE_AUTHOR statements such as:

```
MODULE_AUTHOR ("a_long_email@xx.com" \
               "another_long@yy.com" )
```

and:

```
MODULE_AUTHOR ("a_long_email@xx.com"
               "another_long@yy.com" )
```

which are, both, valid C statements. In the currently implementation neither of these four emails would be printed.

Lincoln-Yuji:
Working on it

PR #1100, opened 2024-05-01
src: maintainers: Enhance print for multiline MODULE_AUTHOR

Lincoln-Yuji:
The previous implementation of print_files_authors only captured and printed authors found in a single line of a MODULE_AUTHOR macro statement. This new modified version allows 'kw m --authors' to also catch and print authors on multiline statements.

Closes: #69


tests/unit/samples/print_file_author_test_dir/code3.c
```diff
@@ -0,0 +1,32 @@
...
+	for (i = 0; i < argc; ++i)
+		printf("%s ", args[i]);
+	return 0;
+}
+
+MODULE_AUTHOR ( "Martha Elsa <martha@opensource.com>" \
+				"Verena Bert <vb@opensource.com>" )
+ORDINARY_MACRO ("Does something...");
+MODULE_AUTHOR ( "Gabi Katinka <gabik@opensource.com>"
+				"Leonie Hildebert <leoni.hildebert@opensource.com>" )
```
rodrigosiqueira, line 32: Btw, I think you should check this file from the kernel:

drivers/media/usb/gspca/sq930x.c

src/maintainers.sh
```diff
@@ -160,9 +160,12 @@ function print_files_authors()
...
   for file in "${files[@]}"; do
-    authors=$(grep --only-matching --extended-regexp 'MODULE_AUTHOR *\(.*\)' "$file" |
-      sed --regexp-extended 's/(MODULE_AUTHOR *\( *\"|\" *\))//g' |
-      sed --expression ':a' --expression 'N' --expression '$!ba' --expression 's/\n/, /g')
+    authors=$(grep --null-data --perl-regexp --only-matching 'MODULE_AUTHOR *\(\K[^)]*' "$file" |
+      tr -d '\0' |
+      sed --expression ':a' --expression 'N' --expression '$!ba' |
+      sed --expression 's/\s*\\//g' --expression 's/^\s*/ /g' |
+      tr -d '\n' |
+      sed --expression 's/\"\s*\"/, /g' --expression 's/\s*\"//g' --expression 's/\s*$//g')
```
rodrigosiqueira, line 168: Alright, this is a huge command, could you help me to unpack it?

First: Could you elaborate a little bit about the `\K` option in this context? Maybe you can consider adding some comments to make it easier to understand these conditions.

Second: Could you break this huge command into a small set of commands? It is hard to review all of these in a single line.

Also, check this code style: https://kworkflow.org/content/codingstyle.html#avoid-short-options

I also noticed that when the MODULE_AUTHOR has `\n,` your change just drops the `\.` Check my comment in the test part.

tests/unit/samples/print_file_author_test_dir/code3.c
rodrigosiqueira, line 1: Maybe rename this file to something more meaningful.

rodrigosiqueira (changes requested):

src/maintainers.sh
```diff
@@ -160,9 +160,24 @@ function print_files_authors()
...
+    # Then we delete the backslash '\' with any possible adjacent white spaces
+    # And then substitute possible whites spaces at the beginning of the lines for a single one
+    authors=$(sed --expression ':a' --expression 'N' --expression '$!ba' <<< "$authors" |
+      sed --expression 's/\\n//g' --expression 's/\s*\\//g' --expression 's/^\s*/ /g' |
+      tr --delete '\n')
+
+    # Finally, we delete the unnecessary quotes and substitute them for commas to separate the authors
+    # The last expression removes any sequence of white spaces at the end of the lines
+    authors=$(sed --expression 's/\"\s*\"/, /g' --expression 's/\s*\"//g' --expression 's/\s*$//g' <<< "$authors")
+
```
Lincoln-Yuji, line 180: Hello. We've added some comments to help with the readability. Hope that makes it easier to undesrtand.

Basically, we get everything inside 'MODULE_AUTHOR' using grep and only match the content inside the parethesis thanks to \K until we find a close parenthesis character.

Then we apply some 'trimming' to the grepped content. Removing white characters such as tabs or spaces and now we also remove possible existing '\n' hard written in some authors' names since they are not necessary for print_files_authors.

Then we have a third block of commands, where we simply remove the quotes and put commas instead to split the authors' names. Sometimes we found out that this sequence would put some white spaces at the of the string. The last expression just removes them.

Lincoln-Yuji: Though we have to be honest. We are not totally sure what the `sed` at line 173 is supposed to do. We decided to keep it since we don't know if it catches some edge cases of authors or special uses of MODULE_AUTHOR or anything else.

We've made some testes without it and apparently it doesn't seem to affect the output. But again, we are not sure what this is doing so we decided to keep it until we find out if it's unnecessary.

src/maintainers.sh
Lincoln-Yuji: Made a comment regarding this. Hope now it's easier to understand what is happening

tests/unit/samples/print_file_author_test_dir/code3.c
Lincoln-Yuji: We've added an use of '\n' in the new version so now I think it handles the '\n' case correctly according to the sq930x.c file.

Lincoln-Yuji:
After multiple tests we didn't manage to find a file or directory which makes this new modified version of print_files_authors fail without the previous piece of code:

```
$ sed --expression ':a' --expression 'N' --expression '$!ba'
```

We are not sure what this is supposed to do yet. We've tried checking the commit when `--authors` was introduced but we didn't find anything documenting this change in details, probably due to the fact it's a very old commit and at the time this was not a big of deal I guess. **We've took the decision to remove it*.*

We squashed all commits into one as suggested by @davidbtadokoro and we are now waiting for feedback. Thank you

src/maintainers.sh
```diff
@@ -160,9 +160,23 @@ function print_files_authors()
   local printed_authors_separator=false
 
   for file in "${files[@]}"; do
-    authors=$(grep --only-matching --extended-regexp 'MODULE_AUTHOR *\(.*\)' "$file" |
-      sed --regexp-extended 's/(MODULE_AUTHOR *\( *\"|\" *\))//g' |
-      sed --expression ':a' --expression 'N' --expression '$!ba' --expression 's/\n/, /g')
+    # Get every content inside the MODULE_AUTHOR's parenthesis (even if it is in multiple lines)
+    # Here grep matches 'MODULE_AUTHOR (' followed by a sequence of non-')' characters
+    # But it only matches the characters to the right of \K thanks to its look-behind nature
+    authors=$(grep --null-data --perl-regexp --only-matching 'MODULE_AUTHOR *\(\K[^)]*' "$file" |
```
davidbtadokoro, line 166: In `MODULE_AUTHOR *`, a safer way to indicate white space for `grep` is `[[:space:]]`. So make it

```bash
... 'MODULE_AUTHOR[[:space:]]* ...
```

src/maintainers.sh
```diff
@@ -160,9 +160,23 @@ function print_files_authors()
...
+      tr --delete '\0')
```
davidbtadokoro, line 167: Either break the command (one line `authors=$(grep ...)` and another `authors=$(printf '%s' "$authors" | tr ...`), or make it one line.

src/maintainers.sh
```diff
@@ -160,9 +160,23 @@ function print_files_authors()
...
+
+    # Now we start trimming the content from the first grep
+    # Delete possible hard written '\n' from authors names
+    # Delete the backslash '\' with any possible adjacent white spaces
+    # Substitute possible white spaces at the beginning of the lines for a single one
+    authors=$(sed --expression 's/\\n//g' --expression 's/\s*\\\s*//g' --expression 's/^\s*/ /g' <<< "$authors" |
+      tr --delete '\n')
```
davidbtadokoro, line 174: Either break the command (one line `authors=$(sed ...)` and another `authors=$(printf '%s' "$authors" | tr ...`), or make it one line.

src/maintainers.sh
```diff
@@ -160,9 +160,23 @@ function print_files_authors()
...
+    # Get every content inside the MODULE_AUTHOR's parenthesis (even if it is in multiple lines)
+    # Here grep matches 'MODULE_AUTHOR (' followed by a sequence of non-')' characters
+    # But it only matches the characters to the right of \K thanks to its look-behind nature
+    authors=$(grep --null-data --perl-regexp --only-matching 'MODULE_AUTHOR *\(\K[^)]*' "$file" |
+      tr --delete '\0')
+
+    # Now we start trimming the content from the first grep
+    # Delete possible hard written '\n' from authors names
+    # Delete the backslash '\' with any possible adjacent white spaces
+    # Substitute possible white spaces at the beginning of the lines for a single one
```
davidbtadokoro, line 172: If possible, don't break it in phrases and say it directly in the imperative way to make it more succinct. For example,

```bash
# Trim contents from first grep deleting `\n` and `\` with any adjacent
# white spaces, and substitute leading white spaces to a single one
```

Again, this is a rough draft, and just a suggestion

src/maintainers.sh
```diff
@@ -160,9 +160,23 @@ function print_files_authors()
...
+    authors=$(sed --expression 's/\\n//g' --expression 's/\s*\\\s*//g' --expression 's/^\s*/ /g' <<< "$authors" |
+      tr --delete '\n')
+
+    # Finally, delete the unnecessary quotes and substitute them for commas to separate the authors
+    # The last expression removes any sequence of white spaces at the end of the lines
```
davidbtadokoro, line 177: Make it a one setence if possible.

davidbtadokoro (changes requested):
Hi guys, and thank you for the update! The PR looks legit!

It may seem like I left to many comments, but the big one is about what I think is the reasoning behind the `sed` `:a`, `N`, and `$!ba` expressions, there is one about using a safer way to indicate a white space for `grep`, and the rest is about the comments (thanks for adding them, btw).

Besides that, I think the commit message subject can be something like `... Handle multi line MODULE_AUTHOR for --authors`, as `Enhance print for multiline` doesn't really express what the change actually does.

Anyway, this is a great PR, and you guys are actively participating in the review process (which is the ideal :smiley:).]

**OBS.:** The best way to see the feedback from my review is going to the tab 'Commits' at the top of the page and seeing the comments inline.

Lincoln-Yuji:
Hello! Thank you for the feedback.

We've just finished applying your suggestions and we believe everything is fine now. But there is one thing that I forgot to mention
previously because it's something that the new code already covers though it might be good to let registered here.

Regarding the `grep` line:

```
authors=$(grep --null-data --perl-regexp --only-matching 'MODULE_AUTHOR\s*\(\K[^)]*' "$file" | tr --delete '\0')
```

We decided to put this entire command in a single line as it was one of the previous suggestions @davidbtadokoro made. We know the recommendation is to avoid long lines as much as possible, but the reason we have to pipe `sed` into `tr` is because, otherwise, grep will print something like:

```
./src/maintainers.sh: line 166: warning: command substitution: ignored null byte in input
```

every time at the terminal as an error/warning message, even though the content will be printed correctly without any failing test if we decide to break this line into 2 different ones. This message is not given if we pipe one after the other.

PR #1101, opened 2024-05-02
src: plugins: subsystems: drm: check enabled connectors

### LINKED ISSUE #1096: Improve --conn-available to share more helpful information about the connector usage
**Description:**

`kw drm` has the `--conn-available` option, which lists all the connectors available in the dGPU/APU. For example, if your GPU has three DPs and one HDMI you should see something like this:
```
[local] Card0 supports:
 DP
 DP
 DP
 HDMI
```
To make this information more meaningful, we should add a mark to tell the user which connector it is in use. From the above example, suppose that I have one display connected to the second DP and another one to the HDMI; we should have an output that looks like this:

```
[local] Card0 supports:
 DP
 DP *
 DP
 HDMI *
```

This information can be achieved by using the command:

```
cat /sys/class/drm/card0-<CONNECTOR-ID>/enabled
```

FelipeAnibal:
Make "kw drm --conn-available" more informative, marking enabled connectors with an '*'. Do this by checking the "/sys/class/drm/\<CONNECTOR\>/enabled" file.

Closes #1096

docs: add new pattern for the flag --conn-available in kw drm

tests: change tests for drm.sh

### DIFF
--- a/src/plugins/subsystems/drm/drm.sh
+++ b/src/plugins/subsystems/drm/drm.sh
@@ -1,6 +1,6 @@
-. "${KW_LIB_DIR}/lib/kw_config_loader.sh" --source-only
-. "${KW_LIB_DIR}/lib/remote.sh" --source-only
-. "${KW_LIB_DIR}/lib/kwlib.sh" --source-only
+include "${KW_LIB_DIR}/lib/kw_config_loader.sh"
+include "${KW_LIB_DIR}/lib/remote.sh"
+include "${KW_LIB_DIR}/lib/kwlib.sh"
 
 declare -gr UNLOAD='UNLOAD'
 declare -gA options_values
@@ -26,9 +26,9 @@ function drm_main()
 
   parse_drm_options "$@"
   if [[ "$?" -gt 0 ]]; then
-    complain "Invalid option: ${options_values['ERROR']} $target $gui_on $gui_off ${remote_parameters['REMOTE_IP']} ${remote_parameters['REMOTE_PORT']}"
+    complain "Invalid option: ${options_values['ERROR']} ${target} ${gui_on} ${gui_off} ${remote_parameters['REMOTE_IP']} ${remote_parameters['REMOTE_PORT']}"
     drm_help
-    return 22
+    return 22 # EINVAL
   fi
 
   target="${options_values['TARGET']}"
@@ -56,7 +56,7 @@ function drm_main()
   if [[ -n "$load_module" ]]; then
     module_control 'LOAD' "$target" "$remote" "$load_module" "$flag"
     if [[ "$?" != 0 ]]; then
-      return 22
+      return 22 # EINVAL
     fi
   fi
 
@@ -100,7 +100,7 @@ function module_control()
   local unformatted_remote="$3"
   local parameters="$4"
   local flag="$5"
-  local module_cmd=""
+  local module_cmd=''
   local remote
   local port
 
@@ -109,12 +109,12 @@ function module_control()
   module_cmd=$(convert_module_info "$operation" "$parameters")
   if [[ "$?" != 0 ]]; then
     complain 'Wrong parameter in --[un]load-module='
-    return 22
+    return 22 # EINVAL
   fi
 
   case "$target" in
     2) # LOCAL
-      cmd_manager "$flag" "sudo bash -c \"$module_cmd\""
+      cmd_manager "$flag" "sudo bash -c \"${module_cmd}\""
       ;;
     3) # REMOTE
       remote=$(get_based_on_delimiter "$unformatted_remote" ':' 1)
@@ -138,10 +138,10 @@ function convert_module_info()
   local unload="$1"
   shift
   local raw_modules_str="$*"
-  local parameters_str=""
-  local final_command=""
-  local remove_flag=""
-  local module_str=""
+  local parameters_str=''
+  local final_command=''
+  local remove_flag=''
+  local module_str=''
   local first_time=1
 
   if [[ "$unload" == "$UNLOAD" ]]; then
@@ -154,21 +154,21 @@ function convert_module_info()
   # Target event. e.g.: amdgpu_dm or amdgpu
   for module in "${modules[@]}"; do
     parameters_str=''
-    module_str="modprobe $remove_flag $module"
+    module_str="modprobe ${remove_flag} ${module}"
 
     if [[ "$module" =~ .*':'.* ]]; then
-      module_str="modprobe $remove_flag "
-      module_str+=$(cut -d ':' -f1 <<< "$module")
+      module_str="modprobe ${remove_flag} "
+      module_str+=$(cut --delimiter=':' --fields=1 <<< "$module")
 
       if [[ "$unload" != "$UNLOAD" ]]; then
         # Capture module parameters
-        specific_parameters_str=$(cut -d ':' -f2 <<< "$module")
+        specific_parameters_str=$(cut --delimiter=':' --fields=2 <<< "$module")
         IFS=',' read -r -a parameters_array <<< "$specific_parameters_str"
         for specific_parameter in "${parameters_array[@]}"; do
           parameters_str+="$specific_parameter "
         done
 
-        module_str+=" $parameters_str"
+        module_str+=" ${parameters_str}"
       fi
     fi
 
@@ -177,7 +177,7 @@ function convert_module_info()
       first_time=0
       continue
     fi
-    final_command+=" && $module_str"
+    final_command+=" && ${module_str}"
   done
 
   if [[ -z "$final_command" ]]; then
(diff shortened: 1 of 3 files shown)


FelipeAnibal:
The warnings from shellcheck were already happening before our changes and we don't understand how to fix them.

We changed the drm.sh file line 271 to pass the shfmt tests. In the shellcheck tests the warning may be related to this issue: https://github.com/koalaman/shellcheck/issues/2544.

FelipeAnibal:
The warnings from shellcheck were already happening before our changes and we don't understand how to fix them.
We changed the drm.sh file line 271 to pass the shfmt tests.

src/plugins/subsystems/drm/drm.sh
```diff
@@ -280,14 +282,27 @@ function get_available_connectors()
...
 
     if [[ -n "$key" && -n "$value" ]]; then
       list_of_values="${cards[$key]}"
+      case "$target" in
+        2) # LOCAL TARGET
+          connector_enabled=$(< "$card_path"/enabled)
+          ;;
+        3) # REMOTE TARGET
+          get_enabled_cmd="cat $card_path/enabled"
+          connector_enabled=$(cmd_remotely "$get_enabled_cmd" "$flag")
```
rodrigosiqueira, line 299: This approach will generate n SSH connections, which I think is not optimal and is also avoidable.
In the code before this part, we use the find command to identify all the connectors. We can leverage the `find` command to get the connector information and the enabled option in a single command, we can use something like this:

`find /sys/class/drm -name 'card*' -exec printf '%s - ' {} \; -exec cat {}/enabled \;`

P.S.: The above command is just a draft; you can probably improve it and maybe optimize it.

tests/unit/drm_plugin_test.sh
```diff
@@ -284,15 +287,25 @@ function test_get_available_connectors_remote()
   output=$(
     function cmd_remotely()
     {
-      printf '/sys/class/drm/card0-%s-1\n' 'DP'
-      printf '/sys/class/drm/card0-%s-1\n' 'eDP'
+      cmd_run=$(printf "%s" "$1" | cut -d ' ' -f1)
```
rodrigosiqueira, line 290: Why not use static results? Personally, I don't like the idea of inserting complex code into a mock function like this. Consider reworking this test to make it more straightforward while you address these shellchecks.

rodrigosiqueira (changes requested):
Hi,
First of all, thanks a lot for your PR.

Since you are working on a single specific function that is a bit outdated for the kw patterns, could you make one commit that adjusts basic things in this function and the next one that enables the feature that you are working on?

Also, you can see other comments in the PR.

Thanks

src/plugins/subsystems/drm/drm.sh
```diff
@@ -266,7 +268,7 @@ function get_available_connectors()
     2) # LOCAL TARGET
       cards_raw_list=$(find "$SYSFS_CLASS_DRM" -name 'card*' | sort -d)
       if [[ -f "$SYSFS_CLASS_DRM" ]]; then
-        ret="$?"
+        ret=$?
```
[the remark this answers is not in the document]
FelipeAnibal: In our first tests, before we changed this, the shfmt tests failed. But testing again now, it passed the tests! We will revert this change.

FelipeAnibal:
Thank you for reviewing the PR and for the feedback! We will work on it.

> Since you are working on a single specific function that is a bit outdated for the kw patterns, could you make one commit that adjusts basic things in this function and the next one that enables the feature that you are working on?

Just to clarify, in this quote what you are suggesting is that we update the function to match the current kw coding style? Or should we change some feature too?

rodrigosiqueira:
> Thank you for reviewing the PR and for the feedback! We will work on it.
> 
> > Since you are working on a single specific function that is a bit outdated for the kw patterns, could you make one commit that adjusts basic things in this function and the next one that enables the feature that you are working on?
> 
> Just to clarify, in this quote what you are suggesting is that we update the function to match the current kw coding style? Or should we change some feature too?

Code style. However, if you see anything else, feel free to update.

FelipeAnibal:
Hello! We've made the changes to coding style and updated the get_available_connectors function to avoid having multiple ssh connections.

Since the code style changes inside the function were very little, we update the whole file.

Thank you for the feedback and for the suggestion to use the "find" command. It is a lot better indeed.

FelipeAnibal:
We made some changes to pass the shfmt and shellchecker tests. 

We are now ignoring a check in the tests file, because an exception applies to it as described at https://www.shellcheck.net/wiki/SC2317#exceptions.

src/plugins/subsystems/drm/drm.sh
```diff
@@ -258,13 +258,14 @@ function get_available_connectors()
...
   local find_conn_cmd
+  local connector_enabled
   declare -A cards
 
   flag=${flag:-'SILENT'}
 
   case "$target" in
     2) # LOCAL TARGET
-      cards_raw_list=$(find "$SYSFS_CLASS_DRM" -name 'card*' | sort -d)
+      cards_raw_list=$(find "$SYSFS_CLASS_DRM" -name 'card*-*' -exec printf '%s,' {} \; -exec cat {}/enabled \; -exec printf '\n' \; | sort -d)
```
rodrigosiqueira, line 268: Add a small comment in this line that explains what's going on in the exec command. Also, use `--dictionary-order` instead of `-d` in the `sort` command. Check:
https://kworkflow.org/content/codingstyle.html#avoid-short-options

src/plugins/subsystems/drm/drm.sh
```diff
@@ -273,21 +274,28 @@ function get_available_connectors()
       target_label='local'
       ;;
     3) # REMOTE TARGET
-      find_conn_cmd="find ${SYSFS_CLASS_DRM} -name 'card*'"
+      find_conn_cmd="find ${SYSFS_CLASS_DRM} -name 'card*-*' -exec printf '%s,' {} \; -exec cat {}/enabled \; -exec printf '\n' \;"
```
rodrigosiqueira, line 277: This line and the one in the previous part of the `case` are very similar; to avoid duplication, put both of them in a single variable.

src/plugins/subsystems/drm/drm.sh
```diff
@@ -273,21 +274,28 @@ function get_available_connectors()
...
 
       cards_raw_list=$(cmd_remotely "$find_conn_cmd" "$flag" | sort -d)
       target_label='remote'
       ;;
   esac
 
-  while read -r card; do
-    card=$(basename "$card")
+  while read -r card_info; do
+    card_path=$(printf '%s' "$card_info" | cut -d ',' -f1)
```
rodrigosiqueira, line 285: https://kworkflow.org/content/codingstyle.html#avoid-short-options

rodrigosiqueira (changes requested):
Hi,
First of all, thanks a lot for the update. Overall, this PR looks good; I just added a few comments.

After using this feature, I realized that the below output is not very informative:

```
kw drm --conn-available 
[remote] Card0 supports:
 DP *
 DP
 DP *
 HDMI
 Writeback
```

It would be better if the output looks like this:

```
kw drm --conn-available 
[remote] Card0 supports:
 DP-1 *
 DP-2
 DP-3 *
 HDMI-A-1
 Writeback-1
```
In summary, you just need to change the cut section to the next `' '` instead of `-`.

src/plugins/subsystems/drm/drm.sh
FelipeAnibal: In the localt target, should we use "eval" to execute the command that we store in the variable?

davidbtadokoro: Hi @FelipeAnibal. Sorry for parachuting into the conversation, but I may be able to help :sweat_smile: 

In the local target, FWIU, I think you can store the whole command inside `$()` inside a variable, say, `foo`, and execute it with `cmd_manager` like `cards_raw_list=$(cmd_manager 'SILENT' "$foo")`. Underneath, this function uses `eval`, but it is an abstraction layer that allow us to control its execution (the first argument `SILENT` tells the function to evaluate it in non-verbose mode). For more info in `cmd_manager` you can see its use throughout the project and look at its [implementation](https://github.com/kworkflow/kworkflow/blob/af4d2e3100bd33c8dd07c8dba2c42cd7af020c9b/src/lib/kwlib.sh#L70).

otavioolsilva:
We followed the suggestions and made the changes!

tests/unit/drm_plugin_test.sh
```diff
@@ -259,23 +262,24 @@ function test_get_available_connectors_local()
...
-    'DP'
-    'DP'
-    'DP'
-    'DVI'
-    'HDMI'
+    "DP-1"
+    "DP-2"
+    "DP-3"
+    "DVI-D-1 *"
+    "HDMI-A-1"
```
davidbtadokoro, line 274: I don't think we need to use single-quotes for any of these.

davidbtadokoro (changes requested):
Hi @FelipeAnibal and @otavioolsilva, and thanks for the great job on updating and evolving this PR. It is looking really great!

To complement Siqueira's review, I've only added one inline comment about single-quotes, and have two (nitpick) remarks about the commit message of the second commit:

1. Limit the commit message body to 72 chars maximum.
2. I am digging the fine-grained description of each files/directories collateral changes, but, as you are adopting this approach, make `docs: [...]` to `documentation: update kw drm --conn-available option description` to reflect the correct path (there is no dir `docs` in the project). Also, I think you can adapt `tests: change tests for drm.sh` to something like `tests: update tests for kw drm --conn-available`.

Besides that, really nice PR so

Acked-by: David Tadokoro \<davidbtadokoro@usp.br\>

FelipeAnibal:
Thanks for the review! We've made the changes.

rodrigosiqueira:
Hi @FelipeAnibal ,
Thanks a lot for your contribution! I merged your change into the unstable branch.

Thanks!

Issue #298, opened 2021-06-03
Add policy about single and double quotes on codestyle


joaoseckler:
Our `codingstyle.rst` should include a topic about using single and double quotes, basically, saying that double quotes are only to be used when there are expansions inside the string. Else, single quotes are to be used.

PR #1103, opened 2024-05-06
Add documentation to single quotes use and fix codestyle for files in /src

gustavofukunaga:
Add documentation for use of single quotes (') instead of double quotes when there are no Bash expansions. Also updated files in `/src` that didn't follow this rule.


gustavofukunaga:
Closes #298

PR #1106, opened 2024-05-06
documentation: man: update man pages to use correct config files

Kaos9001:
Most man pages still mentioned a root level kworkflow.config file as the only source of user-adjustable settings. This commit replaces outdated references to kworkflow.config with the specific .kw/<command>.config format currently used by kw config and also recommends the use of kw config to adjust these options.

In doing so, this commit also removes the list of adjustable options from the main kw.rst man page, as these were outdated, are now mostly command-specific and could cause confusion in new users. Further changes are needed to redistribute these explanations (when needed) to more adequate locations.

### DIFF
--- a/documentation/man/features/kw-deploy.rst
+++ b/documentation/man/features/kw-deploy.rst
@@ -31,22 +31,23 @@ executed:
 
 You can specify the deploy target via command line by using the flag
 ``--remote <remote>:<port>`` (e.g., ``--remote 172.16.254.1:22``); however, if
-you do this frequently you will probably prefer to add this information to your
-local **kworkflow.config**. See the example below::
+you plan on deploying to the same remote frequently can benefit from using the
+``kw remote`` feature to save the SSH information in a configuration file
+used by kw, for example::
 
-  default_deploy_target=remote
-  ssh_user=root
-  ssh_ip=172.16.254.1
-  ssh_port=22
+  kw remote --add origin root@172.16.254.1
+
+For more information, check ``kw remote --help``
 
 If you want to install a new kernel version in your host machine, you can use
 the flag ``--local``; you will need to use your root password.
 
-Another typical operation when deploying a new kernel to a test machine, it is
-the reboot after the update. You can explicitly say it for **kw** by adding the
-flag ``--reboot``, add this to the **kworkflow.config** with::
+Another typical operation when deploying a new kernel to a test machine is
+rebooting after the update. You can add the ``--reboot`` flag to a command to
+explicitly make **kw** reboot the machine afterwards, or you can set this to 
+always happen by modifying ``reboot_after_deploy`` flag in **deploy.config** with::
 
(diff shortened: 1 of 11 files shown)


documentation/man/features/kw-deploy.rst
```diff
@@ -31,20 +31,17 @@ executed:
...
 You can specify the deploy target via command line by using the flag
 ``--remote <remote>:<port>`` (e.g., ``--remote 172.16.254.1:22``); however, if
-you do this frequently you will probably prefer to add this information to your
-local **kworkflow.config**. See the example below::
-
-  default_deploy_target=remote
-  ssh_user=root
-  ssh_ip=172.16.254.1
-  ssh_port=22
+you plan on deploying to the same target frequently you'll probably want to use
```
rodrigosiqueira, line 34: Replace `target` with `remote`.

documentation/man/features/kw-deploy.rst
```diff
@@ -31,20 +31,17 @@ executed:
...
+``kw remote`` to setup and save your target's information to your local
```
rodrigosiqueira, line 35: How about something like:

You can benefit from the feature ``kw remote`` to save the ssh information in a configuration file used by kw; for more information, check ``kw remote --help``.

Also, I think you should keep the example but use the `kw remote`

documentation/man/features/kw-deploy.rst
```diff
@@ -31,20 +31,17 @@ executed:
...
 
 If you want to install a new kernel version in your host machine, you can use
 the flag ``--local``; you will need to use your root password.
 
 Another typical operation when deploying a new kernel to a test machine, it is
 the reboot after the update. You can explicitly say it for **kw** by adding the
-flag ``--reboot``, add this to the **kworkflow.config** with::
+flag ``--reboot``, add this to the **deploy.config** with::
 
   reboot_after_deploy=yes
```
rodrigosiqueira, line 46: How about replacing this example with: `kw config deploy.reboot_after_deploy yes`

rodrigosiqueira (changes requested):
Thanks a lot for this great PR! I did not realize how inconsistent our documentation is.

documentation/man/features/kw-deploy.rst
Kaos9001: Done!

documentation/man/features/kw-deploy.rst
Kaos9001: Both done.

documentation/man/features/kw-deploy.rst
Kaos9001: Ended up rewriting the whole paragraph as it was very confusing, but I also added the proposed example.

rodrigosiqueira:
Thanks a lot! This is a super important improvement!
Your change is merged into the unstable branch.

Thanks again.

PR #1108, opened 2024-05-07
setup: install kernel build dependencies

aquilamacedo:
This update simplifies the process of installing necessary kernel build dependencies, tailored to the distribution.


setup.sh
```diff
@@ -110,6 +110,37 @@ function check_dependencies()
...
+# This function installs the kernel build dependencies.
+# It supports three Linux distributions: Debian, Arch Linux, and Fedora.
+function install_build_kernel_dependencies()
+{
+  local distro
+  local kernel_build_dependencies
+  local dependency_packages
+
+  distro=$(detect_distro '/')
+  kernel_build_dependencies="documentation/dependencies/kernel_build/${distro}.dependencies"
```
rodrigosiqueira, line 122: We may not find the user distro (e.g., NixOS); you must handle that before you try to concatenate the string.

setup.sh
```diff
@@ -222,16 +253,17 @@ function usage()
...
+  say '--help                              | -h    Display this usage message'
+  say "--install                           | -i    Install $app_name"
+  say "--uninstall                         | -u    Uninstall $app_name"
+  say '--skip-checks                       | -C    Skip checks (use this when packaging)'
+  say '--skip-docs                         | -D    Skip creation of man pages (use this when installing)'
+  say '--verbose                           | -v    Explain what is being done'
+  say '--force                             | -f    Never prompt'
+  say "--completely-remove                 | -r    Remove $app_name and all files under its responsibility"
+  say "--docs                              | -d    Build $app_name's documentation as HTML pages into ./build"
+  say "--install-build-kernel-dependencies | -k    Installs all necessary dependencies to build the kernel according to your distribution"
```
rodrigosiqueira, line 265: How about `install-kernel-dev-deps`? Any other ideas?

setup.sh
```diff
@@ -222,16 +253,17 @@ function usage()
...
-  say "--enable-tracing     | -t    Install ${app_name} with tracing enabled (use it with --install)"
+  say '--help                              | -h    Display this usage message'
+  say "--install                           | -i    Install $app_name"
+  say "--uninstall                         | -u    Uninstall $app_name"
+  say '--skip-checks                       | -C    Skip checks (use this when packaging)'
+  say '--skip-docs                         | -D    Skip creation of man pages (use this when installing)'
+  say '--verbose                           | -v    Explain what is being done'
+  say '--force                             | -f    Never prompt'
+  say "--completely-remove                 | -r    Remove $app_name and all files under its responsibility"
+  say "--docs                              | -d    Build $app_name's documentation as HTML pages into ./build"
```
rodrigosiqueira, line 264: Since you changed all of these lines, also use `{}` for `$app_name`.

rodrigosiqueira (changes requested):

setup.sh
```diff
@@ -110,6 +110,44 @@ function check_dependencies()
...
+{
+  local distro
+  local kernel_build_dependencies
+  local dependency_packages
+
+  distro=$(detect_distro '/')
+
+  # Check if distro is equal to 'none'
+  if [ "$distro" = 'none' ]; then
+    complain "Support for this distro is not available yet."
```
rodrigosiqueira, line 125: single-quotes

setup.sh
```diff
@@ -110,6 +110,44 @@ function check_dependencies()
...
+function install_kernel_dev_deps()
+{
+  local distro
+  local kernel_build_dependencies
+  local dependency_packages
+
+  distro=$(detect_distro '/')
+
+  # Check if distro is equal to 'none'
+  if [ "$distro" = 'none' ]; then
```
rodrigosiqueira, line 124: prefer `[[ ]]` and `==`.

setup.sh
```diff
@@ -263,6 +263,7 @@ function usage()
   say '--help                     | -h    Display this usage message'
   say "--install                  | -i    Install $app_name"
   say '--install-kernel-dev-deps  | -k    Installs all necessary dependencies to build the kernel according to your distribution'
+  say "--full-instalation         | -F    Install $app_name and its kernel development dependencies"
```
rodrigosiqueira: {}

setup.sh
```diff
@@ -555,6 +556,14 @@ function install_home()
   warning '-> For a better experience with kw, please, open a new terminal.'
 }
 
+function full_instalation()
+{
+  say 'Installing ...'
```
rodrigosiqueira, line 561: Maybe: `Installing kw ...

setup.sh
```diff
@@ -555,6 +556,14 @@ function install_home()
...
+  install_home
+  say 'Installing dependencies to build the kernel according to your distribution ...'
```
rodrigosiqueira, line 563: Maybe: `Installing kernel dependencies for build...`

rodrigosiqueira (changes requested):

setup.sh
aquilamacedo: This was already covered in the last commit of this PR

davidbtadokoro (changes requested):
Hi @aquilamacedo, and thank you for evolving this PR!

I have tested the use of the `--full-instalation` option in Fedora, Debian, and Arch through containers, and it worked as intended on my end :+1: 

I have just three comments:

1. There is a typo in the option, as `--full-installation` is the correct term instead of `--full-instalation`.
2. There is a duplication of the `Installing...` message as you can see in the print below. Maybe we should issue a generic installing message (make the first something like `Starting kw installation...`, then a dedicated message to each type of installation (the second something like `Installing kw base dependencies...`, and the third the way it is). 
2024-05-22-173915_963x450_scrot
3. The `setup.sh` script tries to install the full dependencies even if they are already installed, as you can see in the print below. You should do it like in the normal installation, i.e., check to see if the dependencies are already installed in the system.
2024-05-22-173928_805x674_scrot

aquilamacedo:
Hi @davidbtadokoro, nice catch! I did a step to check the deps, could you check if everything is in order?

setup.sh
```diff
@@ -110,6 +110,100 @@ function check_dependencies()
...
+        installed=$(rpm -q "$package" &> /dev/null)
+        [[ "$?" -ne 0 ]] && package_list="${package} ${package_list}"
+      done < "$kernel_build_deps_file"
+      cmd="dnf install -y ${package_list}"
+      ;;
+    *)
+      complain "Unsupported OS type: ${distro}"
+      return 1 # EPERM
+      ;;
+  esac
```
davidbtadokoro, line 157: I see that you took part of the `check_dependencies` as inspiration for this block that I am highlighting. Maybe we can extract a function to reduce duplication that accepts a distro and a dependencies file as arguments (in this case, the line `distro=$(detect_distro '/')` wouldn't be included in the function), and returns the `package_list`. The con of  this, is that we would still have a (smaller) switch-case in both `check_and_install_kernel_build_dependencies` and `check_dependencies` to build the correct `cmd` value.

This is just a rough idea, nevertheless, I think somehow extracting this duplicated switch-case (if-elif-else in `check_dependencies`) is cleaner.

setup.sh
```diff
@@ -110,6 +110,100 @@ function check_dependencies()
...
+
+  # Check if distro is equal to 'none'
+  if [[ "$distro" == 'none' ]]; then
+    complain 'Support for this distro is not available yet.'
+    exit 22 # EINVAL
+  fi
+
+  # Use check_and_install_kernel_build_dependencies() to handle dependency
+  # installation
+  check_and_install_kernel_build_dependencies
```
davidbtadokoro, line 204: As you are detecting the distro above and using the same expression to obtain it again inside `check_and_install_kernel_build_dependencies`, can't we pass `distro` as an argument here?

davidbtadokoro (changes requested):
Hi @aquilamacedo, and thanks for the quick update!

I've tested again the PR with containers, and it looks to me that you've tackled all my previous comments :+1: 

I left two more comments about the implementation, so see if you can address them. Otherwise, it looks legit :smile:

rodrigosiqueira:
Hi, this changes lgtm.
@davidbtadokoro , can you merge it if you are ok with this change?

davidbtadokoro:
Hey @aquilamacedo, and thank you for keeping on improving on this PR. I validated once again the changes by building kernels in debian, arch, and fedora containers, and everything still works like a charm! Legitimate, so merged into the unstable branch :+1:

PR #1110, opened 2024-05-08
src: plugins: kernel_install: utils: fix coding style error

fernando38954:
Replace the cat command by < which is equivalent but faster.
    
Substitute @ to * to explicitly join the elements on spaces.

### DIFF
--- a/src/plugins/kernel_install/utils.sh
+++ b/src/plugins/kernel_install/utils.sh
@@ -40,7 +40,7 @@ function cmd_manager()
       ;;
   esac
 
-  eval "$@"
+  eval "$*"
 }
 
 function command_exists()
--- a/src/plugins/kernel_install/utils.sh
+++ b/src/plugins/kernel_install/utils.sh
@@ -197,7 +197,7 @@ function collect_deploy_info()
   bootloader="[bootloader]=$bootloader"
 
   # Get distro
-  distro=$(cat /etc/*-release | grep -w 'ID\(_LIKE\)\?' | cut -d = -f 2 | xargs echo)
+  distro=$(cat /etc/*-release | grep --word-regexp 'ID\(_LIKE\)\?' | cut --delimiter = --fields 2 | xargs printf '%s ')
   distro="[distro]='$distro'"
 
   # Build associative array data


src/plugins/kernel_install/utils.sh
```diff
@@ -40,7 +40,7 @@ function cmd_manager()
       ;;
   esac
 
-  eval "$@"
+  "$@"
```
rodrigosiqueira: You dropped the eval here and re-introduced it in the next commit; in this sense, just remove this change from this commit.

src/plugins/kernel_install/utils.sh
```diff
@@ -197,7 +197,7 @@ function collect_deploy_info()
   bootloader="[bootloader]=$bootloader"
 
   # Get distro
-  distro=$(cat /etc/*-release | grep -w 'ID\(_LIKE\)\?' | cut -d = -f 2 | xargs echo)
+  distro=$(grep < /etc/*-release -w 'ID\(_LIKE\)\?' | cut -d = -f 2 | xargs echo)
```
rodrigosiqueira, line 200: Since you are adjusting codestyle in this part, also address:

https://kworkflow.org/content/codingstyle.html#avoid-short-options
https://kworkflow.org/content/codingstyle.html#use-printf-instead-of-echo

rodrigosiqueira (changes requested):

src/plugins/kernel_install/utils.sh
```diff
@@ -197,7 +197,7 @@ function collect_deploy_info()
   bootloader="[bootloader]=$bootloader"
 
   # Get distro
-  distro=$(cat /etc/*-release | grep -w 'ID\(_LIKE\)\?' | cut -d = -f 2 | xargs echo)
+  distro=$(grep < /etc/*-release --word-regexp 'ID\(_LIKE\)\?' | cut --delimiter = --fields 2 | xargs printf)
```
davidbtadokoro, line 200: Using `< /etc/*-release` results in the following error:

```bash
bash: /etc/*-release: ambiguous redirect
```

Try copying the command inside `$()` to see the error.

Te reason is that `<` works differently from `cat`, as `<` is a redirect that assumes a single file source, while `cat` accepts multiple file sources. That being said, I don't know if there is a way to use `<` along with the wildcard `*` , so maybe we will have to stick with `cat`...

davidbtadokoro, line 200: I think the correct way would be to maintain the order `cat` -> `grep` -> `cut`, i.e.

`distro=$(cat /etc/*-release | grep --word-regexp 'ID\(_LIKE\)\?' | cut --delimiter = --fields 2)`

davidbtadokoro, line 200: The use of `xargs printf` results in the following error:

```
printf: missing operand
```

If I am not mistaken, the use of `xargs` is unnecessary, but please confirm if this is the case.

davidbtadokoro (changes requested):
Hey guys, and thank you for the great update! You've nailed the use of `git rebase` and `git commit --amend` to update the PR :+1: 

I left some comments, so please see if you can address them. Also, when implementing changes, validate if the commands work as intended. As we are using Bash script, we have the advantage to "quick-validate" by running the commands on a Bash shell.

rodrigosiqueira:
@fernando38954 any updates here?

fernando38954:
Sorry for not replying for a long time, I've been a bit busy recently and didn't have much time to solve these reported errors.

Thanks for reminding me!

src/plugins/kernel_install/utils.sh
fernando38954: I notice that without xargs, the output of `printf '%s' "$bootloader $distro"` in the line 204 (where the `distro` variable is used) will have a break line in the middle of string that is not "human readable". So I decided to continue using the xargs but fixing the printf error and make it been must similar as possible as echo, but I didn't find a solution for add the break-line in the final unless adding a `&& printf '\n'` in the final of command of line 200. So I`d like to ask if addind this extra `printf` will violate some coding style error or not.

davidbtadokoro:
Hi @fernando38954 and Eduardo, and thank you guys for the PR and update! LGTM, so merged into the unstable branch :+1:

PR #1113, opened 2024-05-13
tests: integration: refactor kw_version_test to run entirely in container

aquilamacedo:
Refactored the kw version test to run entirely inside the container to
ensure compatibility with different fork and upstream repo labels. Moved
the `kw_version_check_version` function to dynamically determine branch
name, commit hash, and base version within the container. Simplified and
centralized the version check logic to avoid assumptions about the
developer's host repository setup.

### DIFF
--- a/tests/integration/kw_version_test.sh
+++ b/tests/integration/kw_version_test.sh
@@ -4,19 +4,21 @@ include './src/lib/kwio.sh'
 include './tests/unit/utils.sh'
 include './tests/integration/utils.sh'
 
-declare -g expected_output
-
-function oneTimeSetUp()
+# This function gets the commit hash and base version of the specified branch,
+# constructs the expected output.
+#
+# @container: The name of the container.
+#
+# Return:
+# Return 0: On success.
+# Return 1: If it fails to get the branch name, head commit hash, or base version.
+function kw_version_check_version()
 {
-  local kw_git_dir
-  local kw_dir
+  local container="$1"
   local head_hash
   local branch_name
   local base_version
-
-  # git directory path
-  kw_dir="${KWROOT_DIR}"
-  kw_git_dir="${kw_dir}/.git"
+  local expected_output
 
   # In order to check correctness of `kw --version`, we collect some information
   # from the git repo:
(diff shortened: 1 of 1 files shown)


tests/integration/utils.sh
```diff
@@ -431,3 +431,83 @@ function container_inspect()
...
+
+# This function sets up the kworkflow environment within a specified container.
+# It clones the kworkflow repository, checks out the specified branch and
+# optionally a specific commit, and then executes the setup script to install
+# kworkflow.
+#
+# @container: The name of the container.
+# @kw_tmp_dir: The temporary directory for kworkflow.
+# @kw_branch: The branch of kworkflow to be used.
+# @commit_point: The specific commit to be checked out.
```
rodrigosiqueira, line 443: Don't forget to document the resturn type.

tests/integration/utils.sh
```diff
@@ -431,3 +431,83 @@ function container_inspect()
...
+  local kw_branch="$3"
+  local commit_point="$4"
+  local kw_repo_url='https://github.com/kworkflow/kworkflow'
+
+  clone_repository_into_container "$container" "$kw_repo_url" "$kw_tmp_dir"
+
+  container_exec "$container" "git -C ${kw_tmp_dir} checkout ${kw_branch} > /dev/null 2>&1"
+  if [[ "$?" -ne 0 ]]; then
+    complain 'Git checkout operation failed'
+    complain "git -C ${kw_tmp_dir} checkout remotes/origin/${kw_branch}"
```
rodrigosiqueira, line 457: Use a single complain, something like this:
`    complain "Git checkout operation failed: git -C ${kw_tmp_dir} checkout remotes/origin/${kw_branch}"`

tests/integration/utils.sh
```diff
@@ -431,3 +431,83 @@ function container_inspect()
...
+{
+  local container="$1"
+  local kw_tmp_dir="$2"
+  local kw_branch="$3"
+  local commit_point="$4"
+  local kw_repo_url='https://github.com/kworkflow/kworkflow'
+
+  clone_repository_into_container "$container" "$kw_repo_url" "$kw_tmp_dir"
+
+  container_exec "$container" "git -C ${kw_tmp_dir} checkout ${kw_branch} > /dev/null 2>&1"
```
rodrigosiqueira, line 454: Why do we need `/dev/null ...`? Why not use `--quiet`?

tests/integration/utils.sh
```diff
@@ -431,3 +431,83 @@ function container_inspect()
...
+# @commit_point: The specific commit to be checked out.
+function setup_kworkflow_environment()
+{
+  local container="$1"
+  local kw_tmp_dir="$2"
+  local kw_branch="$3"
+  local commit_point="$4"
+  local kw_repo_url='https://github.com/kworkflow/kworkflow'
+
+  clone_repository_into_container "$container" "$kw_repo_url" "$kw_tmp_dir"
```
rodrigosiqueira, line 452: This function may fail, you should check its return.

tests/integration/utils.sh
```diff
@@ -431,3 +431,83 @@ function container_inspect()
...
+
+  container_exec "$container" "git -C ${kw_tmp_dir} checkout ${kw_branch} > /dev/null 2>&1"
+  if [[ "$?" -ne 0 ]]; then
+    complain 'Git checkout operation failed'
+    complain "git -C ${kw_tmp_dir} checkout remotes/origin/${kw_branch}"
+    return 1
+  fi
+
+  if [[ -n "$commit_point" ]]; then
+    container_exec "$container" "git -C ${kw_tmp_dir} checkout ${commit_point} > /dev/null 2>&1"
```
rodrigosiqueira, line 462: Again, why not use `--quiet`?

tests/integration/utils.sh
```diff
@@ -431,3 +431,83 @@ function container_inspect()
...
+    if [[ "$?" -ne 0 ]]; then
+      complain 'Git checkout to specific commit failed'
+      complain "git -C ${kw_tmp_dir} checkout ${commit_point}"
```
rodrigosiqueira, line 465: Same as my previous comment.

tests/integration/utils.sh
```diff
@@ -431,3 +431,83 @@ function container_inspect()
...
+  fi
+
+  container_exec "$container" "cd ${kw_tmp_dir} && ./setup.sh --install --force --skip-checks --skip-docs > /dev/null 2>&1"
+  if [[ "$?" -ne 0 ]]; then
+    complain 'kw setup failed'
+    complain "./setup.sh --install --force --skip-checks --skip-docs"
+    return 1
+  fi
+
+  return 0
```
rodrigosiqueira, line 477: After reading this function, I'm under the impression that you could separate each step into a single function. In this function, you just call the new functions in the sequence that you implemented here.

tests/integration/utils.sh
```diff
@@ -431,3 +431,83 @@ function container_inspect()
...
+}
+
+# This function clones a repository into a specified directory within a
+# container.
+#
+# @container: Name or ID of the container.
+# @repository_url: URL of the repository to clone inside the container.
+# @path_to_clone: Path to clone the repository inside the container.
```
rodrigosiqueira, line 485: Don't forget to document the return.

tests/integration/utils.sh
```diff
@@ -431,3 +431,83 @@ function container_inspect()
...
+function clone_repository_into_container()
+{
+  local container="$1"
+  local repository_url="$2"
+  local path_to_clone="$3"
+
+  container_exec "$container" "git clone ${repository_url} ${path_to_clone}"
+
```
rodrigosiqueira, line 493: drop this extra line.

tests/integration/kw_version_test.sh
```diff
@@ -6,61 +6,69 @@ include './tests/integration/utils.sh'
...
-  # commands directly on the host instead of running in the container.
-  head_hash=$(git --git-dir "${kw_git_dir}" rev-parse --short HEAD)
-  branch_name=$(git --git-dir "${kw_git_dir}" rev-parse --short --abbrev-ref HEAD)
-  base_version=$(head --lines 1 "${kw_dir}/src/VERSION")
-
-  # using the gathered information, we build the expected output
-  expected_output=$(printf '%s\nBranch: %s\nCommit: %s' "$base_version" "$branch_name" "$head_hash")
+  branch_name=$(container_exec "$container" "git -C ${kw_tmp_dir} rev-parse --short --abbrev-ref HEAD")
+  head_commit_hash=$(container_exec "$container" "git -C ${kw_tmp_dir} rev-parse --short origin/${branch_name}")
+  base_version=$(container_exec "$container" "git -C ${kw_tmp_dir} show origin/${branch_name}:./src/VERSION | head --lines 1")
```
rodrigosiqueira, line 31: All of these commands can fail. Add a check for each one.

tests/integration/kw_version_test.sh
```diff
@@ -6,61 +6,69 @@ include './tests/integration/utils.sh'
...
+
+  expected_output=$(printf '%s\nBranch: %s\nCommit: %s' "$base_version" "$branch_name" "$head_commit_hash")
+
+  printf '%s\n' "$expected_output"
```
rodrigosiqueira, line 35: Where is the check? You just printed the result.

tests/integration/kw_version_test.sh
```diff
@@ -6,61 +6,69 @@ include './tests/integration/utils.sh'
...
-  kw_version_test_helper 'debian'
-}
+    # clone the kworkflow repository inside the container.
+    clone_repository_into_container "$container" "$kw_repo_url" "$kw_tmp_dir"
 
-function test_kw_version_on_fedora()
-{
-  kw_version_test_helper 'fedora'
+    # switch to the 'unstable' branch for version checking.
+    container_exec "$container" "git -C ${kw_tmp_dir} checkout unstable > /dev/null 2>&1"
```
rodrigosiqueira, line 62: `--quiet` and this command may fail.

rodrigosiqueira (changes requested):
Thanks a lot for updating your PR.

tests/integration/kw_version_test.sh
aquilamacedo: Yes, because this function will return the expected output value.

davidbtadokoro:
Hi @aquilamacedo, and thank you for finding a great way to remove setup bias! The integration test suite for `kw version` looks better now. Everything looks legit to me, so merged into the unstable branch :+1:

Issue #83, opened 2019-03-05
Add code preview option


rodrigosiqueira:
When I'm exploring code in the Linux Kernel, I usually make this two steps incrementally:

1. kw e <SOMETHING> --> Show me something like:
```
path/file:117:<SOMETHING>
path/file:206:<SOMETHING>
```
2. Then I open the file to take a look at the implementation with the following command:
```
vim +117 path/file
```
I usually do it several times until I find the right code. It could be nice if kw has a feature to show the snippet code based on file and line. Something like:
```
kw snippet path/file 115
110
112 ==> 5 lines of code above 115
113
114
115 --> the target line with the code we want
116
117
118 ==> 5 lines of code below 115
119
120
```

matheustavares:
It doesn't solve the issue but git-grep (the backend of kw e) can be invoked with `-O` which will open the pager for each matched file at the matched line. It's even possible to give a custom pager with `-Ovi` or `-Oless` for example.

rodrigosiqueira:
For printing the line interval, we could use the `sed` command. Try it:

```
sed -n BEGIN,ENDp PATH/TO/FILE
```

rodrigosiqueira:
Ideas:

https://github.com/git/git/tree/master/contrib/git-jump

Lincoln-Yuji:
Is this feature still desirable? May I proceed working on it?

PR #1114, opened 2024-05-16
Add --snippet option to 'kw explore'

Lincoln-Yuji:
### Commit:
Often when exploring code in the Linux Kernel, it's usually desirable to not only see the paths and matched lines, but also the content from near lines. This option allows 'kw explore' to show small previews of the matched files interactively, helping the process of finding the right code.

### Inspiration
We got inspired by issue #83. However, we implemented something different from what was actually suggested in the issue's thread so we are not sure if this a solution to the problem. We believe this is an interesting addition to `kw explore` and gives it more power, but of course we are open to feedback and to know if this feature, in the current state, is good enough to be integrated into kw.

### DIFF
--- a/documentation/man/features/kw-explore.rst
+++ b/documentation/man/features/kw-explore.rst
@@ -6,7 +6,7 @@ kw-explore - Explore folder
 
 SYNOPSIS
 ========
-*kw* (*e* | *explore*) [(-l | \--log) | (-g | \--grep) | (-a | \--all) | \--verbose]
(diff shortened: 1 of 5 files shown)


Lincoln-Yuji:
Hmm we've just checked that this is actually more similar to this issue other issue #1052 

And this seems to be solved by this recent contribution: https://github.com/kworkflow/kworkflow/commit/1f050b5696bfd4e9aa0b48cf7a6056bf1d44a111

If that's the case, then I guess this PR can closed

PR #1118, opened 2024-05-23
tests: unit: maintainers_test: Add failing tests

Lincoln-Yuji:
The kw 'maintainers' feature lacks tests for its multiple corner cases where it should fail. Adding such tests ensures a larger test coverage for expected behaviors of this feature, possibly preventing future accidental undesirable changes.

### DIFF
--- a/tests/unit/maintainers_test.sh
+++ b/tests/unit/maintainers_test.sh
@@ -44,6 +44,9 @@ Maintainers already in 'To:' field of update_patch_test.patch"
 
 FAKE_KERNEL="tests/.tmp"
 
+# Original directory path to go back before each test function
+ORIGINAL_DIR="$PWD"
+
 function oneTimeSetUp()
 {
   # This creates tests/.tmp which should mock a kernel tree root. A .git
@@ -55,8 +58,9 @@ function oneTimeSetUp()
   cp -f tests/unit/samples/MAINTAINERS "$FAKE_KERNEL"/MAINTAINERS
   cp -f tests/unit/samples/external/get_maintainer.pl "$FAKE_KERNEL"/scripts/
   cp -f tests/unit/samples/update_patch_test{_model,}{,2}.patch "$FAKE_KERNEL"/
+  cp --force 'tests/unit/samples/cover_letter_test.patch' "$FAKE_KERNEL"
   cd "$FAKE_KERNEL" || {
-    fail "($LINENO) It was not possible to move to temporary directory"
+    fail "(${LINENO}) It was not possible to move to temporary directory"
     return
   }
   touch fs/some_file
@@ -66,7 +70,7 @@ function oneTimeSetUp()
   git config user.email kw@kw
   git commit --quiet -m "Test message"
   cd "$original_dir" || {
-    fail "($LINENO) It was not possible to move back from temp directory"
+    fail "(${LINENO}) It was not possible to move back from temp directory"
     return
   }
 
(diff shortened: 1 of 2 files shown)


rodrigosiqueira (changes requested):
Hi @Lincoln-Yuji ,
Thanks for this PR, I added some minor comments. Btw, this test is very old and probably requires some code refactor (e.g., split tests in small functions) and code style adjustment. Additionally, you can check the coverage report to see which part of the code has not been tested yet:

https://app.codecov.io/gh/kworkflow/kworkflow/blob/unstable/src%2Fmaintainers.sh

tests/unit/maintainers_test.sh
```diff
@@ -76,6 +77,8 @@ function oneTimeSetUp()
 function oneTimeTearDown()
 {
   rm -rf "$FAKE_KERNEL"
+  rm -rf "$SHUNIT_TMPDIR"
+  mkdir -p "$SHUNIT_TMPDIR"
```
rodrigosiqueira, line 81: https://kworkflow.org/content/codingstyle.html#avoid-short-options

Also, before every rm command, we must check if the variable is not empty or pointing to a weird path. This is very important to keep the code safe and avoid catastrophic removals. Check the function is_safe_path_to_remove.

Also, I know the old code just uses rm, but that is legacy code with some bad practices, and it must be updated.

tests/unit/maintainers_test.sh
```diff
@@ -109,16 +112,157 @@ function test_maintainers_main()
...
   ret="$(maintainers_main .)"
-  multilineAssertEquals "$CORRECT_TMP_FS_MSG" "$ret"
+  multilineAssertEquals "$ret" "$CORRECT_TMP_FS_MSG"
+
+  cd "$original_dir" || {
+    fail "(${LINENO}) It was not possible to move back from temp directory"
+    return
+  }
+  rm -rf "$SHUNIT_TMPDIR"
+  mkdir -p "$SHUNIT_TMPDIR"
```
rodrigosiqueira, line 129: Maybe this should be in the tearDown function, and we cleanup everything per test.

tests/unit/maintainers_test.sh
```diff
@@ -109,16 +112,157 @@ function test_maintainers_main()
...
   multilineAssertEquals "$CORRECT_TMP_MSG" "$ret"
 
   ret="$(maintainers_main .)"
-  multilineAssertEquals "$CORRECT_TMP_FS_MSG" "$ret"
+  multilineAssertEquals "$ret" "$CORRECT_TMP_FS_MSG"
+
+  cd "$original_dir" || {
+    fail "(${LINENO}) It was not possible to move back from temp directory"
+    return
+  }
```
rodrigosiqueira, line 127: Maybe you can move this `cd` to a specific folder to the `setUp` function and go back to the `tearDown` function; that way, each test can be simplified.

tests/unit/maintainers_test.sh
```diff
@@ -109,16 +112,157 @@ function test_maintainers_main()
...
+{
+  local ret
+  local -r original_dir="$PWD"
+
+  cd "$FAKE_KERNEL" || {
+    fail "(${LINENO}) It was not possible to move to temporary directory"
+    return
+  }
+
+  msg="$(maintainers_main --update-patch 'fs/some_file')"
```
rodrigosiqueira, line 179: Declare this variable as local.
Also, use a more intuitive name, something like output_from_function_xpto...

P.S.: Please review this entire commit with this comment in mind.

rodrigosiqueira (changes requested):

rodrigosiqueira:
> Hello, @rodrigosiqueira we have finished applying your requested changes and we are now waiting for new feedbacks. Thank you very much for reviewing our code and suggesting some changes.
> 
> Regarding the string concatenation style for `LINENO`, we ended up applying it for all occurences of this pattern in this file, not only the new lines we made. We did this to keep it consistent accross the entire file.
> 
> However, we had some issues adding quotes to this line:
> 
> ```shell
> cp -f tests/unit/samples/update_patch_test{_model,}{,2}.patch "$FAKE_KERNEL"/
> ```
> 
> It seems that adding either single or double quotes makes this command not copy the patch files and we are not sure if this expression is expanded by the shell itself or if the `cp` command handles this format.
> 
> Because of that we decided to only add the quotes to the new line we made and leave the others untouched.

Hi @Lincoln-Yuji , ignore my comment about cp.

Lincoln-Yuji:
Hi @rodrigosiqueira, we've finished applying your new suggestions and we hope everything is fine now. Thank you for the feedback

rodrigosiqueira:
Thanks! Change merged into the unstable branch.

Issue #664, opened 2022-08-06
Add support for git bisect under kw bd


rodrigosiqueira:
**Description/Why:**

As a kernel developer, I constantly use git bisect together with `kw bd`. It would be nice if we integrate kw bd with git bisect.

**Targeted users:**

* Developers/Testers trying to find a bad commit.

**API draft:**

```
kw bd --good SHA [--bad SHA]
kw bd --good
kw bd --bad
```

lorenzoberts:
Briza and I will try to solve this issue.

PR #1120, opened 2024-05-27
src: build_and_deploy: add support for git bisect under kw bd

lorenzoberts:
We integrated kw bd with git bisect. This involves adding specific flags to the command: --good and --bad. There wasn't a logic for this command, so it was necessary to create a separate file for kw bd and its tests.

Closes: #664

### DIFF
--- a/src/build_and_deploy.sh
+++ b/src/build_and_deploy.sh
@@ -1,0 +1,106 @@
+#
+# The `build_and_deploy.sh` file centralizes functions related to kw bd.
+#
+
+include "${KW_LIB_DIR}/deploy.sh"
+include "${KW_LIB_DIR}/build.sh"
+
+# kw bd is a shortcut to build and deploy a new kernel.
+# It also supports arguments related to git bisect usage.
+#
+# The idea of using git bisect along with kw bd is to find the
+# first commit where building and/or deploying the kernel failed.
+#
+# @raw_options String with all user options
+#
+function build_and_deploy_main()
+{
+
+  local raw_options="$*"
+
+  if [[ "$1" =~ -h|--help ]]; then
+    build_and_deploy_help "$1"
+    exit 0
+  fi
+
+  last_commit=$(eval git rev-parse HEAD)
+  parse_build_and_deploy_options "$last_commit" "$raw_options"
+
+  if [[ "$?" -gt 0 ]]; then
+    complain "${options_values['ERROR']}"
+    exit 22 # EINVAL
+  fi
+
+  run_bisect="${options_values['GOOD_OR_BAD']}"
+  good="${options_values['GOOD']}"
+  bad="${options_values['BAD']}"
+  deploy_main_args="${options_values['DEPLOY_MAIN_ARGS']}"
+
+  if [[ $run_bisect == 0 ]]; then
+    eval "build_kernel_main && deploy_main 1 ""$deploy_main_args"""
+  else
+    git bisect start &&
+      git bisect bad "$bad" &&
+      git bisect good "$good" &&
+      git bisect run eval "kw b && kw d ""$deploy_main_args"""
+  fi
+}
+
+# Build and deploy arguments parser
+#
+# @commit_hash The given commit hash
+# @raw_options Options other than @commit_hash
+#
+function parse_build_and_deploy_options()
+{
+  local commit_hash="$1"
+  shift
+  local raw_options="$*"
+
+  options_values['GOOD']=$commit_hash
+  options_values['BAD']=$commit_hash
+  options_values['GOOD_OR_BAD']=0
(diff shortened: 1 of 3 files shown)


src/build_and_deploy.sh
```diff
@@ -0,0 +1,106 @@
...
+{
+
+  local raw_options="$*"
+
+  if [[ "$1" =~ -h|--help ]]; then
+    build_and_deploy_help "$1"
+    exit 0
+  fi
+
+  last_commit=$(eval git rev-parse HEAD)
```
rodrigosiqueira, line 26: Why do you need to use `eval` here? Why not use the git command directly?

rodrigosiqueira: Also, running this command at this stage looks weird to me since the standard `kw bd` will be the main option. I guess it will be way less common for the end user to use the bisect option when compared with the normal build and deploy.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,106 @@
...
+  parse_build_and_deploy_options "$last_commit" "$raw_options"
+
+  if [[ "$?" -gt 0 ]]; then
+    complain "${options_values['ERROR']}"
+    exit 22 # EINVAL
+  fi
+
+  run_bisect="${options_values['GOOD_OR_BAD']}"
+  good="${options_values['GOOD']}"
+  bad="${options_values['BAD']}"
```
rodrigosiqueira, line 36: Usually, the way we approach the specific feature handling in the main function is something like this:

```bash
if [[ ${options_values['SPECIFIC_FEATURE_LABEL'] == 1} ]]; then
  specific stuff ...
fi
```

In other words, you should have a bisect option and put those specific parameters inside the if condition.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,106 @@
...
+  deploy_main_args="${options_values['DEPLOY_MAIN_ARGS']}"
+
+  if [[ $run_bisect == 0 ]]; then
+    eval "build_kernel_main && deploy_main 1 ""$deploy_main_args"""
```
rodrigosiqueira, line 40: Why do you need to eval here? Since you created a new file, it is better if you handle each function and check the output of each command.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,106 @@
...
+    complain "${options_values['ERROR']}"
+    exit 22 # EINVAL
+  fi
+
+  run_bisect="${options_values['GOOD_OR_BAD']}"
+  good="${options_values['GOOD']}"
+  bad="${options_values['BAD']}"
+  deploy_main_args="${options_values['DEPLOY_MAIN_ARGS']}"
+
+  if [[ $run_bisect == 0 ]]; then
```
rodrigosiqueira, line 39: Add double quotes around all bash variables access.

rodrigosiqueira: Also, kw bd should be the default behavior.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,106 @@
...
+    eval "build_kernel_main && deploy_main 1 ""$deploy_main_args"""
+  else
+    git bisect start &&
+      git bisect bad "$bad" &&
+      git bisect good "$good" &&
+      git bisect run eval "kw b && kw d ""$deploy_main_args"""
```
rodrigosiqueira, line 45: All of these commands can fail, and you cannot trust the SHA passed by the users.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,106 @@
...
+}
+
+# Build and deploy arguments parser
+#
+# @commit_hash The given commit hash
+# @raw_options Options other than @commit_hash
+#
+function parse_build_and_deploy_options()
+{
+  local commit_hash="$1"
```
rodrigosiqueira, line 56: We don't want to make parser functions receive specific parameters like this.

rodrigosiqueira (changes requested):
Hi @lorenzoberts @BrizaMel

Thanks for your PR. I like the idea of creating a specific file for `bd,` but I have a couple of comments.

1. Start this PR with a single commit that introduces the bd file and keeps the current functionality similar to what we have. Since you are making this change, you can also add a check after calling the build function and after the deploy function.
2. You can also start a new file just for handling git operations as described in this issue: https://github.com/kworkflow/kworkflow/issues/1123
3. Your implementation must handle good and bad as separate commits. You also need to implement validation in case the bisect process fails.

My overall suggestion:
1. Close this PR
2. Open one PR that just creates the bd file
3. Work on the issue #1123 as a separate PR (you can do it in parallel with 2)
4. After 2 and 3 get done, come back to the bd --bisect.

Issue #1051, opened 2024-02-28
Create a command to add "Reviewed-by:"/"Acked-by:" tag to N commits


melissawen:
**Description:**

In the past, I've asked to @matheustavares a way to add "Reviewed-by: Author name <author e-mail>" in the commit message of a sequence of N commits. He has suggested something like: `git rebase --exec="git commit --trailer='Reviewed-by: ... <...>'" HEAD~N`

Besides that, we can also add other tags like "Acked-by:", "Reported-by:", "Suggested-by:", "Co-developed-by:"... (?) that follow the same pattern.

**Why:**

When we send a patchset, getting a single message of acceptance is common (i.e. "This series is Reviewed-by: ...", but we need to add the reviewer tag to each commit of the patchset manually before resending a next version or applying them to upstream. Having this command will automate a daily developer/maintainer task. 

**Targeted users:**

Developers/Maintainers

**API draft:**

No idea. I think it may be under patch-hub umbrella since it sounds something useful in the patch handling context too.

rodrigosiqueira:
This is a really good idea, but I'm wondering where we can implement this. I have some ideas:

1. kw mail: Maybe we can add an option to add the reviewed-by/acked-by on the fly. I can only see this being useful in this scenario for cases in which someone already gave the reviewed/acked-by tag, and the developer just resends it. Do you see another case?
2. kw patch-hub: We can add an option in the patch-hub that the user can add these tags, but in that case, we might want to send it and not keep it locally.
3. kw maintainer: Maybe we can expand the maintainer feature to support this type of operation. Something like `kw m --add-reviewed-by <SHA>`

@melissawen @davidbtadokoro @aquilamacedo any thoughts?

Lincoln-Yuji:
Luiza Soezima, Sabrina Araujo and I have started working on this

Issue #1049, opened 2024-02-27
Create an alias to add the "Fixes" tag 


melissawen:
**Description:**

From time to time we need to add a "Fixes" tag to a commit before sending it upstream, to indicate which commit introduced the bug we are fixing. This tag follows a specific format, as described at: https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#describe-your-changes

"use the ‘Fixes:’ tag with the first 12 characters of the SHA-1 ID, and the one line summary"

We don't have a git command (similar to signed-of-by) for it, but we can user a [trailer](https://git-scm.com/docs/git-interpret-trailers) that does it consistently for us. I've set this trailer in my work branch, the next step should be add it to kw.

```
melissawen@killbill:~/linux-dev/drm-kernel/drm-misc(drm-misc-next) $ git config --get-regex trailer
trailer.fixes.key Fixes:
trailer.fixes.ifexists doNothing
trailer.fixes.ifmissing add
trailer.fixes.cmd .kw/glog-fix-commit.sh
 
melissawen@killbill:~/linux-dev/drm-kernel/drm-misc(drm-misc-next) $ cat .kw/glog-fix-commit.sh
#!/bin/sh
test -n "$1" && git log -1 $1 --oneline --format="%h (\"%s\")" --abbrev-commit --abbrev=12 || true
 
melissawen@killbill:~/linux-dev/drm-kernel/drm-misc(drm-misc-next) $ git commit --amend --trailer="fixes:9ca6bc246035"
```

**Why:**

We are automating a recurring task

**Targeted users:**

Developers/Maintainers

**API draft:**

No idea. Maybe we need to create a new command for handling commit messages

rodrigosiqueira:
Another feature that might be a good candidate for the maintainer feature... @davidbtadokoro @melissawen what do you think about that? Does it make sense for you?

PR #1121, opened 2024-05-27
src: signature: Add new feature 'kw signature'

Lincoln-Yuji:
### Purpose
Often a kernel developer has to perform some operations over commit or patch trailer lines. Adding tags such as 'Reviewed-by' or 'Acked-by' is a common task and sometimes it might be required to do this over multiple commits or patches. Also another common one is the 'Fixes' tag, that must follow a specific format.

This feature helps on these cases. It uses 'git rebase', 'git commit --amend' and 'git interpret-trailers' as its backend in order to make these tasks quicker to complete.

### Inspiration
Both #1049 and #1051 represent similar issues, regarding trailer line manipulation of commits and patches. The idea is to deliver a new feature that can make such tasks easier and quicker to deal with.

### Some known limitations
There is no way to "revert" any change made by this command through kw. For example, if the commit log changes, there is no command to undo the previous operation. User will have to manually use `git reset --hard` if changes are not what they expected.

This feature currently does not provide ways to configure conditional actions, which is possible via `git config` as shown by Melissa here: #1049

### DIFF
--- a/documentation/man/features/kw-signature.rst
+++ b/documentation/man/features/kw-signature.rst
@@ -1,0 +1,172 @@
+===================================================
+kw-signature - Write commit and patch trailer lines
+===================================================
+
+.. _signature-doc:
+
+SYNOPSIS
+========
+| *kw* *signature* (-s(<string>) | \--add-signed-off-by=(<string>)) [\--verbose] [<sha> | <patch>]
+| *kw* *signature* (-r(<string>) | \--add-reviewed-by=(<string>)) [\--verbose] [<sha> | <patch>]
+| *kw* *signature* (-a(<string>) | \--add-acked-by=(<string>)) [\--verbose] [<sha> | <patch>]
+| *kw* *signature* (-t(<string>) | \--add-tested-by=(<string>)) [\--verbose] [<sha> | <patch>]
+| *kw* *signature* (-C(<string>) | \--add-co-developed-by=(<string>)) [\--verbose] [<sha> | <patch>]
+| *kw* *signature* (-R(<string>) | \--add-reported-by=(<string>)) [\--verbose] [<sha> | <patch>]
+| *kw* *signature* (-f(<sha>) | \--add-fixes=(<sha>)) [\--verbose] [<sha> | <patch>]
+
+DESCRIPTION
+===========
+This feature adds Linux kernel tags to either commits or patches and helps
+to make this kind of task quicker to complete. It's a wrapper to some useful
+usage of the ``git commit --amend``, ``git rebase`` and ``git interpret-trailers``
+commands. By default, it uses the commit currently pointed by `HEAD`
+as the target of the operation if user does not specify one. At least one **--add**
+option is required.
+
+Most kernel tags contain a signature like:
+
+.. code-block:: text
+
+  TAG: NAME <EMAIL>
+
+By default, options that add these types of tags will use your current `user.name`
+and `user.email` from git configuration if nothing is specified in the command line.
+
+Using multiple options combined is possible. For instance, the following
+command will add **Tested-by** and **Signed-off-by**, in this order,
+to the commit's trailers pointed by `HEAD` using `user.name` and
+`user.email` from git config::
+
+  $ kw signature -t -s
+
+The order the trailer lines are added follows the general sequence:
+
+.. code-block:: text
+
+   Reported-by
+   Co-developed-by
+   Acked-by
+   Tested-by
+   Reviewed-by
+   Signed-off-by
+   Fixes
+
+At the same time, new lines are grouped by their signatures. For example:
+
+.. code-block:: text
+
+   $ kw signature -s'Joe Doe <joe@mail.xyz>' -s -t -t'Joe Doe <joe@mail.xyz>'
+
+   Tested-by: Some Name <some@mail.xyz>
+   Signed-off-by: Some Name <some@mail.xyz>
+   Tested-by: Joe Doe <joe@mail.xyz>
+   Signed-off-by: Joe Doe <joe@mail.xyz>
+
+However, it's important to note that every new tag is added **after** the old ones
+that were already present in the commit or patch message.
+
+Another important aspect of this command is that, since it uses ``git rebase``
+as a part of its backend, if we run it using a SHA like ``kw signature -s HEAD~3``
+then only commits refered by ``HEAD~2``, ``HEAD~1`` and ``HEAD`` will be affected.
+This has the same behavior when you execute ``git rebase --interactive HEAD~3``.
+
+OPTIONS
+=======
+-s, \--add-signed-off-by:
+  Adds a **Signed-off-by** kernel tag line to either commits or patchsets.
+  By default it uses `user.name` and `user.email` from git config to
+  write these lines if no argument is given to this option.
+  
+-r, \--add-reviewed-by:
+  Adds a **Reviewed-by** kernel tag  line to either commits or patchsets.
+  By default it uses `user.name` and `user.email` from git config to
+  write these lines if no argument is given to this option.
+
+-a, \--add-acked-by:
+  Adds a **Acked-by** kernel tagline to either commits or patchsets.
+  By default it uses `user.name` and `user.email` from git config to
+  write these lines if no argument is given to this option.
+
+-t, \--add-tested-by:
+  Adds a **Tested-by** kernel tag line to either commits or patchsets.
+  By default it uses `user.name` and `user.email` from git config to
+  write these lines if no argument is given to this option.
+
+-C, \--add-co-developed-by:
+  Adds a **Co-developed-by** immediately followed by a **Signed-off-by**
+  kernel tag line to either commits or patchsets. By default it uses
+  `user.name` and `user.email` from git config to write these lines if
+  no argument is given to this option.
+
+-R, \--add-reported-by:
+  Adds a **Reported-by** kernel tag line to either commits or patchsets.
+  By default it uses `user.name` and `user.email` from git config to
+  write these lines if no argument is given to this option. Also, the
+  user has the option to pass a **Closes** or **Link** as an additional
+  tag that will be added immediately after for cases when the bug report
+  is available in the web.
+
+-f, \--add-fixes:
+  Adds a **Fixes** kernel tag line to either commits or patchsets.
+  It requires an argument, which must be a valid commit reference, to
+  define the fixed commit's hash that follows this tag.
+
+\--verbose:
+  Displays commands executed under the hood.
+
+EXAMPLES
+========
+Adding a **Reviewed-by** line to commit pointed by ``HEAD``::
+
+  $ kw signature --add-reviewed-by='Reviewer Name <reviewer@mail.org>'
+
+Adding an **Acked-by** line starting from ``HEAD~2`` until ``HEAD``::
+
+  $ kw signature --acked-by='Some Name <example@mail.com>' HEAD~3
+
+Adding **Fixes** line to commit pointed by ``HEAD``, which fixes another
+previous commit ``90a7ba23340d``::
+
+  $ kw signature --add-fixes='90a7ba23340d'
+
+All the above options can be used to perform operations over `.patch` files as well::
+
+  $ kw signature -r'Reviewer Name <reviewer@mail.org>' example.patch
+
+  $ kw signature -a'Some Name <example@mail.org>' example.patch
+
+  $ kw signature -f'90a7ba23340d' example.patch
+
+This command accepts multiples arguments, which means that multiple files
+(both names and globs) and commits can be passed with one single command.
+
+Adding **Signed-off-by** line to multiple `.patch` files::
+
+  $ kw signature -s'Some Name <example@mail.org>' file1.patch file2.patch
+
+This command also accepts globs to reference multiple `.patch` files::
+
+  $ kw signature -s'Some Name <example@mail.org>' *.patch
+
+One more complex example than the one seen in **DESCRIPTION** is::
+
+  $ kw signature -s'Jane Doe <janedoe@mail.xyz>' \
+    -t'Jane Doe <janedoe@mail.xyz>' \
+    -R'Michael Doe <michaeldoe@mail.xyz>;Closes=https://link-to-bug.xyz' \
+    -C'John Doe <johndoe@mail.xyz>' \
+    -C'Michael Doe <michaeldoe@mail.xyz>' \
+    -r'Jane Doe <janedoe@mail.xyz>'
+
+That will write these trailers like so:
+
+.. code-block:: text
+
+  Reported-by: Michael Doe <michaeldoe@mail.xyz>
+  Closes: https://link-to-bug.xyz
+  Co-developed-by: Michael Doe <michaeldoe@mail.xyz>
+  Signed-off-by: Michael Doe <michaeldoe@mail.xyz>
+  Co-developed-by: John Doe <johndoe@mail.xyz>
+  Signed-off-by: John Doe <johndoe@mail.xyz>
+  Tested-by: Jane Doe <janedoe@mail.xyz>
+  Reviewed-by: Jane Doe <janedoe@mail.xyz>
+  Signed-off-by: Jane Doe <janedoe@mail.xyz>
--- a/kw
+++ b/kw
@@ -286,6 +286,13 @@ function kw()
         vm_main "$@"
       )
       ;;
+    signature)
+      (
+        include "${KW_LIB_DIR}/signature.sh"
+
+        signature_main "$@"
+      )
+      ;;
     version | --version | -v)
       (
         include "${KW_LIB_DIR}/help.sh"
--- a/src/_kw
+++ b/src/_kw
@@ -41,6 +41,7 @@ _kw()
     'remote:Manage the set of test machines ("remotes")'
     'report:Show kw pomodoro reports and kw usage statistics'
     'self-update:kw self-update mechanism'
+    'signature:Write trailers in commits or patches'
     'ssh:SSH support'
     'patch-hub: Open UI with lore.kernel.org archives'
     'version:Show kw version'
@@ -407,8 +408,8 @@ _kw_man()
 {
   _values 'kw commands' \
     'backup' 'bd' 'build' 'codestyle' 'config' 'debug' 'deploy' 'device' 'diff' 'drm' 'env' \
-    'explore' 'init' 'send-patch' 'maintainers' 'pomodoro' 'remote' 'report' 'self-update' 'ssh' \
-    'patch-hub' 
+    'explore' 'init' 'send-patch' 'maintainers' 'pomodoro' 'remote' 'report' 'self-update' \
+    'signature' 'ssh' 'patch-hub' 
 }
 
 _kw_p() { _kw_pomodoro }
@@ -484,6 +485,20 @@ _kw_self-update()
   _arguments : \
     '(--verbose)--verbose[enable verbose mode]' \
     '(-u --unstable)'{-u,--unstable}'[update kw based on the unstable branch]'
+}
+
+_kw_signature()
+{
+  _arguments : \
+    '(--verbose)--verbose[enable verbose mode]' \
+    {-s-,--add-signed-off-by=-}'[add Signed-off-by tag line]' \
+    {-r-,--add-reviewed-by=-}'[add Reviewed-by tag line]' \
+    {-a-,--add-acked-by=-}'[add Acked-by tag line]' \
+    {-t-,--add-tested-by=-}'[add Tested-by tag line]' \
+    {-C-,--add-co-developed-by=-}'[add Co-developed-by and Signed-off-by tag lines]' \
+    {-R-,--add-reported-by=-}'[add Reported-by and optional Closes|Link tag lines]' \
+    {-f-,--add-fixes=-}'[add Fixes tag line]' \
+    '1: :_files'
 }
 
 _kw_s() { _kw_ssh }
--- a/src/signature.sh
+++ b/src/signature.sh
@@ -1,0 +1,639 @@
+include "${KW_LIB_DIR}/lib/kw_config_loader.sh"
+include "${KW_LIB_DIR}/lib/kwlib.sh"
+
+declare -gA options_values
+
+# Define a constant to store all kernel tags
+declare -gA KERNEL_TAGS
+
+KERNEL_TAGS['REPORTED_BY']='Reported-by:'
+KERNEL_TAGS['CO_DEVELOPED_BY']='Co-developed-by:'
+KERNEL_TAGS['ACKED_BY']='Acked-by:'
+KERNEL_TAGS['TESTED_BY']='Tested-by:'
+KERNEL_TAGS['REVIEWED_BY']='Reviewed-by:'
+KERNEL_TAGS['SIGNED_OFF_BY']='Signed-off-by:'
+KERNEL_TAGS['FIXES']='Fixes:'
+
+# This structure organizes all the parsed trailer lines by their tags.
+# For example, TRAILER_BUFFERS['SIGNED_OFF_BY'] contains all trailers
+# tagged with 'Signed-off-by'. It's a string where trailers are separated
+# by commas like:
+# "Signed-off-by: Joe Doe <joedoe@mail.xyz>,Signed-off-by: Jane Doe <janedoe@mail.xyz>"
+#
+# It acts as an auxiliary variable to sort the trailers that will be written later
+declare -gA grouped_trailers_by_tag
+
+# This structure groups different tags based on unique signatures.
+# For example, grouped_tags['blabla@mail.xyz'] is a string which contains
+# trailer lines associated to this email and separated by commas like:
+# "Reviewed-by: Joe Doe <joedoe@mail.xyz>,Signed-off-by: Joe Doe <joedoe@mail.xyz>"
+#
+# It acts as an auxiliary variable to sort the trailers that will be written later
+declare -gA grouped_trailers_by_email
+
+# This variable holds all the emails from signatures. The order of such emails
+# represent the order each email entry in `grouped_trailers_by_email` was created
+# during this feature's execution.
+declare -ga sorted_emails
+
+# Variable to store all trailers that will be added. If `sort_all_trailers`
+# function is used, then the order obeys the following general sequence:
+# - Reported-by
+# - Co-developed-by
+# - Acked-by
+# - Tested-by
+# - Reviewed-by
+# - Signed-off-by
+# - Fixes
+# As trailer lines with the same signature are grouped together. Example:
+# Reviewed-by: Joe Doe <joedoe@mail.xyz>
+# Signed-off-by: Joe Doe <joedoe@mail.xyz>
+# Reviewed-by: Bob Doe <bobdoe@mail.xyz>
+# Signed-off-by: Bob Doe <boobdoe@mail.xyz>
+declare -g all_trailers
+
+# This function performs operations over trailers in
+# either patches or commits. It checks if given argument
+# is a valid commit reference or patch path and uses the
+# correct command to perform the task.
+# If that's not the case, a warning message will tell
+# the user this argument was ignored.
+#
+# Also, if no operation option is given, then an error message
+# followed by a helper message is printed to the user.
+#
+# @patch_or_sha_args Holds either patch paths or commit references.
+function signature_main()
+{
+  local patch_or_sha_args
+  local flag
+
+  if [[ "$1" =~ -h|--help ]]; then
+    signature_help "$1"
+    exit 0
+  fi
+
+  # Ensure all trailer arrays are empty
+  for key in "${!grouped_trailers_by_tag[@]}"; do
+    unset "grouped_trailers_by_tag[$key]"
+  done
+  for key in "${!grouped_trailers_by_email[@]}"; do
+    unset "grouped_trailers_by_email[$key]"
+  done
+  sorted_emails=()
+  all_trailers=''
+
+  # Parse all command line options
+  parse_signature_options "$@"
+  if [[ "$?" -gt 0 ]]; then
+    complain "${options_values['ERROR']}"
+    return 22 # EINVAL
+  fi
+
+  [[ -n "${options_values['VERBOSE']}" ]] && flag='VERBOSE'
+  flag=${flag:-'SILENT'}
+
+  read -ra patch_or_sha_args <<< "${options_values['PATCH_OR_SHA']}"
+
+  if [[ -n "${options_values['NO_ADD_OPTION']}" ]]; then
+    complain 'At least one --add option is required to use this command'
+    signature_help
+    return 22 # EINVAL
+  fi
+
+  sort_all_trailers
+
+  for arg in "${patch_or_sha_args[@]}"; do
+    write_all_trailers "$arg" "$flag"
+  done
+}
+
+# This function validates the signature. If the passed signature
+# does not follow the format 'NAME <EMAIL>' with a valid EMAIL, then
+# it returns an error.
+#
+# @signature Holds the string that is validated
+#
+# Return:
+# Returns 0 if signature is valid; returns 22 otherwise.
+function is_valid_signature()
+{
+  local signature="$1"
+
+  if [[ "$signature" =~ ^[^\<]+\ [^\<]+\<([^\>]+)\>$ ]]; then
+    validate_email "${BASH_REMATCH[1]}"
+    if [[ "$?" -gt 0 ]]; then
+      return 22 # EINVAL
+    fi
+  else
+    return 22 # EINVAL
+  fi
+}
+
+# This functions extracts the NAME from a trailer line with a valid signature.
+# Valid signatures can be verified with `is_valid_signature`.
+#
+# @trailer A string that follows the pattern "TRAILER_TAG: NAME <EMAIL>"
+function extract_name_from_trailer()
+{
+  local trailer="$1"
+  printf '%s' "$trailer" | grep --only-matching --perl-regexp '(?<=: ).*?(?= <)'
+}
+
+# This functions extracts the EMAIL from a trailer line with a valid signature.
+# Valid signatures can be verified with `is_valid_signature`.
+#
+# @trailer A string that follows the pattern "TRAILER_TAG: NAME <EMAIL>"
+function extract_email_from_trailer()
+{
+  local trailer="$1"
+  printf '%s' "$trailer" | grep --only-matching --perl-regexp '<\K[^>]+'
+}
+
+# # TODO: To move this function to a 'gitlib' library
+#
+# This function validates if a commit reference is valid.
+#
+# @sha Holds a string that can be a commit hash or pointer.
+#
+# Returns:
+# True if given argument is a valid commit reference and false otherwise.
+function is_valid_commit_reference()
+{
+  local sha="$1"
+
+  if [[ $(git cat-file -t "$sha" 2> /dev/null) != 'commit' ]]; then
+    return 1 # EPERM
+  fi
+}
+
+# This function parses and adds a new trailer line into
+# @all_trailers that will be properly written later, however
+# the trailer isn't added if there's another identical one.
+# It attempts to use the user's name and email configured
+# if a signature is not passed as argument and gives
+# an error if they are not properly set with git config.
+#
+# @keyword Holds a string that is a keyword for a specific operation.
+# @signature Holds a string like 'NAME <EMAIL>' defining the signature.
+#
+# Return:
+# In case of successful return 0 adding the parsed operation;
+# It returns 61 if either user.name or user.email are not configured properly;
+# It returns 22 if the signature does not follow 'NAME <EMAIL>' format;
+function parse_and_add_trailer()
+{
+  local keyword="$1"
+  local signature="$2"
+  local formated_output
+  local parsed_trailer
+  local parsed_trailer_name
+  local parsed_trailer_email
+  local trailer_name
+  local trailer_email
+  local repeated_trailer=false
+  local duplicated_email=false
+
+  parsed_trailer="${KERNEL_TAGS[$keyword]} "
+
+  # Use default from git config if no argument was given
+  if [[ ! "$signature" ]]; then
+    formated_output="$(format_name_email_from_user)"
+    if [[ "$?" -gt 0 ]]; then
+      return 61 # ENODATA
+    fi
+    parsed_trailer+="$formated_output"
+  else
+    signature="$(str_strip "$signature")"
+    if [[ "$keyword" != 'FIXES' ]] && ! is_valid_signature "$signature"; then
+      return 22 # EINVAL
+    fi
+    parsed_trailer+="$signature"
+  fi
+
+  # Check if this trailer has the same email, but a different name compared to
+  # others already read and saved in `all_trailers`.
+  while read -d ',' -r trailer; do
+    # If it's a (Closes|Link) trailer, then there is no need to check email duplication.
+    if [[ "${trailer}" =~ ^(Closes|Link) ]]; then
+      continue
+    fi
+    parsed_trailer_name=$(extract_name_from_trailer "$parsed_trailer")
+    parsed_trailer_email=$(extract_email_from_trailer "$parsed_trailer")
+    trailer_name=$(extract_name_from_trailer "$trailer")
+    trailer_email=$(extract_email_from_trailer "$trailer")
+    if [[ "$trailer_email" == "$parsed_trailer_email" && "$trailer_name" != "$parsed_trailer_name" ]]; then
+      duplicated_email=true
+      warning "Same email used with different names: ${trailer_name}, ${parsed_trailer_name}, ${trailer_email}"
+      warning "Skipping the following trailer line: '${parsed_trailer}'"
+      break
+    fi
+  done <<< "$all_trailers"
+
+  # Check for trailer duplication
+  while read -d ',' -r item; do
+    if [[ "$item" == "$parsed_trailer" ]]; then
+      repeated_trailer=true
+      warning "Skipping duplicated trailer line: '${parsed_trailer}'"
+      break
+    fi
+  done <<< "${grouped_trailers_by_tag[$keyword]}"
+
+  if [[ "$repeated_trailer" = false ]] && [[ "$duplicated_email" = false ]]; then
+    grouped_trailers_by_tag["$keyword"]+="${parsed_trailer},"
+    all_trailers+="${parsed_trailer},"
+  fi
+}
+
+# This function receives a commit SHA and verifies if it's
+# a valid commit reference. If it is, then it outputs the
+# appropriate formatted message. Else it returns an error code.
+#
+# @sha Holds either a commit hash or pointer
+#
+# Return:
+# In case of successful return 0 and prints the formatted message,
+# otherwise, return 22.
+function format_fixes_message()
+{
+  local sha="$1"
+  local formatted_message
+
+  # Check if given value is a valid commit reference
+  if ! is_valid_commit_reference "$sha"; then
+    return 22 # EINVAL
+  fi
+
+  # The 'Fixes:' trailer line must follow a format defined by Linux Kernel developers.
+  # Fixes: e21d2170f366 ("video: remove unnecessary platform_set_drvdata()")
+  formatted_message=$(git log -1 "$sha" --oneline --abbrev-commit --abbrev=12 \
+    --format="%h (\\\"%s\\\")")
+
+  printf '%s' "$formatted_message"
+}
+
+# It gets user.name and user.email from git's configuration.
+# If either the name or email are not configured then this
+# function will return an error code. Otherwise it will output
+# the formatted name and email.
+#
+# Return:
+# In case of successful return 0 and prints a properly formatted
+# output, otherwise, return 61.
+function format_name_email_from_user()
+{
+  local user_name
+  local user_email
+
+  user_name="$(git config user.name)"
+  user_email="$(git config user.email)"
+
+  # If user doesn't have either a name or email configured with
+  # git then they must provide an argument
+  if [[ -z "$user_name" || -z "$user_email" ]]; then
+    return 61 # ENODATA
+  fi
+
+  printf '%s' "${user_name} <${user_email}>"
+}
+
+# This function gets all trailers saved in a `grouped_trailers_by_tag`
+# entry and associates each trailer to a `grouped_trailers_by_email` entry.
+#
+# @specified_tag Holds the tag group that will be re-grouped by their emails.
+function group_by_email()
+{
+  local specified_tag="$1"
+  local email
+
+  while read -d ',' -r trailer; do
+    # Extract email from trailer
+    email=$(extract_email_from_trailer "$trailer")
+    # Save email if no signature with it was saved so far
+    if [[ -z "${grouped_trailers_by_email[$email]}" ]]; then
+      sorted_emails+=("$email")
+    fi
+    # Remove the associative email from (Closes|Link) trailer lines
+    if [[ "${trailer}" =~ ^(Closes|Link) ]]; then
+      trailer=$(printf '%s' "$trailer" | sed --regexp-extended 's/ <[^>]*>$//')
+    fi
+    # Append the signature to the associative array based on the email
+    grouped_trailers_by_email["$email"]+="${trailer},"
+  done <<< "${grouped_trailers_by_tag[$specified_tag]}"
+}
+
+# This function adds every parsed trailer in `grouped_trailers_by_tag`
+# entries to `all_trailers`, following the general order proposed at the
+# `all_trailers` declaration.
+#
+# The strategy to achieve that is to sort all trailers by their tags first,
+# then sort them by their emails. By doing so, this function can enforce that
+# general order, and additionally it forces signatures from the same person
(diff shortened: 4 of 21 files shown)


src/handle_trailer.sh
```diff
@@ -0,0 +1,169 @@
...
+          options_values['ERROR']='Invalid commit reference with --add-fixes option: '
+          options_values['ERROR']+="${options_values['TRAILER_VALUE']}"
+          return 22 # EINVAL
+        fi
+        # The 'Fixes:' trailer line must follow a format defined by
+        # Linux Kernel developers. Example:
+        # Fixes: e21d2170f366 ("video: remove unnecessary platform_set_drvdata()")
+        options_values['TRAILER_VALUE']=$(git log -1 "${options_values['TRAILER_VALUE']}" \
+          --oneline --format="%h (\\\"%s\\\")" \
+          --abbrev-commit --abbrev=12)
```
rodrigosiqueira, line 125: It would be better if this manipulation were moved to a single function and called from the main functions instead of here.

src/handle_trailer.sh
```diff
@@ -0,0 +1,169 @@
...
+        shift
+        ;;
+    esac
+  done
+}
+
+function handle_trailer_help()
+{
+  if [[ "$1" == --help ]]; then
+    include "$KW_LIB_DIR/help.sh"
```
rodrigosiqueira, line 157: String concatenation.

tests/unit/handle_trailer_test.sh
```diff
@@ -0,0 +1,260 @@
...
+include './tests/unit/utils.sh'
+
+# Correct trailers for --add-reviewed-by and -r
+CORRECT_REVIEWED_COMMIT="Signed-off-by: kw <kw@kw>
+Reviewed-by: John Doe <johndoe@community.com>
+
+Signed-off-by: kw <kw@kw>"
+
+CORRECT_REVIEWED_LOG="Signed-off-by: kw <kw@kw>
+Reviewed-by: John Doe <johndoe@community.com>
```
rodrigosiqueira, line 13: I think `community.com` is a valid email; just use some random name.

tests/unit/handle_trailer_test.sh
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env bash
+
+include './src/handle_trailer.sh'
+include './tests/unit/utils.sh'
+
+# Correct trailers for --add-reviewed-by and -r
+CORRECT_REVIEWED_COMMIT="Signed-off-by: kw <kw@kw>
+Reviewed-by: John Doe <johndoe@community.com>
+
```
rodrigosiqueira, line 9: Why do we need these extra spaces (for all the options), and why is the signed-off-by duplicate? This does not look correct.

tests/unit/handle_trailer_test.sh
```diff
@@ -0,0 +1,260 @@
...
+Acked-by: Michael Doe <michaeldoe@community.com>
+Reviewed-by: John Doe <johndoe@community.com>
+
+Signed-off-by: kw <kw@kw>"
+
+function setUp()
+{
+  local -r current_path="$PWD"
+
+  cp -r tests/unit/samples/handle_trailer_patch_samples/*.patch "$SHUNIT_TMPDIR"/
```
rodrigosiqueira, line 71: Use full parameter names and add quotes around the path.

tests/unit/handle_trailer_test.sh
```diff
@@ -0,0 +1,260 @@
...
+  # Setup git repository for test
+  mk_fake_git
+
+  # Start repository
+  git config user.name kw
+  git config user.email kw@kw
+  mkdir fs
+  touch fs/some_file
+  git add fs/
+  git commit --quiet -s -m "fs: some_file: Fill file" -m "First"
```
rodrigosiqueira, line 86: https://kworkflow.org/content/codingstyle.html#avoid-short-options

tests/unit/handle_trailer_test.sh
```diff
@@ -0,0 +1,260 @@
...
+
+  # Simulate wrong change
+  echo 'Wrong text' > fs/some_file
```
rodrigosiqueira, line 89: https://kworkflow.org/content/codingstyle.html#use-printf-instead-of-echo

tests/unit/handle_trailer_test.sh
```diff
@@ -0,0 +1,260 @@
...
+  cd "$current_path" || {
+    fail "($LINENO) It was not possible to move back from temp directory"
+    return
+  }
+}
+
+function tearDown()
+{
+  rm -rf "$SHUNIT_TMPDIR"
+  mkdir -p "$SHUNIT_TMPDIR"
```
rodrigosiqueira, line 112: https://kworkflow.org/content/codingstyle.html#avoid-short-options

tests/unit/handle_trailer_test.sh
```diff
@@ -0,0 +1,260 @@
...
+
+# This function tests 'handle_trailer_main' for
+# cases where the given argument is a patch.
+function test_handle_trailer_main_patch()
+{
+  local original_dir="$PWD"
+  cd "$SHUNIT_TMPDIR" || {
+    fail "($LINENO) It was not possible to move to temporary directory"
+    return
+  }
```
rodrigosiqueira, line 123: Why not move this cd to the setup and the cd at the end of this test to the teardown?

Also: https://kworkflow.org/content/codingstyle.html#string-concatenation

tests/unit/handle_trailer_test.sh
```diff
@@ -0,0 +1,260 @@
...
+  git restore patch_model.patch
+
+  handle_trailer_main -r "John Doe <johndoe@community.com>" patch_model.patch
+  assertFileEquals 'patch_model.patch' 'patch_model_reviewed.patch'
+  git restore patch_model.patch
+
+  # Testing --add-acked-by and -a with patch
+  handle_trailer_main --add-acked-by "Michael Doe <michaeldoe@community.com>" patch_model.patch
+  assertFileEquals 'patch_model.patch' 'patch_model_acked.patch'
+  git restore patch_model.patch
```
rodrigosiqueira, line 137: Maybe split this test into small tests where each part test a specif argument, for example, tests for acked-by and others for reviewed-by

tests/unit/samples/handle_trailer_patch_samples/patch_model.patch
```diff
@@ -0,0 +1,33 @@
...
+To: Jane Doe <jane@email.com>,kernel@list.org
+Subject: [PATCH] handle_trailer test
+
+Lorem ipsum dolor sit amet, consectetur adipisci elit,
+sed eiusmod tempor incidunt ut labore et dolore magna aliqua.
+Ut enim ad minim veniam, quis nostrum exercitationem
+ullam corporis suscipit laboriosam, nisi ut aliquid ex ea
+commodi consequatur.
+
+Signed-off-by: kw <kw@example.net>
```
rodrigosiqueira, line 13: example.net sounds like a real email; try to use something more random.

rodrigosiqueira (changes requested):
Hi,
Sorry for the delay tbh; I was thinking a lot about this feature... It is a very good feature, but I'm still not sure if the name is good enough or if we can put it under other feature... Anyway, since I don't have any good options for this, I decided to go on and review it.

Again, thanks a lot for your PR.

documentation/man/features/kw-handle-trailer.rst
```diff
@@ -0,0 +1,86 @@
...
+
+  $ kw handle-trailer -r "Reviewer Name <mail>" -a "Some Name <mail>"
+
+To perform both operations, the user has to call this command 2 times
+using different options.
+
+OPTIONS
+=======
+-r, \--add-reviewed-by:
+  Adds a **Reviewed-by** trailer line to either commits or patch files.
```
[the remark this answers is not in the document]
Lincoln-Yuji: If I understood correctly, we need to change this line (and every other with the same problem I assume) explaining the option more precisely:
"Adds a <tag> trailer line to either patch files, patchsets or series"

Is that it?

kw
```diff
@@ -287,6 +287,13 @@ function kw()
         vm_main "$@"
       )
       ;;
+    handle-trailer)
```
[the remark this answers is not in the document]
Lincoln-Yuji: This name is kind of bad in fact. We couldn't come up with a better name while making this feature.

Not sure if this will be put into another already existing feature in KW or if it's going to be a different one. If the latter, then a better name is welcome.

We are open to suggestions.

src/handle_trailer.sh
```diff
@@ -0,0 +1,169 @@
...
+    # then build the correct command.
+    if [[ $(git cat-file -t "$arg" 2> /dev/null) == 'commit' ]]; then
+      cmd="git commit --quiet --amend --no-edit --trailer \"${TRAILER_TAG} ${TRAILER_VALUE}\""
+      # Only call 'git rebase' if user is trying to handle multiple commits
+      if [[ "$(git rev-parse "$arg")" != "$(git rev-parse HEAD)" ]]; then
+        cmd="git rebase ${arg} --exec '${cmd}' 2> /dev/null"
+      fi
+    elif is_a_patch "$arg"; then
+      cmd="git interpret-trailers ${arg} --in-place --trailer"
+      cmd+=" \"${TRAILER_TAG} ${TRAILER_VALUE}\""
```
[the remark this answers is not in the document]
Lincoln-Yuji: I'm afraid this can't be done, at least we don't know how to build this command correctly without `\"`. The problem with using single-quotes is that we have multiple nested strings. Replacing the `\"` for `'` messes up the cmd's building process and generates commands that are not correctly executed.

We've tried to think of ways to build the command without it, but it messes up the format.

src/handle_trailer.sh
```diff
@@ -0,0 +1,169 @@
...
+    complain 'An option is required to use this command.'
+    handle_trailer_help
+    return 22 # EINVAL
+  fi
+
+  for arg in "${PATCH_OR_SHA[@]}"; do
+    # Check if given argument is either a patch or valid commit reference,
+    # then build the correct command.
+    if [[ $(git cat-file -t "$arg" 2> /dev/null) == 'commit' ]]; then
+      cmd="git commit --quiet --amend --no-edit --trailer \"${TRAILER_TAG} ${TRAILER_VALUE}\""
```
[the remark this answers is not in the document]
Lincoln-Yuji: I'm afraid this can't be done, at least we don't know how to build this command correctly without `\"`. The problem with using single-quotes is that we have multiple nested strings. Replacing the `\"` for `'` messes up the cmd's building process and generates commands that are not correctly executed.

We've tried to think of ways to build the command without it, but it messes up the format.

src/handle_trailer.sh
```diff
@@ -0,0 +1,169 @@
...
+      # Only call 'git rebase' if user is trying to handle multiple commits
+      if [[ "$(git rev-parse "$arg")" != "$(git rev-parse HEAD)" ]]; then
+        cmd="git rebase ${arg} --exec '${cmd}' 2> /dev/null"
```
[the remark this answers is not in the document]
Lincoln-Yuji: Using `--quiet` does not seem to suppress the following message from `git rebase --exec`:

```
Executing: ...
Executing: git commit --quiet --amend --no-edit --trailer "Name <mail>"
Executing: ...
```

For each commit, the rebase executes the above command and prints that message. This can be visualized when executing the tests. However, redirecting with `2> /dev/null` suppresses this print. Using `git rebase --quiet --exec <cmd>` doesn't seem to solve the issue. There might be a better to solve this though?

documentation/man/features/kw-handle-trailer.rst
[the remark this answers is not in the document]
Lincoln-Yuji: We've talked to David Tadokoro. He explained that this means 'substitute patch file for either patchset or series'.

Similar to substitute format using `sed`. Our doubts were clarified

documentation/man/features/kw-handle-trailer.rst
```diff
@@ -0,0 +1,86 @@
...
+previous commit ``90a7ba23340d``::
+
+  $ kw handle-trailer --add-fixes 90a7ba23340d
+
+All the above options can be used to perform operations over patch files
+as well::
+
+  $ kw handle-trailer -r "Reviewer Name <reviewer@mail.org>" example.patch
+
+  $ kw handle-trailer -a "Some Name <example@mail.com>" example.patch
```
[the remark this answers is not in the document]
Lincoln-Yuji: That's a really cool default behavior for this flags. We recently made a force pull with this request in mind and improved the tests to also take this case in consideration.

tests/unit/handle_trailer_test.sh
Lincoln-Yuji: I believe the confusion here happened because of our poor choice of names for the variables and lack of better comments to explain what these variables are holding. Those variables are holding the trailer lines of multiple commits. For instance, when using the command `git local --max-count 2 --format="%(trailers)"` the output follows this format:

```
Signed-off-by: Name <mail>                              <- Most recent commit
Reviewed-by: Another Name <another mail>
                                                        <- Empty line splitting trailers from different commits
Signed-off-by: Name <mail>                              <- Second most recent commit trailers
Reviewed-by: Some Name <some mail>
```

This is why the `Signed-off` appears duplicated in multiple outputs and there are these extra spaces in all the options. We hope this clarified the issue. We've added some comments and changed the name of some of these variables to make it easier to understand.

Also please tell us if we misunderstood the problem here. Perhaps we could be missing something important

Lincoln-Yuji:
Hello @rodrigosiqueira

We've implemented your suggestions and now we will be waiting for more feedbacks. I noticed you mentioned something about adding `Signed-off-by` as well. Do you think adding a new option such as `--add-signed-off-by | -s` is worth it?

Also, there were somethings related to code style that we didn't manage to change like the `\"` usage and the `2> /dev/null` redirection because it seems that changing those is harder than we expected or maybe we are missing something. They are explained above in this thread.

tests/unit/handle_trailer_test.sh
Lincoln-Yuji: I believe the confusion here happened because of our poor choice of names for the variables and lack of better comments to explain what these variables are holding. Those variables are holding the trailer lines of multiple commits. For instance, when using the command `git log --max-count 2 --format="%(trailers)"` the output follows this format:

```
Signed-off-by: Name <mail>                  <- Most recent commit
Reviewed-by: Another Name <another mail>
                                            <- Empty line splitting trailers from different commits
Signed-off-by: Name <mail>                  <- Second most recent commit trailers
Reviewed-by: Some Name <some mail>
```

This is why `Signed-off` appears duplicated in multiple outputs and there are these extra spaces in all the options. We hope this clarified the issue. We've added some comments and changed the name of some of these variables to make it easier to understand.

Also please tell us if we misunderstood the problem here. Perhaps we could be missing something important

Lincoln-Yuji:
I've asked some suggestions for better names from some friends and ideas they told me were:

- `kw annotate`
- `kw amend`
- `kw tag`
- `kw signature`

Does any of these names sound better or make more sense? I'm also studying the possibility to inject these options into an already implemented feature, but not sure which one and how.

Lincoln-Yuji:
Hello @rodrigosiqueira and @davidbtadokoro. I just want to tell you that I will be marking this PR as Draft for now. There are some minor issues I've noted with the code and other additional changes. Also, I'll be renaming this feature to either `annotate` or `signature`, not sure which one yet. I'll try to transition this PR back to Ready to review within the next 2 weeks.

I have intentions to add new options as well:
- `--add-signed-off-by | -s`
- `--add-reported-by | -R`
- `--add-suggested-by | -S`

These long options sound, you know... too long? Don't know if the `add` prefix or the `by` suffix are needed. I'm keeping them
because it sort of acts as documentation. Plus, more experienced users can always use the short options anyway.

Finally, neither Luiza nor Sabrina will keep working on this PR. I will be working on it alone from now, so it might have longer delays to deliver future requested changes. 

If there's any new additional reviews for the current code, you can write them here and I'll work on them as soon as possible.

Lincoln-Yuji:
I'm re-opening this PR. Fortunately I had some spare time today to finish making the changes I had in mind.

I believe I managed to fix the issues I found. Also, I added the new option `--add-signed-off-by | -s` I mentioned previously, but I didn't implement any of the other two since they seem to be less necessary to make this feature minimally useful and I'm still not sure if this code will require heavy further changes/refactoring.

I also changed the name to `signature`. Hope this name makes more sense. Waiting for new feedbacks

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,94 @@
...
+===================================================
+
+.. _signature-doc:
+
+SYNOPSIS
+========
+| *kw* *signature* (-s(<string>) | \--add-signed-off-by=(<string>)) [\--verbose] [<sha> | <patch>]
+| *kw* *signature* (-r(<string>) | \--add-reviewed-by=(<string>)) [\--verbose] [<sha> | <patch>]
+| *kw* *signature* (-a(<string>) | \--add-acked-by=(<string>)) [\--verbose] [<sha> | <patch>]
+| *kw* *signature* (-f(<sha>) | \--add-fixes=(<sha>)) [\--verbose] [<sha> | <patch>]
```
rodrigosiqueira, line 12: Can we also have:
* Tested-by
* Reported-by
* Co-developed-by

I think with those extra tags, we covered everything :)

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,94 @@
+===================================================
+kw-signature - Write commit and patch trailer lines
```
rodrigosiqueira, line 2: I was thinking about the name, and at the end of the day, you are just adding kernel tags to the commit. We can try to find a name around it. Some brainstorm ideas:
* add-commit-tags
* add-tag
* add-signature-tag
* Link -> This is a special case requiring a link, not an email.

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,94 @@
...
+| *kw* *signature* (-f(<sha>) | \--add-fixes=(<sha>)) [\--verbose] [<sha> | <patch>]
+
+DESCRIPTION
+===========
+This is a wrapper to some useful usage of the 'git commit --amend',
+'git rebase' and the 'git interpret-trailers' commands. This kw command
+is able to perform usual operations over trailer lines of commits and patches.
+By default, it uses the commit currently pointed by ``HEAD`` as the target
+of the operation if user does not specify one. Also, no operation is used by
+default. At least one operation option must be given.
```
rodrigosiqueira, line 21: Just add: At least one operation ...

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,94 @@
...
+
+This command does not allow repetition of options. For instance, the
+following command will give an error::
+
+  $ kw signature -r"Reviewer Name <mail>" -r"Some Name <mail>"
```
rodrigosiqueira, line 26: What if I want to add Reviewed-by from different developers?
For example, I got a Reviewed-by from Joe and another one from Bob, but for some reason, I have to do it manually; why not allow:

```bash
kw signature -r "Joe Doe <somethingxpto@lala.com" -r "Bob Lala <abcd@1234s.com"
```

This scenario looks valid to me. I think you can check for name/email duplication and avoid this case:

```bash
kw signature -r "Joe Doe <somethingxpto@lala.com"  -r "Joe Doe <somethingxpto@lala.com"
```

You can just add a warning in the above case.

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,94 @@
...
+
+However, using different options combined is possible. For instance,
+the following command will add **Reviewed-by** and **Signed-off-by**, in
+this order, to the commit's trailers pointed by ``HEAD`` using the
+user's git config::
+
+  $ kw signature -r -s
```
rodrigosiqueira, line 33: In this case, where will it get the name and email information? Git config?

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Changing the order of the options affects the order the lines are written.
```
rodrigosiqueira, line 35: Btw, google for this kernel page: Documentation/process/submitting-patches.rst. We can enforce the order suggested by the documentation.

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,94 @@
...
+
+OPTIONS
+=======
+-s, \--add-signed-off-by:
+  Adds a **Signed-off-by** trailer line to either commits or patchsets.
+  By default it uses `user.name` and `user.email` from git config to
+  write these lines if no argument is given to this option.
```
rodrigosiqueira, line 42: Add this information to the command description.

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,94 @@
...
+
+EXAMPLES
+========
+Adding a **Reviewed-by** line to commit pointed by ``HEAD``::
+
+  $ kw signature --add-reviewed-by='Reviewer Name <reviewer@mail.org>'
+
+Adding a **Acked-by** line to last 3 commits starting from ``HEAD``::
+
+  $ kw signature --acked-by='Some Name <example@mail.com>' HEAD~3
```
rodrigosiqueira, line 70: Does this work if I use a SHA commit? For example, in my daily life as a kernel developer, I usually want to add my Signed-off-by from the SHA until the HEAD.

rodrigosiqueira, line 70: Also, do the tag options support multiple emails? For example:

```bash
 kw signature --add-reviewed-by='Reviewer Name <reviewer@mail.org>, Another Name <anothername@ljkjsdfal.com>'
```

That results in:

```
Reviewed-by: Reviewer Name <reviewer@mail.org>
Reviewed-by: Another Name <anothername@ljkjsdfal.com>
```

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Adding **Fixes** line to commit pointed by ``HEAD``, which fixes another
+previous commit ``90a7ba23340d``::
+
+  $ kw signature --add-fixes=90a7ba23340d
```
rodrigosiqueira, line 75: Do you know if this command checks if the commit is valid?

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,94 @@
...
+
+All the above options can be used to perform operations over patches as well::
```
rodrigosiqueira, line 77: Maybe use `.patch` files.

src/_kw
```diff
@@ -486,6 +487,17 @@ _kw_self-update()
...
 
+_kw_signature()
+{
+  _arguments : \
+    '(--verbose)--verbose[enable verbose mode]' \
+    '(-s --add-signed-off-by)'{-s-,--add-signed-off-by=-}'[add Signed-off-by trailer line]' \
+    '(-r --add-reviewed-by)'{-r-,--add-reviewed-by=-}'[add Reviewed-by trailer line]' \
+    '(-a --acked-by)'{-a-,--add-acked-by=-}'[add Acked-by trailer line]' \
+    '(-f --add-fixes)'{-f-,--add-fixes=-}'[add Fixes trailer line]' \
+    '1: :_files'
```
rodrigosiqueira, line 498: @davidbtadokoro could you review this part?

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+
+declare -gA options_values
+
+# Define trailer tags
+declare -gA TRAILER_TAGS
+
+TRAILER_TAGS['SIGNOFF']='Signed-off-by:'
+TRAILER_TAGS['REVIEW']='Reviewed-by:'
+TRAILER_TAGS['ACKED']='Acked-by:'
+TRAILER_TAGS['FIXES']='Fixes:'
```
rodrigosiqueira, line 12: For the keys, let's use the full name, e.g., SIGNED_OFF_BY instead of SINGOF.

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+# the user this argument was ignored.
+#
+# Also, if no operation option is given, then an error message
+# followed by a helper message is printed to the user.
+#
+# @patch_or_sha_args Holds either patch paths or commit references.
+function signature_main()
+{
+  local patch_or_sha_args
+
```
rodrigosiqueira, line 31: Remove this extra new line.

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+  acked_line=${options_values['ACKED_LINE']}
+  fixes_line=${options_values['FIXES_LINE']}
+
+  if [[ "${#trailer_operations[@]}" -eq 0 ]]; then
+    complain 'An option is required to use this command.'
+    signature_help
+    return 22 # EINVAL
+  fi
+
+  # Writting the trailers respecting the defined order
```
rodrigosiqueira, line 65: Consider moving this code to its own function.

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+    done
+  done
+}
+
+# This function gets raw data and based on that fill out the options values to
+# be used in another function.
+#
+# Return:
+# In case of successful return 0, otherwise, return 22.
+function parse_signature_options()
```
rodrigosiqueira, line 91: move this function and the help one to the end of this file.

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+}
+
+# This function registers a signature operation into
+# trailer_operations that will be properly excuted later.
+# It attempts to use the user's name and email configured
+# if a signature is not passed as argument and gives
+# an error if they are not properly set with git config.
+#
+# @keyword Holds a string that is a keyword for a specific operation.
+# @signature Holds a string like 'NAME <EMAIL>' defining the signature.
```
rodrigosiqueira, line 244: Don't forget to document the return option for this function.

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+function register_signature_operation()
+{
+  local keyword="$1"
+  local signature="$2"
+
```
rodrigosiqueira, line 249: Remove this extra new line.

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+# This function registers a signature operation into
+# trailer_operations that will be properly excuted later.
+# It attempts to use the user's name and email configured
+# if a signature is not passed as argument and gives
+# an error if they are not properly set with git config.
+#
+# @keyword Holds a string that is a keyword for a specific operation.
+# @signature Holds a string like 'NAME <EMAIL>' defining the signature.
+function register_signature_operation()
```
rodrigosiqueira, line 245: I don't think this is a good name. Basically, you are formatting the input, so maybe you can use a name that better describes this.

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+
+  # Use default from git config if no argument was given
+  if [[ ! "$signature" ]]; then
+    formated_output="$(format_name_email_from_user)"
+    if [[ "$?" -gt 0 ]]; then
+      return 1
+    fi
+    options_values["$signature_key"]+="$formated_output"
+  else
+    options_values["$signature_key"]+="$(str_strip "$signature")"
```
rodrigosiqueira, line 265: You can't trust the input; you should have some level of validation, such as checking for the string size. You can also check if the email has some minimum stuff such as one @ and at least one .; I don't think you need to implement this, iirc, we already have an email validation code.

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+  fi
+
+  trailer_operations+=("$operation_key")
+}
+
+# This function receives a commit SHA and verifies if it's
+# a valid commit reference. If it is, then it outputs the
+# appropriate formatted message. Else it returns an error code.
+#
+# @sha Holds either a commit hash or pointer
```
rodrigosiqueira, line 275: Again, don't forget to document the return (review all comments in this PR).

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+function format_fixes_message()
+{
+  local sha="$1"
+  local formatted_message
+
+  # Check if given value is a valid commit reference
```
rodrigosiqueira, line 281: Nice :) This answers my previous question.

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+  if [[ $(git cat-file -t "$sha" 2> /dev/null) != 'commit' ]]; then
```
rodrigosiqueira, line 282: btw, it would be awesome if you could work on the below issue since you wrote a lot of git code:

https://github.com/kworkflow/kworkflow/issues/1123

src/signature.sh
```diff
@@ -0,0 +1,348 @@
...
+# If either the name or email are not configured then this
+# function will return an error code. Otherwise it will output
+# the formatted name and email.
+function format_name_email_from_user()
+{
+  local user_name
+  local user_email
+
+  user_name="$(git config user.name)"
+  user_email="$(git config user.email)"
```
rodrigosiqueira, line 304: @matheustavares do you think we have a better way to get this information?

rodrigosiqueira (changes requested):
Hi @Lincoln-Yuji ,
First of all, sorry for my huge delay in responding to this change. However, I'm glad that you insisted on it because this feature has great potential and is awesome!

I added many comments and suggestions on how we can proceed with this feature. I did not review the unit test because I think there are already many comments.

Btw, would you consider working on this issue https://github.com/kworkflow/kworkflow/issues/1123 ?

src/signature.sh
matheustavares: No, I think that's the intended way for scripts too :)

Lincoln-Yuji:
Thank you @rodrigosiqueira for the review! I'll take some time to address all these changes. I'll try getting it done within 2 weeks.

I'll keep the name `signature` for now since it sounds low priority compared to everything else. But I'll be revisiting this topic in the future of course.

Plus, I'm not sure if I will have the time to work on issue #1123 right now, but I can take a look at it no problem. Since it's a new library, I believe documenting, testing and effectively applying it into multiple places in the project might be very demanding. I can try opening a new PR in the future to address this issue, but I'll be focusing on this PR for now and then give this **git lib** a shot. Sounds like a natural development, unless this PR can not be accepted without such library already implemented.

documentation/man/features/kw-signature.rst
Lincoln-Yuji: I've updated the code to allow repetition of options. Though I'm not checking name/email duplication yet.
By default the following command:

```sh
kw signature -r "Joe Doe <somethingxpto@lala.com>"  -r "Joe Doe <somethingxpto@lala.com>"
```

It only writes `Reviewed-by: Joe Doe <somethingxpto@lala.com>` once because it seems that both `git amend --trailer` and `git-interpret-trailers` will not write a duplicate line if the last trailer line is equals to the line it's trying to write right now.

Though it's not hard to come up with ways the user can still somehow force a duplicate line. For example:

```sh
kw signature -r -s -r -s
```

The above command will write:

```
Reviewed-by: Name <Mail>
Signed-off-by: Name <Mail>
Reviewed-by: Name <Mail>
Signed-off-by: Name <Mail>
```

I'm still thinking of ways to solve this issue. Maybe I'll have to, in fact, check all names/emails from each option and check if there are repeating options with the same signature.

documentation/man/features/kw-signature.rst
Lincoln-Yuji: I believe what you described is the actual behavior of this feature, though this was not properly documented in the **DESCRIPTION**. I've added an explanation regarding this now in the man page. If there's anything else weirdly presented presented in the documentation please let me know.

Lincoln-Yuji: Since the command now accepts repetition of options, I'm not sure if supporting multiple signatures in the same argument is required. Also, with the current implementation this will not work beacause the program will accuse `Reviewer Name <reviewer@mail.org>, Another Name <anothername@ljkjsdfal.com>` as an invalid signature format.

However, if we define a **separator**, we can split this argument into multiple signatures and then validate each one.

Supporting multiple emails in the same argument is desirable?

rodrigosiqueira:
> Thank you @rodrigosiqueira for the review! I'll take some time to address all these changes. I'll try getting it done within 2 weeks.
> 
> I'll keep the name `signature` for now since it sounds low priority compared to everything else. But I'll be revisiting this topic in the future of course.
> 
> Plus, I'm not sure if I will have the time to work on issue #1123 right now, but I can take a look at it no problem. Since it's a new library, I believe documenting, testing and effectively applying it into multiple places in the project might be very demanding. I can try opening a new PR in the future to address this issue, but I'll be focusing on this PR for now and then give this **git lib** a shot. Sounds like a natural development, unless this PR can not be accepted without such library already implemented.

Don't worry, it was just a suggestion. If you can finish this feature, it is more than enough.

documentation/man/features/kw-signature.rst
rodrigosiqueira: > I've updated the code to allow repetition of options. Though I'm not checking name/email duplication yet. By default the following command:
> 
> ```shell
> kw signature -r "Joe Doe <somethingxpto@lala.com>"  -r "Joe Doe <somethingxpto@lala.com>"
> ```

I suppose the second Joe was a duplication, right? I'll assume you meant Bob.

> 

Now that I'm rereading my comment and checking your reply, I think that a better interface would be:

```sh
kw signature -r "Joe Doe <somethingxpto@lala.com>, Bob Lala <abcd@1234s.com"
```

In other words, just use commas to separate different names.

> It only writes `Reviewed-by: Joe Doe <somethingxpto@lala.com>` once because it seems that both `git amend --trailer` and `git-interpret-trailers` will not write a duplicate line if the last trailer line is equals to the line it's trying to write right now.
> 

I guess this is what we want. However, you can check this yourself in the code and send a warning to the user.

> Though it's not hard to come up with ways the user can still somehow force a duplicate line. For example:
> 
> ```shell
> kw signature -r -s -r -s
> ```
> 

I'm sorry. I guess I was not clear in my original comments. In summary, we don't want duplication of tags with the same name and email. If the tags are different, it is okay to have them, but if they are equal, we don't want them.

> The above command will write:
> 
> ```
> Reviewed-by: Name <Mail>
> Signed-off-by: Name <Mail>
> Reviewed-by: Name <Mail>
> Signed-off-by: Name <Mail>
> ```

From your example, we don't want the duplicated Reviewed-by and Signed-off-by, we just want:

```
Reviewed-by: Name <Mail>
Signed-off-by: Name <Mail>
```

documentation/man/features/kw-signature.rst
rodrigosiqueira: > I'm very confused about how exactly such order should be enforced. The documentation states that the **Signed-off-by** chain should represent chronologically the real route a patch took as it was propagated. Enforcing an order does not sound feasible to me because it would require the feature to predict the route and interactions between submitters and maintainers.
> 

We don't need to be perfect - maybe just close enough. With this idea in mind, we can try to create a heuristic for dealing with the sequence. For example, you can assume that the dev using the `kw signature` at the moment is the last person in the SoB chain.

> However, there're some details I've noticed while reading this document. It says that every `Co-developed-by` tag indicates authorship, which means that this tag must be immediately followed by a `Signed-off-by` of the same co-author. Since this a general rule, is it ok to implement it such that the option `--add-co-developed-by | -C` also adds an immediate `Signed-off-by` tag? Like:
> 
> ```shell
> $ kw signature -C'Developer Name <developer@mail.xyz>'
> # Written tags
> # Co-developed-by: Developer Name <developer@mail.xyz>
> # Signed-off-by: Developer Name <developer@mail.xyz>
> ```
> 
> Or would using `kw signature -C'blabla' -s'blabla'` still be preferable? 

The option added in your first example is the best way to handle the Co-developed-by tag. However, you might use the second example that you shared, and I think it should not be a problem as well. You just need to handle the duplicated Signed-off-by. 

>Also, it seems that the `Reported-by` flag should be followed by a `Closes` or `Link` tag if the bug report is available in the web. Should this somehow be passed via some argument in the command line?
> 

Keep in mind that in kw, we want to enforce the kernel way of doing things as much as possible. In this sense, we want to incentivize the users to follow the rules, which means we want them to use the Link/Close; however, from my experience, it is common not to have such a link. In this case, we need to define a heuristic. As a first suggestion, we can have something like this:

1. Reported-by might be used with a different (and optional) syntax. Something like this: `kw signature --reported-by "Name <email>;[link|close]=LINK"[,Name <email>...]`.
2. If the user does not provide a link: `kw signature --reported-by "Name <email>"` you can prompt a question asking if they have a link and give the option to insert it in case they forget. If they don't have a link, it should be ok. We probably want to have a force option to avoid asking.

> Sorry if I'm too lost with this. I've been thinking of how to enforce an order according to the documentation but I can't come up with a good idea right now.

documentation/man/features/kw-signature.rst
rodrigosiqueira: yeah, and this behavior looks similar to what we have when using git-send-email.

src/signature.sh
rodrigosiqueira: We can take this route but add a TODO comment to refactor this later. Ideally, we want to extract that function from a generic place like `kw_string.sh`.

src/signature.sh
```diff
@@ -0,0 +1,402 @@
...
+  local cmd
+
+  for trailer in "${all_trailers[@]}"; do
+    # Check if given argument is either a patch or valid commit reference,
+    # then build the correct command.
+    if [[ "$(git cat-file -t "$patch_or_sha" 2> /dev/null)" == 'commit' ]]; then
+      cmd="git commit --quiet --amend --no-edit --trailer \"${trailer}\""
+      # Only call 'git rebase' if user is trying to write multiple commits
+      if [[ "$(git rev-parse "$patch_or_sha")" != "$(git rev-parse HEAD)" ]]; then
+        cmd="git rebase ${patch_or_sha} --exec '${cmd}' 2> /dev/null"
```
rodrigosiqueira, line 196: Instead of using /dev/null, can we just use `--quiet`.

src/signature.sh
```diff
@@ -0,0 +1,402 @@
...
+      fi
+    elif is_a_patch "$patch_or_sha"; then
+      cmd="git interpret-trailers ${patch_or_sha} --in-place --trailer \"${trailer}\""
```
rodrigosiqueira, line 199: I didn't know this command before, but I just realized that this is a powerful tool. I thought we could use it with the `--where` parameter to define some heuristics. For example, I really want that at least one Signed-off-by be at the end when we use tags like Tested-by, Reviewed-by, etc. This week, when I used your feature to add the tag `tested-by`, it was added at the end; if it had been added before the last Signed-off-by, it would have been perfect for my use case.

src/signature.sh
```diff
@@ -0,0 +1,402 @@
...
+
+  while [[ "$#" -gt 0 ]]; do
+    case "$1" in
+      --add-signed-off-by | -s)
+        parse_and_add_trailer 'SIGNED_OFF_BY' "$2"
+        return_status="$?"
+
+        if [[ "$return_status" -eq 1 ]]; then
+          options_values['ERROR']='You must configure your user.name and user.email with git '
+          options_values['ERROR']+='to use --add-signed-off-by or -s without an argument'
```
rodrigosiqueira, line 245: Just use `--add-singed-off-by|-s`.

src/signature.sh
```diff
@@ -0,0 +1,402 @@
...
+  local user_name
+  local user_email
+
+  user_name="$(git config user.name)"
+  user_email="$(git config user.email)"
+
+  # If user doesn't have either a name or email configured with
+  # git then they must provide an argument
+  if [[ -z "$user_name" || -z "$user_email" ]]; then
+    return 1
```
rodrigosiqueira, line 170: How about return 61 (ENODATA - No data available), instead of 1.

src/signature.sh
```diff
@@ -0,0 +1,402 @@
...
+  # Use default from git config if no argument was given
+  if [[ ! "$signature" ]]; then
+    formated_output="$(format_name_email_from_user)"
+    if [[ "$?" -gt 0 ]]; then
+      return 1
+    fi
+    trailer+="$formated_output"
+  else
+    signature="$(str_strip "$signature")"
+    if [[ "$keyword" != 'FIXES' ]] && ! is_valid_signature "$signature"; then
```
rodrigosiqueira, line 115: Wait... why is FIXES a factor here but not in the above ones? I did not get it.

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,112 @@
...
+
+  $ kw signature --add-fixes='90a7ba23340d'
+
+All the above options can be used to perform operations over `.patch` files as well::
+
+  $ kw signature -r'Reviewer Name <reviewer@mail.org>' example.patch
+
+  $ kw signature -a'Some Name <example@mail.org>' example.patch
+
+  $ kw signature -f'90a7ba23340d' example.patch
```
rodrigosiqueira, line 101: btw, when I tried -f in a specific branch, I got this error:
Screenshot from 2024-07-27 10-54-58

rodrigosiqueira (changes requested):

Lincoln-Yuji:
Hello @rodrigosiqueira, thank you for your new feedbacks. I have been very busy recently so I'm sorry for the long wait.
I too have noticed that the biggest issue so far is lack of tests for git branch setups or bad inputs, since I also had some of these issues while making further tests. I'll try making this feature more robust to avoid such annoying problems.

documentation/man/features/kw-signature.rst
Lincoln-Yuji: I don't know how to reproduce this error, but if no warning nor errors were reported by the feature itself before reaching this sequence of git errors then I'm almost sure the command assembled in `write_all_trailers` is getting messed up somehow.

Does using `--verbose` prints anything unusual/weird? Or nothing happens at all? By looking at the command used it should print something like:
```bash
git commit --quiet --amend --no-edit --trailer "Fixes: 51fc8079c75e (\"<message>\")"
```

src/signature.sh
Lincoln-Yuji: The FIXES factor is not necessary above because the `--add-fixes` options demands an argument, which means that it's not possible (or at least it shouldn't) for the user to call this option with empty content. When I wrote this code I assumed that if `"$signature"` is not given then the caller is not trying to parse a `Fixes` tag line, unless the developer working on this feature (in that case, me) messed up and forgot to pass the second parameter.

In this line we need the FIXES factor because this tag's content follows a different format and it was already validated and parsed previously in `format_fixes_message`.

I'm open to suggestions on how to improve this code's flow though. I agree it might be confusing.

src/signature.sh
```diff
@@ -0,0 +1,402 @@
...
+# This function gets raw data and based on that fill out the options values to
+# be used in another function.
+#
+# Return:
+# In case of successful return 0, otherwise, return 22.
+function parse_signature_options()
+{
+  local long_options='add-signed-off-by::,add-reviewed-by::,add-acked-by::,add-fixes::'
+  long_options+=',add-tested-by::,add-reported-by::,add-co-developed-by::,verbose'
+  local short_options='s::,r::,a::,f::,t::,R::,C::'
```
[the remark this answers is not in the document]
Lincoln-Yuji: That actually happened to me multiple times when I was testing it myself. I had to use `git reset --hard <SHA>` to backup every time that happened. I've been thinking of how to properly handle this, but didn't have too much free time recently to work on it. But I'm going to try come up with ideas to improve robustness and avoid this really annoying issue.

src/signature.sh
Lincoln-Yuji: I've tried this before, but the `git rebase` command prints multiple lines like:
```
Executing: git commit --quiet --amend --no-edit --trailer "Tag: Joe Doe <mail>"
Executing: git commit --quiet --amend --no-edit --trailer "Tag: Bob Doe <mail>"
Executing: git commit --quiet --amend --no-edit --trailer "Tag: Michael Doe <mail>"
```
Even when we use `--quiet`, but redirecting `2` to `/dev/null` solves it. I don't know why it behaves like this to be honest and I didn't take the time to investigate it properly. I'll be coming back and taking a look at this in the future, since this is something that has been bothering me as well since I wrote this code.

src/signature.sh
Lincoln-Yuji: The `--where` option completely went under my radar. Looks like a nice tool to use. Unfortunately this seems to be limited to `git interpret-trailers` while `git commit --amend` can't benefit from it. However, I've been working on some heuristics to sort the tags **before** writing them using this command, but as you mentioned all new trailers are written at the end by default and only `interpret-trailers` can change this behavior. I'll be revisiting this topic later because this is a very powerful feature in fact.

I've looked up some solutions to work with commits but all of them comes down to using `sed` in very obscure ways, directly editing the commit messages automatically with `rebase` and `amend`. But I'm too afraid to try this, because I have no idea if I'll be able to come up with a good way to implement that, and also it sounds extremely dangerous.

documentation/man/features/kw-signature.rst
Lincoln-Yuji: Ohh ok, I think I understand that now. Each error is associated to a token in the commit message. The message is being expanded by the shell and causing that to happen when `cmd_manager` evals the command. Can you please give the message that is forcing those errors. I believe I overlooked some important details while making this code...

src/signature.sh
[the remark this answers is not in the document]
Lincoln-Yuji: The solution I'll be using for now is giving an error to the user and stop the feature from running any further if an invalid commit or patch is detected within the arguments. Not sure if it's the best solution, but at least it will help us to avoid this catastrophic scenario where many undesirable changes are made and the user might not have a proper way to back it all up.

src/signature.sh
Lincoln-Yuji: This can be tested running the following command inside a git project:
```bash
git rebase HEAD~3 --quiet --exec "printf 'blabla' > /dev/null" 
```
You will see that even with `--quiet`, the command still outputs the following:
```text
Executing: printf 'blabla' > /dev/null
Executing: printf 'blabla' > /dev/null
Executing: printf 'blabla' > /dev/null
```

Lincoln-Yuji:
I'll try making a mini changelog for some changes so far:
- Moved email validation functions from `send_patch` to `kwlib` so they can be used by other features.
- Support for more kernel tags: `Tested-by`, `Co-developed-by` and `Reported-by`.
- New tags are sorted before getting added to the commit or patch (some details already included in the man page). However, they are all added at the end of old tags already present. Maybe we can use `--where` for **patchsets** specifically since `interpret-trailers` offers this very nice tool.
- If an argument is neither a valid commit nor a patch, then the feature stops running and prompts an error, avoiding the annoying and dangerous scenario when we might forget to use `=` delimiter for long options or to put the argument glued to the short options (e.g `--add-fixes='123abc123'` and `-f'123abc123'`).

Lincoln-Yuji:
Sorry for the additional force-pushes. Some unit tests were failing in the CI but locally they were all passing. So I did this to test if something on my branches were possibly out of date even after synchronizing my repo branches with upstream.

documentation/man/features/kw-signature.rst
rodrigosiqueira: Ok, it is time for us to end this PR. In this sense, let's rename this feature to `kernel-tags`.

documentation/man/features/kw-signature.rst
```diff
@@ -0,0 +1,172 @@
...
+   Tested-by
+   Reviewed-by
+   Signed-off-by
+   Fixes
+
+At the same time, new lines are grouped by their signatures. For example:
+
+.. code-block:: text
+
+   $ kw signature -s'Joe Doe <joe@mail.xyz>' -s -t -t'Joe Doe <joe@mail.xyz>'
```
rodrigosiqueira, line 58: Maybe move -s and -t to the beginning, it is weird to see it in the middle. Or this was intentional?

documentation/man/features/kw-signature.rst
rodrigosiqueira: I used this repository:
https://gitlab.freedesktop.org/agd5f/linux
The commit SHA is: 51fc8079c75ea89500a2953d02dcd304719f6929

The commit message is:
```
commit 51fc8079c75ea89500a2953d02dcd304719f6929
Author: Asad Kamal <asad.kamal@amd.com>
Date:   Mon Jan 8 18:34:55 2024 +0800

    Revert "drm/amdgpu: Update irq disable flow during unload"
    
    Revert the commit as the original workaround:
    commit f5c7e7797060 ("drm/amdgpu: Adjust removal control
    flow for smu v13_0_2") that caused the issue needs re-evaluation
    
    This reverts commit 2e8e792e6a51e8cb7f5f96148146b6525dbb9cef.
    
    Signed-off-by: Asad Kamal <asad.kamal@amd.com>
    Reviewed-by: Lijo Lazar <lijo.lazar@amd.com>
```

kw
```diff
@@ -286,6 +286,13 @@ function kw()
         vm_main "$@"
       )
       ;;
+    signature)
```
rodrigosiqueira, line 289: As I commented in the documentation part, let's rename this feature now.

src/signature.sh
rodrigosiqueira: Interesting. In that case, we can live with that. Could you add a tiny comment about that to prevent someone from changing it in the future?

rodrigosiqueira (changes requested):
Hi @Lincoln-Yuji,

Sorry for the delay again. I just added a few more comments, but aside from that, let's start to wrap up this feature; we can fix other issues later.

rodrigosiqueira:
Hi @Lincoln-Yuji , why did you close this PR? Is it because you don't want to work on it anymore?

Issue #657, opened 2022-08-06
Mail: Create group interface


rodrigosiqueira:
**Description:**

When dealing with emails, it is common to have a group of people that we always want to add to our patches. For example, you might want to add people from your work in a public discussion even though the get_maintainers does not include them. For these reasons, supporting groups under kw mail would be desirable.

**Why:**

With the group feature, we want to add more flexibility for developers to always include specific people to become part of the review.

**Targeted users:**

* Developers that constantly send patches to the public mailing list.

**API draft:**

For this feature, we need to manage:
* List groups
* Create groups
* Remove groups
* Rename groups
* Add a person to a group
* Remove person from a group

As a first API draft, I suggest:

* List groups:
```bash
$kw mail --groups
Work:
  Joe Doe <Joe@la.uu>
  Fulano <Fulano@la.uu>

Abc:
 ...
```

* Create a group:
```bash
$kw mail --groups --create Work
Group Work created, but it is empty. Append new people to this group
```
Or
```bash
$kw mail --groups --create Work --append "Joe Doe <Joe@la.uu>, Fulano <Fulano@la.uu>"
```

* Remove a group:
```bash
$kw mail --group --remove Work
```

* Rename group:
```bash
$kw mail --group --rename Work AnotherName
```
-> We might want to check if the "AnotherName" is already create and if "Work" exists.

* Append people to the group:
```bash
$kw mail --group Work --append "New Person <New.Person@la.uu>"
```
-> We might want to check if the target name is already registered

* Remove someone from the group
```bash
$kw mail --group Work --remove-email "New.Person@la.uu"
Are you sure you want to remove "New.Person"?
```

SharmaDhruv2511:
Do I have to write exact same as you mention??

JGBSouza:
As discussed in a previous meeting. We can move this new implementation to a new kw_manager.sh file and adapt the commands as follows:

`kw manager --create-group <name> or -c <name>`
`kw manager --rename-group '<old_name>:<new_name>' or -r '<old_name>:<new_name>'`
`kw manager --remove-group'<name> or -rm '<name>'`
`kw manager --group-append <group_name:ctt1_name <ctt1_email>, ctt2_name <ctt2_email>, ...>`
`kw manager --remove-email <group_name:ctt1_email, ctt2_email, ...>'`

[...]

Issue #550, opened 2022-01-21
Add contacts and group support to kw mail


kwy95:
**Description:**

We could have some sort of email address storing and grouping capabilities to `kw mail`. This would allow the user to send patches to preset groups of people.

**Type:**

Usability.

**Why:**

This allows users greater flexibility with regards to whom they are sending their patches to.

PR #1122, opened 2024-05-31
src: mail: kw_mail_groups

JGBSouza:
Add basic solutions for #657 and #550.


src/mail.sh
```diff
@@ -20,6 +20,13 @@ declare -ga optional_config_options=('sendemail.smtpencryption' 'sendemail.smtpp
 
 declare -gr email_regex='[A-Za-z0-9_\.-]+@[A-Za-z0-9_-]+(\.[A-Za-z0-9]+)+'
 
+# Mail group database tables
+declare -g DATABASE_TABLE_GROUP='mail_group'
+declare -g DATABASE_TABLE_CONTACT='mail_contact'
+declare -g DATABASE_TABLE_CONTACT_GROUP='mail_contact_group'
```
rodrigosiqueira, line 26: I think this should have -r as well.

tests/unit/samples/db_files/init.sql
```diff
@@ -27,3 +27,26 @@ CREATE TABLE IF NOT EXISTS "fake_table" (
   "rank" INTEGER NOT NULL,
   PRIMARY KEY("id")
 );
+
+CREATE TABLE IF NOT EXISTS "mail_group" (
```
rodrigosiqueira, line 31: wait... are we copying the database from database to samples? We should use the file from the database folder; otherwise, we might create differences between these two files.

database/kwdb.sql
```diff
@@ -20,6 +20,33 @@ CREATE TABLE IF NOT EXISTS "command_label" (
   PRIMARY KEY("id")
 );
 
+-- Table containing the kw_mail groups
```
rodrigosiqueira, line 23: @davidbtadokoro please, review the database.

src/mail.sh
```diff
@@ -53,6 +53,11 @@ function mail_main()
     return 0
   fi
 
+  if [[ -n "${options_values['GROUPS']}" ]]; then
+    kw_mail_groups "$flag"
```
rodrigosiqueira, line 58: Since this function might fail, you must handle the error here.

src/mail.sh
```diff
@@ -357,6 +362,94 @@ function find_commit_references()
...
+# @group_name: The name of the group that will be created.
+#
+# Returns:
+# returns 0 if suc cessful, non-zero otherwise
+function create_kw_mail_group()
+{
+  local group_name="$1"
+  local values
+
+  validate_group_name "$group_name" || exit_msg 'Error while creating new group'
```
rodrigosiqueira, line 460: I recommend using the traditional if command here, also, avoid exit here and just return some error code. The same idea applies to the below line.

src/mail.sh
```diff
@@ -357,6 +362,94 @@ function find_commit_references()
...
+  fi
+
+  name_length="$(str_length "$group_name")"
+
+  if [[ "$name_length" -ge 50 ]]; then
+    complain 'The group name must be less than 50 characters'
+    return 22 #EINVAL
+  fi
+
+  if [[ "$group_name" =~ ['!'@#\$%^\&*\(\)+,] ]]; then
```
rodrigosiqueira, line 493: Create a function in the kw_string file to handle special characters.

src/mail.sh
```diff
@@ -959,7 +1052,7 @@ function parse_mail_options()
   local commit_count=''
   local short_options='s,t,f,v:,i,l,n,'
   local long_options='send,simulate,to:,cc:,setup,local,global,force,verify,verbose,'
-  long_options+='template::,interactive,no-interactive,list,private,rfc,'
+  long_options+='template::,interactive,no-interactive,list,private,rfc,groups::,create,'
```
rodrigosiqueira, line 1609: I did not understand why groups have an optional parameter. What is the idea behind it? I guess if you add documentation, it can be easier to understand your intention about groups and create them.

src/mail.sh
```diff
@@ -450,6 +455,96 @@ function check_existent_group()
...
+# @group_name: The name of the removed group
+#
+# Returns:
+# returns 0 if successful, non-zero otherwise
+function remove_kw_mail_group()
+{
+  local group_name="$1"
+  local group_id
+
+  validate_group_name "$group_name" || exit_msg 'Error while removing new group'
```
rodrigosiqueira, line 535: use the standard `if` in this function and others. Also, do not exit here.

src/mail.sh
```diff
@@ -450,6 +455,96 @@ function check_existent_group()
...
+
+  check_existent_group "$group_name" && exit_msg $'This group doesnt exist \nError while removing new group'
+
+  group_id="$?"
+
+  remove_group_contacts_association "$group_id" || exit_msg 'Error while removing contacts associations'
+
+  remove_contacts_without_group "$group_id" || exit_msg 'Error while removing contacts without group'
+
+  remove_group "$group_name" || exit_msg 'Error while removing group'
```
rodrigosiqueira, line 545: Don't we have a cascade option that handles this type of situation? For example, remove from the group tables removes all the elements from the associated tables.

@davidbtadokoro

src/mail.sh
```diff
@@ -1207,6 +1304,10 @@ function parse_mail_options()
         options_values['GROUPS_CREATE']=1
         shift
         ;;
+      --remove)
+        options_values['GROUPS_REMOVE']=1
+        shift
```
rodrigosiqueira, line 1781: Again, document these features on the man page. Also, these names are too generic. You should use something like remove-group.

src/mail.sh
rodrigosiqueira, line 1: The commit title and message are a little bit confusing.

rodrigosiqueira (changes requested):
Hi @JGBSouza ,
Thanks a lot for your PR, it looks nice.

I added some comments that can be transferred to all your commits; please take a look. Also, reviewdog is complaining about some old stuff... since you are working on the mail feature, could you address those issues in a different PR?

@davidbtadokoro could you also check this PR?

rodrigosiqueira:
@JGBSouza any update here?

JGBSouza:
> @JGBSouza any update here?

I'm working on it. I recently moved the kw mail groups feature to kw manager and added the new feature/integration with kw mail to save a group based on the current email recipients, as well as updating the syntax of some commands like rename, create, remove, show, add, ..., to fit what we discussed in a past meeting. I still need to adjust some tests and fix the reported errors

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+# returns 0 if successful, non-zero otherwise
+function add_kw_manager_contacts()
+{
+  local contacts_list="$1"
+  local group_name="$2"
+  local group_id
+  declare -A _contacts_array
+  declare -A _new_contacts_array
+
+  if [[ -z "$contacts_list" ]]; then
```
rodrigosiqueira, line 321: Don't you need to add a check for `group_name` as well?

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+    complain 'The contacts list is empty'
+    return 22 #EINVAL
+  fi
+
+  check_existent_group "$group_name" && exit_msg 'Error while checking group existence'
+
```
rodrigosiqueira, line 327: Remove the new line.

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+  local group_id
+  declare -A _contacts_array
+  declare -A _new_contacts_array
+
+  if [[ -z "$contacts_list" ]]; then
+    complain 'The contacts list is empty'
+    return 22 #EINVAL
+  fi
+
+  check_existent_group "$group_name" && exit_msg 'Error while checking group existence'
```
rodrigosiqueira, line 326: Are you sure you want to exit here? Maybe you want to return and let the main function dealing with the exit.

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+
+  group_id="$?"
+
+  if [[ $(split_contact_infos "$contacts_list" '_contacts_array') ]]; then
+    complain 'Error while parsing contacts list'
```
rodrigosiqueira, line 331: Is it okay to move on to the other activities if there is an error? Don't you need to handle this case?

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+  fi
+
+  if [[ $(add_contacts '_contacts_array') ]]; then
+    complain 'Error while adding new contacts to the database'
```
rodrigosiqueira, line 335: Should you proceed if it fails?

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+
+    validate_contact_infos "$email" "$name" || return 22 #EINVAL
+
+    contacts_array["$email"]="$name"
+    ((added_contacts++))
+  done
+
+  if [[ "$added_contacts" -ne "${#contacts_array[@]}" ]]; then
+    complain 'Some of the contacts must have a repeated email'
+    return 22 #EINVAL
```
rodrigosiqueira, line 392: Is this a problem if you did not add them? I think if you did not add a duplication, it should be expected that the users will accidentally add a repeated email.

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+}
+
+# This function validate if the string with the contact infos is valid
+#
+# @contact_info: The string with the contact info,
+# formed as: "CONTACT_NAME <CONTACT_EMAIL>"
+#
+# Returns:
+# returns 0 if successful, non-zero otherwise
+function check_infos_sintaxe()
```
rodrigosiqueira, line 405: Drop the `e` at the end of `sintaxe`, and replace `i` with `y`.

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+{
+  local contact_info="$1"
+  local lt_pos
+  local lt_count
+  local gt_pos
+  local gt_count
```
rodrigosiqueira, line 411: gt and gl are not descriptive; please use a better variable name.

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+
+# This function split the columns in the given contact string passed
+# as parameter and create an associative array with the infos.
+#
+# @contacts_list: An string formed as "CONTACT_NAME <CONTACT_EMAIL>, CONTACT_NAME <CONTACT_EMAIL>, ..."
+# @contacts_array: An associative array formed as "<CONTACT_EMAIL>:<CONTACT_NAME>"
+#
+# Returns:
+# returns 0 if successful, non-zero otherwise
+function split_contact_infos()
```
rodrigosiqueira, line 353: I feel this could be a generic function that other parts of kw can use, such as the email and signature code.

src/kw_manager.sh
rodrigosiqueira, line 405: I feel this could be a generic function that other parts of kw can use, such as the email and signature code.

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+{
+  local email="$1"
+  local name="$2"
+
+  if [[ -z "$email" || -z "$name" ]]; then
+    complain 'Some of the contact names or emails must be empty'
+    return 22 #EINVAL
+  fi
+
+  validate_email "$email" || return 22 #EINVAL
```
rodrigosiqueira: this function should be generic, maybe you can make a PR that makes this part of the kw_str.

src/kw_manager.sh
```diff
@@ -265,15 +282,245 @@ function rename_kw_manager_group()
...
+  local values
+  local result
+  local email
+  local contact_id
+  local contact_name
+  local message
+  local entries
+  local existent_contact
+
+  for email in "${!contacts_array[@]}"; do
```
rodrigosiqueira, line 483: Don't you need to validate `contacts_array` first?

src/kw_manager.sh
```diff
@@ -328,5 +580,6 @@ function manager_help()
   fi
   printf '%s\n' 'kw manager:' \
     '  manager (-c | --group-create) [<name>] - create new group' \
-    '  manager (-r | --group-remove) [<name>] - remove existent group'
+    '  manager (-r | --group-remove) [<name>] - remove existent group' \
```
rodrigosiqueira, line 1004: don't forget to update the documentation.

src/send_patch.sh
```diff
@@ -6,6 +6,7 @@
 include "${KW_LIB_DIR}/lib/kw_config_loader.sh"
 include "${KW_LIB_DIR}/lib/kwlib.sh"
 include "${KW_LIB_DIR}/lib/kw_string.sh"
+include "./src/kw_manager.sh"
```
rodrigosiqueira, line 9: Can we handle this commit after we figure out the contact part? It could be a new PR. Right now, let's just focus on the group.

documentation/man/features/kw-manager.rst
```diff
@@ -0,0 +1,93 @@
+===============================================
+kw-manager - Manage groups of emails 
+===============================================
+
```
rodrigosiqueira, line 4: I see... you created a single commit for this. Ideally, you want to grow the documentation in the PR as you add new things, but for simplicity's sake, you can ignore my previous comments about the documentation and just keep this final commit with that.

documentation/man/features/kw-manager.rst
```diff
@@ -0,0 +1,93 @@
+===============================================
```
rodrigosiqueira, line 1: Keep these marks at the same size as the title.

documentation/man/features/kw-manager.rst
```diff
@@ -0,0 +1,93 @@
...
+
+.. _manager-doc:
+
+SYNOPSIS
+========
+  | *kw manager* (-c | \--group-create) [<name>]
+  | *kw manager* (-r | \--group-remove) [<name>]
+  | *kw manager* \--group-rename "[<old_name>]:[<new_name>]"
+  | *kw manager* \--group-add "[<group_name>]:[<ctt1_name>] <[<ctt1_email>]>, ..."
+  | *kw manager* \--group-remove-email "[<group_name>]:[<ctt1_name>] <[<ctt1_email>]>, ..."
```
rodrigosiqueira, line 13: `ctt1_`? This could be more intuitive.

src/kw_manager.sh
```diff
@@ -135,15 +146,98 @@ function check_existent_group()
...
+  local group_id
+
+  if [[ "$(validate_group_name "$group_name")" ]]; then
+    return 22 #EINVAL
+  fi
+
+  if ! [[ "$(check_existent_group "$group_name")" ]]; then
+    complain 'This group doesnt exist'
+    return 22 #EINVAL
+  fi
```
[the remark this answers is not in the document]
JGBSouza: is not exactly the same, in this case the function would fail if the group doesn't exist, while when we are creating it, it would fail if the group exist. But i suppose i can remove the validate_group_name cause if the group exists then it doesn't matter if the name is valid cause it's already in the database.

src/kw_manager.sh
```diff
@@ -135,15 +146,98 @@ function check_existent_group()
...
+}
+
+# This function removes all the contacts without group association
+# from the database
+#
+# @group_name: Name of the group which the contacts will be removed
+#
+# Return:
+# returns 0 if successful, non-zero otherwise
+function remove_contacts_without_group()
```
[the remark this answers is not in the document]
JGBSouza: Based on what i searched, the Cascade only works in foreign keys, wich is the case for the 'manager_email_contact_group'. So, when we remove the group, all of it's references would be erased from this table, but the contacts in the 'manager_email_contact' will remain in the database. But as as solution i could use a trigger in the 'manager_email_contact_group' to check whenever an entry is removed if the respective user has any remaining association and remove it if doesnt

PR #1124, opened 2024-06-02
doc: build save-log-to path does not use '='

auyer:
The doc for the save-log-to option is outdated here, showing a sintax '--save-log-to=path', but the correct is '--save-log-to path'


davidbtadokoro:
Hi @auyer, and thank you for the nice catch! Legit, so merged into the unstable branch :+1:

PR #1130, opened 2024-06-04
run_tests: dedicate a container per test file for integration tests

aquilamacedo:
This PR refactors the integration test setup to dedicate a separate
container for each test file. The `prepare_testing_environment` function
is introduced in individual test scripts to set up the environment
specifically for each test, ensuring isolated test environments.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -6,9 +6,10 @@ include './tests/integration/utils.sh'
 include './src/lib/kwio.sh'
 
 declare -a TESTS
-declare TESTS_DIR
+declare TESTS_DIR='./tests'
 declare TESTS_UNIT=1
 declare TESTS_INTEGRATION=1
+declare VERBOSE=0
 
 function show_help()
 {
@@ -21,6 +22,8 @@ function show_help()
     '         Limit tests to integration tests' \
     '  -u, --unit' \
     '         Limit tests to unit tests' \
+    '  -v, --verbose' \
+    '         Increase verbosity of the output' \
     '' \
     'COMMANDS' \
     '  clear-cache - clears tests cache' \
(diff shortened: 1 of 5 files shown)


run_tests.sh
```diff
@@ -125,12 +113,10 @@ function run_tests()
...
-  if [[ "${integration_tests_setup}" == 1 ]]; then
-    printf '\n' # add new line after the last "OK"
-    say 'Tearing down containers used in integration tests...'
-    teardown_containers
-    printf '\n'
-  fi
+  printf '\n' # add new line after the last "OK"
+  say 'Tearing down containers used in integration tests...'
+  teardown_containers
+  printf '\n'
```
rodrigosiqueira, line 119: Can we try moving the first `printf` into the `say` function? Also, consider moving the last `printf` to the `teardown_containders`  function.

tests/integration/device_test.sh
```diff
@@ -4,6 +4,11 @@ include './tests/unit/utils.sh'
 include './tests/integration/utils.sh'
 include './src/device_info.sh'
 
+function oneTimeSetUp()
+{
+  prepare_testing_environment 'device'
```
rodrigosiqueira, line 9: With this function, we display the setup for every test, for example:

```
Preparing environment for kw config integration tests...
[1/6] Using cached container image for archlinux.
[2/6] Creating archlinux container.
[3/6] Using cached container image for debian.
[4/6] Creating debian container.
[5/6] Using cached container image for fedora.
[6/6] Creating fedora container.
```

Can we add a silent option for the `setup_container_environment` function?

rodrigosiqueira (changes requested):

tests/integration/device_test.sh
aquilamacedo: I've included a `--verbose` option for situations where we need to see this information.

rodrigosiqueira:
Hi @aquilamacedo , thanks for your PR!

Issue #952, opened 2023-11-04
Standardize cmd_remotely based on cmd_manager


rodrigosiqueira:
**Description:**

Even though the function `cmd_manager` and `cmd_remotely` are very similar in terms of goals, they have a different API:

* `cmd_manager FLAG CMD`
* `cmd_remote CMD FLAG`

Let's use the `cmd_manager` pattern for the `cmd_remotely` for consistency sake.
 
**Type:**

Refactor

shikharish:
Hello @rodrigosiqueira, I would like to work on this.

dudszzz:
Hello, I can work on this issue.

PR #1133, opened 2024-06-06
src: lib: remote: standarization of the function cmd_remotely

duzaao:
The problem is the different standards in the functions cmd_remotely and cmd_manager.
In this sense, flipped the positions to fix the problem.
#952


rodrigosiqueira:
Thanks @duzaao ,
I changed the commit message, and I merged it into the unstable branch.

Issue #1123, opened 2024-05-31
Implement a `git` lib in `src/lib/git.sh`


MarceloSpessoto:
**Description:**

kworkflow uses `git` for many different cases. It would be great if a specific lib was created to provide git easy git functionalities for kworkflow code.

**Type:**

Refactor

**Why:**

Many kworkflow features, such as `--from-sha` option for `kw build` rely on git functionalities and use it directly, generating a lot of repetitive code. Providing common specific `git` use cases functionalities would improve code organization.

rodrigosiqueira:
From build.sh:

```bash

  if [[ -n "$from_sha_arg" ]]; then
    # Check if there is a rebase in process.
    if [[ -d .git/rebase-merge ]]; then
      warning 'ERROR: Abort the repository rebase before continuing with build from sha (use "git rebase --abort")!'
      return 125 # ECANCELED
    elif [[ -f .git/MERGE_HEAD ]]; then
      warning 'ERROR: Abort the repository merge before continuing with build from sha (use "git rebase --abort")!'
      return 125 # ECANCELED
    elif [[ -f .git/BISECT_LOG ]]; then
      warning 'ERROR: Stop the repository bisect before continuing with build from sha (use "git bisect reset")!'
      return 125 # ECANCELED
    elif [[ -d .git/rebase-apply ]]; then
      printf 'ERROR: Abort the repository patch apply before continuing with build from sha (use "git am --abort")!'
      return 125 # ECANCELED
    fi

  # Check if given SHA represents real commit
    cmd_manager 'SILENT' "git cat-file -e ${from_sha_arg}^{commit} 2> /dev/null"
    if [[ "$?" != 0 ]]; then
      complain "ERROR: The given SHA (${from_sha_arg}) does not represent a valid commit sha."
      return 22 # EINVAL
    fi

    # Check if given SHA is in working tree.
    sha_base=$(git rev-parse --verify "$from_sha_arg")
    merge_base=$(git merge-base "$from_sha_arg" HEAD)
    if [[ "$sha_base" != "$merge_base" ]]; then
      complain "ERROR: Given SHA (${from_sha_arg}) is invalid. Check if it is an ancestor of the branch head."
      return 22 # EINVAL
    fi
```

This could be encapsulated in a library.

AbhinavMir:
@rodrigosiqueira had a look, spent some time thinking about the git commands one could need. Would these make sense?

- Check for ongoing rebase: .git/rebase-merge directory exists
[...]

PR #1136, opened 2024-06-11
src: lib: git: Implement a git lib in src/lib/git.sh

BrizaMel:
Create lib that provides git easy functionalities for kworkflow code.

#1123

### DIFF
--- a/src/lib/git.sh
+++ b/src/lib/git.sh
@@ -1,0 +1,51 @@
+# The git.sh file provides git easy functionalities for kworkflow code.
+
+# Run git bisect
+#
+# @good: good commit hash
+# @bad: bad commit hash
+# @command: test performed by git bisect
+function run_git_bisect()
+{
+  local good="$1"
+  local bad="$2"
+  local command="$3"
+
+  git bisect start
+  if [[ $? != 0 ]]; then
+    printf '%s\n' 'git bisect start failed'
+    exit 22 # EINVAL
+  fi
+  git bisect good "$good"
+  if [[ $? != 0 ]]; then
+    printf '%s\n' 'git bisect good failed'
+    exit 22 # EINVAL
+  fi
+  git bisect bad "$bad"
+  if [[ $? != 0 ]]; then
+    printf '%s\n' 'git bisect bad failed'
+    exit 22 # EINVAL
+  fi
+  git bisect run "$command"
+  if [[ $? != 0 ]]; then
+    printf '%s\n' 'command run failed'
+    exit 22 # EINVAL
+  fi
+  return 0
+}
+
+# Checks if the command is being run inside a git work-tree
+#
+# @flag: How to display (or not) the command used
+#
+# Returns:
+# 0 if is inside a git work-tree root and 128 otherwise.
+function is_inside_work_tree()
+{
+  local flag="$1"
+  local cmd='git rev-parse --is-inside-work-tree &> /dev/null'
+
+  flag=${flag:-'SILENT'}
+
+  cmd_manager "$flag" "$cmd"
+}
--- a/src/lib/kwlib.sh
+++ b/src/lib/kwlib.sh
@@ -2,6 +2,7 @@
 include "${KW_LIB_DIR}/lib/kw_string.sh"
 include "${KW_LIB_DIR}/lib/kw_db.sh"
 include "${KW_LIB_DIR}/lib/kw_time_and_date.sh"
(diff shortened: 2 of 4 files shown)


src/lib/git.sh
rodrigosiqueira, line 1: Please update the commit message to explain why we want to create this file and which features will be affected by this library. Additionally, add a reference to the commit message.

src/lib/git.sh
```diff
@@ -0,0 +1,51 @@
+# The git.sh file provides git easy functionalities for kworkflow code.
```
rodrigosiqueira, line 1: Just one overall comment for this file. Keep in mind that the bar for lib files should be as high as possible since many other files will use it, and it is easy to add tests. Remember, a library is only useful if it is reliable; try to be very rigorous about validations and corner cases here.

src/lib/git.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Run git bisect
```
rodrigosiqueira, line 3: Please expand this documentation to explain how this function is expected to work and possible pitfalls that a developer may encounter.
P.S.: Review all your documentation for this file. Remember, this is a library file and we want to keep it easy to understand.

src/lib/git.sh
```diff
@@ -0,0 +1,51 @@
...
+#
+# @good: good commit hash
+# @bad: bad commit hash
+# @command: test performed by git bisect
```
rodrigosiqueira, line 7: Document returns

src/lib/git.sh
```diff
@@ -0,0 +1,51 @@
...
+function run_git_bisect()
+{
+  local good="$1"
+  local bad="$2"
+  local command="$3"
+
+  git bisect start
+  if [[ $? != 0 ]]; then
```
rodrigosiqueira, line 15: Add double quotes around `$?`. Also, it is worth exploring if git bisect has different return errors that we can handle here.

src/lib/git.sh
```diff
@@ -0,0 +1,51 @@
...
+    printf '%s\n' 'git bisect start failed'
```
rodrigosiqueira, line 16: Use the `complain` function.
The same comment is valid for this entire PR.

src/lib/git.sh
```diff
@@ -0,0 +1,51 @@
...
+    exit 22 # EINVAL
```
rodrigosiqueira, line 17: Avoid exit in a library function since this makes the debug very hard. Since this is a library, we should return an error code and let the caller handle the possible failure.

src/lib/git.sh
```diff
@@ -0,0 +1,51 @@
...
+  fi
+  git bisect good "$good"
```
rodrigosiqueira, line 19: Add a new line between these two lines.

src/lib/git.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Run git bisect
+#
+# @good: good commit hash
+# @bad: bad commit hash
+# @command: test performed by git bisect
+function run_git_bisect()
+{
+  local good="$1"
+  local bad="$2"
```
rodrigosiqueira, line 11: You can't trust the user input. Please validate it before using it.

src/lib/git.sh
```diff
@@ -0,0 +1,51 @@
...
+    exit 22 # EINVAL
+  fi
+  git bisect run "$command"
+  if [[ $? != 0 ]]; then
+    printf '%s\n' 'command run failed'
+    exit 22 # EINVAL
+  fi
+  return 0
+}
+
```
rodrigosiqueira, line 36: iirc, we have some validation for the SHA in the build function. That validation should be part of this lib.

rodrigosiqueira (changes requested):
Hi,
Thanks for this PR, I added some comments. However, when working on this feature, keep these things in mind:
1. You are writing a library component that must be useful in many different places. In this sense, you should be very picky about all your changes to ensure a good consistency for the lib.
2. Documentation is crucial for libs.
3. Check where we use the git command in the entire code from kw, and, based on that, try to abstract our usage of those commands as a library where it makes sense.
4. After you create the library, also add commits that replace the use of git in the kw source code with the new library functions.
5. Test is crucial here!

Thanks

rodrigosiqueira:
@BrizaMel any update here?

rodrigosiqueira:
Closed due to the lack of activity.

Issue #1097, opened 2024-04-27
Display current display resolution


rodrigosiqueira:
**Description:**

`kw drm` has the option `--modes` that exposes all supported modes by the driver. It would be very useful if this option also shows the current display resolution. For example:

```
Modes per card
/card0-DP-2:
3840x2160*
3840x2160
3840x2160
3840x2160
3840x2160
3840x2160
3840x2160
2560x1440
1920x1200
1920x1080
1920x1080
1920x1080
1920x1080
1600x1200
1680x1050
1600x900
1280x1024
1440x900
1280x800
1280x720
1280x720
1280x720
1280x720
1024x768
800x600
720x576
720x576
720x480
720x480
720x480
720x480
640x480
640x480
640x480
720x400
```
We can probably get this information by parsing the file `/sys/kernel/debug/dri/0/state`. It would be even better if this command also shows the refresh rate.

viniciusfl:
I am currently working on this.

PR #1137, opened 2024-06-11
src: plugins: subsystems: drm: show current resolution and refresh rate

viniciusfl:
Solves [issue 1097](https://github.com/kworkflow/kworkflow/issues/1097).

### DIFF
--- a/src/plugins/subsystems/drm/drm.sh
+++ b/src/plugins/subsystems/drm/drm.sh
@@ -6,6 +6,8 @@ declare -gr UNLOAD='UNLOAD'
 declare -gA options_values
 
 declare -g SYSFS_CLASS_DRM='/sys/class/drm'
+declare -g SYSFS_KERNEL_DEBUG_DRI='/sys/kernel/debug/dri'
+declare -g USER_ID="$EUID"
 
 function drm_main()
 {
@@ -345,8 +347,32 @@ function get_supported_mode_per_connector()
   modes=${modes//\/modes/}
   modes=${modes//sys\/class\/drm\//}
 
+  # get current resolution and hz
+  regex='mode: "[0-9]+x[0-9]+": [0-9]+'
+
   say 'Modes per card'
-  printf '%s\n' "$modes"
+  if [[ "$USER_ID" -eq 0 ]]; then
+    for file in "$SYSFS_KERNEL_DEBUG_DRI/"*"/state"; do
+      matches=$(grep -E "${regex}" "$file")
+      while IFS= read -r line; do
+        if [[ $line =~ $regex ]]; then
+          resolution=$(sed -n 's/.*"\(.*\)".*/\1/p' <<< "${BASH_REMATCH[0]}")
+          hz=$(awk -F': ' '{print $3}' <<< "${BASH_REMATCH[0]}")
+          break
+        fi
+      done <<< "$matches"
+    done
+
+    for line in $modes; do
+      if [[ $line == "$resolution" ]]; then
+        printf '%s* %sHz\n' "$resolution" "$hz"
+      else
+        printf '%s\n' "$line"
+      fi
+    done
+  else
+    printf '%s\n' "$modes"
+  fi
 }
 
 function parse_drm_options()
--- a/tests/unit/drm_plugin_test.sh
+++ b/tests/unit/drm_plugin_test.sh
@@ -8,10 +8,13 @@ function setUp()
 {
   # Create a temporary directory for holding different config file
   export FAKE_DRM_SYSFS="${SHUNIT_TMPDIR}/sys/class/drm"
+  export FAKE_DRI_SYSFS="${SHUNIT_TMPDIR}/sys/kernel/debug/dri"
   export original_dir="$PWD"
   export SYSFS_CLASS_DRM="$FAKE_DRM_SYSFS"
+  export SYSFS_KERNEL_DEBUG_DRI="$FAKE_DRI_SYSFS"
 
   mkdir -p "$FAKE_DRM_SYSFS"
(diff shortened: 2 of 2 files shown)


viniciusfl:
@rodrigosiqueira @davidbtadokoro

src/plugins/subsystems/drm/drm.sh
```diff
@@ -345,8 +347,32 @@ function get_supported_mode_per_connector()
...
+  # get current resolution and hz
+  regex='mode: "[0-9]+x[0-9]+": [0-9]+'
+
   say 'Modes per card'
-  printf '%s\n' "$modes"
+  if [[ "$USER_ID" -eq 0 ]]; then
+    for file in "$SYSFS_KERNEL_DEBUG_DRI/"*"/state"; do
+      matches=$(grep -E "${regex}" "$file")
+      while IFS= read -r line; do
+        if [[ $line =~ $regex ]]; then
```
rodrigosiqueira, line 358: Add double-quotes around those bash variables.

src/plugins/subsystems/drm/drm.sh
```diff
@@ -345,8 +347,32 @@ function get_supported_mode_per_connector()
...
+          resolution=$(sed -n 's/.*"\(.*\)".*/\1/p' <<< "${BASH_REMATCH[0]}")
```
rodrigosiqueira, line 359: Add a simple comment before your sed that describes (i.e., add an example) what kind of output we expected for `BASH_REMATCH`.

src/plugins/subsystems/drm/drm.sh
```diff
@@ -345,8 +347,32 @@ function get_supported_mode_per_connector()
...
+          hz=$(awk -F': ' '{print $3}' <<< "${BASH_REMATCH[0]}")
+          break
+        fi
+      done <<< "$matches"
+    done
+
+    for line in $modes; do
+      if [[ $line == "$resolution" ]]; then
```
rodrigosiqueira, line 367: double quotes around $line

src/plugins/subsystems/drm/drm.sh
```diff
@@ -345,8 +347,32 @@ function get_supported_mode_per_connector()
...
+
   say 'Modes per card'
-  printf '%s\n' "$modes"
+  if [[ "$USER_ID" -eq 0 ]]; then
+    for file in "$SYSFS_KERNEL_DEBUG_DRI/"*"/state"; do
+      matches=$(grep -E "${regex}" "$file")
+      while IFS= read -r line; do
+        if [[ $line =~ $regex ]]; then
+          resolution=$(sed -n 's/.*"\(.*\)".*/\1/p' <<< "${BASH_REMATCH[0]}")
+          hz=$(awk -F': ' '{print $3}' <<< "${BASH_REMATCH[0]}")
```
rodrigosiqueira, line 360: by hz, I think you mean refresh_rate.

tests/unit/drm_plugin_test.sh
```diff
@@ -97,6 +101,33 @@ END
 END
 }
 
+function mk_fake_sys_kernel_debug_dri()
```
rodrigosiqueira, line 104: This function can be useful in other tests; move it to tests/unit/utils.sh.

tests/unit/drm_plugin_test.sh
```diff
@@ -97,6 +101,33 @@ END
...
+{
+  declare -a fake_dirs=(
+    "1"
+    "128")
```
rodrigosiqueira, line 108: use single quotes.

tests/unit/drm_plugin_test.sh
```diff
@@ -97,6 +101,33 @@ END
...
+
+  for dir in "${fake_dirs[@]}"; do
+    mkdir -p "${FAKE_DRI_SYSFS}/${dir}"
```
rodrigosiqueira, line 111: avoid short options.

tests/unit/drm_plugin_test.sh
```diff
@@ -332,6 +363,39 @@ function test_get_supported_mode_per_connector()
   export SYSFS_CLASS_DRM="$FAKE_DRM_SYSFS"
   output=$(get_supported_mode_per_connector 2)
   compare_command_sequence '' "$LINENO" 'expected_output' "$output"
+
+  declare -a expected_output=(
```
rodrigosiqueira, line 367: Replace all double quotes with single quotes where we don't have a bash variable.

rodrigosiqueira (changes requested):
Hi,
Thanks for this PR; I added some comments. Also, your feature does not work as expected for remote machines or even for local machines since it is expected to be executed as root, which is not correct.

cmhpedro:
> Hi, Thanks for this PR; I added some comments. Also, your feature does not work as expected for remote machines or even for local machines since it is expected to be executed as root, which is not correct.

Hi! Thanks for the response. We are accessing a kernel file, that needs `sudo` permission. The expected way is executing `kw ...` and requesting `sudo` permission (similar to `setup`)? If this is not the way, is there a suggestion on how to do it?

rodrigosiqueira:
> > Hi, Thanks for this PR; I added some comments. Also, your feature does not work as expected for remote machines or even for local machines since it is expected to be executed as root, which is not correct.
> 
> Hi! Thanks for the response. We are accessing a kernel file, that needs `sudo` permission. The expected way is executing `kw ...` and requesting `sudo` permission (similar to `setup`)? If this is not the way, is there a suggestion on how to do it?

You don't need sudo to deploy to remote since it runs as root.

Also, are there any updates to my other comments?

rodrigosiqueira:
Closed due to the lack of activity.

PR #1138, opened 2024-06-11
src: create build_and_deploy file

BrizaMel:
Create build_and_deploy utilitary. 

[1120](https://github.com/kworkflow/kworkflow/pull/1120)

### DIFF
--- a/documentation/man/features/kw-bd.rst
+++ b/documentation/man/features/kw-bd.rst
@@ -1,0 +1,35 @@
+=======================================
+kw-bd - Build the kernel and deploy it
+=======================================
+
+.. _bd-doc:
+
+SYNOPSIS
+========
+
+*kw* (*bd*)
+
+DESCRIPTION
+===========
+
+Kw bd is expected to run inside a kernel tree. It executes the build,
+and if there are no errors, it proceeds with deploying the new kernel.
+
+Currently, all arguments invoked along with kw bd will be redirected to kw deploy.
+
+See the kw build and kw deploy man pages for more information.
+
+.. note::
+  This command must be run inside of a kernel tree.
+
+EXAMPLES
+========
+
+For simply executing build and deploy use::
+
+	kw bd
+
+If you want to build and deploy the kernel with 
+--create-package option use::
+
+	kw bd --create-package
--- a/kw
+++ b/kw
@@ -138,10 +138,9 @@ function kw()
       ;;
     bd)
       (
-        include "${KW_LIB_DIR}/deploy.sh"
-        include "${KW_LIB_DIR}/build.sh"
+        include "${KW_LIB_DIR}/build_and_deploy.sh"
 
-        build_kernel_main && deploy_main 1 "$@"
+        build_and_deploy_main "$@"
         local ret="$?"
         alert_completion 'kw bd' "$1"
         return "$ret"
--- a/src/build_and_deploy.sh
+++ b/src/build_and_deploy.sh
@@ -1,0 +1,78 @@
+#
+# The `build_and_deploy.sh` is the entrypoint for `kw bd`, i.e.,
+# `kw build` followed by `kw deploy`.
+#
+
+include "${KW_LIB_DIR}/deploy.sh"
+include "${KW_LIB_DIR}/build.sh"
+include "${KW_LIB_DIR}/lib/kwio.sh"
+
+# `kw bd` is a shortcut for `kw build` followed `kw deploy`, to run the entire workflow
+# of building a kernel from source and deploying it to a target machine.
+#
+# @raw_options String with all deploy options
+function build_and_deploy_main()
+{
+  if [[ "$1" =~ -h|--help ]]; then
+    build_and_deploy_help "$1"
+    exit 0
+  fi
+
+  parse_build_and_deploy_options "$@"
+  if [[ "$?" -gt 0 ]]; then
+    complain "${options_values['ERROR']}"
+    build_and_deploy_help
+    return 22 # EINVAL
+  fi
+
+  build_kernel_main
+  if [[ "$?" != 0 ]]; then
+    complain 'kernel build failed\n'
+    exit 22 # EINVAL
+  fi
+
+  # parameter 1 indicates that deploy is being called within kw bd context.
+  deploy_main 1 "$@"
+  if [[ "$?" != 0 ]]; then
+    complain 'kernel deploy failed\n'
+    exit 22 # EINVAL
+  fi
+  return 0
+}
+
+function parse_build_and_deploy_options()
+{
+  local long_options=''
+  local short_options=''
+
+  options="$(kw_parse "$short_options" "$long_options" "$@")"
+  if [[ "$?" != 0 ]]; then
+    options_values['ERROR']="$(kw_parse_get_errors 'kw bd' "$short_options" \
+      "$long_options" "$@")"
+    return 22 # EINVAL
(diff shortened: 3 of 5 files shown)


kw
rodrigosiqueira, line 1: Please, add a commit message.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,34 @@
+#
+# The `build_and_deploy.sh` file centralizes functions related to kw bd.
```
rodrigosiqueira, line 2: Maybe something like [..] file handles the sequence build followed by deploy.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+#
+
+include "${KW_LIB_DIR}/deploy.sh"
+include "${KW_LIB_DIR}/build.sh"
+
+# kw bd is a shortcut to build and deploy a new kernel.
+#
+# @raw_options String with all deploy options
+#
```
rodrigosiqueira, line 11: Drop this extra comment

src/build_and_deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+function build_and_deploy_main()
+{
+
```
rodrigosiqueira, line 14: drop this extra line

src/build_and_deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+  local raw_options="$*"
+
+  if [[ "$1" =~ -h|--help ]]; then
+    build_and_deploy_help "$1"
+    exit 0
+  fi
+
+  build_kernel_main && deploy_main 1 "$raw_options"
```
rodrigosiqueira, line 22: It is better to call each function, check the return type, and then execute the next function.

rodrigosiqueira (changes requested):

davidbtadokoro (changes requested):
Hi @BrizaMel and @lorenzoberts, and thanks for the initiative on extracting `kw bd` to a dedicated file.

I've left some inline comments, so please see if you can address them. There are also two comments that I that don't fit in the inline review:

1. As you are essentially adding a "new feature", you need to add a manpage, new completions, and an authorship of the new feature. See this PR #1121 from Lincoln, Luiza, and Sabrina for reference (pay attention the modifications in the `documentation` dir, and `src/_kw` and `src/bash_autocompletion.sh` files).
2. I think you can modify the commit message to express this adding of a "new feature", and explain the motivation in creating a dedicated file/feature for `kw bd` (again, I think the PR mentioned is a good reference).

@rodrigosiqueira, as a last discussion, how do we feel about the name `kw bd`? I think it can be easily mistaken with `kw db`, and maybe we have an opportunity to create a more expressive name with this PR.

**UPDATE 1:**

I remembered that `kw bd` completions are already implemented, so you don't need to change them. However, in case we change the feature name, you will need to change these files indeed.

**UPDATE 2:**

Also, I've tested the feature, and we need to use `include "${KW_LIB_DIR}/lib/kwio.sh"` for running kw outside the repository, and we will need to use `$@` instead of `$raw_options` as the deploy will fail, otherwise. Both problems are illustrated in the print below.

2024-06-19-160846_899x412_scrot

lorenzoberts:
Hello @davidbtadokoro, thanks for the comments. We tried to address them all, see what you think.
We didn't know if we were supposed to consider ourselves the authors of kw bd since we are just creating its file and man page. Let us know if we should change `documentation/conf.py`

documentation/man/features/kw-bd.rst
```diff
@@ -0,0 +1,42 @@
...
+========
+
+*kw* (*bd*)
+
+DESCRIPTION
+===========
+
+If you are in a kernel directory, this command will first try to
+execute kw build, which builds the kernel, and then execute kw deploy,
+which deploys the kernel.
```
rodrigosiqueira, line 17: Maybe something more direct like:

Kw bd is expected to run inside a kernel tree. It executes the build, and if there are no errors, it proceeds with deploying the new kernel.

documentation/man/features/kw-bd.rst
```diff
@@ -0,0 +1,42 @@
...
+
+Currently, kw bd does not support any arguments for itself. All arguments invoked 
+along with kw bd will be redirected to kw deploy.
```
rodrigosiqueira, line 20: The first phrase contradicts the second phrase.

documentation/man/features/kw-bd.rst
```diff
@@ -0,0 +1,42 @@
...
+
+See the kw build and kw deploy man pages for more information.
+
+.. note::
+  This command must be run inside of a kernel tree.
+
+OPTIONS
+=======
+
+Currently kw bd does not support any options for itself.
```
rodrigosiqueira, line 30: you don't need this if there is no option.

documentation/man/features/kw-bd.rst
```diff
@@ -0,0 +1,42 @@
...
+========
+
+For simply executing build and deploy use::
+
+	kw bd
+
+If you want to build and deploy the kernel with 
+--create-package option use::
+
+	kw bd --create-package
```
rodrigosiqueira, line 42: Again, this contradicts the information you provided above.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,76 @@
...
+    build_and_deploy_help
+    return 22 # EINVAL
+  fi
+
+  build_kernel_main
+  if [[ "$?" != 0 ]]; then
+    complain '%s\n' 'kernel build failed'
+    exit 22 # EINVAL
+  fi
+  deploy_main 1 "$@"
```
rodrigosiqueira, line 33: Add a new line after the `fi` part. Also, add a comment that explains what  `1` parameter means.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,76 @@
...
+  parse_build_and_deploy_options "$@"
+  if [[ "$?" -gt 0 ]]; then
+    complain "${options_values['ERROR']}"
+    build_and_deploy_help
+    return 22 # EINVAL
+  fi
+
+  build_kernel_main
+  if [[ "$?" != 0 ]]; then
+    complain '%s\n' 'kernel build failed'
```
rodrigosiqueira, line 30: You don't need to use  `%s` in this context, you can just add the string directly in the first parameter.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,76 @@
...
+    exit 22 # EINVAL
+  fi
+  deploy_main 1 "$@"
+  if [[ "$?" != 0 ]]; then
+    complain '%s\n' 'kernel deploy failed'
```
rodrigosiqueira, line 35: You don't need to use %s in this context, you can just add the string directly in the first parameter.

src/build_and_deploy.sh
```diff
@@ -0,0 +1,76 @@
...
+        shift
+        ;;
+    esac
+  done
+}
+
+function build_and_deploy_help()
+{
+  if [[ "$1" == --help ]]; then
+    include "$KW_LIB_DIR/help.sh"
```
rodrigosiqueira, line 70: https://kworkflow.org/content/codingstyle.html#string-concatenation

rodrigosiqueira (changes requested):

rodrigosiqueira:
Thanks for the improvements! Change merged into the unstable branch.

PR #1140, opened 2024-06-14
setup: Allow ZDOTDIR paths for zsh autocompletion setup

Lincoln-Yuji:
I was wondering why kw's zsh completions were not working at all on my system. It turns out that I forgot I had the `ZDOTDIR` variable defined to `~/.config/zsh/` directory inside my `~/.profile`:

```bash
export ZDOTDIR="${HOME}/.config/zsh" # ZSH dot directory
```
It's been a while since I had this setup, but if remember correctly I was told that I could do that because zsh partially follows the XDG Base Directory specification: [XDG Base Directory](https://wiki.archlinux.org/title/XDG_Base_Directory).

I believe some users might have the `ZDOTDIR` variable to alternative paths on their systems as well and checking this possibility helps the setup process to cover this possible corner case.

The motivation to use this variable rather than simply using symbolic links is to help "cleaning" the home directory. I know most users probably don't care about this, but since this is a possibility I believe it would be nice to check that during the setup.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -292,14 +292,14 @@ function remove_kw_from_PATH_variable()
 
 function update_path()
 {
-  local shellrc=${1:-'.bashrc'}
+  local shellrc="$1"
 
   IFS=':' read -ra ALL_PATHS <<< "$PATH"
   for path in "${ALL_PATHS[@]}"; do
     [[ "$path" -ef "$binpath" ]] && return
   done
 
-  safe_append "PATH=${HOME}/.local/bin:\$PATH # kw" "${HOME}/${shellrc}"
+  safe_append "PATH=${HOME}/.local/bin:\$PATH # kw" "$shellrc"
 }
 
 function update_current_bash()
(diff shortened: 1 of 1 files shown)


rodrigosiqueira:
@davidbtadokoro could you review this change?

davidbtadokoro:
Hi @Lincoln-Yuji, and thank you so much for spotting this bias! I didn't know about `ZDOTDIR` and, after searching further, it feels natural to support this type of setup for Zsh users.

I've reviewed the change and it looks legit to me. I've also tested it on my setup, but, although I am a Zsh user too, I emulated the described scenario by defining `ZDOTDIR` and moving the `.zshrc` to it and it seemed to work. I've also tested for regressions with `.zshrc` on `HOME` and without Zsh (only Bash user), and it seems alright :+1: 

Because of this, merged into the unstable branch (**great** commit message by the way!)

P.s.: @rodrigosiqueira, thanks for the heads up!

Issue #162, opened 2020-02-25
Create a feature to check code style by function or line interval


melissawen:
Improve "kw c" to receive the name of a function or line-start/line-end as a parameter and search for code-style issues in this interval.

emanuellima1:
Working on this with @kamorst

gustavofukunaga:
I'm working on this with @Caio-Dantas.

PR #1144, opened 2024-06-18
Create options to kw codestyle to check by function or line interval

gustavofukunaga:
Add options to  "kw c" to receive the name of a function using `--function` as a parameter to search for code-style issues in this function definition, or in a line interval defined by `--start-line` and `--end-line` options.

Closes: #162

### DIFF
--- a/src/bash_autocomplete.sh
+++ b/src/bash_autocomplete.sh
@@ -85,7 +85,7 @@ function _kw_autocomplete()
 
   kw_options['clear-cache']='--verbose'
 
-  kw_options['codestyle']='--verbose --help'
+  kw_options['codestyle']='--verbose --help --start-line --end-line'
   kw_options['c']="${kw_options['codestyle']}"
 
   mapfile -t COMPREPLY < <(compgen -W "${kw_options[${previous_command}]} " -- "${current_command}")
--- a/src/bash_autocomplete.sh
+++ b/src/bash_autocomplete.sh
@@ -85,7 +85,7 @@ function _kw_autocomplete()
 
   kw_options['clear-cache']='--verbose'
 
-  kw_options['codestyle']='--verbose --help --start-line --end-line'
+  kw_options['codestyle']='--verbose --help --start-line --end-line --function'
   kw_options['c']="${kw_options['codestyle']}"
 
   mapfile -t COMPREPLY < <(compgen -W "${kw_options[${previous_command}]} " -- "${current_command}")
--- a/src/codestyle.sh
+++ b/src/codestyle.sh
@@ -76,12 +76,14 @@ function codestyle_main()
   [[ -n "${options_values['START_LINE']}" ]] && start_line=${options_values['START_LINE']}
   [[ -n "${options_values['END_LINE']}" ]] && end_line=${options_values['END_LINE']}
 
+  [[ -n "${options_values['FUNCTION']}" ]] && function=${options_values['FUNCTION']}
+
   if [[ -n "$start_line" || -n "$end_line" ]]; then
     range_option_flag='true'
   fi
 
-  # Check if --start-line and --end-line options are being used with a single file
-  if [[ ! -f "$path" && -n "$range_option_flag" ]]; then
+  # Check if --start-line --end-line and --function options are being used with a single file
+  if [[ ! -f "$path" && -n "$range_option_flag" && -n "$function" ]]; then
     complain "Invalid path using start-line and end-line option: ${path}"
     return 2 # ENOENT
   fi
@@ -96,9 +98,15 @@ function codestyle_main()
     end_line=$(wc -l < "$FLIST")
   fi
 
-  # If --start-line and --end-line options are being used, assign a temporary file
-  # to FLIST with the line interval and SPDX License Identifier
-  if [[ -n "$range_option_flag" ]]; then
+  # For --function option, find the start line of the function definition
+  # and assign it to start_line.
+  if [[ -n "$function" ]]; then
+    start_line=$(awk "/[a-zA-Z]+[[:space:]]+${function}[[:space:]]*\\(.*[^;]$/ {print FNR}" "$FLIST")
+  fi
+
+  # If --start-line, --end-line or --function options are being used, assign a temporary file
+  # to FLIST with blank lines and SPDX License Identifier
+  if [[ -n "$range_option_flag" || -n "$function" ]]; then
     suffix=".${FLIST#*.}"
     temp_file=$(mktemp --suffix "$suffix" --tmpdir="$PWD")
 
@@ -116,6 +124,30 @@ function codestyle_main()
     FLIST="$temp_file"
   fi
 
+  # For --function option, write the function definition to the temporary file
+  # and assign it to FLIST
+  if [[ -n "$function" ]]; then
+    awk '
+      $0 ~ "[a-zA-Z]+[[:space:]]+"FUNCTION_NAME"[[:space:]]*\\(.*[^;]$" {
+        in_function_header = 1
+      }
+      in_function_header {
+        if ($0 ~ /{/) {
+          in_function_header = 0
+          in_function = 1
+        }
+        if (!in_function) print
+      }
+      in_function {
+        if (!in_function_header) print
+        if ($0 ~ /{/) open_braces++
+        if ($0 ~ /}/) open_braces--
+        if (open_braces == 0) exit
+      }
+    ' FUNCTION_NAME="$function" "$FLIST" >> "$temp_file"
+    FLIST="$temp_file"
+  fi
+
   for current_file in $FLIST; do
     file="$current_file"
 
@@ -128,7 +160,7 @@ function codestyle_main()
     [[ "$?" != 0 ]] && say "$SEPARATOR"
   done
 
-  if [[ -n "$range_option_flag" ]]; then
+  if [[ -n "$range_option_flag" || -n "$function" ]]; then
     rm "$temp_file"
   fi
 }
@@ -140,7 +172,7 @@ function codestyle_main()
 # In case of successful return 0, otherwise, return 22.
 function parse_codestyle_options()
 {
-  local long_options='verbose,help,start-line:,end-line:'
+  local long_options='verbose,help,start-line:,end-line:,function:'
   local short_options='h'
   local options
 
@@ -157,6 +189,7 @@ function parse_codestyle_options()
   options_values['TEST_MODE']=''
   options_values['START_LINE']=''
   options_values['END_LINE']=''
+  options_values['FUNCTION']=''
 
   eval "set -- $options"
 
@@ -172,6 +205,10 @@ function parse_codestyle_options()
         ;;
       --end-line)
         options_values['END_LINE']="$2"
+        shift 2
+        ;;
+      --function)
+        options_values['FUNCTION']="$2"
         shift 2
         ;;
       TEST_MODE)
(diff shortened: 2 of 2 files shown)


src/bash_autocomplete.sh
```diff
@@ -85,7 +85,7 @@ function _kw_autocomplete()
 
   kw_options['clear-cache']='--verbose'
 
-  kw_options['codestyle']='--verbose --help'
+  kw_options['codestyle']='--verbose --help --start-line --end-line'
```
rodrigosiqueira: You also need to update _kw

src/codestyle.sh
```diff
@@ -73,6 +73,49 @@ function codestyle_main()
...
+    range_option_flag='true'
+  fi
+
+  # Check if --start-line and --end-line options are being used with a single file
+  if [[ ! -f "$path" && -n "$range_option_flag" ]]; then
+    complain "Invalid path using start-line and end-line option: ${path}"
+    return 2 # ENOENT
+  fi
+
+  # When no --start-line is specified, assign the first line of the file to start_line
```
rodrigosiqueira, line 91: All of the comments that you added here should also be added to the Codestyle man page.
Also, consider to expand this tutorial: https://kworkflow.org/tutorials/codestyle.html

src/codestyle.sh
```diff
@@ -73,6 +73,49 @@ function codestyle_main()
   checkpatch=$(join_path "$kernel_root" 'scripts/checkpatch.pl')
   cmd_script="perl ${checkpatch} ${checkpatch_options}"
 
+  [[ -n "${options_values['START_LINE']}" ]] && start_line=${options_values['START_LINE']}
+  [[ -n "${options_values['END_LINE']}" ]] && end_line=${options_values['END_LINE']}
+
```
rodrigosiqueira, line 78: All the start_line and end_line manipulation should be encapsulated in its own functions.

src/codestyle.sh
```diff
@@ -73,6 +73,49 @@ function codestyle_main()
...
+  if [[ -n "$start_line" || -n "$end_line" ]]; then
```
rodrigosiqueira, line 81: What if both variables are empty? Don't you need to return?

src/codestyle.sh
```diff
@@ -73,6 +73,49 @@ function codestyle_main()
...
+    range_option_flag='true'
```
rodrigosiqueira, line 82: You should declare this variable as local and also initialize it with an empty string.

src/codestyle.sh
```diff
@@ -73,6 +73,49 @@ function codestyle_main()
...
+  fi
+
+  # When no --start-line is specified, assign the first line of the file to start_line
+  if [[ -n "$range_option_flag" && -z "$start_line" ]]; then
+    start_line=1
+  fi
+
+  # When no --end-line is specified, assign the last line of the file to end_line
+  if [[ -n "$range_option_flag" && -z "$end_line" ]]; then
+    end_line=$(wc -l < "$FLIST")
```
rodrigosiqueira, line 98: https://kworkflow.org/content/codingstyle.html#avoid-short-options

rodrigosiqueira: Are you sure this is correct? Iirc, FLIST can be a list of c/h files.

src/codestyle.sh
```diff
@@ -73,6 +73,49 @@ function codestyle_main()
...
+  fi
+
+  # If --start-line and --end-line options are being used, assign a temporary file
+  # to FLIST with the line interval and SPDX License Identifier
+  if [[ -n "$range_option_flag" ]]; then
+    suffix=".${FLIST#*.}"
+    temp_file=$(mktemp --suffix "$suffix" --tmpdir="$PWD")
```
rodrigosiqueira, line 111: temp_file is too generic; try to use a more meaningful name.
Also, don't forget to declare it as local.

src/codestyle.sh
```diff
@@ -73,6 +73,49 @@ function codestyle_main()
...
+
+  # When no --end-line is specified, assign the last line of the file to end_line
+  if [[ -n "$range_option_flag" && -z "$end_line" ]]; then
+    end_line=$(wc -l < "$FLIST")
+  fi
+
+  # If --start-line and --end-line options are being used, assign a temporary file
+  # to FLIST with the line interval and SPDX License Identifier
+  if [[ -n "$range_option_flag" ]]; then
+    suffix=".${FLIST#*.}"
```
rodrigosiqueira, line 110: Don't forget to declare it as local.

src/codestyle.sh
```diff
@@ -73,6 +73,49 @@ function codestyle_main()
...
+    temp_file=$(mktemp --suffix "$suffix" --tmpdir="$PWD")
+
+    if [[ "$start_line" != '1' ]]; then
+      printf '%s\n' '// SPDX-License-Identifier: TEMPFILE' > "$temp_file"
```
rodrigosiqueira, line 114: Why do we need this?

src/codestyle.sh
```diff
@@ -73,6 +73,49 @@ function codestyle_main()
...
+      ((blank_lines = start_line - 2))
```
rodrigosiqueira, line 115: Is this because of the SPDX line?

src/codestyle.sh
```diff
@@ -73,6 +73,49 @@ function codestyle_main()
...
+      yes '' | head -n "$blank_lines" >> "$temp_file"
```
rodrigosiqueira, line 116: https://kworkflow.org/content/codingstyle.html#avoid-short-options

src/codestyle.sh
```diff
@@ -84,6 +127,10 @@ function codestyle_main()
     cmd_manager "$flag" "$cmd_script $file"
     [[ "$?" != 0 ]] && say "$SEPARATOR"
   done
+
+  if [[ -n "$range_option_flag" ]]; then
+    rm "$temp_file"
```
rodrigosiqueira, line 164: Never use `rm` with a bash variable without checking if the variable has the correct value. Check `is_safe_path_to_remove`.

src/bash_autocomplete.sh
rodrigosiqueira, line 1: Split your implementation into small functions and also add tests.

Also, consider elaborating more about the temporary file idea in your commit message.

src/codestyle.sh
```diff
@@ -76,12 +76,14 @@ function codestyle_main()
   [[ -n "${options_values['START_LINE']}" ]] && start_line=${options_values['START_LINE']}
   [[ -n "${options_values['END_LINE']}" ]] && end_line=${options_values['END_LINE']}
 
+  [[ -n "${options_values['FUNCTION']}" ]] && function=${options_values['FUNCTION']}
```
rodrigosiqueira, line 79: FUNCTION sounds too generic; try something like TARGET_FUNCTION_FROM_USER or just TARGET_FUNCTION.

src/bash_autocomplete.sh
rodrigosiqueira, line 1: The code you have in this commit is easily testable and requires heavy validation to ensure it works generically.

src/codestyle.sh
```diff
@@ -96,9 +98,15 @@ function codestyle_main()
...
   fi
 
-  # If --start-line and --end-line options are being used, assign a temporary file
-  # to FLIST with the line interval and SPDX License Identifier
-  if [[ -n "$range_option_flag" ]]; then
+  # For --function option, find the start line of the function definition
+  # and assign it to start_line.
+  if [[ -n "$function" ]]; then
+    start_line=$(awk "/[a-zA-Z]+[[:space:]]+${function}[[:space:]]*\\(.*[^;]$/ {print FNR}" "$FLIST")
+  fi
```
rodrigosiqueira, line 105: Check the `src/lib/kw_string` to see if we already have something like that; if not, this is a great candidate to become a generic function in the `kw_string` file.

src/codestyle.sh
```diff
@@ -116,6 +124,30 @@ function codestyle_main()
     FLIST="$temp_file"
   fi
 
+  # For --function option, write the function definition to the temporary file
+  # and assign it to FLIST
+  if [[ -n "$function" ]]; then
```
rodrigosiqueira, line 129: This should have its own function.

src/codestyle.sh
```diff
@@ -116,6 +124,30 @@ function codestyle_main()
...
+    awk '
```
rodrigosiqueira, line 130: A good documentation for this part in the code would be very welcoming.

src/codestyle.sh
```diff
@@ -204,7 +241,8 @@ function codestyle_help()
     '  codestyle [<dir>|<file>|<patch>] - Use checkpatch on target' \
     '  codestyle (--verbose) [<dir>|<file>|<patch>] - Show detailed output' \
     '  codestyle (--start-line <line>) - Set line where the script will start checkpatch' \
-    '  codestyle (--end-line <line>) - Set line where the script will end checkpatch'
+    '  codestyle (--end-line <line>) - Set line where the script will end checkpatch' \
+    '  codestyle (--function <function-name>) - Define function to apply checkpatch'
```
rodrigosiqueira, line 245: Again, you need to update the kw man documentation.

rodrigosiqueira (changes requested):
Thanks for your PR, I added some comments.
Please, invest some time reworking your code and also adding unit tests.

src/codestyle.sh
gustavofukunaga: These variables only get a value when `--start-line` or `--end-line` are being used, so if both variables are empty it means that neither options are being used, returning in this case would affect the other uses of the command, when `--start-line` and `--end-line` aren't utilized.

But this comment made me think of when the option is used but no value is specified like:
```
kw c --start-line {path_to_file}
```
In this case the `start_line` variable would receive `{path_to_file}` as the value, which is not desirable. I think checking if the variables are numbers would fix this issue.

src/codestyle.sh
gustavofukunaga: From what I understand FLIST is a a list of c/h files only when the command receives a directory as an argument.  Before this section, we check if FLIST is a single file with:

```bash
# Check if --start-line and --end-line options are being used with a single file
  if [[ ! -f "$path" && -n "$range_option_flag" ]]; then
    complain "Invalid path using start-line and end-line option: ${path}"
    return 2 # ENOENT
  fi
```

src/codestyle.sh
gustavofukunaga: Every time we would run the command with --start-line greater than 1, checkpatch would give us the warning:

```
WARNING: Missing or malformed SPDX-License-Identifier tag in line 1
```

Even if the original file didn't have this issue. That's because a SPDX-License-Identifier is required in the first line of all files, if --start-line is greater than 1, the temporary file with the code snippet where we run checkpatch would not have this identifier, showing us this warning. 

To solve this issue we add this SPDX-License-Identifier to the first line of the temporary file every time --start-line is greater than 1. Even if the original file had this warning, it wouldn't make sense to show it in this case, because it's a warning that only happens in the first line.

src/codestyle.sh
gustavofukunaga: Yes, the blank lines are added to the temporary file so that checkpatch would show the issue in the right line of the original file. The `start_line - 2` is because e.g. if start_line = 5, it would be necessary to add 4 lines before start_line in the temporary file so that start_line is actually the 5th line in the temp file, and also we have to add the SPDX line as the first line, so we would start the temp file with the SPDX line and 3 blank lines.

rodrigosiqueira:
@gustavofukunaga any update here?

gustavofukunaga:
I'm sorry we're still working on the changes, when we finish everything I'll update the PR branch.

rodrigosiqueira:
Ok, I'll close this PR for now. Feel free to open a new PR when you have a new version.

PR #1147, opened 2024-06-21
Enable kw mail file and directory sending

JGBSouza:


src/lib/kwlib.sh
```diff
@@ -291,6 +291,33 @@ function is_a_patch()
...
+  fi
+
+  file_content=$(< "$FILE_PATH")
+
+  local -ar COVER_LETTER_EXPECTED_STRINGS=(
+    'Subject:'
+    'To:'
+    'From:'
+    'Date:'
+  )
```
rodrigosiqueira: Are you sure those strings are enough to determine whether it is a cover letter? Those are strings available in all patches generated by git. I think you need some specific heuristics, such as the patch number (it is usually 0) or another thing.
@davidbtadokoro do you know if there is a clear pattern for the cover letter? I'm asking you because you worked a lot in the lore part.

src/lib/kwlib.sh
```diff
@@ -291,6 +291,33 @@ function is_a_patch()
...
+
+  for expected_str in "${COVER_LETTER_EXPECTED_STRINGS[@]}"; do
+    if [[ ! "$file_content" =~ $expected_str ]]; then
+      return 1
```
rodrigosiqueira, line 307: maybe: ENOKEY 126 Required key not available

tests/unit/lib/kwlib_test.sh
```diff
@@ -22,6 +22,11 @@ function setupPatch()
   cp -f tests/unit/samples/test.patch "$SHUNIT_TMPDIR"
 }
 
+function setupCoverLetter()
+{
+  cp -f tests/unit/samples/patch_files/0000-cover-letter.patch "$SHUNIT_TMPDIR"
```
rodrigosiqueira, line 27: - Remember: https://kworkflow.org/content/codingstyle.html#avoid-short-options
- Add quotes around the path.
- I think you did not add this sample file.

tests/unit/lib/kwlib_test.sh
```diff
@@ -418,6 +423,14 @@ function test_is_a_patch()
   true # Reset return value
 }
 
+function test_is_a_cover_letter()
+{
+  setupCoverLetter
+  is_a_cover_letter "$SHUNIT_TMPDIR/0000-cover-letter.patch"
+  [[ "$?" != 0 ]] && fail "Failed to check if a file is a patch."
+  true # Reset return value
```
rodrigosiqueira, line 431: why not use something like `assert_equal`?

src/lib/kwlib.sh
```diff
@@ -265,30 +265,23 @@ function get_kernel_version()
...
-  local file_content
+  local cmd='git apply --check '
+  local cmd_return
 
   if [[ ! -f "$FILE_PATH" ]]; then
     return 1
   fi
 
-  file_content=$(< "$FILE_PATH")
+  cmd+=" ${FILE_PATH} > /dev/null 2>&1"
```
rodrigosiqueira, line 275: Can we just use `--quiet` instead of sending everything to /dev/null?

src/lib/kwlib.sh
```diff
@@ -265,30 +265,23 @@ function get_kernel_version()
...
+  cmd_return="$?"
 
-  for expected_str in "${PATCH_EXPECTED_STRINGS[@]}"; do
-    if [[ ! "$file_content" =~ $expected_str ]]; then
-      return 1
-    fi
-  done
+  if [[ "$cmd_return" == 0 || "$cmd_return" == 1 ]]; then
+    return 0
+  fi
```
rodrigosiqueira, line 282: I'm confused here, 0 should be also considered? Also, update this function documentation to make it easier to follow why you return 0 and 1.

src/mail.sh
rodrigosiqueira, line 1: Notice that this commit has the message: join with firest commit.

src/mail.sh
```diff
@@ -340,16 +349,54 @@ function find_commit_references()
       fi
     done <<< "$parsed"
     parsed=''
+
   done <<< "$(git rev-parse -- $args 2> /dev/null)"
 
-  if [[ -n "$commit_range" ]]; then
+  if ! validate_patch_dir_reference "$patch_dir_references"; then
```
rodrigosiqueira, line 355: Move this function call from outside the if, and just check `?`. Finally, use the `[[ ]]` pattern.

src/mail.sh
```diff
@@ -340,16 +349,54 @@ function find_commit_references()
...
+
+    printf '%s' "$(str_strip "$patch_dir_references")"
+
     return 0
   fi
 
   return 22 # EINVAL
 }
 
+# This function checks if the patches files or directorys given as references
```
rodrigosiqueira, line 372: /directororys/directories/

tests/unit/mail_test.sh
```diff
@@ -22,6 +25,9 @@ function oneTimeSetUp()
   touch "$KW_ETC_DIR/mail_templates/test2"
   printf '%s\n' 'sendemail.smtpserver=smtp.test2.com' > "$KW_ETC_DIR/mail_templates/test2"
 
+  mkdir "${FAKE_GIT}/${FAKE_PATCHES_DIRS[0]}"
+  cp --recursive tests/unit/samples/patch_files/* "${FAKE_GIT}/${FAKE_PATCHES_DIRS[0]}"
```
rodrigosiqueira, line 29: Add quotes around the path.

tests/unit/samples/patch_files/0000-cover-letter.patch
rodrigosiqueira, line 1: I think this file should be in the first commit.

rodrigosiqueira (changes requested):

rodrigosiqueira:
I'm going to close this PR; we might get back to it at some point.

PR #1143, opened 2024-06-17
tests: integration: build_test: add the kw build test

aquilamacedo:
This PR Introduce integration tests for the kw build feature with randomized
distribution selection. The tests select a random distribution from Debian,
Archlinux, and Fedora, optimizing resource usage by running tests on a single
selected distro.

PR #1148, opened 2024-06-24
run_tests: streamline test execution logic

aquilamacedo:
Refactored the run_tests script to default to unit tests and improved variable naming (TESTS_DIR, RUN_INTEGRATION, RUN_UNIT). Adjusted script logic for clarity and maintainability.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -6,9 +6,9 @@ include './tests/integration/utils.sh'
 include './src/lib/kwio.sh'
 
 declare -a TESTS
-declare TESTS_DIR='./tests'
-declare TESTS_UNIT=1
-declare TESTS_INTEGRATION=1
+declare TESTS_DIR=''
+declare CLEAR_INTEGRATION_CACHE=0
+declare CLEAR_UNIT_CACHE=0
 declare VERBOSE=0
 
 function show_help()
@@ -139,7 +139,9 @@ function set_tests()
   local file
   TESTS=()
   for file in "$@"; do
-    TESTS+=("${file}")
+    if [[ "$file" == *'_test.sh' ]]; then
+      TESTS+=("$file")
+    fi
   done
 }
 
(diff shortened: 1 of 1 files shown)


run_tests.sh
```diff
@@ -232,11 +236,17 @@ case "$action" in
     list_tests
     ;;
   test)
+    RUN_UNIT=1
+    run_user_provided_tests "$@"
+    ;;
+  test-integration)
+    TESTS_DIR='./tests/integration'
+    RUN_INTEGRATION=1
```
rodrigosiqueira, line 244: Why do we need this new option?
I'm able to run a single integration test with `./run_test --integration test <test_name>`

rodrigosiqueira (changes requested):

run_tests.sh
aquilamacedo: That's true. As we discussed in the call, I hadn't noticed this option. I updated the PR with some improvements.

aquilamacedo:
Hi @rodrigosiqueira. Could you review this PR? it is important to continue the build PR (#1143 ).

run_tests.sh
```diff
@@ -29,7 +29,8 @@ function show_help()
     '  clear-cache - clears tests cache' \
     '  help - displays this help message' \
     '  list - lists all test files under tests/' \
-    '  test - runs the given test'
+    '  test - runs the given test' \
+    '  test-integration - runs only the integration tests for the given test'
```
davidbtadokoro, line 33: I think this change is outdated with the last update

davidbtadokoro (changes requested):
Hey @aquilamacedo, and thank so much for this great refinement.

I tested it, and it works as intended. Just left a minor comment. It seems to also make sense conceptually, and it doesn't conflict with the GitHub Actions jobs. 

Just one thing: do you think it is valid to have a `--all` flag? @rodrigosiqueira what are your thoughts on this?

run_tests.sh
```diff
@@ -219,11 +221,13 @@ while [[ "$1" =~ ^- && ! "$1" == '--' ]]; do
...
 done
 
-if [[ "$1" == '--' ]]; then shift; fi
-
 action=${1:-all}
 shift
 
+if [[ "$RUN_UNIT" -eq 0 && "$RUN_INTEGRATION" -eq 0 ]]; then
+  RUN_UNIT=1
+fi
```
davidbtadokoro, line 229: Either remove this if or also set `TESTS_DIR='./tests/unit'`.

run_tests.sh
```diff
@@ -6,9 +6,9 @@ include './tests/integration/utils.sh'
 include './src/lib/kwio.sh'
 
 declare -a TESTS
-declare TESTS_DIR='./tests'
-declare TESTS_UNIT=1
-declare TESTS_INTEGRATION=1
+declare TESTS_DIR='./tests/unit'
+declare RUN_INTEGRATION=0
+declare RUN_UNIT=0
```
davidbtadokoro, line 11: As these booleans are only used for the clear-cache, maybe rename these two.

aquilamacedo:
Hi @davidbtadokoro and @rodrigosiqueira.

I noticed that when executing ./run_tests.sh --all, the script always starts with the integration tests due to the alphabetical order of the directories. Since "integration" comes before "unit", the integration tests run first.

However, I'm wondering if you see any issue with running the integration tests first. The reason is that if someone wants to see all the tests executed, they will eventually have to wait for the integration tests anyway.

Any thoughts?

rodrigosiqueira:
> Hi @davidbtadokoro and @rodrigosiqueira.
> 
> I noticed that when executing `./run_tests.sh --all`, the script always starts with the integration tests due to the alphabetical order of the directories. Since "integration" comes before "unit", the integration tests run first.
> 
> However, I'm wondering if you see any issue with running the integration tests first. The reason is that if someone wants to see all the tests executed, they will eventually have to wait for the integration tests anyway.
> 
> Any thoughts?

I'm not worried about that, it is the `--all` option.

rodrigosiqueira:
Thanks for your PR, change merged into the unstable branch.

PR #1155, opened 2024-08-07
src: patch-hub: Use new `patch-hub` implementation

davidbtadokoro:
[What]

Make `kw patch-hub` use the new `patch-hub` implementation, which is
being developed as a standalone sub-project of the KWorfkflow
organization [1]. The `patch-hub` executable is installed by running
`curl` to an endpoint to the latest release.

Also, add instructions to users on installing the new `patch-hub`
executable when running `setup.sh --install` fails to install the latest
release or when invoking `kw patch-hub` without having the executable
accessible in the system `PATH`.

Finally, make adaptations to the manpage.

[Why]

With the growth of the `patch-hub` implementation inside the `kworkflow`
project, it became increasingly apparent that it basically reached its
ceiling. More notably, these were the main issues we were facing:

  1. Even though we dedicated much time to parallelizing the fetching of
     the latest patchsets from target mailing lists, the performance was
     (being generous) unacceptable half of the time. This stemmed from
     the interpreted nature of `Bash`, which was aggravated by the
     gigantic number of context changes by the heavy use of `sed`,
     `grep`, and other building blocks for processing.
  2. `Dialog` is an excellent tool to bootstrap a TUI, but, on the other
     hand, its simplicity and "black-box essence" made it absurdly hard
     (practically impossible) to implement some of the not-so-complex
     TUI elements we needed.
  3. Difficulty in testing the TUI components, lowering the feature
     robustness and making it unreliable.

Although we felt that it reached its ceiling, we were nowhere close to
the functionalities and overall look and feel we aimed for. So,
I decided to completely redo the feature as a dedicated sub-project
using Rust. The new `patch-hub` is promising, as its current release
easily surpasses the old one in all the points mentioned above and
doesn't show signs that it will be so complex that we will hit another
ceiling in the future. Besides, it already progressed farther in terms
of functionalities with the action to reply a patchset with the
`Reviewed-by` tag.

[1]: https://github.com/kworkflow/patch-hub


rodrigosiqueira (changes requested):

rodrigosiqueira:
Change merged into the unstable branch! Thanks a lot.

PR #1158, opened 2024-08-15
Treat db returns

JGBSouza:
The return values in the database functions did not differentiate the errors between an empty argument and an execution error in the command execution, making it harder to deal with these errors properly. so i updated the values of the missing arguments to 61 # ENODATA. Also i took advantage of this commit to move the kw_db_test to the /tests/unit/lib directory wich i think that is the right one as the kw_db is also in the src/lib directory.


rodrigosiqueira:
Very nice changes! Changes merged into the unstable branch.

PR #1160, opened 2024-08-16
tests: unit: mail_test: rename mail_test to send_patch_test

JGBSouza:
The previous mail feature was renamed to `send_patch`, all files that refer to it should be renamed to, althought, this was not the case for it's file test, wich remain named as mail_test. In this sense, rename the file `mail_test` to `send_patch_test` in order to fix it.


rodrigosiqueira:

Issue #1165, opened 2024-12-12
Add fish support


superm1:
**Description:**

Currently there is bash completion support, but the shell I use (fish) doesn't seem to have support.  Would you consider adding it?

**Type:**

Enhancement

**Why:**

It's becoming a popular shell; particular for interactive use cases.  Some distros even install it by default (Like CachyOS).

PR #1166, opened 2024-12-17
Add fish support

superm1:
This reverts some of 8ea478a ("fish: Remove fish support") but fixes the unit tests and adds coverage for all remaining commands

Fixes: https://github.com/kworkflow/kworkflow/issues/1165

### DIFF
--- a/tests/unit/fish_completion_test.sh
+++ b/tests/unit/fish_completion_test.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+. ./tests/unit/utils.sh --source-only
+
+function suite()
+{
+  suite_addTest "testKwCompletion"
+}
+
+function getSortedCompletions()
+{
+  local env_path=".:/bin"
+  local env_fish_complete_path=""
+  local completion="./src/kw.fish"
+  local cmd=$1
+  local fish_cmd="set PATH $env_path;
+                    set fish_complete_path $env_fish_complete_path;
+                    source $completion;
+                    complete -C'$1 '"
+
+  fish -c "$fish_cmd" | awk '{print $1}' | sort
+}
+
+function testKwCompletion()
+{
+  local kw_options="init build b deploy d bd diff df ssh s self-update u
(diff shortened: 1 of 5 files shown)


tests/unit/fish_completion_test.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+. ./tests/unit/utils.sh --source-only
```
rodrigosiqueira, line 3: We have a helper for that, you can just use:

`include './tests/unit/utils.sh'`

superm1:
(Will close this and open a new one targeting correct branch)

tests/unit/fish_completion_test.sh
```diff
@@ -0,0 +1,41 @@
...
+
+function suite()
+{
+  suite_addTest "testKwCompletion"
```
rodrigosiqueira, line 7: Maybe use a more meaningful name for this test, for example: `testKwCompletionForFish`.

P.s.: I never used this suite_addTest, do we really need it here?

tests/unit/fish_completion_test.sh
```diff
@@ -0,0 +1,41 @@
...
+}
+
+function getSortedCompletions()
```
rodrigosiqueira, line 10: We use snake case in kw.

rodrigosiqueira (commented):
I just publish my review because I noticed that you close the PR when I was reviewing it.

superm1:
> I just publish my review because I noticed that you close the PR when I was reviewing it.

Oh sorry!

I figured I should target right branch before you got to it.  I'll adjust the feedback for the new branch too.

rodrigosiqueira:
> > I just publish my review because I noticed that you close the PR when I was reviewing it.
> 
> Oh sorry!
> 
> I figured I should target right branch before you got to it. I'll adjust the feedback for the new branch too.

np, I'll review the new PR after you update it.

Thanks

PR #1170, opened 2024-12-20
setup: fix PATH variable for fish

superm1:
The PATH variable was treating the kw binary as a directory. Instead it should be using the binpath that it's "in".


rodrigosiqueira:
Hi @superm1 ,
Your change is merged into the unstable branch. Thanks

PR #1184, opened 2025-01-30
lib: remote: fix fetch error due to missing output path

melissawen:
is_ssh_connection_configured erases $output data in get_config_from_proc due to undeclared local variable used in both function. This issue prevented kw config-kernel-manager --fetch from work properly.

Fixes: 68a2b3a ("src: remote: Improve ssh handling in case of failure")
Co-developer-by: Rodrigo Siqueira <siqueirajordao@riseup.net>


rodrigosiqueira:
Thanks a lot for this fix.

PR #1197, opened 2025-02-19
Deploy refactor

rodrigosiqueira:
This commit enables the support for kw to use operations available in the src/lib folder.


melissawen:
Hi,

I tested on steamdeck (remote) and local (amd-laptop) so far so good.
The reboot to custom kernel is working too, however I noticed on steamdeck that instead of acting as grub-reboot, it looks like a grub-set-default, since it boots the same custom kernel all time, instead of the steamOS kernel.
Local deploy only boots once on custom kernel, and from the second reboot, it boots in the first kernel of the list (as expected)

melissawen:
BTW, I didn't experience any regression so far.

melissawen:
Hey, in current unstable branch, there are two issues related permissions on local deploy. As you are working on deploy files, I don't want to handle conflicts: do you mind to add check this possible solution: https://github.com/melissawen/kworkflow/commit/8b7fe8eb39da9f8b51f5b7d1183993ca54491a02 and merge to your PR?

PR #1204, opened 2025-03-15
setup: port kw for MacOS

davidbtadokoro:
Some kernel developers use a MacOS host as a development environment to leverage the Apple Silicon ARM chips to build for Raspberry Pi, for example. kw can support this context as it is entirely written in Bash, and we mostly need to map dependencies from Homebrew [1].

To support MacOS, change the `setup.sh` script to detect a MacOS system and install dependencies using Homebrew. In `setup.sh`, the adaptation needed was only to detect MacOS using `uname --kernel-name` and direct the correct `brew` commands. The list of base dependencies (at `documentation/dependencies`) is minimal to packages that seem core, while the build dependencies (at
`documentation/dependencies/kernel_build`) are empty.

[NOTE] I tested this in an "old" 2016 MacBookPro and I am unable (for some reason) to update to a MacOS version supported by Homebrew, which adverts that it is expected that "formulae will fail to build" and that "Homebrew will be slow and buggy". Due to this, I couldn't install most packages and essentially tested if the `setup.sh` script was reliable in detecting MacOS and running the correct `brew` commands.

[1]: https://brew.sh/


rodrigosiqueira:
@mairacanal iirc, you use Mac, right? Could you help us improve this PR to enable kw to work on MacOS systems?

mairacanal:
> @mairacanal iirc, you use Mac, right? Could you help us improve this PR to enable kw to work on MacOS systems?

I've been dreaming about this PR for a while 😅

Unfortunately, things didn't work out as expected here as you can see in the [1]. I'm using an M3 Pro with MacOS Sequoia and ran kw in a case-sensitive volume to avoid issues. From what I could understand, it didn't identify my OS as MacOS.

[1] [setup.txt](https://github.com/user-attachments/files/19274017/setup.txt)

With this change, I managed to get it working:

```diff
diff --git a/setup.sh b/setup.sh
index badf834..5091d2c 100755
--- a/setup.sh
+++ b/setup.sh
@@ -126,11 +126,11 @@ function check_dependencies()
   local distro
   local ret

-  distro=$(detect_distro '/')
-
   # If distro is equal to 'none', it can still be MacOS
-  if [[ "$distro" == 'none' && "$(uname --kernel-name)" == 'Darwin' ]]; then
+  if [[ "$(uname --kernel-name)" == 'Darwin' ]]; then
     distro='macos'
+  else
+    distro=$(detect_distro '/')
   fi

   package_list=$(get_missing_packages "$distro" "${DOCUMENTATION}/dependencies/${distro}.dependencies")
diff --git a/src/lib/kwlib.sh b/src/lib/kwlib.sh
index 1112e2f..f67ccc6 100644
--- a/src/lib/kwlib.sh
+++ b/src/lib/kwlib.sh
@@ -339,11 +339,6 @@ function detect_distro()
   local os_release_process
   declare -a os_family=('debian' 'arch' 'fedora')

-  # If the system is MacOS, don't bother to do the other checks
-  if [[ "$(uname --kernel-name)" == 'Darwin' ]]; then
-    printf '%s\n' 'none'
-  fi
-
   etc_path=$(join_path "$root_path" '/etc')

   if [[ -d "$etc_path" && -z "$str_check" && -z "$raw_os_release" ]]; then
```

But, I still get this error a couple of times:

```
sed: can't read /Users/mairacanal/.bashrc: No such file or directory
```

I'm using a zsh shell.

---

This is just the first step in porting kw to MacOS as there are some differences in the MacOS and Linux bash. They are mostly compatible, but a few things differ. For example, when running `kw init` in a Linux repository (also in the case-sensitive partition), I get:

```
  kw init
getopt: illegal option -- o
getopt: illegal option -- p
getopt: illegal option -- t
getopt: illegal option -- i
getopt: illegal option -- o
getopt: illegal option -- n
getopt: illegal option -- s
```

---

About the build dependencies, I recommend reading about Bee Headers [2]. I'm using them to build the kernel in my machine and it works great.

[2] https://github.com/bee-headers/homebrew-bee-headers

rodrigosiqueira:
@davidbtadokoro Do you have time to apply @mairacanal suggestions? If not, and if it is ok for you, I can continue this effort.

davidbtadokoro:
Hi, @mairacanal and @rodrigosiqueira. I am deeply sorry for not replying to you and have no excuses other than letting this PR slip through my organization. Even though I couldn't continue the effort, I should have replied to you and not abandoned it, so I apologize for both...

Thank you so much, Maira, for testing the PR and detecting (and fixing some of) its issues! The detection part of macOS seems to be working, and the necessary dependencies, including those for building, appear to be available in Homebrew. FWIU, the bash in macOS vs the bash in Linux differ in version (in macOS, it is outdated), and because kw uses newer bash features, adding it as a dependency for Homebrew to update should do it.

However, as `getopt`, differently from `getopts` (with an 's'), is not in-built, the error you've found may indicate that some packages can have different APIs for macOS. It may also be that the version installed on your setup could be outdated because I didn't add `getopt` as a dependency.

In any case, I think that the effort now will be to find these incompatible packages and hope that they are reasonably solvable, like in the case of `getopt`, in which we can try using the "built-in version" :sweat_smile: 

Beyond that, I don't see any other setbacks in the port.

Once again, sorry guys for the inexcusable delay and thanks for the attention.

PR #1205, opened 2025-03-30
plugins/drm: fix syntax error in local gui_control

melissawen:
There is a syntax error when running `kw drm --gui-(on/off)` for local target:

$ kw drm --gui-off --verbose --local
sudo systemctl isolate multi-user.target
sudo for i in /sys/class/vtconsole/*/bind; do printf "%s\n" 0 > $i; done; sleep 0.5 /.local/lib/kw/lib/kwlib.sh: eval: line 122: syntax error near unexpected token `do' /.local/lib/kw/lib/kwlib.sh: eval: line 122: `sudo for i in /sys/class/vtconsole/*/bind; do printf "%s\n" 0 > $i; done; sleep 0.5'

The cause is that sudo prefix is only applied to the first command in a sequence of multiple commands separated by semicolons. Fix that by using `sudo -- sh -c '$cmd'`, which ensures multiple commands are executed as sudo.

### DIFF
--- a/src/plugins/subsystems/drm/drm.sh
+++ b/src/plugins/subsystems/drm/drm.sh
@@ -264,8 +264,8 @@ function gui_control()
 
   case "$target" in
     2) # LOCAL TARGET
-      gui_control_cmd="sudo ${gui_control_cmd}"
-      bind_control_cmd="sudo ${bind_control_cmd}"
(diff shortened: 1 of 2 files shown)


rodrigosiqueira:
Merged! Thanks

PR #1213, opened 2025-05-08
src: kw_env: Enhancement create enviroment message

juliacalixtorosa:
Change the warning message from 'kw env --create' to make it less confusing regarding the config file.

### DIFF
--- a/src/kw_env.sh
+++ b/src/kw_env.sh
@@ -242,7 +242,8 @@ function create_new_env()
     return
   fi
 
-  warning "You don't have a config file, get it from default paths"
+  say "Environment was created without a kernel .config file. Use kw env --use ${env_name} to switch to the new env."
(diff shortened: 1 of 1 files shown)


src/kw_env.sh
```diff
@@ -242,7 +242,7 @@ function create_new_env()
     return
   fi
 
-  warning "You don't have a config file, get it from default paths"
+  warning "Environment was created, use it with kw env --use and don't forget to set the .config file."
```
rodrigosiqueira, line 245: Maybe something like this:

```sh
say "Environment was created without a kernel .config file. Use kw env --use ${env_name} to switch to the new env."
warning "The new env does not have a default .config; you must provide it for a correct kernel compilation. It is recommended to use kw kernel-config-manager."
```

juliacalixtorosa: We implemented it according to your suggestion. Thank you.

rodrigosiqueira:
Thanks, the change merged with modifications into the unstable branch.

PR #1215, opened 2025-05-14
src: build_and_deploy: make kw bd accept kw d arguments

nattan447:
Previously, to build and deploy the kernel, users had to use kw bd, but it did not accept the same arguments as kw d, which was inconsistent and unintuitive.

this refactor ensures that kw bd now accepts the same arguments as kw d, providing a more coherent and flexible interface for users.

### DIFF
--- a/src/build_and_deploy.sh
+++ b/src/build_and_deploy.sh
@@ -18,13 +18,6 @@ function build_and_deploy_main()
     exit 0
   fi
 
-  parse_build_and_deploy_options "$@"
-  if [[ "$?" -gt 0 ]]; then
-    complain "${options_values['ERROR']}"
-    build_and_deploy_help
-    return 22 # EINVAL
-  fi
-
   build_kernel_main
   if [[ "$?" != 0 ]]; then
     complain 'kernel build failed\n'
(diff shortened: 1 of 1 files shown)


rodrigosiqueira (changes requested):
Hi @nattan447,

Thanks for addressing this issue.

Actually, fixing this issue is a little bit harder in the sense that not all of the deploy and build options should be valid here since this is "build & deploy", i.e., a particular sequence. I checked the deploy options, and the following should not be part of `bd`:

- `--setup`
- `--uninstall`
- `--list`
- `--ls-line`
- `--list-all`

From build, I think the following should not be part of `bd`:
- `--menu`
- `--info`
- `--doc`
- `--clean`
- `--full-cleanup`
- `--from-sha` --> We need to think more about this one.

nattan447:
> Hi @nattan447,
> 
> Thanks for addressing this issue.
> 
> Actually, fixing this issue is a little bit harder in the sense that not all of the deploy and build options should be valid here since this is "build & deploy", i.e., a particular sequence. I checked the deploy options, and the following should not be part of `bd`:
> 
> * `--setup`
> * `--uninstall`
> * `--list`
> * `--ls-line`
> * `--list-all`
> 
> From build, I think the following should not be part of `bd`:
> 
> * `--menu`
> * `--info`
> * `--doc`
> * `--clean`
> * `--full-cleanup`
> * `--from-sha` --> We need to think more about this one.

nattan447:
> > Hi @nattan447,
> > Thanks for addressing this issue.
> > Actually, fixing this issue is a little bit harder in the sense that not all of the deploy and build options should be valid here since this is "build & deploy", i.e., a particular sequence. I checked the deploy options, and the following should not be part of `bd`:
> > 
> > * `--setup`
> > * `--uninstall`
> > * `--list`
> > * `--ls-line`
> > * `--list-all`
> > 
> > From build, I think the following should not be part of `bd`:
> > 
> > * `--menu`
> > * `--info`
> > * `--doc`
> > * `--clean`
> > * `--full-cleanup`
> > * `--from-sha` --> We need to think more about this one.

Therefore, would it make more sense for me to open a pull request that filters the options related to deploy first? I can create a pull request with these changes, and then open a separate one to filter the build options.

Issue #1185, opened 2025-02-05
Using the same name for environments in different work trees mixes settings and overwrites compiled files.


melissawen:
**Description:**

If an user creates two environments with the same name, but in different repositories and therefore, different config files, compiled files and kw settings, the latest environment used replaces everything from the previous one.

**How to Reproduce:**

1. Clone two different kernel repositories
2. Create a kw env with the same name, for example, LOCAL
3. Setup and compile file in repository 1 under LOCAL env.
4. In another tab, use the LOCAL env in repository 2. it will use everything from LOCAL env in repository 1.

**Expected behavior:**

Isolate environments by name and repositories.

**System information:**

 - OS: Debian testing
 - kw Version: unstable
 - 
**Additional context:**

The problem is the way in which kw cached environment files, only by name. it should create a subfolder or sufix according to repository, or does not allow to create environments with the same name in different repositories.

PR #1223, opened 2025-06-09
Add a unique suffix to kw environments

geraldino2:
kw currently stores environment data in `${XDG_CACHE_HOME:-"${HOME}/.cache"}/${KWORKFLOW}/${ENV_DIR}/${env_name}`, and allows users to create non-distinct environment names in different work trees. When the new environment name isn't unique, data gets overwritten.

This PR implements a suffix to `env_name`: `$PWD`, encoded using base64 (available in Arch/Debian/Fedora through `coreutils`, already in `*.dependencies`).

It should fix issue #1185.

### DIFF
--- a/src/kw_env.sh
+++ b/src/kw_env.sh
@@ -194,6 +194,7 @@ function create_new_env()
   local local_kw_build_config="${local_kw_configs}/build.config"
   local local_kw_deploy_config="${local_kw_configs}/deploy.config"
   local env_name=${options_values['CREATE']}
+  local encoded_pwd=$(get_encoded_pwd)
   local cache_build_path="$KW_CACHE_DIR"
   local current_env_name
   local output
@@ -237,18 +238,18 @@ function create_new_env()
   done
 
   # Handle build and config folder
-  cmd="mkdir -p ${cache_build_path}/${ENV_DIR}/${env_name}"
+  cmd="mkdir --parents ${cache_build_path}/${ENV_DIR}/${encoded_pwd}/${env_name}"
   cmd_manager "$flag" "$cmd"
 
   current_env_name=$(get_current_env_name)
   ret="$?"
   # If we already have an env, we should copy the config file from it.
   if [[ "$ret" == 0 ]]; then
-    cmd="cp ${cache_build_path}/${ENV_DIR}/${current_env_name}/.config ${cache_build_path}/${ENV_DIR}/${env_name}/.config"
+    cmd="cp ${cache_build_path}/${ENV_DIR}/${encoded_pwd}/${current_env_name}/.config ${cache_build_path}/${ENV_DIR}/${encoded_pwd}/${env_name}/.config"
     cmd_manager "$flag" "$cmd"
     return
   elif [[ -f "${PWD}/.config" ]]; then
-    cmd="cp ${PWD}/.config ${cache_build_path}/${ENV_DIR}/${env_name}"
+    cmd="cp ${PWD}/.config ${cache_build_path}/${ENV_DIR}/${encoded_pwd}/${env_name}"
     cmd_manager "$flag" "$cmd"
     return
   fi
@@ -258,10 +259,10 @@ function create_new_env()
   warning 'It is recommended to use kw kernel-config-manager.'
 
   if [[ -e /proc/config.gz ]]; then
-    cmd="zcat /proc/config.gz > ${cache_build_path}/${ENV_DIR}/${env_name}/.config"
+    cmd="zcat /proc/config.gz > ${cache_build_path}/${ENV_DIR}/${encoded_pwd}/${env_name}/.config"
     cmd_manager "$flag" "$cmd"
   elif [[ -e "/boot/config-$(uname -r)" ]]; then
-    cmd="cp /boot/config-$(uname -r) ${cache_build_path}/${ENV_DIR}/${env_name}/.config"
+    cmd="cp /boot/config-$(uname -r) ${cache_build_path}/${ENV_DIR}/${encoded_pwd}/${env_name}/.config"
     cmd_manager "$flag" "$cmd"
   else
     warning 'kw was not able to find any valid config file for the new env'
@@ -276,6 +277,7 @@ function destroy_env()
   local cache_build_path="$KW_CACHE_DIR"
   local current_env
   local env_name=${options_values['DESTROY']}
+  local encoded_pwd=$(get_encoded_pwd)
   local cmd
 
   flag=${flag:-'SILENT'}
@@ -302,7 +304,7 @@ function destroy_env()
     fi
   fi
 
-  cmd="rm -rf ${local_kw_configs:?}/${ENV_DIR}/${env_name} && rm -rf ${cache_build_path:?}/${ENV_DIR}/${env_name}"
+  cmd="rm --recursive --force ${local_kw_configs:?}/${ENV_DIR}/${env_name} && rm -rf ${cache_build_path:?}/${ENV_DIR}/${encoded_pwd}/${env_name}"
   cmd_manager "$flag" "$cmd"
   success "The \"${env_name}\" environment has been destroyed."
 }
@@ -344,14 +346,14 @@ function list_env_available_envs()
   if [[ -f "${local_kw_configs}/${ENV_CURRENT_FILE}" ]]; then
     current_env=$(< "${local_kw_configs}/${ENV_CURRENT_FILE}")
     say 'Current env:'
-    printf ' -> %s: %s\n\n' "$current_env" "${KW_CACHE_DIR}/${ENV_DIR}/${current_env}"
+    printf ' -> %s: %s\n\n' "$current_env" "${KW_CACHE_DIR}/${ENV_DIR}/$(get_encoded_pwd)/${current_env}"
   fi
 
   warning 'Other kw environments:'
   # For the below loop, we want to split the array
   # shellcheck disable=SC2068
   for env in ${all_envs[@]}; do
-    printf ' * %s: %s\n' "$env" "${KW_CACHE_DIR}/${ENV_DIR}/${env}"
+    printf ' * %s: %s\n' "$env" "${KW_CACHE_DIR}/${ENV_DIR}/$(get_encoded_pwd)/${env}"
   done
 }
 
--- a/src/kw_env.sh
+++ b/src/kw_env.sh
@@ -19,6 +19,7 @@ declare -gr ENV_CURRENT_FILE='env.current'
 
 function env_main()
 {
+
   parse_env_options "$@"
   if [[ $? -gt 0 ]]; then
     complain "Invalid option: ${options_values['ERROR']}"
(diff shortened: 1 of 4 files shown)


src/lib/kwlib.sh
```diff
@@ -695,6 +695,25 @@ function get_current_env_name()
...
 }
 
+# This function returns "${current_env}-base64(${PWD})".
+#
+# @env_name: The name of the environment to encode
+#
+# Return:
+# Returns 0 and prints ${current_env}-base64(${PWD}) if env_name is provided.
+# Returns 1 if env_name is not provided.
+get_env_name_encoded_with_pwd() {
```
rodrigosiqueira: This is not the codestyle adopted by kw, please, read the below page and review your entire PR.

 https://kworkflow.org/content/codingstyle.html

src/lib/kwlib.sh
```diff
@@ -695,6 +695,25 @@ function get_current_env_name()
...
+  local env_name="$1"
+
+  if [[ -z "$env_name" ]]; then
+    return 1
```
rodrigosiqueira, line 710: Return 22 makes more sense here (EINVAL).

src/lib/kwlib.sh
```diff
@@ -695,6 +695,25 @@ function get_current_env_name()
...
+  fi
+
+  local encoded_pwd=$(printf '%s' "$PWD" | base64)
```
rodrigosiqueira, line 713: move the `local encoded_pwd` declaration to the top of the function.

rodrigosiqueira, line 713: Regarding base64, this seems excessive for this problem.  Additionally, I think base64 can return a multiple-line output, which can be a problem here.  I suppose if you want to use it, you need to add a wrapper to ensure the output is not displayed on multiple lines.

However, a simple md5sum would be better here.

src/lib/kwlib.sh
```diff
@@ -695,6 +695,25 @@ function get_current_env_name()
...
+
+  printf '%s-%s' "$env_name" "$encoded_pwd"
```
rodrigosiqueira, line 715: I don't think you need to add the env_name; you can use the encoded_pwd as a subfolder, and inside it, you keep the env name. In other words, to solve this problem, you just create a new subfolder that is the encoded version of the path.

src/lib/kwlib.sh
rodrigosiqueira, line 1: The unit test should be part of this commit.

src/lib/kwlib.sh
```diff
@@ -714,6 +714,25 @@ get_env_name_encoded_with_pwd() {
...
 
+# Extracts the environment name from a string of the format "env_name-base64(PWD)".
+# Assumes env_name may contain hyphens.
+#
+# @encoded: The encoded string (e.g., "my-env-L2hvbWUvYXBvbG8y")
+#
+# Return:
+# Returns 0 if successful, 1 if input is empty or malformed.
+# Prints the env_name (everything before the last hyphen) if successful.
+decode_env_name_with_pwd() {
```
rodrigosiqueira: from my comments in the first commit, I'm not convinced that we need this function.

src/kw_env.sh
```diff
@@ -109,13 +109,14 @@ function validate_env_before_switch()
...
   local target_env="${options_values['USE']}"
+  local encoded_target_env=$(get_env_name_encoded_with_pwd "$target_env")
   local local_kw_configs="${PWD}/.kw"
   local tmp_trash
   local cmd
 
   flag=${flag:-'SILENT'}
 
-  if [[ ! -d "${local_kw_configs}/${ENV_DIR}/${target_env}" ]]; then
+  if [[ ! -d "${local_kw_configs}/${ENV_DIR}/${encoded_target_env}" ]]; then
```
rodrigosiqueira, line 119: iirc, the original issue reported by Melissa was about issues with the env collision in the cache, not in the local config. In this sense, you should focus on things that touch `KW_CACHE_DIR`, not local `.kw` config files. In other words, you should not change these parts. Am I missing something?

src/kw_env.sh
```diff
@@ -213,42 +217,42 @@ function create_new_env()
...
-  cmd="mkdir -p ${cache_build_path}/${ENV_DIR}/${env_name}"
+  cmd="mkdir -p ${cache_build_path}/${ENV_DIR}/${encoded_env_name}"
   cmd_manager "$flag" "$cmd"
 
   current_env_name=$(get_current_env_name)
   ret="$?"
   # If we already have an env, we should copy the config file from it.
   if [[ "$ret" == 0 ]]; then
-    cmd="cp ${cache_build_path}/${ENV_DIR}/${current_env_name}/.config ${cache_build_path}/${ENV_DIR}/${env_name}/.config"
+    cmd="cp ${cache_build_path}/${ENV_DIR}/${current_env_name}/.config ${cache_build_path}/${ENV_DIR}/${encoded_env_name}/.config"
```
rodrigosiqueira, line 251: As I commented before, it would be better if we just create a new folder with the encoded name (without the env name), and inside it we create the env name. For example, suppose the encoded_pwd is `845d87776ab0bcd0c17f051d264f1cf2`, and the env names are `RPI64` and `x86_64`, we would have something like this:

`${cache_build_path}/${ENV_DIR}/${encoded_pwd}/${current_env_name}/`

tests/unit/kw_env_test.sh
rodrigosiqueira, line 1: unit test should be in the same commit that introduces the tested modifications.

rodrigosiqueira (changes requested):
Hi,
First of all, thanks for your PR. I think you are in the right direction, but you need to improve some things:
1. Organize your commits. Usually, PR reviews happen on a commit basis, which means that we review from the first commit to the last one. Every commit should be self-contained, and tests should be introduced in the same commit as the feature. For example, if you added function X in the first commit, then you should add the test for X in the same commit. Also, if one commit improves something introduced in a previous commit that has not been merged yet, then you should squash your new commit into the previous one. Anyway, please take the time to improve your commits.
2. Keep in mind that envs have two types of metadata: local configs (.kw) and global folders (.cache). For this problem afaik, you just need to focus on the global part.
3. Finally, take some time to read the kw codestyle guideline.

src/lib/kwlib.sh
geraldino2: Nice catch, I wasn't used to this multiple-line output. base64 on MacOS doesn't have it, but coreutils has, indeed, a default line limit of 76 chars. I'll disable line wrapping.

Is there any reason to use md5sum? base64 seems to do well: reversible and unique.

src/lib/kwlib.sh
geraldino2: Good idea.

src/kw_env.sh
geraldino2: You are correct, changes on `.kw` aren't necessary. The goal was to have consistent naming between local/global contexts. However, as you suggested, I'm going to use `base64(path)/env_name` in `KW_CACHE_DIR`, so I'll revert changes in `.kw`.

geraldino2:
Hi, thanks for the feedback. The PR is looking better now. I've adjusted the code to focus only on the global part and organized the commits. About the codestyle, I think the problem was originally addressed in the `chore` commits; I don't see any other issue.

Do you mind taking a look again?

rodrigosiqueira (changes requested):
Hi,

Thanks a lot for your update. I think you are close to finishing it. I added some minor comments, but there is one final issue to handle.

One thing to keep in mind is that we already have users who use the kw env feature. In this sense, you also need to provide one mid-step to migrate the env path to the new one. You can do it in a new commit. To validate what I mean, you can try the following:
1. Outside your branch, create two or three new envs. Ensure that you add a .config file to each one; in other words, create a realistic environment.
2. Move back to your env, and try to create a new env. You will encounter your first error.
My suggestion is:
1. Create a function named `migrate_old_env_to_base64` (or something better). Add a comment in this function to indicate that we can remove it in the future.
2. This function should check the cache folder to see if the expected level of directories is present, i.e., your modification introduced a folder new level, If it is not found, you can assume it is the old approach.
3. Add this function at the beginning of every env function to ensure that the migration happens as soon as possible.

Thanks

geraldino2:
Hi,

I created `migrate_old_envs_to_base64`, it's being called on `env_main`. Unless someone is in the case reported in #1185, migration should go well.

By the way, since I'm already modifying these files, do you want me to add the changes below to the current PR?
- src/kw_env.sh:list_env_available_envs: array is loaded with `find "${local_kw_configs}/${ENV_DIR}" -maxdepth 1` and its size is compared to `0`. If `find` has no `mindepth`, the base path is also printed. It's an easy bug that can be fixed by adding `mindepth 1`. The second `no_env_message` on `list_env_available_envs` is unreachable right now;
- tests/unit/kw_env_test.sh: `$LINENO` is being used instead of `${LINENO}`.

Do you mind taking a look again?

Thanks

rodrigosiqueira:
> Hi,
> 
> I created `migrate_old_envs_to_base64`, it's being called on `env_main`. Unless someone is in the case reported in #1185, migration should go well.
> 
> By the way, since I'm already modifying these files, do you want me to add the changes below to the current PR?
> 
>     * src/kw_env.sh:list_env_available_envs: array is loaded with `find "${local_kw_configs}/${ENV_DIR}" -maxdepth 1` and its size is compared to `0`. If `find` has no `mindepth`, the base path is also printed. It's an easy bug that can be fixed by adding `mindepth 1`. The second `no_env_message` on `list_env_available_envs` is unreachable right now;
> 
>     * tests/unit/kw_env_test.sh: `$LINENO` is being used instead of `${LINENO}`.
> 
> 
> Do you mind taking a look again?
> 
> Thanks

yes, please, I appreciate that. Add all of them in their own commit.

tests/unit/kw_env_test.sh
```diff
@@ -346,4 +346,95 @@ function test_create_and_destroy_env()
...
+  assertTrue "(${LINENO}) env1 config should be preserved" '[[ -f "${new_env_path}/env1/build.config" ]]'
+  assertTrue "(${LINENO}) env2 config should be preserved" '[[ -f "${new_env_path}/env2/build.config" ]]'
+
+  # Verify old structure is removed
+  assertFalse "(${LINENO}) Old env1 should be moved" '[[ -d "${old_env_path}/env1" ]]'
+  assertFalse "(${LINENO}) Old env2 should be moved" '[[ -d "${old_env_path}/env2" ]]'
+
+  # Verify config content is preserved
+  local content1=$(cat "${new_env_path}/env1/build.config")
+  local content2=$(cat "${new_env_path}/env2/build.config")
```
rodrigosiqueira, line 386: https://kworkflow.org/content/codingstyle.html#command-substitution-and-arithmetic-expression

Also, move the local declaration to the beginning of the function.

rodrigosiqueira (changes requested):
Hi,
That is a very nice update! I tested it locally, and it looks like it is working as expected. I added some minor comments; I'm looking forward to the new version.

Thanks a lot.

rodrigosiqueira:
Hi,
I made some changes myself, and merged it into the unstable branch. Thanks.

PR #1231, opened 2025-07-16
documentation: minor fix on notification.alert settings

melissawen:
To change notification settings we should set notification.alert, not kworkflow.alert, update documentation to match it.


rodrigosiqueira:

Issue #1229, opened 2025-07-06
Shouldn't `KBUILD_BUILD_TIMESTAMP=""` be set in the build environment?


superm1:
**Description:**

For `ccache` to work properly build timestamps shouldn't be embedded otherwise it will be a cache miss every time.

I believe that `kw` should be calling something like `export KBUILD_BUILD_TIMESTAMP=""` in the env before building.

**Type:**

[Others]

**Why:**

So that `ccache` can have more cache hits.

PR #1232, opened 2025-08-13
Set KBUILD_BUILD_TIMESTAMP when using ccache

superm1:
Otherwise there will be constant cache misses

Closes: https://github.com/kworkflow/kworkflow/issues/1229


rodrigosiqueira:
Thanks a lot for this patch! Change merged into the unstable branch.

PR #1241, opened 2025-09-12
rpi: copy overlay files for downstream kernels to Rapberry Pi 4 and 5

melissawen:
KW packages overlay folder but don't copy it from the kw_pkg to remote machine. Overlay files are not present in upstream kernel, but it's in downstream kernels. Also the support for RaspberryPi 4 can be extended to Raspberry Pi 5. Rename the template accordingly.


rodrigosiqueira:
Hi @melissawen , thanks for your PR. Merged into the unstable branch.

PR #1243, opened 2025-09-15
src: deploy: fix alert completion for general kernel deploy and list

melissawen:
If the function exits instead of returns the alert_completion is never called in kw main.


rodrigosiqueira:

PR #1244, opened 2025-09-17
src/deploy: correct order of parameters in function documentation

melissawen:
run_kernel_uninstall() takes first force and then flag as parameters, but their positions are swapped in the documentation and can lead to some confusion.


rodrigosiqueira:
Merged, thanks

Issue #1208, opened 2025-04-13
`kw bd` doesn't support `--remote`


superm1:
**Description:**

Using `kw b` followed by `kw d --remote foo` works, but not `kw bd --remote foo`

**How to Reproduce:**

```
❯ kw bd --remote dock
kw bd: unrecognized option '--remote'
kw bd:
  bd - build and deploy kernel and modules:
```
**Expected behavior:**

Works

**System information:**

```
❯ kw --version 
alpha
Branch: rodrigosiqueira-decomposing_deploy_files
Commit: ce1d591
```

rodrigosiqueira:
Thanks for reporting this issue; multiple people complained about deploy options not working with `bd`. I'll look into this issue (I think it is not super hard to fix) as soon as I have the first version to address #576 (hopefully this week).

melissawen:
+1 to make bd accept at least deploy options - IIRC, it was the original behavior and removed for some reason.
Better if it accepts both: kw b and kw d options :)

BTW, I don't think it's a bug, it's just very bad UX :)

rodrigosiqueira:
Fix merged into the unstable branch.

PR #1245, opened 2025-09-20
Make bd accept some of the deploy options

rodrigosiqueira:
This commit addresses the issue #1208 pointed out by @superm1 and @melissawen .
With this change, kw bd accepts the following deploy options: --reboot, --no-reboot, --create-package, --boot-into-new-kernel-once, --verbose, and --remote.


rodrigosiqueira:
Cange merged into the unstable branch,.

PR #1247, opened 2025-10-21
Fix ccache related issues

Carton:
- currently enable_ccache option in build.config not used.
- also fix cross compilation related issues when ccache is enabled.

runned run_tests.sh to verify.

### DIFF
--- a/src/build.sh
+++ b/src/build.sh
@@ -123,7 +123,13 @@ function build_kernel_main()
   optimizations="-j$((parallel_cores * cpu_scaling_factor / 100))"
 
   if [[ -n "${options_values['CCACHE']}" ]]; then
-    [[ -n "$llvm" ]] && compiler='clang' || compiler='gcc'
+    if [[ -n "$cross_compile" ]]; then
+      # Cross-compilation: use cross-compiler with ccache
+      [[ -n "$llvm" ]] && compiler="${cross_compile}clang" || compiler="${cross_compile}gcc"
+    else
+      # Native compilation: use host compiler with ccache
+      [[ -n "$llvm" ]] && compiler='clang' || compiler='gcc'
+    fi
     optimizations="KBUILD_BUILD_TIMESTAMP= CC=\"ccache ${compiler} -fdiagnostics-color\" ${optimizations}"
   fi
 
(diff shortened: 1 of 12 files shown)


src/build.sh
```diff
@@ -335,7 +335,7 @@ function parse_build_options()
   options_values['ARCH']="${build_config[arch]:-$arch_fallback}"
   options_values['MENU_CONFIG']=''
   options_values['CROSS_COMPILE']="${build_config[cross_compile]}"
-  options_values['CCACHE']="${build_config[ccache]}"
+  options_values['CCACHE']="${build_config[enable_ccache]:-false}"
```
rodrigosiqueira: Replace 'false' with 'no'.

src/build.sh
```diff
@@ -355,6 +355,13 @@ function parse_build_options()
     options_values['USE_LLVM_TOOLCHAIN']=''
   fi
 
+  # Check ccache option
+  if [[ ${options_values['CCACHE']} == 'true' ]]; then
```
rodrigosiqueira: Replace 'true' with 'yes'

rodrigosiqueira (changes requested):
First of all, sorry for the delay in reviewing your PR. Thanks a lot for your contribution!

Everything lgtm, except the use of 'true' and 'false'. Most of the config files use 'yes' and 'no', let's keep this pattern.

Thanks again!

Carton:
Hi,
I checked the code, there are many sample configs (not in my change) that have already used "enable_ccache=true" or "enable_ccache=false". But yes there are many more configs that uses "=yes/no".

Do I need to change all these sample configs from "enable_ccache=true/false" to "enable_ccache=yes/no" along with other code changes?

rodrigosiqueira:
> Hi, I checked the code, there are many sample configs (not in my change) that have already used "enable_ccache=true" or "enable_ccache=false". But yes there are many more configs that uses "=yes/no".
>

Yeah, that was a mistake we made a long time ago. Actually, you will find many other inconsistencies in the old code that we need to catch up on at some point.
 
> Do I need to change all these sample configs from "enable_ccache=true/false" to "enable_ccache=yes/no" along with other code changes?

Could you do that? If you don't have time, that's okay; just make the changes I suggested in your PR.

Thanks

Carton:
OK. I'll try to clean up related configs.

rodrigosiqueira:
Hi @Carton Thanks a lot for your contribution!

PR merged into the unstable branch.

Thanks again.

PR #1259, opened 2026-03-21
src: lib: kwlib: Remove '=' from base64

rodrigosiqueira:
When certain environments are created in specific folders, an '=' character is added to the base64 output, which is not part of the base64 output. Unfortunately, leaving this '=' can cause issues when managing folders. This commit removes the '=' character at the end.


melissawen:
Please, merge :)

It fixes a very annoying issue

rodrigosiqueira:
Merged. Thanks

PR #1262, opened 2026-03-29
Rework devices

rodrigosiqueira:
This is a huge rework of the device features. I used an AI tool to review my changes and help with the unit tests in the last commit, so I added the Assisted-by tag.


melissawen:
Hey, Thanks for reworking kw device with inxi.

I'd prefer if you temporary preserve `kw device` command but warning it was renamed. Same issue happened in the past with kw send-patch and I think these interface changes without notice a bit annoying.

I found some issue when testing:
- In a device with kw but without inxi, the command doesn't work. I see the commit that install inxi patch via kw deploy, but could you trigger the same update with `kw get-device-info`?
- `Storage boot partitions` info are never filled. Should it be removed?

PR #1263, opened 2026-05-17
src: lib: kw_db: Create default database on first use

aquilamacedo:
Create the default per-user `kw.db` when the database helpers access the default database and the file is missing.

This fixes packaged installations where setup.sh is not run per user, so commands such as kw build can record statistics without ending with `"Database does not exist"`.

Explicit non-default database paths still report `ENOENT` as before.

### DIFF
--- a/src/lib/kw_db.sh
+++ b/src/lib/kw_db.sh
@@ -32,6 +32,51 @@ function execute_sql_script()
   sqlite3 -init "$KW_DB_DIR/pre_cmd.sql" "$db_path" < "$sql_path"
 }
 
+# This function creates the default user database when it is missing.
+#
+# Return:
+# 0 if the database exists or was created
+# 1 if the database directory could not be created
+# 2 if the database SQL file is missing
+# non-zero from execute_sql_script if database creation fails
+function ensure_default_database_exists()
+{
+  local db_path
+  local sql_path="${KW_DB_DIR}/kwdb.sql"
+
+  db_path="$(join_path "$KW_DATA_DIR" "$DB_NAME")"
+
+  [[ -f "$db_path" ]] && return 0
+
+  if [[ ! -f "$sql_path" ]]; then
+    complain "Could not find the file: ${sql_path}"
+    return 2 # ENOENT
+  fi
+
+  mkdir --parents "$KW_DATA_DIR"
+  if [[ "$?" != 0 ]]; then
+    complain "Could not create database directory: ${KW_DATA_DIR}"
+    return 1 # EPERM
+  fi
+
+  execute_sql_script "$sql_path" "$DB_NAME" "$KW_DATA_DIR" > /dev/null
+}
+
+# This function checks if the given database target is kw's default database.
+#
+# @db:        Name of the database file
+# @db_folder: Path to the folder that contains @db
+#
+# Return:
+# 0 if the target is the default database; non-zero otherwise
+function is_default_database_target()
+{
+  local db="$1"
+  local db_folder="$2"
+
+  [[ "$db" == "$DB_NAME" && "$db_folder" == "$KW_DATA_DIR" ]]
+}
+
 # This function reads and executes a SQL script in the database
 #
 # @sql_cmd:   SQL command to be executed on @db
@@ -49,6 +94,10 @@ function execute_command_db()
   local db_path
 
   db_path="$(join_path "$db_folder" "$db")"
+
+  if [[ ! -f "$db_path" ]] && is_default_database_target "$db" "$db_folder"; then
+    ensure_default_database_exists
+  fi
 
   if [[ ! -f "$db_path" ]]; then
     complain 'Database does not exist'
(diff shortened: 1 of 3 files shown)


aquilamacedo:
Hi @rodrigosiqueira, I opened these PR and the [other one](https://github.com/kworkflow/kworkflow/pull/1264)  because I am trying to get kw in better shape before preparing the next Debian upload :)

Could you review them when you have time?

src/lib/kw_db.sh
```diff
@@ -32,6 +32,41 @@ function execute_sql_script()
   sqlite3 -init "$KW_DB_DIR/pre_cmd.sql" "$db_path" < "$sql_path"
 }
 
+# This function creates the default user database when it is missing.
+#
+# @db:        Name of the database file
+# @db_folder: Path to the folder that contains @db
+#
+# Return:
+# 0 if the database exists or was created; non-zero otherwise
```
rodrigosiqueira: You have more than one type of return in case of failure. In this case, document those returns.

src/lib/kw_db.sh
```diff
@@ -32,6 +32,41 @@ function execute_sql_script()
...
+  local db="${1:-"$DB_NAME"}"
+  local db_folder="${2:-"$KW_DATA_DIR"}"
+  local db_path
+  local sql_path="${KW_DB_DIR}/kwdb.sql"
+
+  db_path="$(join_path "$db_folder" "$db")"
+
+  [[ -f "$db_path" ]] && return 0
+
+  if [[ "$db" != "$DB_NAME" || "$db_folder" != "$KW_DATA_DIR" ]]; then
```
rodrigosiqueira: At the beginning of this function, you added:

```
  local db="${1:-"$DB_NAME"}"
  local db_folder="${2:-"$KW_DATA_DIR"}"
```

You are checking whether the string name passed as a parameter is different from the default one. In this case, why accept a parameter if you will enforce the default name anyway? Did I miss something? I think you can just remove the parameters if you want to enforce the default name.

src/lib/kw_db.sh
```diff
@@ -32,6 +32,41 @@ function execute_sql_script()
...
+    return 2 # ENOENT
+  fi
+
+  if [[ ! -f "$sql_path" ]]; then
+    complain "Could not find the file: $sql_path"
```
rodrigosiqueira: Don't forget to add {} around bash string variables (in this commit, there are multiple instances of that).

src/lib/kw_db.sh
```diff
@@ -32,6 +32,41 @@ function execute_sql_script()
...
+    return 2 # ENOENT
+  fi
+
+  if ! mkdir -p "$db_folder"; then
```
rodrigosiqueira: Remember: https://kworkflow.org/content/codingstyle.html#avoid-short-options

Also, move the mkdir command outside the `if`part, and check "$?" in the if statement.

tests/unit/lib/kw_db_test.sh
```diff
@@ -73,6 +73,25 @@ function test_execute_sql_script()
...
 }
 
+function test_default_database_is_created_on_insert()
+{
+  local columns='("label_name","status","date","time","elapsed_time_in_secs")'
+  local rows="('build','success','2026-05-16','12:00:00','1')"
+  local output
+  local ret
+
+  rm -f "${KW_DATA_DIR}/kw.db"
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#avoid-short-options

tests/unit/lib/kw_db_test.sh
```diff
@@ -73,6 +73,25 @@ function test_execute_sql_script()
...
+
+  output=$(insert_into '"statistics_report"' "$columns" "$rows")
+  ret="$?"
+
+  assert_equals_helper 'Default database should be created on demand' "$LINENO" 0 "$ret"
+  assertTrue "($LINENO) DB file should be created" '[[ -f "$KW_DATA_DIR/kw.db" ]]'
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#string-concatenation

rodrigosiqueira (changes requested):
Hii @aquilamacedo ,
Thanks a lot for your  PR, follows some of my comments.

src/lib/kw_db.sh
aquilamacedo: You are right. I originally passed `$db` and `$db_folder` to avoid changing the behavior of explicit non-default database targets, but I agree that this made `ensure_default_database_exists()` confusing.

I removed those parameters, so the helper now only uses `$DB_NAME` and `$KW_DATA_DIR` directly. The callers check whether the requested target is the default database before calling it, so non-default targets still keep the old ENOENT and "Database does not exist" behavior.

aquilamacedo:
Hi @rodrigosiqueira, I just force pushed the updated version with the review changes. Could you take another look when you have time? 

I also added a regression test for the non-default DB case.

rodrigosiqueira:
Thanks! Change merged into the unstable branch.

PR #1264, opened 2026-05-17
src: lib: kwio: Skip empty alert commands

aquilamacedo:
Avoid evaluating an empty notification command when sound or visual alerts are requested but the corresponding command is unset.

This prevents shell syntax errors such as `eval: syntax error near unexpected token '&'` in packaged or partially configured installations while preserving the existing warning path for configured commands that are unavailable.


rodrigosiqueira:
Merged with tiny modification into the unstable branch. Thanks!

Issue #1257, opened 2026-02-25
kw send-patch fails if setting a version


melissawen:
**Description:**

`kw send-patch -s` stopped to work when passing a version (like -v8). From verbose, I noticed -v8 goes as the last option in the git send-email command generated by kw and looks like this format isn't accepted by git send-email anymore (?). At least on my side. If I move -v8 to the beginning of the list of options, the command is executed as expected.

**How to Reproduce:**

Run `kw send-patch -s --simulate -8 -v8`. You will see the following error message: `No patch files specified!`

**Expected behavior:**

Send patches.

**Screenshots:**

That's the verbose output of `kw send-patch -s -v8 -8` when it doesn't work:

```
git send-email --dry-run --to-cmd='bash /home/melissawen/.local/lib/kw/plugins/kw_mail/to_cc_cmd.sh /home/melissawen/.cache/kw to cover-letter' --cc-cmd='bash /home/melissawen/.local/lib/kw/plugins/kw_mail/to_cc_cmd.sh /home/melissawen/.cache/kw cc cover-letter' --annotate --cover-letter --no-chain-reply-to --thread -8 -v8
```

On the other hand, this command works: 

```
git send-email --dry-run -v8 --to-cmd='bash /home/melissawen/.local/lib/kw/plugins/kw_mail/to_cc_cmd.sh /home/melissawen/.cache/kw to cover-letter' --cc-cmd='bash /home/melissawen/.local/lib/kw/plugins/kw_mail/to_cc_cmd.sh /home/melissawen/.cache/kw cc cover-letter' --annotate --cover-letter --no-chain-reply-to --thread -8
```

**System information:**

 - OS: Debian testing
 - kw Version: unstable
 - Bash Version: 5.3.3

rodrigosiqueira:
Hi @melissawen , we have a fix for that in the unstable branch. When you have time, could you try it?

PR #1265, opened 2026-05-20
src: send-patch: Bug fix send-patch version (Issue #1257) & help update

ciziks:
Without this fix, `kw send-patch -s -8 -v8` places the version flag after the commit range in the `git send-email` call (**-8 -v8**), which causes git to fail with `\"No patch files specified!\"`. Placing the version flag first (**-v8 -8**) resolves the issue. The send-patch test also was changed to check if after this change, the command is working as intended.

Related: [Issue #1257 ](https://github.com/kworkflow/kworkflow/issues/1257)

Also, the short help text was outdated. Updated according to the official documentation [send-patch Documentation - kw](https://kworkflow.org/man/features/kw-send-patch.html).

_Signed-off-by: Lucas Ivars Cadima Ciziks <lucas@ciziks.com>_
_Signed-off-by: Felipe Souza <felipe.souza1506@outlook.com>_

### DIFF
--- a/src/send_patch.sh
+++ b/src/send_patch.sh
@@ -1119,7 +1119,7 @@ function parse_mail_options()
             pass_option_to_send_email+=" ${i}"
           fi
         done
-        options_values['PASS_OPTION_TO_SEND_EMAIL']="$(str_strip "$pass_option_to_send_email ${options_values['PATCH_VERSION']}")"
+        options_values['PASS_OPTION_TO_SEND_EMAIL']="$(str_strip "${options_values['PATCH_VERSION']}${pass_option_to_send_email}")"
         options_values['COMMIT_RANGE']+="$(find_commit_references "${options_values['PASS_OPTION_TO_SEND_EMAIL']}")"
         rev_ret="$?"
         shift "$#"
--- a/tests/unit/send_patch_test.sh
+++ b/tests/unit/send_patch_test.sh
@@ -305,6 +305,18 @@ function test_mail_parser()
   assert_equals_helper 'Set version option' "$LINENO" "$expected" "${options_values['PASS_OPTION_TO_SEND_EMAIL']}"
   assert_equals_helper 'Set version option' "$LINENO" '@^' "${options_values['COMMIT_RANGE']}"
 
+  # version flag must precede commit count so git send-email accepts it
+  parse_mail_options -8 -v3
+  expected='-v3'
+  assert_equals_helper 'Set version with commit count: PATCH_VERSION' "$LINENO" "$expected" "${options_values['PATCH_VERSION']}"
+  expected='-v3 -8'
+  assert_equals_helper 'Set version with commit count: PASS_OPTION_TO_SEND_EMAIL' "$LINENO" "$expected" "${options_values['PASS_OPTION_TO_SEND_EMAIL']}"
+
+  # The trailing space in '-8 ' comes from send_patch.sh appending a space after the commit count
+  # (options_values['COMMIT_RANGE']+="$commit_count "). reposition_commit_count_arg injects '--'
+  # before '-8' when no '--' is explicitly passed, which triggers that code path.
+  assert_equals_helper 'Set version with commit count: COMMIT_RANGE' "$LINENO" '-8 ' "${options_values['COMMIT_RANGE']}"
+
   parse_mail_options '--send'
   assert_equals_helper 'Set send flag' "$LINENO" 1 "${options_values['SEND']}"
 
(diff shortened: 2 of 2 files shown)


src/send_patch.sh
```diff
@@ -1119,7 +1119,7 @@ function parse_mail_options()
             pass_option_to_send_email+=" ${i}"
           fi
         done
-        options_values['PASS_OPTION_TO_SEND_EMAIL']="$(str_strip "$pass_option_to_send_email ${options_values['PATCH_VERSION']}")"
+        options_values['PASS_OPTION_TO_SEND_EMAIL']="$(str_strip "${options_values['PATCH_VERSION']}${pass_option_to_send_email}")"
```
rodrigosiqueira, line 1122: Don't you need the space between these two variables?

tests/unit/send_patch_test.sh
```diff
@@ -305,6 +305,14 @@ function test_mail_parser()
   assert_equals_helper 'Set version option' "$LINENO" "$expected" "${options_values['PASS_OPTION_TO_SEND_EMAIL']}"
   assert_equals_helper 'Set version option' "$LINENO" '@^' "${options_values['COMMIT_RANGE']}"
 
+  # version flag must precede commit count so git send-email accepts it
+  parse_mail_options -8 -v3
+  expected='-v3'
+  assert_equals_helper 'Set version with commit count: PATCH_VERSION' "$LINENO" "$expected" "${options_values['PATCH_VERSION']}"
+  expected='-v3 -8'
+  assert_equals_helper 'Set version with commit count: PASS_OPTION_TO_SEND_EMAIL' "$LINENO" "$expected" "${options_values['PASS_OPTION_TO_SEND_EMAIL']}"
+  assert_equals_helper 'Set version with commit count: COMMIT_RANGE' "$LINENO" '-8 ' "${options_values['COMMIT_RANGE']}"
```
rodrigosiqueira, line 318: I suppose the extra space in  '-8 ' is associated with my previous comment, right?

src/send_patch.sh
```diff
@@ -1152,12 +1152,12 @@ function send_patch_help()
     exit
   fi
   printf '%s\n' 'kw send-patch:' \
-    '  send-patch (-s | --send) [<options>] - Send patches via e-mail' \
```
rodrigosiqueira: Maybe open a PR just for this modification, and others associated with send-patch documentation

rodrigosiqueira (changes requested):
Hi,
First of all, thanks a lot for your PR I added some tiny comments. About the commit message, follow some observations:
1. If you work with somebody else, add their Signed-off-by. You can also add Co-Developed-by.
2. There is some extra spaces in the commit body text.
3. Expand the first commit to describe the problem better.

src/send_patch.sh
ciziks: I would argue that the space is not needed here. `pass_option_to_send_email` is always built with a leading space, since both branches of the loop use **+=" ${i}"**:

 ```
 pass_option_to_send_email+=" ${i@Q}"
  # or
  pass_option_to_send_email+=" ${i}"
```
So when **pass_option_to_send_email** has content, the space between the two variables is already there. Adding an explicit space actually introduces a double space in that case, which `str_strip` does not remove.

File: tests/unit/send_patch_test.sh (hunk 2 above)
Comment (ciziks) [reply tests/unit/send_patch_test.sh:318]: Yes, they are related indirectly. 

The trailing space in `-8 ` comes from line **1165** in `send_patch.sh` where `options_values['COMMIT_RANGE']+="$commit_count "`, which is hit because `reposition_commit_count_arg` automatically injects `--` before `-8` when no `--` is explicitly passed. This test reflects the real value of `COMMIT_RANGE`.

Comment (rodrigosiqueira) [reply tests/unit/send_patch_test.sh:318]: Could you add the explanation you provided above this specific test? I can see this becoming a mysterious thing in the future without a comment.

Comment (ciziks) [reply tests/unit/send_patch_test.sh:318]: Done on commit https://github.com/kworkflow/kworkflow/pull/1265/changes/4fe9364469692b417f0eded437478202e2573703

File: tests/unit/send_patch_test.sh [hunk 4]
```diff
@@ -311,6 +311,10 @@ function test_mail_parser()
   assert_equals_helper 'Set version with commit count: PATCH_VERSION' "$LINENO" "$expected" "${options_values['PATCH_VERSION']}"
   expected='-v3 -8'
   assert_equals_helper 'Set version with commit count: PASS_OPTION_TO_SEND_EMAIL' "$LINENO" "$expected" "${options_values['PASS_OPTION_TO_SEND_EMAIL']}"
+
+  # The trailing space in '-8 ' comes from send_patch.sh appending a space after the commit count
+  # (options_values['COMMIT_RANGE']+="$commit_count "). reposition_commit_count_arg injects '--'
+  # before '-8' when no '--' is explicitly passed, which triggers that code path.
```
rodrigosiqueira, line 317: you don't need a new commit for that, just squash this commit with the previous one.

ciziks: Done, sorry about that

rodrigosiqueira:
Hi @ciziks , I'm so sorry for my huge delay to review this change.

Anyway, it looks good! I already merged it into the unstable branch.

Issue #1163, opened 2024-10-10
kw send-patch: run checkpatch before sending patch


melissawen:
**Description:**

Add checkpatch as a requirement in kw send-patch. Should also add a config/flag option to disable this requirement.

**Why:**

It's in the kernel documentation

**Targeted users:**

Developers

**API draft:**

TBD

rodrigosiqueira:
@melissawen this modification is already in the unstable branch. When you have time, could you try it?

PR #1266, opened 2026-05-27
src: send-patch: New Feature: Run checkpatch before sending patch (Issue #1163)

ciziks:
**Add `checkpatch.pl` validation over all patches before sending them via `git send-email`**. If any patch fails, the operation is aborted with the respective error messages. Add the `checkpatch_before_send` option to `send_patch.config` to allow users to control whether `checkpatch.pl` is run before sending patches. (_default=yes_).

A `--no-checkpatch` flag is also introduced to allow bypassing the check when wanted. Also, added unit tests covering the checkpatch validation introduced in `run_checkpatch_on_patches()` and an assertion for `--no-checkpatch` in `test_mail_parser`.

Related Issue: [Issue #1163](https://github.com/kworkflow/kworkflow/issues/1163)
  
Signed-off-by: Lucas Ivars Cadima Ciziks (lucas@ciziks.com)
Signed-off-by: Felipe Souza (felipe.souza1506@outlook.com)


src/send_patch.sh
```diff
@@ -77,6 +77,44 @@ function send_patch_main()
...
+  if [[ ! -f "$checkpatch" ]]; then
+    warning 'checkpatch.pl not found, skipping style check.'
+    return 0
+  fi
+
+  for patch_path in "${patch_cache}/"*; do
+    if ! is_a_patch "$patch_path"; then
+      continue
+    fi
+    say "Running checkpatch on: $(basename "$patch_path")"
```
rodrigosiqueira: Capture the `basename` in a variable, and just concatenate it here to follow the usual code style.

src/send_patch.sh
```diff
@@ -77,6 +77,44 @@ function send_patch_main()
...
+    local cmd="perl $checkpatch"
```
rodrigosiqueira: move `local cmd` to the beginning of this function. Also, check: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/send_patch.sh
```diff
@@ -77,6 +77,44 @@ function send_patch_main()
...
+    [[ -n "$checkpatch_options" ]] && cmd+=" $checkpatch_options"
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/send_patch.sh
```diff
@@ -77,6 +77,44 @@ function send_patch_main()
...
+    cmd_manager "$flag" "$cmd $patch_path"
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#string-concatenation

src/send_patch.sh
```diff
@@ -77,6 +77,44 @@ function send_patch_main()
...
+    if [[ "$?" != 0 ]]; then
+      ret=1
```
rodrigosiqueira, line 123: iirc, checkpatch can return 3 values: 0, 1, and 2. If it returns 1, some issues were detected, but the execution was ok; if it returns 2, the script crashed. It might be worth to handle those case to make the debug easier.

src/send_patch.sh
```diff
@@ -126,6 +164,16 @@ function mail_send()
   fi
 
   kernel_root="$(find_kernel_root "$PWD")"
+
+  if [[ -n "$kernel_root" && "${send_patch_config[checkpatch_before_send]}" != 'no' &&
```
rodrigosiqueira: I was looking for the configuration option `checkpatch_before_send`, but I didn't find it, so I realized you added it in the next commit. The second commit belongs here, since it is a functional part of this commit. Remember, commits must work atomically, and this commit will exhibit the wrong behavior without the next commit. Squash the next commit into this one.

src/send_patch.sh
```diff
@@ -126,6 +164,16 @@ function mail_send()
...
+      -z "${options_values['NO_CHECKPATCH']}" ]]; then
```
rodrigosiqueira: I think this verification should be above this "if"; the current "if" should be nested below it. Usually, command-line flags have a higher precedence than configuration options.

src/send_patch.sh
```diff
@@ -126,6 +164,16 @@ function mail_send()
...
+    run_checkpatch_on_patches "$kernel_root" "$flag"
+    if [[ "$?" != 0 ]]; then
+      complain 'Patches failed checkpatch, aborting send.'
```
rodrigosiqueira: I was thinking... maybe it is ok to ignore the checkpatch failure and let the developer decide whether to proceed. For example, I can see myself extrapolating the total column count in some of the amdgpu drivers, but that's still ok. In that case, I would like to be asked whether I want to proceed.

@melissawen any thoughts?

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+function test_run_checkpatch_on_patches()
+{
+  local output
+  local fake_checkpatch="${FAKE_KERNEL}/scripts/checkpatch.pl"
+  local patch_cache="${KW_CACHE_DIR}/patches"
+  local fake_patch="${patch_cache}/0001-fake.patch"
+
+  # Isolate from any pre-existing cache and checkpatch_opts config
+  rm -rf "$patch_cache"
+  mkdir -p "$patch_cache"
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#avoid-short-options

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
 
+function test_run_checkpatch_on_patches()
+{
+  local output
+  local fake_checkpatch="${FAKE_KERNEL}/scripts/checkpatch.pl"
+  local patch_cache="${KW_CACHE_DIR}/patches"
+  local fake_patch="${patch_cache}/0001-fake.patch"
+
+  # Isolate from any pre-existing cache and checkpatch_opts config
+  rm -rf "$patch_cache"
```
rodrigosiqueira: This seems dangerous. Did you override this variable in the setUp() part? Could this be messing up the local installation of kw by removing this folder? Maybe a safe alternative would be to create a temporary fake folder and update this variable to point to the new folder.

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+  mkdir -p "$patch_cache"
+  configurations['checkpatch_opts']=''
+
+  # No checkpatch.pl, should warn and return 0
+  run_checkpatch_on_patches "$FAKE_KERNEL" 'TEST_MODE' &> /dev/null
+  assert_equals_helper 'Missing checkpatch.pl should return 0' "$LINENO" 0 "$?"
+
+  # Create a fake checkpatch.pl that always succeeds
+  printf '#!/usr/bin/perl\nexit 0;\n' > "$fake_checkpatch"
+  chmod +x "$fake_checkpatch"
```
rodrigosiqueira: All of these fake checkpatches things feel like something that should be put into the folder `tests/unit/samples/checkpatch`, and you just invoke the version that you want here instead of creating them dynamically.

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+
+  # No patches in cache, should return 0
+  run_checkpatch_on_patches "$FAKE_KERNEL" 'TEST_MODE' &> /dev/null
+  assert_equals_helper 'No patches should return 0' "$LINENO" 0 "$?"
+
+  # Fake patch with required strings for is_a_patch
```
rodrigosiqueira: This should be a new test, isolated in its own function. This makes it easier to debug things when the test breaks. Ideally, keep the tests small and focused.

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+  printf 'Subject: [PATCH] fake patch\n---\ndiff --git a/f b/f\n--- a/f\n+++ b/f\n@@ -1 +1 @@\n' > "$fake_patch"
+  local expected_checkpatch
+  expected_checkpatch="$(join_path "$FAKE_KERNEL" 'scripts/checkpatch.pl')"
```
rodrigosiqueira: Try to keep all local variables in the top of the function.

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+  output=$(run_checkpatch_on_patches "$FAKE_KERNEL" 'TEST_MODE')
+  assert_equals_helper 'Should call checkpatch on patch' "$LINENO" \
+    "Running checkpatch on: 0001-fake.patch
+perl $expected_checkpatch $fake_patch" "$output"
```
rodrigosiqueira: I suppose the perl line is a continuation. Could you put it in the same line?

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+  assert_equals_helper 'Failing checkpatch should return 1' "$LINENO" 1 "$?"
+
+  # Multiple failing patches: all should be processed before aborting
+  local fake_patch2="${patch_cache}/0002-fake.patch"
+  printf 'Subject: [PATCH 2/2] second fake patch\n---\ndiff --git a/g b/g\n--- a/g\n+++ b/g\n@@ -1 +1 @@\n' > "$fake_patch2"
+
+  output=$(run_checkpatch_on_patches "$FAKE_KERNEL" 'SILENT')
+  assert_equals_helper 'Multiple failing patches should return 1' "$LINENO" 1 "$?"
+  assertTrue "($LINENO) First patch should be checked" '[[ "$output" =~ "0001-fake.patch" ]]'
+  assertTrue "($LINENO) Second patch should be checked" '[[ "$output" =~ "0002-fake.patch" ]]'
```
rodrigosiqueira: https://kworkflow.org/content/codingstyle.html#string-concatenation

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+
+  rm -f "$fake_checkpatch" "$fake_patch" "$fake_patch2"
```
rodrigosiqueira: Try to be very careful when using `rm`. You must be sure that you are not running any unexpected variables. I recommend you rely more on temporary folders for this kind of thing. If you really want to use `rm` with a variable, ensuring the target variable has the expected value is crucial.

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+}
+
+function test_mail_send_checkpatch()
```
rodrigosiqueira, line 645: based on my comment above, also review this test.

rodrigosiqueira (changes requested):
Hi,

First of all, thanks a lot for your PR. Follows some of my comments.

src/send_patch.sh
ciziks: thanks, I've implemented it, but do you have any reference for these exit codes?
Just for curiosity, because I've searched on the official documentation (https://docs.kernel.org/dev-tools/checkpatch.html#) and the code, but the only instance I've found is:
```
	if (!defined $root) {
		print "Must be run from the top-level dir. of a kernel tree\n";
		exit(2);
	}
```

ciziks: Anyways, I think `kw` would never get to this point of letting it run outside of a kernel tree, but reassurance is always welcome :)

src/send_patch.sh
ciziks: Applied `fixup` correctly.

src/send_patch.sh
ciziks: Done on commit https://github.com/kworkflow/kworkflow/pull/1266/commits/0cee945ac18d5af1d7cc4090314dad0a251eb83a

tests/unit/send_patch_test.sh
ciziks: Fixed on commit https://github.com/kworkflow/kworkflow/pull/1266/commits/d3b2d8ad632340a3d0e209b2ba4617b14082b271

tests/unit/send_patch_test.sh
ciziks: Fixed on commit https://github.com/kworkflow/kworkflow/pull/1266/commits/fbbbbce2b89ba075742f8343d6ede1d81375bb31

tests/unit/send_patch_test.sh
ciziks: Fixed on commit https://github.com/kworkflow/kworkflow/pull/1266/commits/fbbbbce2b89ba075742f8343d6ede1d81375bb31

tests/unit/send_patch_test.sh
ciziks: Fixed on commit https://github.com/kworkflow/kworkflow/pull/1266/commits/fbbbbce2b89ba075742f8343d6ede1d81375bb31

tests/unit/send_patch_test.sh
ciziks: Done on commit https://github.com/kworkflow/kworkflow/pull/1266/commits/b527d3bfe0ea139c2e6a3137a25df963719b319c

tests/unit/send_patch_test.sh
ciziks: Done on commit https://github.com/kworkflow/kworkflow/pull/1266/commits/b527d3bfe0ea139c2e6a3137a25df963719b319c

tests/unit/send_patch_test.sh
ciziks: Done on commit https://github.com/kworkflow/kworkflow/pull/1266/commits/b527d3bfe0ea139c2e6a3137a25df963719b319c

tests/unit/send_patch_test.sh
ciziks: Done on commit https://github.com/kworkflow/kworkflow/pull/1266/commits/b527d3bfe0ea139c2e6a3137a25df963719b319c

tests/unit/send_patch_test.sh
ciziks: Done on commit https://github.com/kworkflow/kworkflow/pull/1266/commits/b527d3bfe0ea139c2e6a3137a25df963719b319c

src/send_patch.sh
rodrigosiqueira: I'm not sure if you addressed this issue in the subsequent commits (keep in mind that I review from the first commit to the last). Anyway, don't forget to address this issue in this commit.

src/send_patch.sh
rodrigosiqueira: same here.

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+    "Running checkpatch on: 0001-fake.patch
+perl $expected_checkpatch $fake_patch" "$output"
+
+  # checkpatch.pl that always fails, should return 1
+  printf '#!/usr/bin/perl\nexit 1;\n' > "$fake_checkpatch"
+  run_checkpatch_on_patches "$FAKE_KERNEL" 'SILENT' &> /dev/null
+  assert_equals_helper 'Failing checkpatch should return 1' "$LINENO" 1 "$?"
+
+  # Multiple failing patches: all should be processed before aborting
+  local fake_patch2="${patch_cache}/0002-fake.patch"
```
rodrigosiqueira: move the local variable to the top of the function.

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+  printf '#!/usr/bin/perl\nexit 0;\n' > "$fake_checkpatch"
+  chmod +x "$fake_checkpatch"
+
+  # No patches in cache, should return 0
+  run_checkpatch_on_patches "$FAKE_KERNEL" 'TEST_MODE' &> /dev/null
+  assert_equals_helper 'No patches should return 0' "$LINENO" 0 "$?"
+
+  # Fake patch with required strings for is_a_patch
+  printf 'Subject: [PATCH] fake patch\n---\ndiff --git a/f b/f\n--- a/f\n+++ b/f\n@@ -1 +1 @@\n' > "$fake_patch"
+  local expected_checkpatch
```
rodrigosiqueira: move the local variable to the top of the function.

tests/unit/send_patch_test.sh
rodrigosiqueira: Remember my previous review.

tests/unit/send_patch_test.sh
rodrigosiqueira: This feels like something that should be in a TearDown function.
P.s.: Remember my previous review.

tests/unit/send_patch_test.sh
rodrigosiqueira: This feels like a SetUp() thing.

tests/unit/send_patch_test.sh
```diff
@@ -515,6 +519,100 @@ function test_mail_send()
...
+{
+  local output
+  local fake_checkpatch="${FAKE_KERNEL}/scripts/checkpatch.pl"
+  local patch_cache="${KW_CACHE_DIR}/patches"
+  local fake_patch="${patch_cache}/0001-fake.patch"
+
+  # Isolate from any pre-existing cache and checkpatch_opts config
+  rm -rf "$patch_cache"
+  mkdir -p "$patch_cache"
+  configurations['checkpatch_opts']=''
```
rodrigosiqueira, line 530: Don't we need a test where this variable is not empty?

src/send_patch.sh
```diff
@@ -90,23 +90,27 @@ function run_checkpatch_on_patches()
...
       continue
     fi
-    say "Running checkpatch on: $(basename "$patch_path")"
-    local cmd="perl $checkpatch"
-    [[ -n "$checkpatch_options" ]] && cmd+=" $checkpatch_options"
-    cmd_manager "$flag" "$cmd $patch_path"
+    
+    patch_basename=$(basename "${patch_path}")
+    say "Running checkpatch on: ${patch_basename}"
+    [[ -n "$checkpatch_options" ]] && cmd+=" ${checkpatch_options}"
```
rodrigosiqueira: About this check, I think you can move it outside the loop. I also suspect that you can compound the command in the wrong way in the case "$checkpatch_options" has a value. How about add a unit test for this? I suppose if you add a test that sets `configurations['checkpatch_opts']=...` you can validate what I said.

rodrigosiqueira (commented):
Hi,

Thanks for updating the PR. I've added another set of reviews, but it's getting difficult to review your changes because you're adding new commits to fix issues from previous ones. Remember that this change is new, so please correct any mistakes in the commit that introduced it in your PR.

I always review commits in order, from first to last, and your series should clearly show the history. Please, try to clean up your branch.

Thanks!

ciziks:
Hey @rodrigosiqueira , thanks for all the feedback and sorry for the messy commit history, I'm still getting the hang of git rs. I've squashed everything back into the original 4 commits and also addressed the latest comments and suggestions you left. Hope it looks good now, let me know if there's anything else! ;)

rodrigosiqueira:
Hi @ciziks, sorry for the huge delay. Your PR is solid, and I really liked it; since I took too long and I'm interested in this change, I made some modifications on top of your original contribution (I kept your authorship) and merged it into the unstable branch. Thanks again.

Issue #1177, opened 2025-01-14
Apply patches from URL


melissawen:
**Description:**

It's somehow related to patch-hub, but not exactly.
I would like to pass a URL with a patch (for example, from gitlab/github or lore) and got it apply on my work tree.

Basically a wrapper for:

`$ curl URL | git am`

**Why:**

I believe it's a command executed often.

**Targeted users:**

Maintainers and reporters

**API draft:**

`kw patch-from-url <URL>`

PR #1267, opened 2026-05-27
apply-patch feature (issue #1177)

frsouzaa:
This PR introduce 2 changes:

1- Sort help list
When the h or -h commands was passed to kw, the list containing the available command was unsorted, making the readability worse.

2- apply-patch feature
This command downloads a patch from the given URL and applies it to the current work tree using "git am". It requires the command to be executed from within a kernel tree.

Related issue: [https://github.com/kworkflow/kworkflow/issues/1177](https://github.com/kworkflow/kworkflow/issues/1177)


src/help.sh
```diff
@@ -10,32 +10,32 @@ function kworkflow_help()
     '  Remote - machine reachable via the network' \
     '' \
     'Commands:' \
-    '  init - Initialize kworkflow config file' \
```
rodrigosiqueira, line 13: This commit does not belong to this PR. Open a new PR just with this.

src/apply_patch.sh
```diff
@@ -0,0 +1,160 @@
...
+function apply_patch_main()
+{
+  local ret
+  local url
+  local flag
+  local patch_content
+
+  flag=${flag:-'SILENT'}
+
+  while [[ "$#" -gt 0 ]]; do
```
rodrigosiqueira: Usually kw has a function dedicated to parse. Check this page: https://kworkflow.org/content/project_structure.html

src/apply_patch.sh
```diff
@@ -0,0 +1,160 @@
...
+  say "Downloading patch..."
+
+  patch_content=$(download_patch "$flag" "$url")
+  ret="$?"
+  if [[ "$ret" != 0 ]]; then
+    complain "$patch_content"
+    return "$ret"
+  fi
+
+  original_head=$(git rev-parse HEAD | cut -c1-7)
```
rodrigosiqueira: If you are using a git command directly in a file, consider creating a specific function for it in the git.sh file. Ofc, this should be in a separate commit.

src/apply_patch.sh
```diff
@@ -0,0 +1,160 @@
...
+  say 'Applying patch...'
+
+  log=$(cmd_manager "$flag" "git am --quiet 2>&1 <<< \"\$patch_content\"")
```
rodrigosiqueira: same as my previous comment.

src/apply_patch.sh
```diff
@@ -0,0 +1,160 @@
...
+function download_patch()
+{
+  local flag="$1"
+  local url="$2"
+  local response
+  local curl_cmd
+  local http_code
+  local patch_content
+
+  curl_cmd="curl --silent --write-out \"\n%{http_code}\" --user-agent \"kw\" --max-time 30 $url"
```
rodrigosiqueira, line 116: btw, I don't think we should handle the download of patches. b4 already does that, and it is a dependency in kw. I would rely on it.

rodrigosiqueira (changes requested):
Hi, first of all, thanks for your contribution.

I've added a preliminary review. Tbh, I didn't conduct an in-depth analysis since I believe we should have a more detailed discussion about how this feature should be implemented. For instance, I would prefer that we rely more on b4 for handling patch downloads. Perhaps we could use curl to download items for patchwork.

src/help.sh
frsouzaa: https://github.com/kworkflow/kworkflow/pull/1276

src/apply_patch.sh
frsouzaa: https://github.com/kworkflow/kworkflow/pull/1267/changes#diff-0737387c3481dc395b181a679db533e8f5d490c081bbdbbedff8e53ae6da1b6dR136

src/apply_patch.sh
frsouzaa: https://github.com/kworkflow/kworkflow/pull/1267/changes#diff-58c49966f0357db4f52a7712cfdc31eee5a934c63088568d737f844cc25dec16R63

src/apply_patch.sh
frsouzaa: https://github.com/kworkflow/kworkflow/pull/1267/changes#diff-58c49966f0357db4f52a7712cfdc31eee5a934c63088568d737f844cc25dec16R63

src/apply_patch.sh
frsouzaa: I have talked about this with @davidbtadokoro, and our understanding is that b4 is a little bit overkill for this. Using curl makes it simpler and more distro agnostic, what you think?

src/lib/git.sh
```diff
@@ -59,3 +59,20 @@ kw_git_is_ancestor()
     return 22 # EINVAL
   fi
 }
+
+kw_git_get_current_commit_hash()
```
rodrigosiqueira, line 63: Could you add a documentation?

src/lib/git.sh
```diff
@@ -59,3 +59,20 @@ kw_git_is_ancestor()
...
+{
+  git rev-parse HEAD
```
rodrigosiqueira, line 65: Could you add a unit test? I was also thinking this command could fail in different scenarios; one is executing it in a non-git repo. Could you add a function to check if it is a git repository; otherwise, flag that the command cannot be executed?

src/lib/git.sh
```diff
@@ -59,3 +59,20 @@ kw_git_is_ancestor()
...
+}
+
+kw_git_am()
+{
+  local content="$1"
+
+  cmd_manager 'SILENT' 'git am --quiet 2>&1 <<< "$content"'
```
rodrigosiqueira, line 72: Ideally, you should check `$content` before using it. Also, `git am` can fail for multiple reasons; we should not only check those cases, but also have a fallback action.

src/lib/git.sh
rodrigosiqueira, line 1: Tbh, I feel this commit can be a single-dedicated PR because getting the git am implementation correct (with many  tests) requires a lot of attention; this part is crucial, and we need to get it right.

rodrigosiqueira (changes requested):
@hi @frsouzaa ,

I apologize for the delay in providing feedback.

I have a suggestion: how about we close this pull request and open a new one that focuses specifically on the Git portion? I’ve added some comments regarding this. I believe that ensuring a correct implementation of those functions, along with any others, is more valuable at this stage.

PR #1269, opened 2026-05-29
documentation: man: features: kw-send-patch: Fix send patch manpage NAME entry

aquilamacedo:
The **kw send-patch** manpage currently uses `kw send-patch` as the manpage name, but lexgrog rejects names with spaces in the NAME section.

Use `kw-send-patch` as the manpage name so lexgrog can parse the **NAME** section correctly. Keep the real command form, `kw send-patch`, in the synopsis and examples.

I found this while checking a Debian lintian.


rodrigosiqueira:
Thanks! Merged into the unstable branch

PR #1270, opened 2026-06-03
.github: workflows: ci.yaml: unify ci workflow

frsouzaa:
Consolidate five separate CI workflow files (unit_tests.yml, integration_tests.yml, shellcheck_reviewdog.yml, shfmt.yml, and test_setup_and_docs.yml) into a single unified ci.yaml workflow file.

This change improves maintainability by:
- Centralizing all CI jobs in one place for easier updates and modifications;
- Grouping related jobs together with consistent naming and structure.

The new workflow includes:
- Updated action versions for better compatibility and security;
- Refined job triggers and configuration;
- All original functionality preserved from the five separate workflows.


frsouzaa:
Little note: If the changes are accepted, the new file must be included in the unstable branch as well

rodrigosiqueira:
Thanks for the contribution. I have to be honest: I'm having a hard time justifying this change.
  
The maintainability argument doesn't hold up imho. The five original files and the new unified file have roughly the same number of lines, so the complexity reduction is not meaningful. If anything, separate files are easier to maintain: a developer updating unit test dependencies edits a focused, single-purpose file. With the unified file, they navigate past unrelated jobs to find what they need. The same applies when something breaks: a failing job used to point you directly to its own workflow file; now everything lives together.

Could you help me understand the motivation better? Maybe you have experience with unified CI workflows that I'm missing. What problems were you running into with the separate files that led you to this approach?

frsouzaa:
Hello, thanks for the response, actually I do have a background building CI workflows using GitHub Actions. The main problem here isn't the code itself but the way it's represented in GitHub interface. Today a push/pull_request trigger five different workflows that belong to the same CI:

image

What is a bit confusing to follow up, in my opinion.

So unifying the files into a centralized workflow makes it easier to follow what happened in each event:

image

 And inside this unified workflow you can find all the analysis associated to the commit:

image

There are some exemples in [my fork of kw](https://github.com/frsouzaa/kworkflow/actions/workflows/ci.yaml).

rodrigosiqueira:
Hi,

  Thanks for the explanation, and the screenshots. The point about the Actions tab is fair. Seeing five separate rows for what's really one CI run for a commit is a real annoyance, and I get why grouping them under one workflow is appealing from that angle.

  That said, I'd prefer not to change this part of the project right now, for two reasons:

  - I'm weighing a move to Codeberg at some point, and CI config is one of the most GitHub-specific things in this repo. Restructuring it now has a decent chance of being thrown away entirely if that happens, so I'd rather not spend either of our time on it yet.
  - My time is limited these days, and current CI is stable for me. I'd rather put what time I have into other areas right now (your apply-patch PR, for one).

  None of this is a rejection of the idea, just a "not now". I hope that's fair. What do you think?

PR #1277, opened 2026-06-24
bug: Fixes "kw deploy --modules" not running.

Niyudi:
Currently, "kw deploy --modules" fails because it can't find the install_modules functions. It had been moved to "install.sh", but it was not included before running it when using the "--modules" option. This simply includes the file right before running the function.


rodrigosiqueira:
Hi @Niyudi , sorry my huge delay.

Anyway, this is a nice fix. Could you add a commit message and your Signed-off-by? Check: https://kworkflow.org/content/howtocontribute.html

PR #1278, opened 2026-06-27
tests: unit: Fix return value of compare_command_sequence for single …

### LINKED ISSUE #1142: Fix return value from compare_command_sequence
**Description:**

The `compare_command_sequence` function from `tests/unit/utils.sh` is used by many unit tests to compare expected and real output. This function, however, returns value 1 for one-line outputs, even if they match. 

This happens because the function iterates through the lines with a `counter` variable, which starts as zero and is incremented by one for each line. When incrementing its value with `$((counter++))`, this expression returns the `counter` value before the sum. 

This affects test cases with a single line, that iterates a single time and, therefore, return `0` from `$((0++))` (more details [here](https://stackoverflow.com/questions/66824241/do-expressions-enclosed-in-double-parentheses-mean-return-inside-a-bash-function)).

Although the use of `assertEquals` is suggested over `compare_command_sequence` for single line outputs, this is still an unexpected behavior. This function should be fixed to return `0` for single line inputs. It would also be effective to indicate the use of `assertEquals` over `compare_command_sequence` for this specific scenario in kworkflow documentation.

**How to Reproduce:**

Use `compare_command_sequence` and print `$?`.

**Expected behavior:**

`compare_command_sequence` returns 0 for every correct output comparison, including single line outputs.

**System information:**

 - OS: openSUSE Tumbleweed x86_64
 - kw Version: beta
 - Bash Version: 5.2.26(1)-release

LeooCesar:
…line outputs

The compare_command_sequence function was returning an exit status of 1 for single-line outputs even when the comparison matched. This happened because the postfix increment ((count++)) evaluates to 0 on the first iteration, causing bash to return an error status.

This patch changes the increment to a prefix evaluation ((++count)) to ensure a successful exit status.

Additionally, a comment was added to the function header to explicitly recommend the use of assertEquals over compare_command_sequence for single-line output comparisons.

Fixes #1142

### DIFF
--- a/tests/unit/utils.sh
+++ b/tests/unit/utils.sh
@@ -368,6 +368,8 @@ function mk_fake_kw_env()
 # @line $LINENO variable
 # @expected_res Name of the array variable containing expected strings
 # @result_to_compare A raw output from the string
+# Compare a sequence of commands output
+# Note: For single-line outputs, the use of assertEquals is highly recommended over compare_command_sequence.
 function compare_command_sequence()
 {
   local msg="$1"
@@ -386,7 +388,7 @@ function compare_command_sequence()
           "$KW_COLOR_RED" "$KW_COLOR_NONE" "$KW_COLOR_RED" "${f}" "$KW_COLOR_NONE"
       fi
     fi
-    ((count++))
+    ((++count))
   done <<< "$result_to_compare"
 }
 


rodrigosiqueira:
Hi @LeooCesar ,

First of all, thanks a lot for your contribution, and sorry for my huge delay. I made some fixes in your original change, and pushed it to the unstable branch, thanks.

